PR #376, opened 2015-01-15
Contrib/travis fix

wuan:
this PR should fix the current errors when installing postgres on Travis-CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@
 
 language: c
 
+addons:
+ - postgresql: "9.3"
 cache:
  - apt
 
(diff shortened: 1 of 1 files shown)


pguttmann:
Hmmm... This certainly seems to have addressed the postgresql "opportunity" :)
Should some/all of those cppcheck errors/warnings get entered as bugs ?

bobgob:
Yes.  I'm not sure they are all bugs, but they should certainly be looked at and verified.  Please file issues.

pguttmann:
OK, will do for the (error)s - I'll also take a quick peek to see if there are any easy/obvious fixes as well ;)

PR #407, opened 2015-03-26
fix #406 and add groupname option to cp2foss

### LINKED ISSUE #406: cp2foss fails to run after 2.6
larry is working on it.

steffen-weber:
fix #406 and add groupname option to cp2foss

### DIFF
--- a/src/cli/cp2foss.php
+++ b/src/cli/cp2foss.php
@@ -1,6 +1,7 @@
 <?php
 /***********************************************************
  Copyright (C) 2008-2014 Hewlett-Packard Development Company, L.P.
+ Copyright (C) 2015 Siemens AG
 
  This program is free software; you can redistribute it and/or
  modify it under the terms of the GNU General Public License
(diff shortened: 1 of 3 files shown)


pguttmann:
You may wish to confirm that your change to account_check gets addressed in various other cli files that may be impacted by your change ;)

ubuntu@foss262demo:~/fossology$ grep -r 'account_check' .
./src/lib/php/common-auth.php:function account_check(&$user, &$passwd)
./src/cli/fossjobs.php:account_check($user, $passwd); // check username/password
./src/cli/cp2foss.php:account_check($user, $passwd); // check username/password
./src/cli/fo_bucket_list.php:account_check($user, $passwd); // check username/password
./src/cli/fo_copyright_list.php:account_check($user, $passwd); // check username/password
./src/cli/fo_nomos_license_list.php:account_check($user, $passwd); // check username/password

steffen-weber:
Since $group is only an optional argument in account_check, the call ```account_check($user, $passwd)``` behaves as before.

pguttmann:
Ah, ok - so by assigning a "default value" to that argument in the function definition it becomes "optional" I guess.

joshovi:
Merged this in hope that this will fix Paul's tickets.

Issue #319, opened 2015-01-09
tooltips


yaobinshi:
---

Author Name: **Bob Gobeille** 
Original Redmine Issue: 7281, http://www.fossology.org/issues/7281
Original Date: 2014/07/15

---

I think our UI would be enhanced by the judicious use of tool tips.  I'd like to suggest one of the jQuery plugins because there are other jQuery features we might want to use in the future (like file path navigation and menus).

Here is one possibility that I've used and was also recommended by Roger Meier:
http://jqueryui.com/tooltip/

PR #444, opened 2015-05-27
added tooltips in license clearing view

shaheemazmalmmd:
Added tooltips in license clearing view. This is related to #319.


bobgob:
It looks like you added titles instead of tooltips like we discussed in #319 
This is certainly a low overhead way to do it but it has little flexibility and is not a good user experience because of the lag time between hovering and the tip display.  Using something like mentioned in #319 shows the tip  immediately with no lag and allows more formatting and display options.

shaheemazmalmmd:
Replaced title with jquery tooltip.

PR #461, opened 2015-06-15
generate SPDX v2.0 rdf report

steffen-weber:
This generates an SPDX v2.0 rdf report.
The SPDXTools-v2.0.0 (http://spdx.org/sites/spdx/files/SPDXTools-v2.0.0.zip) can be used to transform the report into other formats.
The test in spdx2/agent_tests/Functional/schedulerTest.php uses this official tool for verifying one generated rdf report.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,7 @@ variable.list
 /src/copyright/agent/ecc
 /src/copyright/VERSION-ecc
 /src/copyright/VERSION-copyright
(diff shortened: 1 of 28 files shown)


bufferoverflow:
Steffen, this is cool stuff!
I really like this, a dream becomes true. I've tested it with ```vagrant up```, upload, license identification and generated my first SPDX2 file with FOSSology... fantastic to see this!
+1 ready to merge
all the best!
-roger

PR #478, opened 2015-07-10
Added dual licenses to nomos

shaheemazmalmmd:
Removed the duplicate sendmail license

Changes for DUAL-gpl-commercial license


bobgob:
These changes do not look right to me.  For example, a "DUAL-Gephi" license is returned but there is no "Gephi" license.  I would call this file a "Dual GPL-3.0 - CDDL" license.  Same comment applies to the other licenses introduced.

mcjaeger:
I am not sure if that approach would be beneficial for all of these licenses, because there are general dual license texts, such as 

"If you purchased a commercial license for this product the commercial license agreement applies to this source code. Otherwise the GPL license agreement applies to this source code." (example from stackoverflow)

where is makes sense to call it DUAL-GPL-commercial. But for the Asterisk License, this one is quite special from this particular organization:

http://svnview.digium.com/svn/asterisk/trunk/LICENSE?view=markup

so my idea was just to have this similar to the SPDX naming for "single" licenses, where the issuing organization represents actually the license short name.

I am not strong on this or that scheme. But the current proposal for  - as examples - DUAL-Asterisk, DUAL-Perl, or the "SENDMAIL OPEN SOURCE LICENSE" (http://www.sendmail.org/~ca/email/sm-X/LICENSE) resulting DUAL-Sendmail looks reasonable to me. (for example, instead of DUAL-Sendmail, DUAL-Opensource-Freeware does not sound good to me)

mcjaeger:
shall we have a table and set the Id for each of these?

bobgob:
I shouldn't have written "Same comment applies to the other licenses introduced." since I didn't look at all the other licenses that were added (I only sampled them).  The problem with DUAL-Gephi is that no one knows what that means and (I think) it is nothing more than a Dual GPL-3.0 - CDDL.  "DUAL-Gephi" says nothing more than "Gephi License".  You have to be familiar with it to know what it means, and if you are familiar with it then adding "DUAL" is redundant.  But GPL-3.0 and CDDL are well known licenses.  I'm going to talk to the SPDX folks and get their logic on this.

bobgob:
FYI, http://gephi.github.io/developers/license/
lists their license as dual license CDDL 1.0 and GNU General Public License v3.  Not the Gephi license.

bobgob:
Sendmail isn't in SPDX and it is called "Sendmail License" by fedora (https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing).  So that's what I'd use.
We should probably go through each of these licenses.

bobgob:
Following SPDX Appendix IV (see http://spdx.org/sites/spdx/files/SPDX-2.0.pdf):
The Gephi license should be "(GPL-3.0 AND CDDL)"
This means we should change our license names to remove "Dual" and use their expression syntax.

bobgob:
Yikes, I should have written "(GPL-3.0 OR CDDL)"!

bobgob:
SPDX strongly recommends that instead of adding an Asterisk license that we add an Asterisk-exception.  I'll send out an email on fossology-devel to get a broader audience.

mcjaeger:
Well, OK, but the SPDX does not see individual results as license text or license template but their recommendation is to single existing individual license license text and do a conjunction with logical operators. While i see the advantage that this enables the attention to the involved license texts in the entire form, the disadvantage lies in the fact that we have a license text that we would like to add as a template for enabling full text matches.

Moreover, if we introduce the new entries as license short names, my recommendation would be use dashes or similar characters to avoid spaces between the elements.

So  "GPL-3.0-AND-CDDL"?

bobgob:
I don't see the problem with full text templates.  Can you be more specific? 

If we use dashes instead of their parentheses with spaces we will be breaking compatibility.  Personally, I agree with you and don't like the spaces but if we are going to be compatible it seems like we must use their names.   Standardized names are a big part of SPDX and automated processing of SPDX files.

mcjaeger:
With the full text templates my initial idea was that GPL-3.0-AND-X might be contents of several license templates. There it might not serve as ID for a text.

Since the SPDX it self is using organization names for licenses I felt like if there is a explicit organizational reference it is also consistent with the SPDX license naming scheme. For the Asterisk example, to call this GPL with Asterisk exception sounds inconsistent with the existing GPL exceptions to me. But ... whatever, I have no strong feeling about license names. And being in sync with the SPDX is priority.

These are the list of 

Old License ID | URL | Comments | New
--- | --- | --- | ---
Dual-Gephi | ... | ... | GPL-3.0-OR-CDDL
Dual-Logback | ... | ... | ... 
Open ssl |  ... |  OpenSSL License and the original SSLeay license | ... 
Dual-Qooxdoo |  ... |  GNU Lesser General Public License (LGPL)  and Eclipse Public License (EPL) | ...
Dual-Xmind | .. | the Eclipse Public License v1.0 and Eclipse Public License (EPL)... | ...
Dual-jquery | http://code.jquery.com/jquery-1.7.2.js  | just the header text, opposed to: http://code.jquery.com/jquery-1.8.0.js | ...
Dual-rhinoslider | http://rhinoslider.com/license/ | just the header text | ...
Tri-Mozilla | http://www.mozilla.org/MPL/boilerplate-1.1/mpl-tri-license-txt | old version, also available under: https://hg.mozilla.org/mozilla-central/file/1332fc1c15e1/toolkit/content/license.html
Tri-Qt | https://qt.gitorious.org/qt/qt/source/b12a8d47ba8a803cae88718249b8d3f7d4a85948:src/corelib/thread/qmutex.cpp |  some how tri license |  ...
Dual-Asterisk | http://svnview.digium.com/svn/asterisk/trunk/LICENSE?view=markup | well, dual in the sense of normal and commercial license | ... 
Dual-Perl | http://dev.perl.org/licenses/ | should be the text of perl packages as well | ...
Dual-Sendmail || http://www.sendmail.org/~ca/email/sm-X/LICENSE || well, not two OSS licenses but commercial optiuon... ||... ||
Dual-X3DOM || https://github.com/x3dom/x3dom/blob/master/LICENSE ||  call this xdom, not"Dual-MIT-GPL" for example... ||... ||
Dual-Glassfish | https://java.net/projects/jstl/sources/svn/content/trunk/impl/src/main/java/org/apache/taglibs/standard/functions/Functions.java?rev=97  | is available in different software projects... | ...
Dual-Hapi | http://sourceforge.net/p/hl7api/code/764/tree/trunk/hapi-mvn/hapi-base/src/main/java/ca/uhn/hl7v2/HapiContext.java |  specific for the HAPI project | ...
Dual-Netbeans | https://netbeans.org/about/legal/license.html | please see the header in verbatim. | 
Dual-GPL-Commercial | http://stackoverflow.com/questions/6786664/dual-licensing-in-source-code | example for a classic dual license GPL-commercial statement | ... |
Dual-Magnolia | http://svn.magnolia-cms.com/view/build/maven-plugins/magnolia-build-tools/trunk/src/dual-licensed/resources/LICENSE.txt.vm | just special item ("Magnolia Network Agreement") | ... |...

mcjaeger:
I have sent an e-mail to spdx-legal about this. maybe my e-mail was not well formulated. Let's wait a little for the answer.

Issue #457, opened 2015-06-08
Folder level Add / remove selected license should not affect files whose clearing status is 'green'


rlintu:
Possibility to make a license decision for several files at the same time by selecting [Edit] for a folder is a very good feature. Currently that, however, affects all files in a folder even in case the user has earlier made a license decision for the file (which is noted as a green circle in Edited results). Also files 
for which scanners have not found a license are automatically set to the same decision state.
If the user selects BSD-3-Clause for the folder which has 60 files with BSD-3-Clause  and 10 No_license_found files, then she has to manually correct the license for those 10 No_license_found
files. The situation is even worse if there are files that have been analyzed earlier with 'Apply decision to all future occurrences of this file' option (which I think is a good feature). In this case the user may not even notice that he has accidentally change the license.

[Edit] is also an option for a single file. Because the decision for the file can be and most likely is done just by selecting a file name, this [Edit] option can be removed for a single file.

mcjaeger:
thanks for pointing this out: we actually found it useful this way need to keep it. Sometimes a licensing statement says in a README something like: "all files in this folder are subject to the XY-License"

Then, even the files are not licensed themselves, the publisher of the open source package has licensed them. Accordingly, "Edit"-feature allows for mapping this into clearing decisions. (referring to your statement that also files for which scanner have not found ...)

So, we see sense in this feature. In order to support your case (and your pull request), a checkbox selection or option  appears appropriate like saying that applying the license will not be performed in case something is present with different result?

Regarding the case "Apply decisions to future occurrences": I am not sure that we are using it actually, because of the problem mentioned above that the same file could be licensed differently in different contexts. Please, let's do not go into this, as I am "software" not "legal".

We have added the feature to be consistent with the original behavior of fossology before we have added the concept of clearing decisions. So overall, if this is kind of dependent on a user's selection, it could match both interests?

rlintu:
I guess we agree on that the existing concluded license1 (f.ex. BSD-2-Clause) should not be changed to license2 resulting f.ex. BSD-2-Clause, GPL-2.0 using [Edit]. Right?

Then about setting the concluded license to 'No_license_found file'. You referred that 

>Sometimes a licensing statement says in a README something like: "all files in this folder are subject to the XY-License". 

[...]

PR #520, opened 2015-09-14
#457 - When concluded license is set using Edit option then license is not …

rlintu:
…set to a file that has no license or a file to which concluded license has already set.

This is a correction to 
https://github.com/fossology/fossology/issues/457


rlintu:
I close this one and  I will create a new pull request with a correct procedure. This one was by mistake created from the master branch.

Issue #460, opened 2015-06-09
Please "enhance" the ECC scanner capabilities.


pguttmann:
Currently the ECC scanner strings are embedded in the source code. Please investigate how to dynamically load these from an external/config file or a database table.

mcjaeger:
It think something is in the works to this regard. Not easy, because once ECC is changed, it makes also sense to apply the same change for the copyright regular expressions.

pguttmann:
Perhaps some sort of "simplified" version control along the lines of sha1/md5 for the "collection" of regexes used for either copyright or ECC scans could also be used to track changes in the regexes scanned for ?

PR #525, opened 2015-09-24
Fix issue #460

maximilianhuber:
This improves the copyright and ecc agents and should be sufficient to solve #460.

It introduces a Class which parses the config files (copyright.conf and ecc.conf) on demand and stores them in a singleton.


maximilianhuber:
Sadly the code is not stable at the moment:
```
$ cd src/copyright/agent_tests/Functional
$ while make test; do echo "try another round"; done
```
fails randomly after approx. 10 rounds.

I am currently working on this issue.

@steffen-weber this is exactly the problem you have mentioned.

maximilianhuber:
The travis failures are due to travis internal download problems.

PR #527, opened 2015-10-01
Password in DBConnection string issue#355

shaheemazmalmmd:


pguttmann:
Hmmm... Travis seems to be "stuck" on a bad entry in the CPAN Mirror listings at - http://www.cpan.org/MIRRORED.BY
The correct URL would appear to be - http://mirror.nyi.net/CPAN/
instead of http://mirror.nyi.net/pub/CPAN/

maximilianhuber:
This implementation only works when the password is the last key=value pair in `CMD`. But the ordering of these pairs is given by the order in the Db.conf file.

I think one has to search the key "password" and has to replace the corresponding value.

PR #528, opened 2015-10-10
uniform and fixed micromenu

wuan:
This hopfully makes the micromenu more usable. For example some file viewing plugins do not contain the format part and the order of the items is not the same in all cases.

This PR sets the micromenu `View` to be used for all views showing file contents (e. g. View, Copyright, Audit) and the micromenu `View-Meta` to be used for views showing meta-information of a file.

Currently in the `View` case:
> One-Shot Copyright/Email/URL | One-Shot License   •   Info | View | Audit | Copyright/Email/Url/Author | ECC | Bucket   •   Hex | Text | Formatted   •   Refresh

And in the `View-Meta` case:
> One-Shot Copyright/Email/URL | One-Shot License   •   Info | View | Audit | Copyright/Email/Url/Author | ECC   •   Refresh

Basically, the view settings are missing here.

The menu generating code inside the plugins could still be abstracted. Discussion is welcome.

### DIFF
--- a/src/lib/php/common-menu.php
+++ b/src/lib/php/common-menu.php
@@ -20,6 +20,7 @@
  * \brief common menu functions
  */
 const MENU_PATH_SEPARATOR = "::";
+const MENU_BREAK = "[BREAK]";
 
 /**
  * \brief
@@ -196,11 +197,11 @@ function menu_insert_r(&$menuItems, $path, $pathRemainder, $LastOrder, $Target, 
   if ($menuItemsExist) {
     foreach($menuItems as &$menuItem) {
       // need to escape the [ ] or the string will not match
-      if (!strcmp($menuItem->Name, $pathElement) && strcmp($menuItem->Name, "\\[BREAK\\]")) {
+      if (!strcmp($menuItem->Name, $pathElement) && strcmp($menuItem->Name, MENU_BREAK)) {
         $currentMenuItem = $menuItem;
         break;
       }
-      else if (!strcmp($menuItem->Name, "\\[BREAK\\]") && ($menuItem->Order == $LastOrder)) {
+      else if (!strcmp($menuItem->Name, MENU_BREAK) && ($menuItem->Order == $LastOrder)) {
         $currentMenuItem = $menuItem;
         break;
       }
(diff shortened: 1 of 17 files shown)


src/lib/php/common-menu.php
```diff
@@ -196,11 +197,11 @@ function menu_insert_r(&$menuItems, $path, $pathRemainder, $LastOrder, $Target,
   if ($menuItemsExist) {
     foreach($menuItems as &$menuItem) {
       // need to escape the [ ] or the string will not match
-      if (!strcmp($menuItem->Name, $pathElement) && strcmp($menuItem->Name, "\\[BREAK\\]")) {
```
wuan: Did this really work as intended before? Before the change only one [BREAK] element per submenu was possible.

src/www/ui/ui-browse-license.php
```diff
@@ -77,6 +77,7 @@ public function __construct() {
    */
   function RegisterMenus()
   {
+    return 0;
```
maximilianhuber: Do you really want to `return 0;` here?

wuan: No, thanks!

maximilianhuber:
When I try your changes, I get the following errors:
tmp

wuan:
Can you try a full install? The service.xml file needs to be updated.

maximilianhuber:
I have used
```
$  sudo make uninstall && make && sudo make install
```
on root level to install your changes.

Where is the service.xml located?

maximilianhuber:
After `vagrant up` with your fork, I get:
tmp
after adding the following lines to `src/www/ui/index.php`
```
ini_set('display_errors', 'On');
error_reporting(E_ALL | E_STRICT);
```

wuan:
The script called by `fo-postinstall` seemed to dislike the dash in the dependency injection key. That is fixed now and the error message during `vagrant up` should be gone.

The depenency injection configuration file is located at `src/lib/php/services.xml`. It should be regenerated by make from the template file `src/lib/php/services.xml` through the following lines in the Makefile:
```
# generate services.xml
services.xml: services.xml-process
```
To be sure, that the file is updated, you can try
```
$  sudo make uninstall && make clean && make && sudo make install
```

steffen-weber:
Other service ids are written in snake_case to solve the trouble with dashes.
The fatal error mentioned by @maximilianhuber was probably caused by caching which can be solved with ``sudo make empty-cache``.

wuan:
Ah, I forgot the cache, what about cleaning the cache during `make install`?

maximilianhuber:
I would support the idea with `empty-cache` while `make install`.

maximilianhuber:
With the new modifications i am able to start FOSSology, but i have found an new problem with your current branch:
tmp
and the content of the table does not load via ajax:
tmp

### what i have done:
1. vagrant up with a virgin copy of your branch.
2. setup php (for the upload)
3. upload `openssl-1.0.2d.tar` and start some agents for it
4. click on browse -> license-browser

### modifications:
I still use 
```
ini_set('display_errors', 'On');
error_reporting(E_ALL | E_STRICT);
```

wuan:
The missing variable is defined now. Please note that Ajax requests don't work because of the error display.

Issue #325, opened 2015-01-09
wget_agent redirects STDERR to /dev/null and so user doesn't see errors


yaobinshi:
---

Author Name: **Gervase Markham** 
Original Redmine Issue: 7341, http://www.fossology.org/issues/7341
Original Date: 2014/07/23
Original Assignee: Dong Ma

---

wget_agent just failed on me. It turned out that this was because I was using Upload from Server and tar couldn't read one of the files in the target directory. However, the UI didn't tell me that. All I got was:

2014-07-23 14:23:38 wget_agent [0] :: JOB[40].wget_agent[4386.localhost]: "FATAL wget_agent.c.673: rc_system is:512, CMD is:tar -cvvf  '/srv/fossology/repository/hare/wget/wget.4386' -C '/usr/src/b2g-fossology' ./  >/dev/null 2>&1"
2014-07-23 14:23:38 wget_agent [0] :: JOB[40].wget_agent[4386.localhost]: "FATAL main.c.247: Failed to archieve. GlobalURL, GlobalTempFile, TempFileDir are: /usr/src/b2g-fossology, /srv/fossology/repository/hare/wget/wget.4386, /srv/fossology/repository/hare/wget, Mode is: 40755 (octal)"
2014-07-23 14:23:38 wget_agent [0] :: JOB[40].wget_agent[4386.localhost]: agent failed with error code 50
2014-07-23 14:23:38 wget_agent [0] :: JOB[40].wget_agent[4386.localhost]: agent failed, code: 50

(Also, we need to fix the spelling of "archive" in that error.) 

We should log all output that goes to STDERR instead of throwing it away. That would have shown me:

tar: ./external/aac/documentation/.aacDecoder.pdf.sUQw6j: Cannot open: Permission denied

which explains the problem well.

Here are the sites to change:

src/wget_agent/agent/wget_agent.c
404:        snprintf(CMD,MAXCMD-1, "tar -cvvf  '%s' -C '%s' ./ >/dev/null 2>&1", TempFile, TempFilePath);
505:  snprintf(command,MAXCMD-1, "tar -cvvf  '%s' -C '%s' ./ >/dev/null 2>&1", GlobalTempFile, TempFileDirectory);
669:    snprintf(CMD,MAXCMD-1, "tar -cvvf  '%s' -C '%s' ./ %s >/dev/null 2>&1", TempFile, Path, GlobalParam);
[...]

Issue #349, opened 2015-01-09
cp2foss fails to upload a directory using '*' option


yaobinshi:
---

Author Name: **Paul Guttmann** 
Original Redmine Issue: 7677, http://www.fossology.org/issues/7677
Original Date: 2014/09/09
Original Assignee: larry shi

---

This is to correct the 2nd defect reported in Backlog - 7083
http://www.fossology.org/issues/7083

I cloned https://github.com/spdx-tools/fossology-spdx into a test directory, changed to that directory and then attempted -
cp2foss --username fossy --password fossy -f cp2fosstest '*'

This appears to fail in a similar way to Bob's reported errors.
Interestingly, when I replaced '*' with 'fossology-spdx' everything was fine...

mcjaeger:
I think that is also the error we have seen when taking files from another application.

mcjaeger:
seems to be related with #512

PR #529, opened 2015-10-12
Contrib/issue 325/349

shaheemazmalmmd:
This branch resolves issue #325 and #349


maximilianhuber:
Instead of removing `>/dev/null 2>&1` (in the commit "Changes WRT issue#325") one could replace it by `1>/dev/null` to suppress only STDOUT (but show STDERR).

PR #530, opened 2015-10-14
Password in DBConnection string issue#355

shaheemazmalmmd:


src/lib/c/libfossdb.c
```diff
@@ -140,8 +140,16 @@ PGconn* fo_dbconnect(char* DBConfFile, char** ErrorBuf)
     *ErrorBuf = malloc(ERRBUFSIZE);
     if (*ErrorBuf)
     {
+      int i = 0;
+      const char pass[9]= "password";
```
maximilianhuber: I think it would be more secure to search for `"password="` since
```
"dbname=passwordProtectedDB password=secretPassword"
```
becomes
```
"dbname=passwordP********** password=secretPassword"
```
in this implementation.

PR #538, opened 2015-10-26
split jq_cmd_args into multiple arguments (in scheduler agent)

maximilianhuber:


maximilianhuber:
This implementation has some problems at the moment and **should not be merged**, yet!

bufferoverflow:
:+1:

maximilianhuber:
Sadly, I have to say that this implementation works but has some problems. Since it is no longer needed for the other pull request (#556) I will close it.

This try might be useful for others, which want to tackel this problem.

PR #549, opened 2015-11-10
Write correct version of DB-scheme to DB

maximilianhuber:
* the scheme is equivalent to version 2.6.3.3
* this value is displayed under `/?mod=foconfig`

This pull request is mentioned in #550

### TODO

update other version strings from `trunk` to `3.0.0`, for example in
* `/VERSION`
* `/VERSIONSTRING`
* `/variable.list`


bufferoverflow:
:+1:

Issue #386, opened 2015-01-30
Upload from Server: remove or disable by default


fogninid:
cp2foss already allows to upload files directly from the server and does not expose the server to users who only have http access.
I would remove the plugin completely or disable it on a default installation.
Or does anybody see a use case which is not covered by cp2foss?

yaobinshi:
usually, cp2foss is used by other script or tool, cp2foss cover the jobs upload from server, upload from url, and upload from version control system do.

cp2foss can not do the job upload from file(you are  accessing fossology  from  another machine).
certainly if you are doing http access fossology from the fossology server, upload from server is able to do the jobs what upload from file does.

sure, you can run sudo mv /usr/{local}/share/fossology/www/ui/upload-*  ./temp_folder

fogninid:
How do you put the file on server if you do not have enough access to run cp2foss?

As it is now, every user can 'upload' a file from the server, you just need that the scheduler has the permission to read it (=everything readable by fossy:fossy).
This has obvious privacy/security implications. For example try uploading /srv/fossology/repository/localhost/files/

I would at least make the installation optional: e.g. `sudo make -C src/www/ui install-fileupload`

bobgob:
@fogninid I agree.  We hadn't considered the security implications before.  Removing it will force people who have this usage model to either use cp2foss, or keep their files in a place where they can be reached with Upload from URL.  That seems reasonable to me.   Making the install optional, would give the user a choice.  But I agree, it should definitely not be installed by default.

steffen-weber:
only files in white-listed directories can be uploaded from server with 382d4f4c2230b6ae6566c829ec179618ab5bc59b

Issue #435, opened 2015-05-18
the options 'Decider' and 'reuse' are missing on upload from url/server/vcs pages


yaobinshi:
any particular reasons?

steffen-weber:
The options for uploads from url/vcs will be added soon.
I'm not sot sure about uploads from server since security issue #386.

steffen-weber:
fixed with #559

PR #559, opened 2015-11-19
Rewrite upload pages

### LINKED ISSUE #543: The upload-pages should be consistend and based on the same template
The upload-pages:
* Upload **File**
  - `src/www/ui/page/UploadFilePage.php`  (based on `UploadPageBase`)
* Upload from **URL**
  - `src/www/ui/page/UploadUrlPage.php`  (based on `UploadPageBase`)
* Upload from **Server**
  - `src/www/ui/upload-srv-files.php` (based on `FO_Plugin`)
* Upload from **version control system**
  - `src/www/ui/upload-vcs.php` (based on `FO_Plugin`)

should be rewritten in such a way, that
* all pages are located in `src/www/ui/page/`
* all pages derive from `UploadPageBase`
* all pages use Twig with the same base-template
* all strings, which are used to build a shell command, should be escaped
* the upload permissions of the "Upload from Server" plugin should be restricted

maximilianhuber:
This pull request should close #543.

#### What was done
All upload pages are now based on `src/www/ui/page/UploadPageBase.php` and are placed in `src/www/ui/page/`.

On the **Admin -> Customize** page are new options, to configure the server upload:
itm
By default are only uploads from `localhost` and the folder `/tmp/` allowed (maybe there is also another / better canonical choice).


maximilianhuber:
This pull request will also close #386 (since one is now able to restrict the folders from which uploads are allowed) and #435

maximilianhuber:
I am thinking about renaming the upload pages. This would **prevent name clashes** for the case, when someone does the update via `sudo make install` without an preceding `sudo make uninstall`.

PR #565, opened 2015-11-26
Small changes, Added DTD to index file to prevent phpunit test case failure

shaheemazmalmmd:


maximilianhuber:
adding this line will add **a second doctype definition on every regular page**:

tmp

I will close this pull request.

PR #569, opened 2015-11-28
feat(vagrant): use proxy from host_ip:3128

bufferoverflow:
just create a file named PROXY within src root folder to use it


bufferoverflow:
included in #572

mcjaeger:
hm I have some hard time testing cntlm because I would not have linux hosts so far (and do not fancy installing on win) - would you be Ok with getting the proxy host and port as configuration as well?

bufferoverflow:
I did so many rework on this via #572 and I'm ready to do more as soon as the other stuff is committed, there is NO side effect when PROXY env variable is not set.

PR #572, opened 2015-11-29
Fix/vagrant ninka

bufferoverflow:
This enables ninka on vagrant, based on PR #569

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,6 +4,21 @@
 # Copyright Siemens AG, 2014
 # SPDX-License-Identifier:  GPL-2.0 LGPL-2.1
 
+$add_proxy_settings = <<PROXYSCRIPT
+# get gateway of running VM and set proxy info within global profile
+echo 'export host_ip="`netstat -rn | grep "^0.0.0.0 " | cut -d " " -f10`"'  > /etc/profile.d/proxy.sh
+echo 'export ftp_proxy="http://$host_ip:3128"'                             >> /etc/profile.d/proxy.sh
+echo 'export http_proxy="http://$host_ip:3128"'                            >> /etc/profile.d/proxy.sh
+echo 'export https_proxy="https://$host_ip:3128"'                          >> /etc/profile.d/proxy.sh
+
+# load the proxy within current shell
+. /etc/profile.d/proxy.sh
+
+if ! sudo grep -q http_proxy /etc/sudoers; then
+  sudo su -c 'sed -i "s/env_reset/env_reset\\nDefaults\\tenv_keep = \\"http_proxy https_proxy ftp_proxy\\"/" /etc/sudoers'
+fi
+PROXYSCRIPT
+
 $build_and_test = <<SCRIPT
 export DEBIAN_FRONTEND=noninteractive
 
@@ -50,6 +65,13 @@ Vagrant.configure("2") do |config|
   end
 
   config.vm.network :forwarded_port, guest: 80, host: 8081
+
+  # use proxy from host if environment variable PROXY=true
+  if ENV['PROXY']
+    if ENV['PROXY'] == 'true'
+      config.vm.provision :shell, :inline => $add_proxy_settings
+    end
+  end
 
   # call the script
   config.vm.provision :shell, :inline => $build_and_test
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -28,10 +28,13 @@ sudo dpkg-reconfigure locales
 
 sudo apt-get update -qq -y
 
-sudo apt-get install -qq curl php5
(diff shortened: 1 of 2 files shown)


maximilianhuber:
This creates a folder `/home/vagrant/ninka` which is owned by root. I would further suggest to put the script to install ninka in a file `install/scripts/install-ninka.sh`.
(I had to install `DBD::SQLite` and `Spreadsheet::WriteExcel` bevore `perl Makefile.pl` worked.)

I would also suggest to call in the Vagrantfile the other scripts in `install/scripts` to allow uploads larger then 2MB and optimize the postgres settings by default.

For testing purposes would also suggest to supply a script `install/scripts/install-spdx-tools.sh` install the **spdx-tools**:
```bash
#!/usr/bin/env bash

cd "$(basename $0)/../../src/spdx2/agent_tests/Functional/"
wget https://github.com/spdx/tools/releases/download/V2.0.2/spdx-tools-2.0.2-jar-with-dependencies.jar
cd ~-
```

I was not yet able to test the PROXY settings, but the dummy file should be added to the `.gitignore`.

bufferoverflow:
I can do a little rework towards ```install/scripts/install-ninka.sh```, my inspiration was from .travis.yml.

I did not add PROXY to .gitignore because people using it should be aware that PROXY is set.

bufferoverflow:
DBD::SQLite and Spreadsheet::WriteExcel is installed via Debian Packages libspreadsheet-writeexcel-perl and libdbd-sqlite3-perl, see also cd65d40

maximilianhuber:
I think adding such small scripts to `install/scripts/` is usefull for users, which install fossology from source and without vagrant.

The installation of the perl modules as debian packages is a good idea.

bufferoverflow:
rework done.
I will provide a spdx-tools install script as soon as this is merged.

maximilianhuber:
#### As an alternative to the dummy file:

pass an argument to vagrant using the following line in bash
```BASH
$ PROXY='true' vagrant up
```
and then check in the Vagrantfile for `ENV['MY_VAR']=='true'` and call another sript defined in the Vagrantfile:
```RUBY
[...]
  if ENV['PROXY']
    if ENV['PROXY'] == 'true'
      config.vm.provision :shell, :inline => $add_proxy_settings
    end
  end
[...]
```

(Sorry, but I do not like such dummy files)

bufferoverflow:
@maximilianhuber everything ok now?

maximilianhuber:
Yes, I think everything looks good now.

At the moment I do same testing and I think it can be merged after them.

Vagrantfile
```diff
@@ -4,6 +4,25 @@
 # Copyright Siemens AG, 2014
 # SPDX-License-Identifier:  GPL-2.0 LGPL-2.1
 
+$add_proxy_settings = <<PROXYSCRIPT
+
+# get gateway of running VM and set proxy info within global profile
+echo 'export host_ip="`netstat -rn | grep "^0.0.0.0 " | cut -d " " -f10`"' >> /etc/profile.d/proxy.sh
```
petermarko: I think it should be > instead of >>

Vagrantfile
```diff
@@ -4,6 +4,25 @@
...
+. /etc/profile.d/proxy.sh
+
+if ! sudo grep -q http_proxy /etc/sudoers; then
+  sudo su -c 'sed -i "s/env_reset/env_reset\\nDefaults\\tenv_keep = \\"http_proxy https_proxy ftp_proxy\\"/" /etc/sudoers'
+fi
+
+
+PROXYSCRIPT
+
+
```
petermarko: Compared styling with original SCRIPT, there are 4 empty lines to many...

bufferoverflow:
@maximilianhuber any news on this or shall I rework again based on comments from  @petermarko ?

maximilianhuber:
I think, adding the suggested changes, would be good.

I was not yet able to test the proxy stettings, since i was not able to find a machine with the fitting proxy setup. But, since this does not interfere as long as the env variable is not set, i think it can be merged.

I will merge it after the reworking.

bufferoverflow:
e voila

maximilianhuber:
Some travis checks are failing? But current master is green and I do not think that this PR is the cause of this.

I am confused and i will restart some tests for master.

Maybe the "new Platform" is the problem:
tmp

bufferoverflow:
travis does now docker based builds but this requires no sudo within .travis.yml and this is not feasible at the moment for fossology. I will restart some builds on the pull request.

bufferoverflow:
the issues where not caused by the change, it seems that it fails because of some travis related issues (they are migrating to a new datacenter as well)

maximilianhuber:
The last thing, i would suggest, is to move the dependencies `git libspreadsheet-writeexcel-perl libdbd-sqlite3-perl` to `nstall-ninka.sh`. But that is not important and can be done later.

Issue #331, opened 2015-01-09
conf files still say SVN_REV


yaobinshi:
---

Author Name: **Bob Gobeille** 
Original Redmine Issue: 7449, http://www.fossology.org/issues/7449
Original Date: 2014/08/15
Original Assignee: Dong Ma

---

Our conf files (e.g. src/nomos/VERSION) still says SVN_REV and VERSION says "trunk" even though we have moved to git.
SVN_REV should probably be called COMMIT
and build versions should probably be the branch name.
Or perhaps we use something like:
VERSION=2.5.0+
BRANCH=master
COMMIT=abcdef

mcjaeger:
related to #550 but not covered entirely by #550

maximilianhuber:
closed by #581

PR #581, opened 2015-12-03
Removed SVN_REV from files and replaced Commit with commit_hash #331

shaheemazmalmmd:


maximilianhuber:
What are the scripts `utils/fo-mktar*` for? They still use/write the `SVN_REV`:
tmp

PR #582, opened 2015-12-06
[#580] improved protocol inference

### LINKED ISSUE #580: Insecure login warning when using HTTPS
I'm getting this message on the home screen on version 3.0.0 using apache 2.4.7 on Ubuntu 14.04, despite using HTTPS: `This login uses HTTP, so passwords are transmitted in plain text. This is not a secure connection.`

I've traced this back to https://github.com/fossology/fossology/blob/61ea9dbadb65b721b50bdfca2c5e81a3b1c979fd/src/www/ui/page/HomePage.php#L51 which is checking `$_SERVER['SERVER_PROTOCOL']`. As explained [on stackoverflow](http://stackoverflow.com/questions/16825243/why-is-phps-server-protocol-showing-http-1-1-even-when-using-https), `$_SERVER['HTTPS']` might be a better way to check for this.

Relevant `$_SERVER` variables from my system:

    ["HTTPS"] => string(2) "on"
    ["SERVER_SOFTWARE"] => string(21) "Apache/2.4.7 (Ubuntu)"
    ["SERVER_PORT"] => string(3) "443"
    ["REQUEST_SCHEME"] => string(5) "https"
    ["SERVER_PROTOCOL"] => string(8) "HTTP/1.1"

timurphy:
`$_SERVER['SERVER_PROTOCOL']` does not reliably provide a means for inferring the use of HTTPS. `$_SERVER['HTTPS']` is set on many web servers if HTTPS is being used, so check that first.

This fixes #580

### DIFF
--- a/src/www/ui/core-auth.php
+++ b/src/www/ui/core-auth.php
@@ -227,7 +227,16 @@ class core_auth extends FO_Plugin
       global $Plugins;
       $this->vars['info'] = $Plugins[$initPluginId]->infoFirstTimeUsage();
     }
-    $this->vars['protocol'] = preg_replace("@/.*@", "", @$_SERVER['SERVER_PROTOCOL']);
+
+    if (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != "off")
+    {
+      $this->vars['protocol'] = "HTTPS";
+    }
+    else
+    {
+      $this->vars['protocol'] = preg_replace("@/.*@", "", @$_SERVER['SERVER_PROTOCOL']);
+    }
+
     $this->vars['referrer'] = $referrer;
     $this->vars['loginFailure'] = !empty($userName) || !empty($password);
     if (!empty($userName) && $userName!='Default User') {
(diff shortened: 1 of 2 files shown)


bufferoverflow:
+1

build is broken because of:
```
FoLibCPPTest::test_runBadCommandQueryCheckIfError

expecting errors

-----

ERROR: ERROR:  syntax error at end of input

LINE 1: CREATE TABLE tbl( integer

                                 ^

On: CREATE TABLE tbl( integer

-----

 : OK

FoLibCPPTest::test_runSchedulerConnectConstructor : OK
```

maximilianhuber:
Is this realy the problem? The line above says `expecting errors` and the line below says `: OK`

maximilianhuber:
The Problem is, that
```
$ phpunit -csrc/phpunit.xml --testsuite="$PHPTESTSUITE"
```
does not finish. Here is a screenshot of a green test:
tmp

timurphy:
phpunit also didn't finish for #572 and #583, so I doubt the error is caused by this PR.

steffen-weber:
The problem was caused by an updated phpUnit which does not support php 5.5; with commit 3c9381ad0a6bf2df0a4147f41762c2094a84042f tests will be fine again.

PR #593, opened 2015-12-16
fix(ui-download): add $filenameFallback solve #589

### LINKED ISSUE #589: SPDX download fails when name contains percent '%' character
Seen in version 3.0.0. Attempting to download generated SPDX when the name contains the '%' character results in this error:

    The filename fallback cannot contain the "%" character.

The file was created successfully and can be copied manually from `/srv/fossology/repository/report`.

Steps to reproduce:
1. Select `Upload from URL`
2. Set URL to https://upload.wikimedia.org/wikipedia/commons/thumb/1/1e/Laterite_-_%C4%90%C3%A1_ong_2.JPG/320px-Laterite_-_%C4%90%C3%A1_ong_2.JPG
3. Click `Upload`
4. Navigate to Browse screen and select `Export SPDX RDF` action

maximilianhuber:
* use `$filename` and replace all critical chars (__'%'__,__'/'__,__'\\'__) with __'_'__ to create a `$filenameFallback`
* closes #589

### DIFF
--- a/src/www/ui/ui-download.php
+++ b/src/www/ui/ui-download.php
@@ -219,8 +219,12 @@ class ui_download extends FO_Plugin
     $session = $container->get('session');
     $session->save();
 
+    $filenameFallback = str_replace('%','_',$filename);
+    $filenameFallback = str_replace('/','_',$filenameFallback);
+    $filenameFallback = str_replace('\\','_',$filenameFallback);
+
     $response = new BinaryFileResponse($path);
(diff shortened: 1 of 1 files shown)


bufferoverflow:
:+1: 

I'm not that familiar with the current PHP stack, but something like this [toAscii](https://doc.nette.org/en/2.3/strings#toc-toascii-s) might be an option as well.

maximilianhuber:
I dont think that `toAscii` would help. We just want to get rid of these three chars since they are not safe in file names (on some systems).

bufferoverflow:
perfect, was just an idea.

Issue #273, opened 2015-01-09
any user who is not the owner can delete any folder via /delagent -F


yaobinshi:
---

Author Name: **larry shi** 
Original Redmine Issue: 6741, http://www.fossology.org/issues/6741
Original Date: 2014/03/18
Original Assignee: Dong Ma

---

None

yaobinshi:
---

Original Redmine Comment
Author Name: **Bob Gobeille** 
Original Date: 2014/03/18 17:42:39 +0000

---

We are past the 2.5 code complete date, so moving to 2.6.

mcjaeger:
should have been made a common ticket with #272

maximilianhuber:
maybe related to #368

shaheemazmalmmd:
Folders are not linked with users in database model. any user above write permission can create a folder and delete it except "root folder" of him from UI, But from command line he can delete his root folder as well. i think we need to link user_fk to folder table or foldercontents to restrict other users. ?? but if the group admin is deleting it then he cant because he is not a owner. 

or the user with top level folder access can delete his child's not all folders he doesn't belongs to ??

maximilianhuber:
related to / same as 
* #186 and
* #1293 (in the old bug-tracker)

PR #602, opened 2015-12-24
fix[#273] any user who is not the owner can delete any folder via /delagent -F  

shaheemazmalmmd:
This pull request also restricts users other than owners and Admins to delete a folder from UI as well.

### DIFF
--- a/src/delagent/agent/delagent.c
+++ b/src/delagent/agent/delagent.c
@@ -157,7 +157,7 @@ int main (int argc, char *argv[])
     DeleteUpload(DelUpload); 
     fprintf(stdout, "The upload '%ld' is deleted by the user '%s'.\n", DelUpload, user_name);
   }
-  if (DelFolder) { DeleteFolder(DelFolder); }
+  if (DelFolder) { DeleteFolder(DelFolder, user_id); }
   if (DelLicense) { DeleteLicense(DelLicense); }
 
   /* process from the scheduler */
@@ -167,7 +167,7 @@ int main (int argc, char *argv[])
     {
       Parm = fo_scheduler_current();
       
-      if (ReadParameter(Parm) < 0)
+      if (ReadParameter(Parm, user_id) < 0)
         exit(-1);
     }
   }
--- a/src/delagent/agent/delagent.c
+++ b/src/delagent/agent/delagent.c
@@ -142,6 +142,11 @@ int main (int argc, char *argv[])
   sprintf(agent_rev, "%s.%s", VERSION, COMMIT_HASH);
   /* Get the Agent Key from the DB */
   fo_GetAgentKey(db_conn, basename(argv[0]), 0, agent_rev, agent_desc);
+
+  if(Scheduler == 1 && user_id == -1){
+    /* get user_id for user who queued the agent */
+    user_id = fo_scheduler_userID();
+  }
   
   if (ListProj) ListUploads(user_id, user_perm);
   if (ListFolder) ListFolders(user_id);
--- a/src/delagent/agent/util.c
+++ b/src/delagent/agent/util.c
@@ -525,7 +525,7 @@ void DeleteUpload (long UploadId)
  * \param int DelFlag 0=no del, 1=del if unique parent, 2=del unconditional
  *
  */
-void ListFoldersRecurse	(long Parent, int Depth, long Row, int DelFlag)
+void ListFoldersRecurse	(long Parent, int Depth, long Row, int DelFlag, int user_id)
 {
   int r,MaxRow;
   long Fid;
@@ -533,6 +533,7 @@ void ListFoldersRecurse	(long Parent, int Depth, long Row, int DelFlag)
   char *Desc;
   PGresult *result;
   char SQL[MAXSQL];
+  int count = 0;
 
   /* Find all folders with this parent and recurse */
   memset(SQL,'\0',sizeof(SQL));
@@ -562,9 +563,9 @@ void ListFoldersRecurse	(long Parent, int Depth, long Row, int DelFlag)
         Desc = PQgetvalue(result,r,3);
         if (Desc && Desc[0]) printf(" (%s)",Desc);
         printf("\n");
-        ListFoldersRecurse(Fid,Depth+1,Parent,0);
+        ListFoldersRecurse(Fid,Depth+1,Parent,0,user_id);
       }
-      else ListFoldersRecurse(Fid,Depth+1,Parent,1);
+      else ListFoldersRecurse(Fid,Depth+1,Parent,1,user_id);
     }
     else
     {
@@ -591,6 +592,17 @@ void ListFoldersRecurse	(long Parent, int Depth, long Row, int DelFlag)
         LOG_FATAL("folder id=%ld will be deleted with flag %d\n",Parent,DelFlag);
       if (DelFlag==0)
         break;
+
+      memset(SQL,'\0',sizeof(SQL));
+      snprintf(SQL,sizeof(SQL),"SELECT count(*) FROM folder join users on (users.user_pk = folder.user_fk or users.user_perm = 10) where folder_pk = %ld and users.user_pk = %d;",Parent,user_id);
+      result = PQexec(db_conn, SQL);
+      if (fo_checkPQresult(db_conn, result, SQL, __FILE__, __LINE__)) exit(-1);
+      count = atol(PQgetvalue(result,0,0));
+      if(count == 0){ 
+        LOG_FATAL("Folder does not exist (or) Permission denied to delete this folder.\n");
+        break;
+      }
+
       if (DelFlag==1 && UnlinkContent(Parent,Row,1))
       {
         break;
@@ -681,7 +693,7 @@ void ListFolders (int user_id)
   snprintf(SQL,sizeof(SQL),"SELECT folder_pk,parent,name,description,upload_pk FROM folderlist ORDER BY name,parent,folder_pk;");
   result = PQexec(db_conn, SQL);
   if (fo_checkPQresult(db_conn, result, SQL, __FILE__, __LINE__)) exit(-1);
-  ListFoldersRecurse(1,1,-1,0);
+  ListFoldersRecurse(1,1,-1,0,user_id);
 
   /* Find detached folders */
   MaxRow = PQntuples(result);
@@ -702,7 +714,7 @@ void ListFolders (int user_id)
       Desc = PQgetvalue(result,i,3);
       if (Desc && Desc[0]) printf(" (%s)",Desc);
       printf("\n");
-      ListFoldersRecurse(Fid,1,i,0);
+      ListFoldersRecurse(Fid,1,i,0,user_id);
     }
   }
 
@@ -780,9 +792,9 @@ void ListUploads (int user_id, int user_perm)
  * 
  * \param long FolderId the fold id to delete
  **/
-void DeleteFolder (long FolderId)
+void DeleteFolder (long FolderId, int user_id)
 {
-  ListFoldersRecurse(FolderId,0,-1,2);
+  ListFoldersRecurse(FolderId,0,-1,2,user_id);
 #if 0
   /** Disabled: Database will take care of this **/
   MyDBaccess(DB,"VACUUM ANALYZE foldercontents;");
@@ -803,7 +815,7 @@ void DeleteFolder (long FolderId)
  * \return 0 on OK, -1 on failure.
  *
  **/
-int ReadParameter (char *Parm)
+int ReadParameter (char *Parm, int user_id)
 {
   char FullLine[MAXLINE];
   char *L;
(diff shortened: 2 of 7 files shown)


src/delagent/agent/delagent.c
```diff
@@ -157,7 +157,7 @@ int main (int argc, char *argv[])
     DeleteUpload(DelUpload); 
     fprintf(stdout, "The upload '%ld' is deleted by the user '%s'.\n", DelUpload, user_name);
   }
-  if (DelFolder) { DeleteFolder(DelFolder); }
+  if (DelFolder) { DeleteFolder(DelFolder, user_id); }
```
maximilianhuber: I think, the permission check should be encapsulated in a function `check_permission_folder_del` similar to `check_permission_del` for uploads. (One might also rename the functions to be consistend)

Then one does not have to pass the `user_id`.

shaheemazmalmmd: If we encapsulate a function for folder check, then also userid is required for ListFoldersRecurse and deletefolder should pass the userid. BTW user permission check for folders are done in ListFoldersRecurse only.

maximilianhuber:
Question: what happens if you try to delete a folder to which you have RW-permissions but which contains a **subfolder where you don't have permissions**. 

I am not sure, whether it is possible to create this situation.

shaheemazmalmmd:
If the user is having permissions to delete *top level folder* he can delete all the child folders.

maximilianhuber:
I am not able to delete any folder at the moment. (Fresh vagrant up with current state)

what i have done:

1. log in as fossy
2. **create** folder
3. try to **delete** the created folder

The corresponding job is green but in its log it states that 
```
2016-01-07 15:21:30 delagent [0] :: JOB[6].delagent[21213.localhost]: "FATAL util.c.591: folder id=8 will be deleted with flag 2"
2016-01-07 15:21:30 delagent [0] :: JOB[6].delagent[21213.localhost]: "FATAL util.c.596: Folder does not exist (or) Permission denied to delete this folder."
```
Btw: i dont think that the log level should be FATAL. It should be something like INFO for the first and WARN for the second line.

#### Screenshot
screenshot

maximilianhuber:
One is able to reach the lines 576 of `util.c` without passing any permission checks and there is the comand `DeleteUpload(atol(PQgetvalue(result,r,4)));` which
* actually delete uploads and
* does not use the given `user_id`.

I still think that the permissions should only be checked in the line [delagent.c:160](https://github.com/fossology/fossology/pull/602/files#diff-05f1dc96f586f173cfec58c500ee82a0R160).

maximilianhuber:
I think you should move `src/www/ui/admin-folder-delete.php` to `src/delagent/ui/admin-folder-delete.php`.

shaheemazmalmmd:
Commit [#ec161b]

Just tested delete upload from delagent but permission check is executed on line delagent.c:160. itself 

Could you tell me how to reproduce that ??

src/delagent/agent/util.c
```diff
@@ -591,6 +591,12 @@ void ListFoldersRecurse	(long Parent, int Depth, long Row, int DelFlag)
         LOG_FATAL("folder id=%ld will be deleted with flag %d\n",Parent,DelFlag);
```
maximilianhuber: You could move this line after line 593. This would remove the cluttering of the output of `./delagent --user fossy --password fossy -f`:

```
# Folders
   1 :: Software Repository
      8 :: FolderToDelete
FATAL util.c.591: folder id=8 will be deleted with flag 0

      3 :: test1 (a first test)
         4 :: test1suba (a first subfolder of test1)
FATAL util.c.591: folder id=4 will be deleted with flag 0

FATAL util.c.591: folder id=3 will be deleted with flag 0
[...]
```

maximilianhuber:
Since the line 576 of util.c is reacheable without passing any permission checks, i am able to delete all its contained uploads.

#### Situation:
```
$ ./delagent --user rwUser --password password -f
# Folders
   1 :: Software Repository
     -- :: Contains: jersey-server-linking-1.8.jar
      9 :: testNotToDelete
        11 :: ToDeleteButWUploads
           -- :: Contains: jersey-server-linking-1.8.jar
     11 :: ToDeleteButWithUploads
        -- :: Contains: jersey-server-linking-1.8.jar
```
where 
* the folder **11** was created by the rwUser with read-write permissions, all other folders were created by fossy
* the **two uploads** were uploaded by fossy and one was copied into the Software Repository
* the folder **11** was copied(*) to 9 by fossy

I think it is clear, that **rwUser** should not be allowed to remove the directory **11**.

#### What was done
delete the folder **11**, i.e.
```
$ ./delagent --user rwUser --password password -F 11
```
#### The result
One gets the situation
```
$ ./delagent --user rwUser --password password -f
# Folders
   1 :: Software Repository
      9 :: testNotToDelete
```
and all uploads are removed.

--

(*) Why is it called a copy if it keeps the same id? I would expect to be able to make a "safety-copy" and then delete one of them.

maximilianhuber:
(on commit e56677e)

#### Problem:
You don't get any return code from `ListFoldersRecurse`, i.e. you go past the lines 565 resp. 567 in any case.

#### Situation:
```
$ ./delagent --user fossy --password fossy -f
# Folders
   1 :: Software Repository
     12 :: FolderOwnedByRWUser
        13 :: FolderOwnedByAdmin
```
where
* **12** was created by rwUser
* **13** was created by fossy

#### What was done
delete folder **12**, i.e.
```
$ ./delagent --user rwUser --password password -F 12
FATAL util.c.591: folder id=13 will be deleted with flag 1

FATAL util.c.596: Folder does not exist (or) Permission denied to delete this folder.

FATAL util.c.591: folder id=12 will be deleted with flag 2
```
(It tried to delete **13** and recognized, that it has no permissions to do that. After that it deleted **12** ignoring subfolders)

#### The result:
```
$ ./delagent --user fossy --password fossy -f
# Folders
   1 :: Software Repository
INFO: Default folder not deleted.
# Unlinked folders
  12 :: FolderOwnedByAdmin
     13 :: FolderOwnedByAdmin
```

maximilianhuber:
I still think that the best solution would be to build a function, which recursively checks for permissions **before** starting the deletion. A similar PHP function could be used to prevent users in the UI from deleting folders/files where the have no permissions.

maximilianhuber:
The delagent job should not return SUCCESS, when permissions were denied. It especialy should not be green in the 'Show Jobs'-view.

maximilianhuber:
#### Problem
If you try to delete a Folder to which you don't have permissions, the delagent will still delete all subfolders where you have permissions.

This means that `$ ./delagent --user rwUser --password password -F 1` will delete all folders owned by **rwUser**.

One might argue that this is correct behaviour but i don't think so.

maximilianhuber:
I have got the following stack trace and **i will try to add more information soon.**

<pre>
<b>Fatal error:</b>
Uncaught exception 'Fossology\Lib\Exception'
with message
    'error executing:
        Fossology\Lib\Dao\FolderDao::insertFolderContents:
            INSERT INTO foldercontents (parent_fk, foldercontents_mode, child_id) VALUES ($1, $2, $3) -- -- Array ( [0] => 1 [1] => 1 [2] => )
            ERROR: <b>null value in column "child_id" violates not-null constraint</b>
            DETAIL: Failing row contains (5, 1, 1, null).'
in /usr/local/share/fossology/lib/php/Db/DbManager.php:120
<b>Stack trace:</b>
#0 /usr/local/share/fossology/lib/php/Db/ModernDbManager.php(71): Fossology\Lib\Db\DbManager->checkResult(false, 'Fossology\Lib\D...') 
#1 /usr/local/share/fossology/lib/php/Dao/FolderDao.php(100): Fossology\Lib\Db\ModernDbManager->execute('Fossology\Lib\D...', Array) 
#2 /usr/local/share/fossology/lib/php/Dao/FolderDao.php(245): Fossology\Lib\Dao\FolderDao->insertFolderContents(1, 1, NULL) 
#3 /usr/local/share/fossology/www/ui/admin-folder-create.php(67): Fossology\Lib\Dao\FolderDao->createFolder('newFolder', '', 1) 
#4 /usr/local/share/fossology/www/ui/ad in /usr/local/share/fossology/lib/php/Db/DbManager.php on line 120
</pre>

#### What was done:
1. Fresh install from **master**
2. checkout **siemens:contrib/issue-273**
3. `make` and `sudo make install` ⁽¹⁾
4. try to add a folder (as fossy)

after that i got
screenshot


--
⁽¹⁾ maybe one should also run `fo-postinstall`. I will try

shaheemazmalmmd:
Yes as user_fk column is additionally added to folders table...

maximilianhuber:
#### Problem:
A user without write-permissions is able to delete folders owned by himself.

#### Situation:
A user with read-write-permissions creates a folder and then gets edited to be
only a read-user (or non permission user).

#### Solution:
Only users, which have at least read-write-permission level should be allowed to
delete folders.

maximilianhuber:
Tested on a fresh install of e635ae1.
#### Problem:
Got a segmentation falt.

#### Situation:
```
$ ./delagent --user=rw --password=pass -f                                                                                                                                                            
# Folders
   1 :: Software Repository
      5 :: ownedByRW
         8 :: ownedByRoot
         6 :: ownedByRWSub
```

#### What was done:
```
$ ./delagent --user=rw --password=pass -F 5
column number 1 is out of range 0..0
Segmentation fault (core dumped)
```

maximilianhuber:
Tested on a fresh install of e635ae1.
#### Problem
Can't delete folder witch contains upload created by someone else even if Admin permissions to the upload are granted.

#### Situation:
```
$ ./delagent --user=rw --password=pass -f
# Folders
   1 :: Software Repository
     10 :: RW1
```
where the folder **10** is owned by **rw** and the file is uploaded by **fossy** and **fossy** has granted **rw** admin access to the upload, i.e. 
screenshot

#### What was done:
```
$ ./delagent --user=rw --password=pass -F 10
WARN: Folder does not exist (or) Permission denied to delete folder '10'.
```

shaheemazmalmmd:
In the above comment you are giving admin permissions to the group and users in the group for that upload. May be in the current code delagent is not linked with groups. it only check for the user permissions.
In this case user still have r/w permissions.

maximilianhuber:
But delagent handles such permission checks, when I delete a file from the UI. But this might be implemented in PHP and not in the c code.

shaheemazmalmmd:
Have you deleted the file form ui with r/w permissions ?? well i need to check this as deleting from ui also calls delagent.

maximilianhuber:
I have used the page `?mod=admin_upload_delete` where I only see uploads to which I have **r/w** permissions (if my group only has **r** permissions the upload does not get shown). And I'm able to delete files.

shaheemazmalmmd:
here ?mod=admin_upload_delete handles deletion of upload not folder.

Folder deletion from UI also uses delagent.

maximilianhuber:
Both, folder and upload deletion, is handled by delagent.

shaheemazmalmmd:
NO i think only folder deletion is handled in delagent.

maximilianhuber:
There is the [HUGE function `DeleteUpload`](https://github.com/fossology/fossology/blob/master/src/delagent/agent/util.c#L127-L618) in `/src/delagent/agent/util.c` and it is called from [`delagent.c`](https://github.com/fossology/fossology/blob/master/src/delagent/agent/delagent.c#L150-L159).

--
Futher, delagent has the following options
```
$ ./delagent
Usage: ./delagent [options]
  List or delete uploads.
  Options
  -i   :: Initialize the DB, then exit.
  -u   :: List uploads IDs.
  -U # :: Delete upload ID.
  -f   :: List folder IDs.
  -F # :: Delete folder ID and all uploads under this folder.
          Folder '1' is the default folder.  '-F 1' will delete
          every upload and folder in the navigation tree.
  -s   :: Run from the scheduler.
  -T   :: TEST -- do not update the DB or delete any files (just pretend)
  -v   :: Verbose (-vv for more verbose)
  -c # :: Specify the directory for the system configuration
  -V   :: print the version info, then exit.
  --user|-n # :: user name
  --password|-p # :: password
```

and you can delete files via
<pre>
<b>$ ./delagent --user=fossy --password=fossy -u</b>
# Uploads
2 :: coreutils-8.24.tar.xz
3 :: exampleData.sh
4 :: lintian_2.5.38.tar.xz
<b>$ ./delagent --user=fossy --password=fossy -U 3</b>
The upload '3' is deleted by the user 'fossy'.
</pre>

src/delagent/agent/util.c
```diff
@@ -986,3 +996,104 @@ void Usage (char *Name)
...
+  memset(SQL,'\0',sizeof(SQL));
+  snprintf(SQL,sizeof(SQL),"SELECT upload_pk FROM folderlist WHERE folder_pk is null AND parent = %ld;",parent);
+  result = PQexec(db_conn, SQL);
+  if (fo_checkPQresult(db_conn, result, SQL, __FILE__, __LINE__)) exit(-1);
+  count = PQntuples(result);
+  if (count == 0){
+    return 1;
+  }
+  for(i = 0; i < count; i++){
+    upload_id = atol(PQgetvalue(result, 0, i));
```
maximilianhuber: The function `PQgetvalue` is of the form
``` 
char *PQgetvalue(const PGresult *res, int row_number, int column_number);
```
this means that you are iterating through the columns. Is that what you want? (`PQntuples` returns the number of rows) (see [link](http://www.postgresql.org/docs/8.4/static/libpq-exec.html))

The same applies to [line 1060](https://github.com/fossology/fossology/pull/602/files#diff-020758a99280d21bcd094cf5f770500bR1060).

shaheemazmalmmd:
Closing This pull Request. we will soon address this issue from other branch.

Issue #290, opened 2015-01-09
agent_desc not being initialized in install


yaobinshi:
---

Author Name: **Bob Gobeille** 
Original Redmine Issue: 6891, http://www.fossology.org/issues/6891
Original Date: 2014/04/09
Original Assignee: Dong Ma

---

I just noticed that the agent.agent_desc field is not being initialized during an install.  I traced this error all the way back to the 2.0.0 release.

mcjaeger:
I think it was corrected at some time, but requires retest.

mcjaeger:
seems to be still an issue, however some agents do seem to work. see screenshot in next commont :D

maximilianhuber:
![tmp](https://cloud.githubusercontent.com/assets/1187050/11275456/f0942458-8edd-11e5-819f-24a9a08c49ba.png)

PR #603, opened 2015-12-30
fix[#290] agent_desc not being initialized in install

shaheemazmalmmd:


maximilianhuber:
Hi, there are several travis jobs which fail, in the step **functional_test** of nomos. See for example https://travis-ci.org/fossology/fossology/jobs/99420265#L1521

I think one has to have a look at that and check, what causes theese problems.

maximilianhuber:
I can confirm, that the  `agent_descs` are initialized after a fresh `vagrant up`:
screenshot

src/lib/c/libfossagent.c
```diff
@@ -160,6 +161,12 @@ FUNCTION int fo_GetAgentKey(PGconn* pgConn, const char* agent_name, long Upload_
   }
 
   Agent_pk = atol(PQgetvalue(result, 0, 0));
+  /* Check If agent_desc is null*/
```
maximilianhuber: You could also update the outdated comment

src/maintagent/agent/maintagent.c
```diff
@@ -56,6 +57,8 @@ int main(int argc, char **argv)
   COMMIT_HASH = fo_sysconfig("maintagent", "COMMIT_HASH");
   VERSION = fo_sysconfig("maintagent", "VERSION");
   snprintf(agent_rev, sizeof(agent_rev), "%s.%s", VERSION, COMMIT_HASH);
+  /* Get the Agent Key from the DB */
```
maximilianhuber: This comment does not match. It should tell something like "update `agent_desc` if necessary"

maximilianhuber:
Test are all ok. I will merge it but i would be happy if you could update the two comments first.

PR #619, opened 2016-01-21
fix(security) SQL injection vulnerability in read_permission

timurphy:
The `read_permission` function used by command line tools does not sanitize inputs before using them in SQL statements. This change uses the DbManager class instead, passing the inputs as parameters. This does not change any functionality; it only removes this SQL injection vulnerability. Style reflects usage in other parts of the file.


timurphy:
Note: #620 discovered while testing this PR.

PR #621, opened 2016-01-21
Do not change decision for files already having a concluded license

rlintu:
See #457

### DIFF
--- a/src/deciderjob/agent/DeciderJobAgent.php
+++ b/src/deciderjob/agent/DeciderJobAgent.php
@@ -121,37 +121,55 @@ class DeciderJobAgent extends Agent {
     return true;
   }
 
-  private function noLicensesCheck(&$detectedLicenses, &$removedLicenses)
+  /**
+   * \brief Check if a new clearing decision can be added
+   * 
+   * Clearing is not allowd if:
+   *   - concluded license has already been added, or
+   *   - agent has detected No_license_found, or
+   *   - user has concluded that there is no valid license in a file
+   *
+   * @param $itemTreeBounds
+   * @param $userId
+   * @param $groupId
+   * @param $eventsFromJob
+   * @return boolean
+   */
+  private function isClearingAllowed($itemTreeBounds, $userId, $groupId, $eventsFromJob)
   {
-    $noLicenses = false;
-    if (count($detectedLicenses) == 0)
+    $detectedLicenses = $this->licenseDao->getAgentFileLicenseMatches($itemTreeBounds);
+    list($concludedLicenses, $removedLicenses) = $this->clearingDecisionProcessor->getCurrentClearings($itemTreeBounds, $groupId, LicenseMap::CONCLUSION);
+    if (!is_null($concludedLicenses))
     {
-      return true;
-  	}
-    if ($detectedLicenses[0]->getLicenseId() == 507)
-    {
-      return true;
-    }
-
-    foreach ($detectedLicenses as $dlic)
-    {
-      $noLicenses = false;
-      foreach ($removedLicenses as $rlic)
+      foreach ($concludedLicenses as $license)
       {
-        if ($dlic->getLicenseId() == $rlic->getLicenseId())
+        $event = $license->getClearingEvent();
+        if (!is_null($event))
         {
-          $noLicenses = true;
-          break;
+          // If enventId related to the concluded lisense is not in current jobs
+          // event list, then the conclusion already exist and new evenst are discarded
+          $eventId = $event->getEventId();
+          if (array_search($eventId, $eventsFromJob) == false)
+          {
+            return false;
+          }
         }
       }
-      if (!$noLicenses)
-      {
-        break;
-      }
     }
-    return $noLicenses;
+    
+    // In some test cases variables can be null which make test cases to fail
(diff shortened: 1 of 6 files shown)


maximilianhuber:
This code breaks some tests in the phpunit suite.

rlintu:
I'm studying those test issues.

maximilianhuber:
Hi @rlintu, do you have any progress here? Do you know what the problem is?

rlintu:
I have looked the unit tests. Corrections have been made so that the tests can be now run properly. However, there are 4 failed cases. That is to my understanding due to the change how it is checked whether decision can be made or not which is the whole point of this patch. I didn't find easy way to modify tests. Probably totally new test cases have to be made. Due to other obligations I haven't had time to start writing new test cases yet.

rlintu:
One unit test has been removed because it didn't use test database as input. The modified deciderjobagent expects valid db structure so it is easier to use test database as input. One test case has been modified and one new has been added. Some clarifications are added which do not affect the test execution. Now deciderjob unit tests should succeed.

src/deciderjob/agent/DeciderJobAgent.php
```diff
@@ -84,7 +88,8 @@ function processClearingEventOfCurrentJob()
       $containerBounds = $this->uploadDao->getItemTreeBounds($uploadTreeId);
       foreach($this->loopContainedItems($containerBounds) as $itemTreeBounds)
       {
-        $this->processClearingEventsForItem($itemTreeBounds, $userId, $groupId, $additionalEventsFromThisJob);
+      	$events = $this->clearingDao->getRelevantClearingEvents($itemTreeBounds, $groupId);
```
maximilianhuber: I am not sure what this line is doing, `$events` is never used.

src/deciderjob/agent/DeciderJobAgent.php
```diff
@@ -117,13 +122,44 @@ function processUploadId($uploadId)
     return true;
   }
 
+  private function noLicensesCheck(&$detectedLicenses, &$removedLicenses)
+  {
+  	$noLicenses = false;
```
maximilianhuber: It might be a good idea to fix the indentation here.

src/deciderjob/agent/DeciderJobAgent.php
```diff
@@ -117,13 +122,44 @@ function processUploadId($uploadId)
...
+  }
+
   /**
    * @param ItemTreeBounds $itemTreeBounds
    * @param int $userId
    */
   protected function processClearingEventsForItem(ItemTreeBounds $itemTreeBounds, $userId, $groupId, $additionalEventsFromThisJob)
   {
-    $this->dbManager->begin();
+  	$this->dbManager->begin();
```
maximilianhuber: Please also think about removing this tab.

src/deciderjob/agent/DeciderJobAgent.php
```diff
@@ -139,11 +175,58 @@ protected function processClearingEventsForItem(ItemTreeBounds $itemTreeBounds,
 
     if ($createDecision)
     {
-      $this->clearingDecisionProcessor->makeDecisionFromLastEvents($itemTreeBounds, $userId, $groupId, DecisionTypes::IDENTIFIED, $this->decisionIsGlobal, $additionalEventsFromThisJob);
+      /* Do not add/change concluded license if:
```
maximilianhuber: Please use spaces instead of tabs

src/deciderjob/agent_tests/Functional/schedulerTest.php
```diff
@@ -225,7 +226,8 @@ private function runnerDeciderScanWithNoEventsAndOnlyNomosShouldNotMakeADecision
     $this->dbManager->queryOnce("INSERT INTO highlight (fl_fk,start,len) VALUES(12222,12,3)");
     $this->dbManager->queryOnce("INSERT INTO highlight (fl_fk,start,len) VALUES(12222,18,3)");
 
-    list($success,$output,$retCode) = $runner->run($uploadId=2, $userId=6, $groupId=4, $jobId=31, $args="");
+    /* uploadId 1 includes pfile 4 */
+    list($success,$output,$retCode) = $runner->run($uploadId=1, $userId=2, $groupId=2, $jobId=31, $args="");
```
maximilianhuber: Why do you change every `$userId` to `2` and every `$groupId` to `2`? I do not see the change in the setup which would explain these changes.

rlintu: The reason was to use user and group ids which have been defined in test database.

wuan:
What is the current status here? The formatting in DeciderJobAgent.php is still broken.

rlintu:
Travis reports 1 failed monk test which I cannot repeat for some reason. I looked quickly bulkTest.php. Probably the reason is that clearing is queued to deciderjob and actually there is a previous clearing for the file. The point of this PR is to prevent adding a new clearing in that case. But that needs more study.

I do not know if anyone has tried this PR to see how it works in practice and whether there is a view that it is a good idea at all. Probably that is why there is no milestone target either. I have used earlier and this current version extensively and it fits to our work flow. However, I understand that others may have different way of working. So it is better to withdraw this?

rlintu:
withdrawn

PR #622, opened 2016-01-21
Fix error which emerges in PHP <= 5.4

rlintu:


maximilianhuber:
This fix does not work. The value `["/tmp/"]` is a list but `("/tmp/") == "/tmp/"` is just a string. The correct solution is to write `array("/tmp/")` as shown in PR #605.

(and please do not delete the trailing newline at the end of the file.)

mcjaeger:
we need to recheck, maybe reassign reviewer?

steffen-weber:
In addition to the fix, a unit test would be nice.

PR #624, opened 2016-01-22
Feature/user edit 198

steffen-weber:

PR #625, opened 2016-01-25
Feature/user edit 198

steffen-weber:


maximilianhuber:
Is this exactly the same code as #624? Does travis create inconsistent VMs or why was the old PR constantly failing with the same error?

--
I will test this feature.

PR #630, opened 2016-02-01
fix(delagent): delagent error message wording

timurphy:
Changing message "You 'username' does not have permission..." to "User 'username' does not have permission..."


maximilianhuber:
Thank you for this commit, but delagent gets rewritten at the moment. I Think I will cherry-pick your changes to the dev-branch and close this PR.

PR #631, opened 2016-02-02
feature(ui): add security check to `user-edit.php`

maximilianhuber:
you are either allowed to do this if you are admin or if you are editing your own account

### DIFF
--- a/src/www/ui/user-edit.php
+++ b/src/www/ui/user-edit.php
@@ -55,6 +55,15 @@ class UserEditPage extends DefaultPlugin
     $SessionUserRec = $this->GetUserRec($user_pk);
     $SessionIsAdmin = $this->IsSessionAdmin($SessionUserRec);
 
+    $user_pk_to_modify = intval($request->get('user_pk'));
+    if (!($SessionIsAdmin or
+          empty($user_pk_to_modify) or
+          $user_pk == $user_pk_to_modify))
+    {
+      $vars['content'] = _("Your request is not valid.");
+      return $this->render('include/base.html.twig', $this->mergeWithDefault($vars));
+    }
+
     $vars = array('refreshUri' => Traceback_uri() . "?mod=" . self::NAME);
 
     /* If this is a POST (the submit button was clicked), then process the request. */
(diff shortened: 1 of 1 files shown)


steffen-weber:
* the global auth-UserId, i.e. Auth::getUser(), instead of $_SESSION['UserId'] should be used: $user_pk is available here
* `$request->get('user_pk')` instead of `GetParm('user_pk', PARM_TEXT)` which internally accesses superglobals directly
* since requestUserId (resp. one of the function call mentioned above) is required to evaluate the condition in every case, using a local var for $request->get('user_pk') might be good

src/www/ui/user-edit.php
```diff
@@ -55,6 +55,14 @@ protected function handle(Request $request)
     $SessionUserRec = $this->GetUserRec($user_pk);
     $SessionIsAdmin = $this->IsSessionAdmin($SessionUserRec);
 
+    if (!($SessionIsAdmin or
```
steffen-weber: Since `or` is evaluated lazy, $user_pk_to_modify is not definined for $SessionIsAdmin==true, but required to compute !(true or anything) or $user_pk == $user_pk_to_modify in that case.

maximilianhuber: Sorry, but I don't get what you mean. I use `$user_pk_to_modify` only in this statement and only after the line, where it will be defined. I don't think that lazy evaluation is a problem here.

steffen-weber: It won't be defined for $SessionIsAdmin==true due to [lazy evaluation](https://en.wikipedia.org/wiki/Lazy_evaluation).
Equivalently, `(!(true or empty($a=2)) or 2==$a)` will be false (and throw a notice) while `(!(true | empty($a=2)) or 2==$a)` is true.
Moreover, [empty()](http://php.net/manual/en/function.empty.php) supporting expressions, rather than only variables, is a new feature of php 5.5.0

maximilianhuber: Thank you for the hint. Is updated...

Issue #608, opened 2016-01-07
Incorrect username in .fossology.rc crashes tools


timurphy:
Using version 3.0.0-107-g0cca7ce, typing an incorrect user into `.fossology.rc` crashes command line tools and prints a stack trace, which contains the username and password in cleartext:

    $ fossjobs -u2
    invalid user namestring(296) "#0 /usr/local/share/fossology/lib/php/common-auth.php(92): Fossology\Lib\Dao\UserDao->getUserAndDefaultGroupByUserName('fossy1')
    #1 /usr/local/lib/fossology/fossjobs.php(131): account_check('fossy1', 'fossy', '')
    #2 /usr/local/lib/fossology/fo_wrapper(77): require('/usr/local/lib/...')
    #3 {main}"

If the username is correct but the password is wrong, a helpful message is printed instead:

    $ fossjobs -u2
    User name or password is invalid.

We probably want that second message for both cases.

PR #632, opened 2016-02-09
fix[#608] handiled exception in common-auth.php for incorrect usernam…

shaheemazmalmmd:
…e in .fossology.rc crashes tools


maximilianhuber:
This looks good, i will merge it after a short test

PR #633, opened 2016-02-11
fix(spdx2): fix several bugs in DEP5  and SPDX2 reports

maximilianhuber:
the used `or` is not the correct operator to combine licenses. The conjunction `and` should have been used.

### DIFF
--- a/src/spdx2/agent/spdx2.php
+++ b/src/spdx2/agent/spdx2.php
@@ -30,6 +30,7 @@ use Fossology\Lib\Data\Upload\Upload;
 use Fossology\Lib\Db\DbManager;
 use Fossology\Lib\Proxy\LicenseViewProxy;
 use Fossology\Lib\Proxy\ScanJobProxy;
+use Fossology\Lib\Proxy\UploadTreeProxy;
 
 include_once(__DIR__ . "/version.php");
 include_once(__DIR__ . "/services.php");
@@ -211,9 +212,16 @@ class SpdxTwoAgent extends Agent
     $itemTreeBounds = $this->uploadDao->getParentItemBounds($uploadId,$uploadTreeTableName);
     $clearingDecisions = $this->clearingDao->getFileClearingsFolder($itemTreeBounds, $this->groupId);
     $this->heartbeat(0);
+
     $filesWithLicenses = $this->getFilesWithLicensesFromClearings($clearingDecisions);
+    $this->heartbeat(0);
+
+    $this->addClearingStatus($filesWithLicenses,$itemTreeBounds);
+    $this->heartbeat(0);
 
     $licenseComment = $this->addScannerResults($filesWithLicenses, $itemTreeBounds);
+    $this->heartbeat(0);
+
     $this->addCopyrightResults($filesWithLicenses, $uploadId);
     $this->heartbeat(0);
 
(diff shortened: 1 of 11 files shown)


maximilianhuber:
The second fix needs to be verified, since it also modifies the output for other formats. (Do not merge)

maximilianhuber:
This change breaks the tests for SPDX2, but they might be desirable. Needs to be discussed with @mcjaeger

mcjaeger:
shall I do the copyright statement testing?

maximilianhuber:
No, we should first talk about the changes indirectly introduced to the SPDX2 format.

maximilianhuber:
**The tests still need to be adapted.** They are testing that the SPDX report contains the email address `condor-admin@cs.wisc.edu` which is not a correct copyright statement and thus should not be contained.

maximilianhuber:
* the commit **	fix(spdx2): make escaping in templates consistend** solves #639
* the commit **feature(spdx2): also use clearing state...** solves issue #640

maximilianhuber:
The custom license texts are at the moment represented in the following way:

If
* it `$TXT` is a custom license text for the license with shortname "SHR"

then
* a new license get entered with the shortname `SHR-md5($TXT)`.

It might be a good idea to use a shorter unique identifier.

maximilianhuber:
TODO: fix also #643

maximilianhuber:
**TODO:** how to represent/show Iicense Comments in the reports?

* SPDX2 supports **License Comments** (and File Comments)
* DEP5 only has file comments

maximilianhuber:
@wuan and @steffen-weber: can one of you review the code in this PR? The functionality will be reviewed by @mcjaeger.

There is still one problem, that the occurrences of `</text>` within a `<text>...</text>` block is not escaped. I am not yet sure how  to escape that thing in the tag:value format file.

maximilianhuber:
@wuan: Is this PR reviewed?

src/spdx2/agent/spdx2.php
```diff
@@ -223,11 +256,14 @@ protected function renderPackage($uploadId)
    * @param ClearingDecision[] $clearingDecisions
    * @return string[][][] $filesWithLicenses mapping item->'concluded'->(array of shortnames)
    */
-  protected function getFilesWithLicensesFromClearings(&$clearingDecisions)
+  protected function getFilesWithLicensesFromClearings($itemTreeBounds)
```
steffen-weber: please add type hint and update `@param`

maximilianhuber: Is done.

maximilianhuber:
@steffen-weber and @wuan: If the code look

src/spdx2/agent/spdx2.php
```diff
@@ -299,25 +365,39 @@ protected function toLicensesWithFiles(&$filesWithLicenses, $treeTableName)
...
+        if($licenses['isCleared'])
+        {
+          $msgLicense = "None";
+        }
+        else
+        {
+          $msgLicense = "NoLicenseConcluded";
+        }
       }
+      $this->toLicensesWithFilesAdder($licensesWithFiles,array($msgLicense),$licenses['copyrights'],$fileId,$fullPath);
```
steffen-weber: [Line 300](https://github.com/fossology/fossology/pull/633/files#diff-7bba19f12bab5e8b7e4b602c33f2e955L300) has to be updated as well, otherwise Files with concluded licenses are listed twice: one time with concluded license and second time with whatever $msgLicense was in previous loop.

maximilianhuber: I will have a look, thank you for revealing this bug.

maximilianhuber: fixed (with fixup commit `feature(spdx2): also use clearing state` to 74dd460)

PR #634, opened 2016-02-11
feature(license-list): improve UI and allow more agents

maximilianhuber:
This PR introduces many new features to the **license-list** UI and CLI.

### UI
It now looks like
tmp
and grants access to many options which were previously only reachable from the cli, i.e.
* excluding of files using a simple pattern and
* ecluding of subfolders.
One is now also able to choose which agents to include. (Previously only Nomos was used)

### CLI
For the CLI there were also two new commands `fo_monk_license_list` and `fo_ninka_license_list` introduced (They all could simply be combined to a single command line tool).

### misc
All interfaces use a common function `getLicensesPerFileNameForAgentId` in **LicenseDao.php** (Previously there were two different implementations in **fo_nomos_lisense_list.php** and **ui-license-list.php**).

### Performance comparison for CLI (for a random example)
the old implementation took:
```
real    0m9.492s
user    0m0.070s
sys     0m1.536s
``` 
the new implementation takes:
```
real    0m1.464s
user    0m0.226s
sys     0m0.073s
```
and a diff shows that the results are (up to order of the lines and up to order of individual licenses) excactly the same.

--
~~This PR will be squashed before merge.~~ DONE

### DIFF
--- a/src/lib/php/Dao/LicenseDao.php
+++ b/src/lib/php/Dao/LicenseDao.php
@@ -324,7 +324,7 @@ class LicenseDao extends Object
                                                    $selectedAgentIds=null,
                                                    $includeSubfolders=true,
                                                    $excluding='',
-                                                   $ignore=0)
+                                                   $ignore=false)
   {
     $uploadTreeTableName = $itemTreeBounds->getUploadTreeTableName();
     $statementName = __METHOD__ . '.' . $uploadTreeTableName;
@@ -399,7 +399,7 @@ ORDER BY lft asc
       {
         continue;
       }
-      
+
       $this->updateStackState($pathStack, $rgtStack, $lastLft, $row);
       $path = implode($pathStack,'/');
       $this->addToLicensesPerFileName($licensesPerFileName, $path, $row, $ignore);
--- a/src/lib/php/Dao/LicenseDao.php
+++ b/src/lib/php/Dao/LicenseDao.php
@@ -33,7 +33,7 @@ use Monolog\Logger;
 class LicenseDao extends Object
 {
   const NO_LICENSE_FOUND = 'No_license_found';
-  
+
   /** @var DbManager */
   private $dbManager;
   /** @var Logger */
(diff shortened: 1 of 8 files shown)


maximilianhuber:
@mcjaeger: you could test (but not merge) the new gui at `?mod=license-list` introduced in this branch. BUT: the order of the files is still bad: folderwise it is sorted correctly but not filewise.

maximilianhuber:
@steffen-weber: If you find enough spare-time, you could have a look at this PR.

src/lib/php/Dao/LicenseDao.php
```diff
@@ -313,6 +314,147 @@ public function getLicenseIdPerPfileForAgentId(ItemTreeBounds $itemTreeBounds, $
...
+    $uploadTreeTableName = $itemTreeBounds->getUploadTreeTableName();
+    $statementName = __METHOD__ . '.' . $uploadTreeTableName;
+    $param = array();
+
+    if ($includeSubfolders)
+    {
+      $param[] = $itemTreeBounds->getLeft();
+      $param[] = $itemTreeBounds->getRight();
+      $condition = " lft BETWEEN $1 AND $2";
+      $condition .= " AND (ufile_mode & (1<<28)) = 0";
```
steffen-weber: This condition means 'ufile is no artifact'. With same condition in else case, the check in [line 413](https://github.com/fossology/fossology/pull/634/files#diff-691d17f7359c7e2f3074a185349fa863R413) could be avoided.

src/lib/php/Dao/LicenseDao.php
```diff
@@ -313,6 +314,147 @@ public function getLicenseIdPerPfileForAgentId(ItemTreeBounds $itemTreeBounds, $
...
+        ufile_name,
+        lft, rgt, ufile_mode, pfile_fk
+      FROM $uploadTreeTableName
+      WHERE $condition) AS subselect1
+LEFT JOIN (SELECT rf_shortname,pfile_fk,agent_fk
+           FROM license_file, license_ref
+           WHERE rf_fk = rf_pk) AS subselect2
+  ON subselect1.pfile_fk = subselect2.pfile_fk
+$agentSelect
+ORDER BY lft asc, rgt desc
```
steffen-weber: If adj2nest agent was successful, `lft asc` should imply `rgt desc`.

steffen-weber:
The $nameRange parameter for LicenseDao::getLicensesPerFileNameForAgentId seems to work strange. I guess, the range ['b','d'] applied to bdir/adir/cfile would yield bdir/cfile. Could you write some unit tests for new methods?

maximilianhuber:
Yes, I think you are right that the `$nameRange` feature is broken. I will remove it since it is unused.

PR #642, opened 2016-02-24
fossupload_status print usage on error or --help

timurphy:
Currently the usage information is coded but never printed. If --uploadId is not
provided, the error 'no uploadId supplied' is printed and the program exits. With
this change, usage is printed after this error, and also if the --help flag is used.

### DIFF
--- a/src/cli/fossupload_status.php
+++ b/src/cli/fossupload_status.php
@@ -29,16 +29,23 @@ global $Plugins;
 error_reporting(E_NOTICE & E_STRICT);
 
 $Usage = "Usage: " . basename($argv[0]) . " [options]
+  --help      = display this help text
   --username  = user name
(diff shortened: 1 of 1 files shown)


timurphy:
Travis build failing with this:

    $ sudo apt-get install $CXX $CC --force-yes || sudo apt-get install $CC --force-yes
    Reading package lists... Done
    Building dependency tree       
    Reading state information... Done
    E: Unable to locate package clang++-3.6
    E: Couldn't find any package by regex 'clang++-3.6'
    E: Unable to locate package clang-3.6
    E: Couldn't find any package by regex 'clang-3.6'

Not related to this PR

maximilianhuber:
I have restarted the travis builds. Maybe they pass this time

PR #644, opened 2016-02-25
fix(dashboard): missing $this-> in method call

timurphy:
Bug introduced in cf2a177f7f731f35e8587bec2e4f1cbe5a251edd. This is breaking the Dashboard page.
Tested on version `3.0.0-167-ge9e4a36`


maximilianhuber:
You are totally right

Issue #647, opened 2016-03-02
Bulk Scan not matching - OK, buthard to see


mcjaeger:
A user submits a bulk scan and the bulk scan dos not match.

Indeed, for some licenses different copyright statements with the same license text prevent a full match of the monk when performing the bulk scan.

However, the user is confused because the bulk scan display (bulk history, lower right in the detailed file view) does not highlight the differences which are just a year in the full license text. (the text is displayed w/o line breaks and therefore it is indeed hard to see).

Maybe, a highlighting of the matches is too much, but what could help is to display the matching degree of each of the bulk items for the file displayed. By this, the user can check that actually monk went over the file, but did not determine a sufficient match. 

In the copyright case, this could lead to a '99%' degree, indicating the user that a tiny section of the license text used for bulk is not matching a file. 

Attached a screenshot, where the "2000-2009" in the middle of the text is not matching "2000, 2001, ..."

![036 oss sw360 20160302 screenshot bulk scan match not visible](https://cloud.githubusercontent.com/assets/3418977/13456868/6ee83cda-e065-11e5-9df2-3324d21ed1a6.png)
 ...

shaheemazmalmmd:
Percentage can be displayed as monk agent easily provides % of match. this can be displayed from table.

Issue #585, opened 2015-12-07
Supported PHP versions


steffen-weber:
PHP 5.3 and PHP 5.4 have reached End of life, i.e., the releases are no longer supported and users of these releases should upgrade as soon as possible, as they may be exposed to unpatched security vulnerabilities, see http://php.net/supported-versions.php
Hence Fossology should not support these releases anymore to avoid responsibility for possible vulnerabilities.

Also the active support for PHP 5.5 is ended and only fixes for critical security issues will be provided. Latest test failures of Travis CI (https://travis-ci.org/fossology/fossology/builds/95226989) were caused by PHPUnit 5 requiring PHP 5.6. Since End of life is in 7 months, end supporting PHP 5.6 soon is highly recommended.

Any opinions? Should this question be send to the mailing list(s)?

bufferoverflow:
Yes, I would say skip 5.3 and 5.4. Debian and Ubuntu is already on 5.6
Here is a list of php versions per distro: http://phpversions.info/operating-systems/

rlintu:
That would mean to drop support for Ubuntu 12.04 LTS. I think there are so many important bug fixes coming for 3.0 that those should be available for Ubuntu 12.04 as well. Maybe the release with new features can then be available only for later releases.

pguttmann:
"Any opinions? Should this question be send to the mailing list(s)?" - Given: "That would mean to drop support for Ubuntu 12.04 LTS. "
I think sending the question to the mailing list(s) would definitely be worthwhile.

maximilianhuber:
Note: If we decide to drop support for 5.3 then some parts of the code introduced in the pull request #649 could be removed.

maximilianhuber:
There is currently a PR (#689) which only works with PHP>=5.5.9 (i.e. not on Ubuntu 12.04, Debian 7 or RHEL 7.1).

@pguttmann: Was this question already discussed on the mailing list?

My opinion: drop support for PHP <5.5.9

bufferoverflow:
according to http://php.net/eol.php we should drop all php versions < 5.6

PR #649, opened 2016-03-08
Create xml coverage

steffen-weber:
Add coverage badged from coveralls.io

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,3 +86,8 @@ script:
  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.5" ]]; then composer require --dev --no-update phpunit/phpunit ~4; fi
  - phpunit -csrc/phpunit.xml --testsuite="$PHPTESTSUITE"
  - set +e
+
+after_script:
+ - mkdir build
+ - mkdir build/logs
+ - php src/vendor/bin/coveralls -vv -x clover.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,6 +88,5 @@ script:
  - set +e
 
 after_script:
- - mkdir build
- - mkdir build/logs
+ - mkdir -p build/logs
  - php src/vendor/bin/coveralls -vv -x clover.xml
--- a/src/composer.json
+++ b/src/composer.json
@@ -1,6 +1,8 @@
 {
+    "name": "fossology/fossology",
+    "description": "FOSSology is a open source license compliance software system and toolkit.",
     "_readme": [
-        "Copyright Siemens AG 2014",
+        "Copyright Siemens AG 2014-2016",
         "",
         "Copying and distribution of this file, with or without modification,",
         "are permitted in any medium without royalty provided the copyright",
@@ -21,7 +23,8 @@
         "mockery/mockery": ">=0.9.1",
         "hamcrest/hamcrest-php": "1.1.*",
         "squizlabs/php_codesniffer": "1.*",
-        "sebastian/phpcpd": "2.*"
+        "sebastian/phpcpd": "2.*",
+        "satooshi/php-coveralls": "1.0.*"
     },
 
     "autoload": {
@@ -31,6 +34,12 @@
             "Fossology\\DeciderJob\\UI\\": "deciderjob/ui",
             "Fossology\\Decider\\UI\\": "decider/ui",
             "Fossology\\UI\\Page\\": "www/ui/page"
+        }
+    },
+    
+    "config": {
+        "platform": {
+            "php": "5.3.3"
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Travis-CI Build Status](https://travis-ci.org/fossology/fossology.png)](https://travis-ci.org/fossology/fossology/)
 [![Stories in Ready](https://badge.waffle.io/fossology/fossology.svg?label=ready&title=Ready)](http://waffle.io/fossology/fossology)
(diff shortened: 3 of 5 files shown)


maximilianhuber:
I don't think that `src/composer.lock` should be added.

I have restarted the travis tests, since some of them died in the `apt-get`-phase

src/composer.json
```diff
@@ -32,5 +35,11 @@
             "Fossology\\Decider\\UI\\": "decider/ui",
             "Fossology\\UI\\Page\\": "www/ui/page"
         }
+    },
+    
+    "config": {
+        "platform": {
+            "php": "5.3.3"
```
maximilianhuber: Why do you have added this part?

.travis.yml
```diff
@@ -86,3 +86,8 @@ script:
  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.5" ]]; then composer require --dev --no-update phpunit/phpunit ~4; fi
  - phpunit -csrc/phpunit.xml --testsuite="$PHPTESTSUITE"
  - set +e
+
+after_script:
+ - mkdir build
+ - mkdir build/logs
```
maximilianhuber: maybe replace with `mkdir -p build/logs`?

maximilianhuber:
Code looks good to me

src/composer.json
steffen-weber: Without this configuration, composer update would install package versions which require php>=5.4
We should answer question https://github.com/fossology/fossology/issues/585 soon.

maximilianhuber: I agree with you and I would suggest to drop 5.3 now. We also have released 3.0.0 for 5.3 as suggested in the issue #585 and the next one will be with new features.

@mcjaeger: what is your opinion?

maximilianhuber: Let us postpone this question, I have created a note in #585.

maximilianhuber:
The merge lead to failing travis tests. Will try to fix them.

#### Note:
* All tests where green before the merge. After restarting them some become red.

* The tests fail with the following output (also see screenshot):
```
[...]
  - Installing symfony/console (v2.6.13)
    Downloading: Connecting...    Failed to download symfony/console from dist: Could not authenticate against github.com
    Now trying to download from source
  - Installing symfony/console (v2.6.13)
    Cloning 0e5e18ae09d3f5c06367759be940e9ed3f568359

  - Installing symfony/config (v2.6.13)
    Downloading: Connecting...    Failed to download symfony/config from dist: Could not authenticate against github.com
    Now trying to download from source
  - Installing symfony/config (v2.6.13)
    Cloning b5063937fab6fdfb7bacc00bc8c0cd7ee0c50070
    b5063937fab6fdfb7bacc00bc8c0cd7ee0c50070 is gone (history was rewritten?)


  [RuntimeException]                                                           
  Failed to execute git checkout 'b5063937fab6fdfb7bacc00bc8c0cd7ee0c50070' -  
  - && git reset --hard 'b5063937fab6fdfb7bacc00bc8c0cd7ee0c50070' --          
  fatal: reference is not a tree: b5063937fab6fdfb7bacc00bc8c0cd7ee0c50070                                                                                 
```
although the referenced commit [symfony/config/b5063937fab6fdfb7bacc00bc8c0cd7ee0c50070](https://github.com/symfony/config/commit/b5063937fab6fdfb7bacc00bc8c0cd7ee0c50070) exists.

* Other tests, which only differ in the version of *g++* do not fail

--
tmp

maximilianhuber:
Sometimes it works:
tmp

I have restarted the travis jobs since one of them worked now

maximilianhuber:
The failing test is is [#647](https://travis-ci.org/fossology/fossology/builds/116097694).

The  failing jobs (even after multiple restarts):
* [#647.3](https://travis-ci.org/fossology/fossology/jobs/116097697)
* [#647.12](https://travis-ci.org/fossology/fossology/jobs/116097708)
* [#647.15](https://travis-ci.org/fossology/fossology/jobs/116097711)

The job [#647.5](https://travis-ci.org/fossology/fossology/jobs/116097699) has failed previously but has recovered after restarting it.

maximilianhuber:
@steffen-weber: also other travis tests are failing, which not yet include the changes introduced in this PR. For example [this test](https://travis-ci.org/siemens/fossology/builds/116130804) of a branch on a fork.

I seams that this is a general infrastructure problem.

steffen-weber:
@maximilianhuber  It could be caused by a github limit:
https://getcomposer.org/doc/articles/troubleshooting.md#api-rate-limit-and-oauth-tokens
With `composer install --prefer-source` as in [this commit](https://github.com/steffen-weber/fossology/commit/1cdf0805c869eb8adc5c941a90475ccb30aa811c), the test set-up was slower (by factor 4) but 
after failing to execute git checkout travis successfully downloaded from dist.

maximilianhuber:
Am I right that the parameter `--prefer-source` only changes the fallback order from
```
    dist > source
```
to 
```
    source > dist
```
and that this makes it slow enough that it don't reaches the api-rate-limit?

Your commit does not look like a clean solution.

steffen-weber:
It might be a dirty solution and I'm not even sure if it will fix the trouble. It was just an idea for further investigations.

maximilianhuber:
> although the referenced commit symfony/config/b5063937fab6fdfb7bacc00bc8c0cd7ee0c50070 exists.

They have rebased their history. The correct hash for this commit is now [0ca496cbe208fc37c4cf3415ebb3056e0963115b](https://github.com/symfony/config/commit/0ca496cbe208fc37c4cf3415ebb3056e0963115b). We should change that in the `composer.lock` file.

BUT: this does not solve the problem, that composer falls back to source download.

This does not only occur in travis, also an local machines which live behind a proxy the `composer install` step has the same problem.

PR #658, opened 2016-03-17
fix(ui): only admin should be able to create groups

### LINKED ISSUE #656: Verify user before creating group
Hello, 

Currently, any user can add/create group 
https://<fossology.url>/?mod=group_add

This action needs to be changed. User with admin rights needs to be able to create a group. Please correct me if I am wrong or if this is a feature not security issue

maximilianhuber:
Has to be discussed, whether this change is necessary
closes #656


maximilianhuber:
A user is also able to manage and delete groups to which he has the correct permissions. This is valid behaviour and does not introduce permission issues.

PR #665, opened 2016-03-31
fix(security) cp2foss SQL injection vulnerability

timurphy:
Two SQL queries in the cp2foss utility insert variables into the SQL string, which creates an SQL injection vulnerability. This PR converts those queries into prepared statements using DbManager. SELECT * has also been replaced with SELECT <fields> for efficiency.

### DIFF
--- a/src/cli/cp2foss.php
+++ b/src/cli/cp2foss.php
@@ -156,7 +156,7 @@ function GetBucketFolder($UploadName, $BucketGroupSize) {
  * This is recursive!
  */
 function GetFolder($FolderPath, $Parent = NULL) {
-  global $PG_CONN;
+  $dbManager = $GLOBALS['container']->get('db.manager');
   global $Verbose;
   global $Test;
   if (empty($Parent)) {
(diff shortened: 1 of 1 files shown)


src/cli/cp2foss.php
```diff
@@ -175,19 +175,16 @@ function GetFolder($FolderPath, $Parent = NULL) {
   }
   /* See if it exists */
   $SQLFolder = str_replace("'", "''", $folderHead);
```
maximilianhuber: Using prepared statements, you do not longer have to have this escaping here

timurphy: Thanks. I'll run more tests for this case before committing the change.

maximilianhuber:
Code looks good to me, I will test the functionallity soon

maximilianhuber:
I have found some problems but I am not yet shure, whether they are related to your PR. I will do some further checks.

timurphy:
Removed the `str_replace` and ran some extra tests containing `'` and `"` characters.

PR #666, opened 2016-03-31
fix(SPDX2): use implodedLicenses in tagValue format

maximilianhuber:
The `OR`s and `AND`s in tag:value were flipped

### DIFF
--- a/src/spdx2/agent/spdx2.php
+++ b/src/spdx2/agent/spdx2.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2015, Siemens AG
+ * Copyright (C) 2015-2016, Siemens AG
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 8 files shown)


src/spdx2/agent/spdx2.php
```diff
@@ -198,24 +198,24 @@ protected function getUri($fileBase,$packageName)
...
    */
-  protected function implodeLicenses($licenses)
+  protected function implodeLicenses($licenses, $prefix="")
   {
     sort($licenses);
     if(count($licenses) == 3 &&
        ($index = array_search("Dual-license",$licenses)) !== false)
     {
-      return $licenses[$index===0?1:0] . " or " . $licenses[$index===1?2:1];
+      return $prefix . $licenses[$index===0?1:0] . " OR " . $prefix . $licenses[$index===1?2:1];
```
steffen-weber: there should be parentheses around this result, shouldn't it?

maximilianhuber: I do not think that they are necessary since they will be added on demand in line https://github.com/fossology/fossology/pull/666/files#diff-7bba19f12bab5e8b7e4b602c33f2e955R212

steffen-weber: I think code will not go to line 212 after return.
Please provide unit tests.

maximilianhuber: In the case that the licenses are "BSD", "GPL" and "Dual-License" the returned string is
```
"BSD OR GPL"
```
without parentheses.

In the other case that the licenses are "Zlib" and "BSD OR GPL" the returned string will be
```
"Zlib AND (BSD OR GPL)"
```
with the added parantheses from line R212

That are the two supported cases.

(I will add unit tests)

maximilianhuber:
@steffen-weber: I have added unit tests

PR #671, opened 2016-04-15
In License Browser, the menu order getting changed if the user clicks on menus like ECC, copyright

shaheemazmalmmd:


maximilianhuber:
Without your PR the menu (with debug information) looks as follows:
tmp
and
tmp

With your PR the most entries will have the value `1` and they will be thus sorted alphabetically.

I would suggest to set these values to be always `10` instead of always `1` and mix them with other entries.

PR #680, opened 2016-04-28
feat(conf): added header/copyright information

shaheemazmalmmd:


maximilianhuber:
If @mcjaeger is okay with the fact, that this PR adds `2016` as copyright date to every file, then I would say, that this PR could be merged.

mcjaeger:
propose to have the copyright line with Siemens removed, but keeping the "as-is"-permission header is a good thing.

shaheemazmalmmd:
Done.

PR #684, opened 2016-06-13
fix(ui): Added recent agent_pk in the place of any agent_pk

shaheemazmalmmd:
1) Added both file counts in ui to indicate users
2) Show all files was not working for special chars in url(like +)


ag4ums:
tested and reviewed by @ag4ums , 
please merged.

mcjaeger:
it was about the differences between file count on the main page and the file count of the license-specific file listing (showed when clicking on the count number of the histogram row)

PR #685, opened 2016-06-16
Updates to legacy runBuild for github, FO 3.0+

dgstangel:
These are the latest customizations to the runBuild script as it was used
in the legacy HP FOSSology package build environment.  This includes
updates to the source code in Github, as well as defaulting to building
3.0.

This still isn't perfect.  It requires that there be a /home/build account on the target
system, and that a /home/build/.pbrc file be set up with sane default values to
run the FO build.  And even having done all this it still has some errors I 
have not been able to track down.

Finally, you still need a set of QEMU VMs to use in building the distro packages.
Bruno Cornec has offered us several newer ones already prepared, but
have not yet received them.

### DIFF
--- a/utils/runBuild_v2.0.php
+++ b/utils/runBuild_v2.0.php
@@ -20,10 +20,10 @@
 /**************************************************************
  runBuild v2.0
 
- Script to create packages use Project-Builder.
+ Script to create packages using Project-Builder (http://project-builder.org)
 
  --------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


mcjaeger:
hi, in case it is ready for review, please use the according label.

dgstangel:
I believe these changes are no longer needed as Linux Foundation migrates the FO build to its native infrastructure. (These refer specifically to an obsolete, in house HP build environment).

Issue #515, opened 2015-09-03
OAuth2 support


bufferoverflow:
Would be great to have OAuth2 support, this would enable us to
- use standard authentication systems on internal deployments
- use a Mozilla or GitHub account to login on public FOSSology instance

This is related to  #193

bobgob:
So what is involved?  common-auth.php (siteminder_check()) simply validates that a specific environment variable is set and returns its value (the fossology user id).  This is the technique used for Siteminder and PingIdentity.

bufferoverflow:
Here is a nice library http://oauth2-client.thephpleague.com/ that supports OAuth2 for PHP.

bufferoverflow:
Here is some sample code to achieve this:

``composer.json`` should contain this:
```json
{
    "require": {
        "league/oauth2-client": "^1.3"
    }
}
```
The example is using [gitlab as a oauth service provider](http://doc.gitlab.com/ce/integration/oauth_provider.html). To test this, do
* register the application [here](https://gitlab.com/profile/applications) 
* add the ApplicationId- and Secret-hash to the ``index.php`` file below
* run the sample by executing: ``php -S localhost:8000``
* point your browser to http://localhost:8000

```php
<?php
require __DIR__ . '/vendor/autoload.php';
$provider = new \League\OAuth2\Client\Provider\GenericProvider([
    'clientId'                => 'hash',
    'clientSecret'            => 'hash',
    'redirectUri'             => 'http://localhost:8000',
    'urlAuthorize'            => 'https://gitlab.com/oauth/authorize',
    'urlAccessToken'          => 'https://gitlab.com/oauth/token',
    'urlResourceOwnerDetails' => 'https://gitlab.com/api/v3/user'
]);

if (!isset($_GET['code'])) {
    $authorizationUrl = $provider->getAuthorizationUrl();
    $_SESSION['oauth2state'] = $provider->getState();
    header('Location: ' . $authorizationUrl. "&scope=api&response_type=code");
    exit;

} else {
    try {
        $accessToken = $provider->getAccessToken('authorization_code', [
            'code' => $_GET['code']
        ]);
        echo "<pre>";
        echo $accessToken->getToken() . "\n";
        echo $accessToken->getRefreshToken() . "\n";

        $resourceOwner = $provider->getResourceOwner($accessToken);

        var_export($resourceOwner->toArray());

        echo "\n";
        echo "name  " . $resourceOwner->toArray()['name']  . "\n";
[...]

PR #689, opened 2016-08-25
Implementation of gitlab-oauth2 service for fossology

shaheemazmalmmd:
closses #515

### DIFF
--- a/src/www/ui/core-auth.php
+++ b/src/www/ui/core-auth.php
@@ -1,7 +1,7 @@
 <?php
 /***********************************************************
  * Copyright (C) 2008-2013 Hewlett-Packard Development Company, L.P.
- * Copyright (C) 2015 Siemens AG
+ * Copyright (C) 2015-2016 Siemens AG
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
@@ -22,6 +22,7 @@ use Fossology\Lib\Dao\UserDao;
 use Fossology\Lib\Db\DbManager;
 use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\HttpFoundation\RedirectResponse;
+use Omines\OAuth2\Client\Provider\Gitlab;
 
 define("TITLE_core_auth", _("Login"));
 
@@ -105,7 +106,9 @@ class core_auth extends FO_Plugin
     if (!empty($_SESSION['time']))
     {
       /* Logins older than 60 secs/min * 480 min = 8 hr are auto-logout */
-      if (@$_SESSION['time'] + (60 * 480) < $Now) $this->updateSession("");
+      if (@$_SESSION['time'] + (60 * 480) < $Now){
+          $this->updateSession("", true);
+      }
     }
 
     $_SESSION['time'] = $Now;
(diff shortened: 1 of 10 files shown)


maximilianhuber:
With this changes most tests fail with `Segmentation fault`, see screenshot:

tmp

Additionally, all travis jobs were restarted, since at least one test failed due to a race condition.

shaheemazmalmmd:
Error_(PHP Parse error:  syntax error, unexpected '[' in /home/travis/build/fossology/fossology/src/vendor/guzzlehttp/psr7/src/functions.php on line 78)_ seems to be with the package guzzlehttp. which is getting downloaded internally by composer. there is some issue created here as well https://github.com/guzzle/guzzle/issues/644
and 
Only PHP 5.5 (>=5.5.9) or greater version will support Oauth2. May be we should remove tests in travis which are executed for php less than 5.5.9 ??.

maximilianhuber:
I think, we first have to finish the discussion in #585 before we drop testing of some php versions.

src/www/ui/core-auth.php
```diff
@@ -214,8 +242,13 @@ public function Output()
         $referrer = Traceback_uri();
       }
     }
+    if(!empty($getEmail) && empty($userName)){
+      $oAuthClient = true; // To test in the server
```
maximilianhuber: This might be a line, which should be removed.

src/www/ui/core-auth.php
```diff
@@ -266,15 +305,15 @@ function OutputOpen()
    *
    * @return boolean
    */
-  function checkUsernameAndPassword($userName, $password)
+  function checkUsernameAndPassword($userName, $password, $oAuthClient=false)
```
maximilianhuber: A function called `checkUsernameAndPassword` should not modify the global state e.g. `$_SESSION`

src/www/ui/core-auth.php
```diff
@@ -202,6 +208,28 @@ public function Output()
     $userName = GetParm("username", PARM_TEXT);
     $password = GetParm("password", PARM_TEXT);
     $referrer = GetParm("HTTP_REFERER", PARM_TEXT);
+    $oAuthClient = GetParm("oAuthClient", PARM_TEXT);
+    $getEmail = GetParm("HTTP_SCMAIL", PARM_TEXT);
+    $providerCheck = GetParm("providerCheck", PARM_TEXT);
+
+    if(!empty($providerCheck) && empty($getEmail)){
```
maximilianhuber: This could be moved to another function

src/www/ui/core-auth.php
```diff
@@ -202,6 +208,28 @@ public function Output()
...
+      $provider = new Gitlab([
+        "clientId"                => $SysConf['SYSCONFIG']['GitlabAppIdOauth'],
+        "clientSecret"            => $SysConf['SYSCONFIG']['GitlabSecretOauth'],
+        "redirectUri"             => $SysConf['SYSCONFIG']['RedirectOauthURL'],
+        "domain"                  => $domainOauth
+      ]);
+      $authorizationUrl = $provider->getAuthorizationUrl();
+      $_SESSION['oauth2state'] = $provider->getState();
+      header('Location: ' . $authorizationUrl);
+      exit;
```
maximilianhuber: Why do you exit here in such a harsh way?

bufferoverflow:
@shaheemazmalmmd please fix the issues mentioned by @maximilianhuber and rebase so this can get merged

mcjaeger:
we will do a relaunch here with a new PR

Issue #692, opened 2016-09-01
avoid downloading composer.phar in build process


bcornec:
While trying to package fossology there is the need to download composer.phar before using it.

I'd recommend to make it a dependency (like boost e.g.), potentially providing a script or using fo-install-deps to proved a version of it.

Having it in the git repository already downloaded would allow for off-Internet builds to work without issue (you cannot build as of now without an Internet connection, nor without correct proxy settings e.g.).

mcjaeger:
related to #688

maybe composer.phar itself / only is downloaded already.

https://github.com/fossology/fossology/blob/d26bccd82a4566b8f6bfaac137db169e4a8835dd/utils/fo-installdeps#L266

but I think the small detail is when is downloaded what: dependencies?

bcornec:
IMHO, there shouldn't be a call to curl in a build process. That put at risk the reproduceability  (no version is mentionned so it would be better to have a stable version) and the build itself as without an Internet access the build fails. And It seems from my tests that this is the only place where we have that.

Managing it as a separate package could be a way forward.

maximilianhuber:
Might be related to #696, although this PR is not concerned with the download of `composer.phar` but with the call to it, which might want to download dependencies from the internet.

maximilianhuber:
@bcornec: What excactly do you mean with 
> Managing it as a separate package could be a way forward.

bcornec:
The FOSSology project could provide a fossology-composer, which deliver in a known place the composer.phar binary that then would be used in the build process of the fossology package.

That way we have a stable version of composer.phar, identified, and delivered in a packaged format, and built separately from FOSSology (Of course, that package has the same requirement of not being Internet dependent, so I think we could just download on a regular basis, for each major version, a new version of that component and rebuild the package with it). Of course, we also need to provide it as a source format, but that's part of the package build anyway so will be covered.

And FOSSology could just have a build require on it. As we do for PHP, tar, ...

If you agree with the approach, I can propose a PR for such a behaviour.

pguttmann:
I like your idea of fossology-composer package updated as part of the package build process.

maximilianhuber:
A stable version of composer.phar could also be obtained by using the url https://getcomposer.org/download/1.2.1/composer.phar. The downloaded composer.phar then also could be checked for the hash to be matching.

Would this improve the Situation?

bcornec:
[...]

PR #696, opened 2016-09-12
(Work in progress) Refactor Makefiles to seperate the composer call

maximilianhuber:
**OUTDATED:**

> in `src/Makefile` the call to `install` is replaced by
> 
> ```
> install: composer_install deploy
> ```
> 
> this allows the user to do the composer related stuff, which wants to talk to
> the internet, before the other install related commands.
> 
> The two commands are:
> 
> ```
> cd src && make composer_install
> ```
> 
> and
> 
> ```
> make deploy
> ```
> 
> This might be useful for packaging.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -20,6 +20,7 @@ fi
 PROXYSCRIPT
 
 $build_and_test = <<SCRIPT
+set -ex
 export DEBIAN_FRONTEND=noninteractive
 
 echo "Provisioning system to compile, test and develop."
@@ -29,9 +30,6 @@ sudo dpkg-reconfigure locales
 sudo apt-get update -qq -y
 
 sudo apt-get install -qq curl php5 git libspreadsheet-writeexcel-perl libdbd-sqlite3-perl
-
-curl -sS https://getcomposer.org/installer | php
-sudo mv composer.phar /usr/bin/composer
 
 # install spdx-tools
 /vagrant/install/scripts/install-spdx-tools.sh
@@ -45,6 +43,9 @@ echo "lets go!"
 cd /vagrant
 
 ./utils/fo-installdeps -e -y
+curl -sS https://getcomposer.org/installer | php
+sudo mv composer.phar /usr/bin/composer
+make phpvendors
 make CFLAGS=-I/usr/include/glib-2.0
 sudo make install
 sudo /usr/local/lib/fossology/fo-postinstall
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,9 @@ $(CLEANDIRS):
 phpvendors:
 	$(MAKE) -C $(FOSRCDIR) phpvendors
 
+phpvendors-no-dev:
+	$(MAKE) -C $(FOSRCDIR) phpvendors-no-dev
+
 # release stuff
 tar: dist-testing
 dist-testing:
--- a/src/Makefile
+++ b/src/Makefile
@@ -77,10 +77,11 @@ $(CLEANDIRS):
 phpvendors:
 	composer install -q
 
+phpvendors-no-dev:
+	composer install -q --no-dev
+
 composer_install:
-	@echo "current dir: '$${PWD}'"
-	$(INSTALL_DATA) composer.json composer.lock $(DESTDIR)$(MODDIR)
-	(cd $(DESTDIR)$(MODDIR); composer install -q --no-dev)
+	cp -r vendor $(DESTDIR)$(MODDIR)
 
 .PHONY: subdirs $(BUILDDIRS)
 .PHONY: subdirs $(DIRS)
--- a/utils/fo-mktar
+++ b/utils/fo-mktar
@@ -1,39 +1,48 @@
-#!/bin/sh
+#!/bin/bash
 # This script packages the directory into a tar file.
 # Copyright (C) 2014 Hewlett-Packard Development Company, L.P.
-# 
+#
 #  This program is free software; you can redistribute it and/or
 #  modify it under the terms of the GNU General Public License
 #  version 2 as published by the Free Software Foundation.
-#  
+#
 #  This program is distributed in the hope that it will be useful,
 #  but WITHOUT ANY WARRANTY; without even the implied warranty of
 #  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 #  GNU General Public License for more details.
-#  
+#
 #  You should have received a copy of the GNU General Public License along
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
+set -e
+
 # make sure we're in a checked out git copy
 if [ ! -d .git ]; then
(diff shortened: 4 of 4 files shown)


maximilianhuber:
This removes the calls to `wget` and `curl` while build time, i.e.
```
# Things that need internet connection
./utils/fo-installdeps -e -y
make phpvendors
cd src
make composer_install
cd ..

# the following lines should not be executed on your system
sudo rm $(which wget) # dangerous!
sudo rm $(which curl) # dangerous!

# Things that do not need internet connection
make CFLAGS=-I/usr/include/glib-2.0
sudo make deploy
```
(Tested in the vagrant environment)

bcornec:
I don't think this will be sufficient. 

Let me recap the goal I think we should have on this:

1/ A build of fossology should be fully standalone once the tar ball is downloaded and all build requirements satisfied
2/ everything which is thus needed to realize 1/ should be performed *before* and provided in the tar ball.

So if the goal is to call make composer_install as a bootstrap command, then create the tar ball with the result, and ship that so people can build SW then yes, that's good. Now this is really creating a separate step which should be invoked by the source maintainers, not the apps builders. 

And I think documentation would need to be updated as well to describe these steps. Or I may have missed something :-)

maximilianhuber:
For point 1/

Maybe modify the makefiles such that one can use:
```
make prefetch
```
to get all dependencies, and then has the options to run
```
make tar
```
to get the tar ball or
```
make
sudo make install
```
to install and develop FOSSology.

Would this be your desired behaviour for the Makefiles?

maximilianhuber:
Currently `make vendors` has the behavior:
- it downloads all dependencies into the `./src/vendosr/` folder which could be added to the tar ball

The [call to composer in the Makefile](https://github.com/fossology/fossology/blob/c3929070ea1c555f0b2137ffcad2e1aea4a84f49/src/Makefile#L80-L83) could then be replaced by
```
composer_install:
    $(INSTALL_DATA) vendors $(DESTDIR)$(MODDIR)
```
which would copy the inflated PHP dependencies to the install dir.

**But:** this does not allow to difference between dev- and runtime-dependencies, i.e. would also install `phpunit`, `mockery`, ... which are only used for testing.

This raises the following question: **should the tar ball also contain all dependencies for testing?**
(this would also include the `spdx-tools.jar` and some testfiles, which are large)

--

If the answer is **no**: The solution is easy, just add a `make phpvendors-testing` to download also testing dependencies and modify `make phpvendors` to use the `--no-dev` flag.

If the answer is **yes**: This could be solved in multiple ways, e.g.

1. simply install also the dev dependencies (makes artifacts larger)
2. have the `./src/vendors` and `./src/vendors-no-dev` folders in the source tree (makes the tar ball bigger)

maximilianhuber:
The spdx-tools are downloaded in line [`./src/spdx2/agent_tests/Functional/schedulerTest.php#L170`](https://github.com/fossology/fossology/blob/c3929070ea1c555f0b2137ffcad2e1aea4a84f49/src/spdx2/agent_tests/Functional/schedulerTest.php#L170).

maximilianhuber:
@bcornec I have pushed changes which remove the previous commit and

- [use the local php dependencies in `./src/vendors` for `make install`](https://github.com/fossology/fossology/pull/696/files#diff-f57f2991a6aa25fe45d8036c51bf8b4dR84) and
- modify `utils/fo-mktar` (called by `make tar`) to [also include the runtime php dependencies](https://github.com/fossology/fossology/pull/696/files#diff-cd8a899e4c4a98415d6293e5c1957f74R80) (and also do some cleanup in the `fo-mktar` script).

--

These changes imply that, for installing FOSSology from source, one has now to call
<pre>
<b>make phpvendors-no-dev</b>
make
sudo make install
</pre>

and for testing
<pre>
<b>make phpvendors</b>
make test
</pre>
(which was also previously required)

(README and webpage then has to be updated)

--

one alternative would be to check in the folder `./src/vendor/`, but i do not think that this is desirable.

maximilianhuber:
These changes do not yet attack the problem with the `composer.phar` file, for potential solutions see https://github.com/fossology/fossology/issues/692#issuecomment-247427858

maximilianhuber:
Adding the runtime php dependencies to the tar ball changes its size only from `11948 B` to `12568 B`. That should not be a problem.

maximilianhuber:
An alternative would be to still allow `make install` to use composer to download the dependencies and modify its behaviour by a flag, i.e.
<pre>
make phpvendors-no-dev
make
<b>OFFLINE=true</b> sudo make install
</pre>

this would developers still allow to use
<pre>
git clone ...
cd fossology
make
sudo make install
</pre>

Vagrantfile
```diff
@@ -45,6 +43,9 @@ echo "lets go!"
 cd /vagrant
 
 ./utils/fo-installdeps -e -y
+curl -sS https://getcomposer.org/installer | php
+sudo mv composer.phar /usr/bin/composer
```
maximilianhuber: > Now I understand this is a Vagrant file, not your Makefile anymore, so maybe this is much more valid here.

Yes, the vagrantfile should not be part of the discussion here.

bcornec:
Humm, lots of questions ;-)

Ok, wrt test: if we want to be able to run them as part of the package build (rpm does support that), then we need the test content in the tar ball.
wrt having the src/vendors* dir in the git tree I think that what I was advocating for !
wrt SPDX tools, is it at run time or at build time ? If the former, then it has no impact on packaging, which is not the case with the latter.
wrt split of composer, I've commented what I've just tried to do on #692 

I think we are converging which is great ! 
Honestly the way you want to manage the life cylce of the composer.phar app is not relevant for me as packaging activity. So do it the way which is the most convenient for you. I'll adapt the build after.

Now having just a simple script doing it could be sufficient. called from make or not, is not so important.

maximilianhuber:
I am not yet sure whether it is a good idea to have `./src/vendors/` in the git tree. This would imply that we need a clean way to upgrade all dependencies. I could imagine that this can become ugly, e.g. if one decides to "fix" something by modifying the source code at `./src/vendors/`.

Further i am not sure whether the dependencies under `./src/vendors/` depend on the used PHP version, although [the PHP version is pinned in `composer.json`](https://github.com/fossology/fossology/blob/master/src/composer.json#L42).

--

SPDX tools are only needed for testing, it us used for verification of the generated SPDX files.

maximilianhuber:
This PR should be tested to work well with the package building by project-builder introduced in #701

bcornec:
Making dependencies explicit or implicit is a good question to ask for a project. 

You can either rely on what the distro provides (and you don't control it yourself), or rely on what you provide (but you need to deal with updates.

In that case, I'm just advocating for the composer tool to be included instead of downloaded wildly during build as you never know in that case which version you'll get. If it would be provided by a distro, at least you'd have some stability. Here not. So my recommendation is just to download it once, to be sure to have the version you want, and then from time to time, you may update it if/when needed.

mcjaeger:
it would require some updated documentation and it seems the package build could go without at the moment (because composer is pulled directly using wget), let's move it to after 3.1.

maximilianhuber:
This is outdated. Will be reopened once necessary.

Issue #695, opened 2016-09-12
Run Docker container as unprivileged user


bufferoverflow:
The Docker container does not run as fossology user today, which is best practice, see https://docs.docker.com/engine/reference/builder/#user

maximilianhuber:
In fact, the scheduler runs as user fossy which is created by the call [/docker-entrypoint.sh#L58](https://github.com/fossology/fossology/blob/c3929070ea1c555f0b2137ffcad2e1aea4a84f49/docker-entrypoint.sh#L58). This is handled by the service.

But apache is running as root.

In fact I am currently refactoring the compose setup to split the apache into its own container. Further is the repository currently not persistent in the container.

--

My suggested changes, which fix some of the above mentioned problems and do not rely on services running within containers, are found at in my repository as branch [dev/refactorDockerSupport](https://github.com/maximilianhuber/fossology/tree/dev/refactorDockerSupport). (There is still apache running as root)

These changes
- split the container into multiple ones:
  - one for the scheduler
  - one for the apache
  - one for the SW360 connection (not implemented yet) 
- (currently) break the support for running docker without compose.

koa:
With some few modifications the apache can be started as user fossy

- chown to fossy of 
  - /var/run/apache2/
  - /var/lock/apache2/
  - /var/log/fossology
- use a non priviledged port (8080) for apache

But there is another problem: the fo-postinstall.sh (needed for initialize the database) need root to start.
Solution: 
- Skip root validation if there is only database initialitation
- do the rest of initialitations while building docker image (as root)

bufferoverflow:
@maximilianhuber any update on this? Enable the Docker image able to run as unprivileged user before splitting into several containers makes sense.

maximilianhuber:
@bufferoverflow: Sorry, but the PR #697 has some problems which could not yet been resolved.

Maybe one could extract the necessary changes as you have suggested. But the splitting of the containers is a necessary step, to enable connection to SW360. It also improves the logging (the scheduler log is captured by docker-logs) and makes it more modular.

maximilianhuber:
the PR #697 also makes the fossology repository a volume, which allows to use the docker instance productively: not all uploaded files get removed after restart.

bufferoverflow:
:+1: for the repository volume

max-wittig:
should be resolved with the merge of PR #758

PR #697, opened 2016-09-12
Refactor docker support

maximilianhuber:
This PR was already announced in a comment on #695.

It is not yet ready and needs further improvements and intense testing.

This PR
- splits the container into multiple ones:
  - one for the scheduler
  - one for the apache
    - which runs as user fossy (as suggested in https://github.com/fossology/fossology/issues/695#issuecomment-246330229)
  - one for the SW360 connection (not implemented yet) 
- moves calls like `/usr/local/lib/fossology/fo-postinstall --common` from entrypoint to Dockerfile
- reorders Dockerfile to prevent rebuilding all dependencies on every changes of the source code
- (currently) breaks the support for running docker without compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ use, since the the standalone image does not take care of data persistency.
 
 A pre-built Docker image is available from [Docker Hub](https://hub.docker.com/r/fossology/fossology/) and can be run using following command:
 ``` sh
-docker run -p 8081:80 fossology/fossology
+docker run -p 8081:8081 fossology/fossology
 ```
 
 Execution with external database container can be done using Docker Compose.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # FOSSology Dockerfile
 # Copyright Siemens AG 2016, fabio.huser@siemens.com
+# Copyright TNG Technology Consulting GmbH 2016, maximilian.huber@tngtech.com
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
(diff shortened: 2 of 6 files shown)


maximilianhuber:
Yes, your solution looks good. Especially moving parts like 
```
RUN mkdir -p /var/log/apache2/
RUN chown -R fossy:fossy /var/log/apache2/

RUN mkdir -p /var/log/fossology
RUN chown fossy:fossy /var/log/fossology
RUN chown fossy:fossy /var/run/apache2/
RUN chown fossy:fossy /var/lock/apache2/
RUN chgrp fossy /usr/local/etc/fossology/Db.conf
￼RUN chown fossy:fossy /usr/local/etc/fossology/Db.conf
```
to the Dockerfile is good. I will adopt some of your changes to this PR.

maximilianhuber:
@koa: I do not understand the following lines in your Dockerfile:
```
RUN sed 's/:80/:8080/' </etc/apache2/sites-enabled/000-default.conf >/tmp/out.conf
RUN mv /etc/apache2/sites-enabled/000-default.conf /tmp/out.conf
```
Both commands only write to `/tmp/out.conf` which is never used? Do you intended to move the file in the other direction? this would be eqivalent to:
```
RUN sed -i 's/:80/:8080/' /etc/apache2/sites-enabled/000-default.conf
```

koa:
Ups this was a error.
Here is the fix: https://github.com/koa/fossology/commit/09f810e0e44e60d1fa124bf856a3e9a86ae0b3c5

maximilianhuber:
Why do you need these change? it works for me without the one modified in koa@09f810e.

koa:
This 2 lines makes something completely different than the developer expects.
If you want to keep the current behavior, take this:
`RUN rm /etc/apache2/sites-enabled/000-default.conf`

maximilianhuber:
I mean: I do not need to touch `000-default.conf` at all. (see my updated branch)

koa:
It is important to have the line 
`USER fossy`
within the Dockerfile.

This tells openshift, that this container works without root priviledges.

maximilianhuber:
Yes. I am working towards this, but I also want do not want to do the `fo-postinstall --web-only` part in the common Dockerfile and vice versa.

The solution might be to use a common base Dockerfile (which makes FOSSology) and split the custom part into separate Dockerfiles (for scheduler, apache, sw360connection i.e. SSH).

maximilianhuber:
I have pushed https://github.com/fossology/fossology/pull/697/commits/7bfb9ac5a686fe6e0fd322db72f77d199eceacab which contains a (currently) not working scetch of a split-up Dockerfile.

maximilianhuber:
The script creating the db is [`install/db/dbcreate`](https://github.com/fossology/fossology/blob/c3929070ea1c555f0b2137ffcad2e1aea4a84f49/install/db/dbcreate.in) which internally calls the `psql` command with `su postgres -c "psql ....."` which does not work with an unprivileged user.

Thus `fo-postinstall --databases` (resp. `dbcreate`) currently has to be called by root and can not be run in an unprivileged entrypoint script.

Maybe one has to add something like `sudo`. But I am not sure whether that adds back the privileges one wants to drop.

maximilianhuber:
Further, the script mentioned above does not init the database at all, if the variable `$FOSSOLOGY_DB_HOST` is set:
```
[...]
if [ "$FOSSOLOGY_DB_HOST" ]; then
  echo "NOTE: using external DB on host $FOSSOLOGY_DB_HOST"
  exit 0
fi
[...]
```

Since this works for the "old" docker setup, where posgres is already on another host, there has to be some other part of the code where the DB is initialized.

maximilianhuber:
There is the infamous [`install/fossinit.php`](https://github.com/fossology/fossology/blob/c3929070ea1c555f0b2137ffcad2e1aea4a84f49/install/fossinit.php) which is called by `fo-postinstall` in the lines 
- 243 if `$LICENSEREF=1` i.e. if
  - `--licenseref` or `--everything` is passed or
  - it is also implied by `--web-only`
- 296 (not reacheable branch)

maximilianhuber:
Ah, the database is automatically created by the starting postgres container and the name of the DB if defined in the line  https://github.com/fossology/fossology/pull/697/files#diff-9c7fd30be857581ab5813b48a107a256R73.

Thus, the `fossology`-db should exist. I will check.

README.md
```diff
@@ -26,16 +26,9 @@ For installation instructions see http://www.fossology.org/projects/fossology/wi
...
 FOSSology comes with a Dockerfile allowing the containerized execution
-both as single instance or in combination with an external PostgreSQL database.
-**Note:** It is strongly recommended to use an external database for production
-use, since the the standalone image does not take care of data persistency.
+in combination with an external PostgreSQL database.
 
-A pre-built Docker image is available from [Docker Hub](https://hub.docker.com/r/fossology/fossology/) and can be run using following command:
-``` sh
-docker run -p 8081:80 fossology/fossology
-```
```
bufferoverflow: We should really keep that, it's great for newbies to play with FOSSology.

maximilianhuber:
@bufferoverflow the dockerfiles here are now completely rewritten and are now
- running as USER
- running scheduler and apache on different containers (when started with docker-compose)

and they do not remove the feature, that the container can run independently (one could add the postgres folder as volume, to make everything persistent).

bufferoverflow:
@maximilianhuber gerat! I just retriggered the build on travis, because they are failing...

maximilianhuber:
They are failing due to failing `git clone...`:
tmp

I do not think that the problem is on our side.

pguttmann:
I think line 113 may be the cause ;)
I see no branch dev/refactorDockerSupport in github

maximilianhuber:
It tries to clone from the wrong remote. The requested branch is in my fork at https://github.com/maximilianhuber/fossology/tree/dev/refactorDockerSupport

bufferoverflow:
It just fails with PHP 5.3 so we should definetly get rif of it, there is PHP 7 available and remove it also from travis.yml, see https://travis-ci.org/fossology/fossology/builds/178777805

bufferoverflow:
just applied your patch on top of #718 and it passed travis

the container build was successfully via `docker-compuse up` and `docker build .`
see https://github.com/bufferoverflow/fossology/commits/chore/docker-refactor

bufferoverflow:
please use `chore` instead of `chores` as per https://github.com/fossology/fossology/blob/master/CONTRIBUTING.md#git-commit-conventions

bufferoverflow:
I have rebased your PR and made a new one with your commit here #720

maximilianhuber:
closed in favor of #720

PR #700, opened 2016-09-17
fix(copyrightandeccview): added tooltip nect to description,Text Find…

### LINKED ISSUE #506: copyright report confusing fields
Tested on #25f4d65 
on mod=copyright-view there are three fields:
  Description
  Text finding
  Comment
What are these for?  Can we have a help mouse over to explain them?

screen shot 2015-08-26 at 11 59 02 am

shaheemazmalmmd:
…ing and comment.

closes #506

### DIFF
--- a/src/copyright/ui/template/ui-cp-view.html.twig
+++ b/src/copyright/ui/template/ui-cp-view.html.twig
@@ -43,6 +43,11 @@
 {% block foot %}
   {{ parent() }}
 
+  <script type="text/javascript">
+    $(document).ready(function () {
+      $('img').tooltip();
+    });
+  </script>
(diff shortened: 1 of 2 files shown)


maximilianhuber:
@mcjaeger: can you check that the added descriptions match your expectations?

The added descriptions (extracted from the source code) are:
- **Description:** Classification of the finding, such as copyright statement or TSU number or ECC warning or author but no copyright statement
- **Text finding:** Exact text finding, if the regular expression match is wrong not matching everything, too much
- **Comment:** A user can place comment for leaving a notice that ma be viewed by other users on Fossology

(code looks good but was not yet deployed)

PR #713, opened 2016-11-07
Contrib/del folder in list

ag4ums:
Unwanted delete of copied folders while listing from command line


ag4ums:
Closing this as its not for here

PR #701, opened 2016-09-21
pkg: first import of a pbconf tree

bcornec:
Hello,

In order to ease the build of packages, I've added a pbconf directory which contains all the files required by project-builder.org to build packages for as many distributions as you want.

Currently I have tested with mageia 5 (rpm), fedora 23 (rpm) and ubuntu 15.04 (deb). 

More work is of course needed, as mentioned in BR #692 and #696 to be able to use it correctly.

Goal is to setup a build machine using Docker containers on the LF infra so anybody in the project can create build environments, launch build of versions for any target distro supported, including doing that as a git hook so we can trap errors related to packaging as early as possible.

HTH,
Bruno.

PR #718, opened 2016-11-28
chore(php): remove 5.3, set 5.6, add 7.0 to travis-ci

### LINKED ISSUE #585: Supported PHP versions

bufferoverflow:
Closes #585

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@
 # build FOSSology on Travis CI - https://travis-ci.org/
 
 language: php
-
+php:
+  - '5.6'
 addons:
  - postgresql: "9.3"
 cache:
@@ -38,13 +39,13 @@ before_script:
  - install/scripts/install-spdx-tools.sh
  - psql -c "CREATE USER fossy WITH PASSWORD 'fossy' CREATEDB;" -U postgres
  - psql -c "create database fossology;" -U postgres
- 
+
 env:
-  global: 
+  global:
    - PHPTESTSUITE='Fossology PhpUnit Agent Test Suite'
    - CHECKBEFORE='cppcheck -q -isrc/nomos/agent_tests/testdata/NomosTestfiles/ -isrc/testing/dataFiles/ src/'
    - MAKETARGETS='all test-lib'
-  matrix: 
+  matrix:
    - CC=gcc CXX=g++ MAKETARGETS='all test-lib test-monk test-nomos'
    - CC=clang-3.5 CXX=clang++-3.5
    - CC=clang-3.6 CXX=clang++-3.6 MAKETARGETS='all test-lib test-monk test-nomos'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -61,23 +62,22 @@ env:
...
-        PHPTESTSUITE='Fossology PhpUnit Test Suite'
-        CHECKBEFORE='src/vendor/bin/phpcs --standard=src/fossy-ruleset.xml src/lib/php/*/ src/www/ui/page/ src/www/ui/async/ src/spdx2 src/monk'
-        MAKETARGETS='build-lib VERSIONFILE build-cli'
     - php: 5.4
       env: PHPTESTSUITE='Fossology PhpUnit Test Suite' CHECKBEFORE='' MAKETARGETS='build-lib VERSIONFILE build-cli'
     - php: 5.5
       env: PHPTESTSUITE='Fossology PhpUnit Test Suite' CHECKBEFORE='' MAKETARGETS='build-lib VERSIONFILE build-cli'
     - php: 5.6
+      env:
+        PHPTESTSUITE='Fossology PhpUnit Test Suite' CHECKBEFORE=''
```
maximilianhuber: I think the `CHECKBEFORE=''` is redundant

.travis.yml
```diff
@@ -61,23 +62,22 @@ env:
...
+        CHECKBEFORE='src/vendor/bin/phpcs --standard=src/fossy-ruleset.xml src/lib/php/*/ src/www/ui/page/ src/www/ui/async/ src/spdx2 src/monk'
+        MAKETARGETS='build-lib VERSIONFILE build-cli'
+    - php: 7.0
```
maximilianhuber: I am not able to find the corresponding line on https://travis-ci.org/fossology/fossology/builds/179358115, only as allowed failure (as defined in line 79). Is this the intended behavior?

Screenshot:
tmp

bufferoverflow: php 7 fails, that's why it is under allowed failures

.travis.yml
bufferoverflow: fixed within new commit

.travis.yml
maximilianhuber: Ok, the semantics of the travis.file confused me.

maximilianhuber:
Would it be an option to tag the commit 414a41c8a59749b0fbfc6e38b2272028a50bf727 as release 3.1?

src/composer.json
```diff
@@ -39,7 +39,7 @@
     
     "config": {
         "platform": {
-            "php": "5.3.3"
```
maximilianhuber: Would it be an option to tag the commit 414a41c as release 3.1?

Or we could revert this PR and wait for packaging to be working

mcjaeger: I am uncertain for myself. In turn I think the practise of organisations using ten year old distros just because their IT department does not support other version of Linux  should be revised too. And upgrading PHP is more urgent than upgrading other components to latest stable.

Also I remember correctly now, only @bcornec has complained on the mailing list about upping PHP to 5.6 (because we need this for the securing login with OAuth).

I think there is the point to use cent 7 and this is maybe not even in the epel 7? Then, it could conflict with this branch not merged, right?

https://github.com/fossology/fossology/pull/701

bcornec: Hello Michael, I don't really complain, I just wanted to underline the fact that doing so you loose support for some older versions. Now, it could just be a matter of stating on the project page, which Linux distributions are supported for FOSSology ! And I'd build packages accordingly as well, so would be indeed useful.

PR #730, opened 2017-01-04
Split composer download and install and use a temp directory to host it

bcornec:
(michael:) this is required for "making make" without moving composer to `/usr/bin` which keeps the make "sudoless".

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -75,12 +75,16 @@ $(CLEANDIRS):
 	$(MAKE) -s -C $(@:clean-%=%) clean
 
 phpvendors:
-	composer install -q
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -75,12 +75,16 @@ $(CLEANDIRS):
 	$(MAKE) -s -C $(@:clean-%=%) clean
 
 phpvendors:
-	composer install -q
+	$(BUILDDIRS)/composer install -q
```
maximilianhuber: `$BUILDDIRS` is a **list** defined at line `38` not a single directory? Does this work?

PR #731, opened 2017-01-05
fix(Dockerfile): solve composer usage

bcornec:
Adapt Dockerfile to the new place where composer is
downloaded and used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,8 @@ RUN set -x \
  && /fossology/utils/fo-installdeps -e -y \
  && $_cleanup
 RUN curl -sS https://getcomposer.org/installer | php && \
-    mv composer.phar /usr/local/bin/composer
+	mkdir /tmp/composer && \
+    mv composer.phar /tmp/composer/composer
 
 ADD . .
 RUN chmod +x /fossology/docker-entrypoint.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,9 +32,7 @@ RUN set -x \
  && /fossology/install/scripts/php-conf-fix.sh --overwrite \
  && /fossology/utils/fo-installdeps -e -y \
  && $_cleanup
-RUN curl -sS https://getcomposer.org/installer | php && \
-	mkdir /tmp/composer && \
-    mv composer.phar /tmp/composer/composer
+RUN make composer_install
 
 ADD . .
 RUN chmod +x /fossology/docker-entrypoint.sh
--- a/README.md
+++ b/README.md
@@ -32,8 +32,10 @@ use, since the the standalone image does not take care of data persistency.
 
 A pre-built Docker image is available from [Docker Hub](https://hub.docker.com/r/fossology/fossology/) and can be run using following command:
 ``` sh
-docker run -p 8081:8081 fossology/fossology
+docker run -p 8081:80 fossology/fossology
 ```
+
+The docker image can then be used using http://IP_OF_DOCKER_HOST:8081/repo user fossy passwd fossy.
 
 Execution with external database container can be done using Docker Compose.
 The Docker Compose file is located under the `/install` folder can can be run using following command:
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -30,8 +30,7 @@ sudo apt-get update -qq -y
 
 sudo apt-get install -qq curl php5 git libspreadsheet-writeexcel-perl libdbd-sqlite3-perl
 
-curl -sS https://getcomposer.org/installer | php
-sudo mv composer.phar /usr/bin/composer
+make composer_install
 
 # install spdx-tools
 /vagrant/install/scripts/install-spdx-tools.sh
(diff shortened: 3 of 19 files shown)


heckj:
Thanks @bcornec

FrenchBen:
Patch works for me and fixes #732  👍

Dockerfile
```diff
@@ -33,7 +33,8 @@ RUN set -x \
  && /fossology/utils/fo-installdeps -e -y \
  && $_cleanup
 RUN curl -sS https://getcomposer.org/installer | php && \
-    mv composer.phar /usr/local/bin/composer
+	mkdir /tmp/composer && \
+    mv composer.phar /tmp/composer/composer
```
maximilianhuber: the path `/tmp/composer/composer` is not a very good choice, since 
- this folder gets cleaned on every reboot on most systems
- it might interfere with other fossoloy development setups
- `/tmp/` might be mounted with `noexec` for security reasons

Note: this comment is to late and is related to https://github.com/fossology/fossology/pull/701 which is already merged. In my opinion one should change [`COMPOSERTMP = /tmp/composer`](https://github.com/fossology/fossology/pull/701/files#diff-f57f2991a6aa25fe45d8036c51bf8b4dR7) to something in the current subtree.

maximilianhuber: The script `fo_installdeps` still installs composer to the "old" place: https://github.com/fossology/fossology/blob/35d95766b6000acefae7b9849f4718f0380c7acf/utils/fo-installdeps#L267

maximilianhuber: If one calls the following
```bash
$ composer_download
$ sudo composer_install
```
one is no longer able to execute `composer_download` again, since root owns some of the files in `~/.composer`. Thus loosing the binary at `/tmp/composer/composer` is hard problem. (tested on vagrant setup)

maximilianhuber: **Idea:** on Projectbuilder setup install composer to `/tmp/composer/composer` and add `/tmp/composer` to `$PATH` before calling `make phpvendors` resp `make composer_install` and remove all changes related to the placement of composer.

bcornec: You points wrt noexec are valid so I agree we should find a better place to host it. 
I didn't want to perturbate your process too much, but indeed downloading it in the fossology tree should bring what we need and solve the issues you mentioned. I'll propose a new patch for that.

maximilianhuber: This does not solve the problem with **"once you have used `sudo make composer_install` and you have lost the composer binary, you are no longer able to install composer via `make composer_download`"** (unless you use admin privileges remove `~/.composer`) see [comment above](https://github.com/fossology/fossology/pull/731#discussion_r96665109).

What speaks against using the `$PATH` variable to tell the makefiles where to look for composer?

maximilianhuber:
See #738 as alternative solution.

Dockerfile
bcornec: Using PATH is relying on environment to be correct in the first place. If another command is named composer and placed earlier in PATH, you lost :-( And if it's a malicious one, it's worse.

What's wrong with calling it with the full path name, once you know where it should be hosted ?

I think the approach we currently have is the right one, except for the location which should be under the fossology tree (which I didn't want to do myself). Also which variable can be used to declare that fossology tree ?

Wrt the .composer dir, I wasn't aware of that. But the solution is rather to remove that directory then in the Makefile at last step of the install phase (run by root as well), as anyway it shouldn't be created as root.

maximilianhuber: But:
- we still do not know where it should be hosted
- the command `make composer_download` fails after composer was called once as root
- you need to download it once for every source tree
  - if you clone in a new location you might not be able to install the necessary composer, unless you remove the `~/.composer` folder. In fact you might not have the necessary permissions to do that and thus are not able build the project
- being forced to remove the folder `~/.composer` is not good
- you then have to adopt
  - the `fo-installdepts` script
  - the documentation: an additional step is needed for the source installation ([the archived documentation](http://archive15.fossology.org/projects/fossology/wiki/SysConfigSrc) which is still used by some people and is easily found will stop working)
  - the docker scripts
  - the vagrant scripts

> Using PATH is relying on environment to be correct in the first place.

- since you are the person defining `PATH=/my/composer/path:$PATH` that should not be a problem
- you might not want to prevent users from using system packages of composer. E.g. my operating system offers them.

> Also which variable can be used to declare that fossology tree ?

Do you search for an environmental variable used for this? I'm not aware of such an variable in the Makefiles

> But the solution is rather to remove that directory then in the Makefile at last step of the install phase (run by root as well)

You can't simply remove the `~/.composer` folder, which might be used for other projects on the machine.

> ...  as anyway it shouldn't be created as root.

That is an composer problem. When downloaded and installed as user, it seems to write the current home into the `composer.phar` which we rename to `composer`. Thus runnig this composer with root touches files in `~/.composer` of the user.

bcornec:
Replacing with #739 for composer and #740  for project-builder.org patches

PR #734, opened 2017-01-17
feat(packaging): vagrant test file and config for httpd 2.4

mcjaeger:
Cleaning up the final version of the vagrant test file in pbconf for
testing and developing package builds after it worked for centos7. In
addition fixing configuration for Apache httpd 2.4 and leaving config for 2.2 as comment

for this PR 734 there is now an updated vagrant file which creates a centos7 machine and creates the packages for the same. Currently there are the following open issues with the created packages:

* apache 2.4 config must be adapated [1], which is also covered by this PR 734 
* SELinux setting for the file in /var/cache is wrong
* It would require ninka, otherwise fossology-ninka (and other parts of fossology) will not work. Work around with uninstalling or not installing fossology-ninka
* The time zone must be set for PHP. This is an issue which I have not seen for my deb trials [2].

[1] https://httpd.apache.org/docs/2.4/upgrading.html
[2] http://stackoverflow.com/questions/16765158/date-it-is-not-safe-to-rely-on-the-systems-timezone-settings

### DIFF
--- a/install/fo-apache.conf
+++ b/install/fo-apache.conf
@@ -5,8 +5,15 @@ Alias /repo /usr/share/fossology/www/ui
 <Directory "/usr/share/fossology/www/ui">
 	AllowOverride None
 	Options FollowSymLinks MultiViews
-	Order allow,deny
-	Allow from all
+	
+    <IfVersion < 2.3>
+        order allow,deny
+        allow from all
+    </IfVersion>
+    <IfVersion >= 2.3>
+        Require all granted
+    </IfVersion>
+	
 	# uncomment to turn on php error reporting
 	#php_flag display_errors on
 	#php_value error_reporting 2039
--- a/install/src-install-apache-example.conf
+++ b/install/src-install-apache-example.conf
@@ -1,8 +1,8 @@
 # FOSSology example apache config
 
-Alias /repo/ /usr/local/share/fossology/www/ui/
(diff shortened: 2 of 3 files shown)


maximilianhuber:
You could use
```
        <IfVersion < 2.3>
                order allow,deny
                allow from all
        </IfVersion>
        <IfVersion >= 2.3>
                Require all granted
        </IfVersion>
```
to have an configuration file which works with the old and new apache.

This was actually already done for the file https://github.com/fossology/fossology/blob/master/install/src-install-apache-example.conf

**btw.:** Why do we have two files 
- `install/src-install-apache-example.conf` and
- `install/fo-apache.conf`?

Other question: Why is fossology by default placed at the suburl `repo/`? I think it would be good to move it to `/`.

install/fo-apache.conf
```diff
@@ -5,8 +5,14 @@ Alias /repo /usr/share/fossology/www/ui
 <Directory "/usr/share/fossology/www/ui">
 	AllowOverride None
 	Options FollowSymLinks MultiViews
-	Order allow,deny
-	Allow from all
+	
+	# Order and Allow are for Apache httpd 2.2
+	# Order allow,deny
+	# Allow from all
```
maximilianhuber: One should prefer the approach working on both version. This is described in the [comment above](https://github.com/fossology/fossology/pull/734#issuecomment-273523560).

PR #738, opened 2017-02-01
fix(make): do not place composer at `/tmp/composer/composer`

### LINKED ISSUE #732: docker-compose up fails to build fossology-scheduler
When trying to bring up Fossology using the docker-compose file and running:
```
cd install
docker-compose up
```
I get:
```
current dir: '/fossology/src'
/usr/bin/install -D -m 644 composer.json composer.lock /usr/local/share/fossology
(cd /usr/local/share/fossology; /tmp/composer/composer install -q --no-dev)
/bin/sh: 1: /tmp/composer/composer: not found
Makefile:87: recipe for target 'composer_install' failed
make[1]: Leaving directory '/fossology/src'
Makefile:45: recipe for target 'install-src' failed
make[1]: *** [composer_install] Error 127
make: *** [install-src] Error 2
ERROR: Service 'fossology-scheduler' failed to build: The command '/bin/sh -c set -x  && make install  && make clean' returned a non-zero code: 2
```
Version:
```
$ docker-compose version
docker-compose version 1.10.0-rc1, build ecff6f1
$ docker version
Client:
 Version:      1.13.0-rc5
 API version:  1.25
 Go version:   go1.7.3
 Git commit:   43cc971
 Built:        Thu Jan  5 03:07:30 2017
 OS/Arch:      darwin/amd64

Server:
 Version:      1.13.0-rc5
 API version:  1.25 (minimum version 1.12)
 Go version:   go1.7.3
 Git commit:   43cc971
 Built:        Thu Jan  5 03:07:30 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

maximilianhuber:
this pull request
- undos the changes to the composer placement introduced in #701
- adds the option to pass the environmental variable `COMPOSER_PHAR` to
  `src/Makefile` to allow placing the composer.phar at arbitrary places
- this fixes the source installation, i.e. the documentation was not up
  to date
- this fixes the problems with `vagrant up` (see #732)
- this fixes the problems with `docker-compose build`

as discussed in #731

closes #732

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -49,7 +49,8 @@ $(BUILDDIRS):
 	$(MAKE) -s -C $(@:build-%=%)
 
 # install depends on everything being built first
-install: all $(INSTALLDIRS) composer_install
+install: install_offline composer_install
+install_offline: all $(INSTALLDIRS)
 $(INSTALLDIRS):
 	$(MAKE) -s -C $(@:install-%=%) install
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -4,7 +4,7 @@
 
 TOP = ..
 VARS = $(TOP)/Makefile.conf
-COMPOSERTMP = /tmp/composer
+COMPOSER_PHAR ?= composer
 include $(VARS)
 
 # the directories we do things in by default
@@ -76,17 +76,12 @@ $(CLEANDIRS):
 	$(MAKE) -s -C $(@:clean-%=%) clean
 
 phpvendors:
-	$(COMPOSERTMP)/composer install -q
-
-composer_download:
-	curl -sS https://getcomposer.org/installer | php
-	mkdir -p $(COMPOSERTMP)
-	mv composer.phar $(COMPOSERTMP)/composer
+	$(COMPOSER_PHAR) install -q
 
 composer_install:
 	@echo "current dir: '$${PWD}'"
 	$(INSTALL_DATA) composer.json composer.lock $(DESTDIR)$(MODDIR)
-	(cd $(DESTDIR)$(MODDIR); $(COMPOSERTMP)/composer install -q --no-dev)
+	(cd $(DESTDIR)$(MODDIR); export COMPOSER_HOME=/dev/null; $(COMPOSER_PHAR) install -q --no-dev)
 
 .PHONY: subdirs $(BUILDDIRS)
 .PHONY: subdirs $(DIRS)
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,13 +27,13 @@ RUN set -x \
 
 ADD utils/fo-installdeps utils/fo-installdeps
 ADD install/scripts/php-conf-fix.sh install/scripts/php-conf-fix.sh
+ADD utils/install_composer.sh utils/install_composer.sh
 RUN set -x \
  && $_update \
- && /fossology/install/scripts/php-conf-fix.sh --overwrite \
  && /fossology/utils/fo-installdeps -e -y \
- && $_cleanup
-RUN curl -sS https://getcomposer.org/installer | php && \
-    mv composer.phar /usr/local/bin/composer
+ && $_cleanup \
+ && /fossology/install/scripts/php-conf-fix.sh --overwrite \
+ && /fossology/utils/install_composer.sh
 
 ADD . .
 RUN chmod +x /fossology/docker-entrypoint.sh
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -30,8 +30,7 @@ sudo apt-get update -qq -y
 
 sudo apt-get install -qq curl php5 git libspreadsheet-writeexcel-perl libdbd-sqlite3-perl
 
-curl -sS https://getcomposer.org/installer | php
-sudo mv composer.phar /usr/bin/composer
+# sudo /vagrant/install/install_composer.sh
 
 # install spdx-tools
 /vagrant/install/scripts/install-spdx-tools.sh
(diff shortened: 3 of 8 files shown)


maximilianhuber:
@bcornec does the commit 011ffe3 fix the composer problem in the packaging context?

src/Makefile
```diff
@@ -76,17 +75,12 @@ $(CLEANDIRS):
 	$(MAKE) -s -C $(@:clean-%=%) clean
 
 phpvendors:
-	$(COMPOSERTMP)/composer install -q
-
-composer_download:
-	curl -sS https://getcomposer.org/installer | php
-	mkdir -p $(COMPOSERTMP)
-	mv composer.phar $(COMPOSERTMP)/composer
+	composer install -q
```
bcornec: Problem is that we're loosing the 2 steps approach here (download target and usage target)

Download target should be invoked during an environment preparation phase, not during build. Usage when we need it during build.

Also using a full path name for calling composer is not necessarily an issue, (instead of using the PATH which can be wrong) as long as we don't use something controversial such as /tmp/something. It could have been stored in the fossology tree as well, but I didn't found a variable describing it in that Makefile, so that's why I didn't use it.

utils/run_with_composer_in_path.sh
```diff
@@ -0,0 +1,16 @@
...
+tmp_composer="${tmp_composer_dir}/composer"
+
+type "composer" &> /dev/null && {
+    exec "$@"
+} || {
+    mkdir -p "$tmp_composer_dir"
+    [[ -e "$tmp_composer" ]] || {
+        curl -sS https://getcomposer.org/installer \
+            | php && mv composer.phar "$tmp_composer"
+    }
```
bcornec: So I think we're back to what existed initially, and the problem related to that approach: there is no split between the download of composer and its usage again.

src/Makefile
maximilianhuber: Ok, that's a problem of the wrapper script. A correct altenrative would be to download composer to some temporary path (e.g. `tmp/composer/`) and later in the next step all the make command with the correct path (e.g. `PATH='tmp/composer/:$PATH make ....`).

Does this fit your 2 step approach?

utils/run_with_composer_in_path.sh
maximilianhuber: You could split it up again (instead of using the wrapper) as discussed in https://github.com/fossology/fossology/pull/738#discussion_r98911383

bcornec:
I think there are coherency issues that need to be solved:

You have a Vagrant file that does:
curl -sS https://getcomposer.org/installer | php
sudo mv composer.phar /usr/bin/composer

You have a Dockerfile that does:
RUN curl -sS https://getcomposer.org/installer | php && \
	mkdir /tmp/composer && \
    mv composer.phar /tmp/composer/composer

You have a Makefile that does:
composer_download:
	curl -sS https://getcomposer.org/installer | php
	mkdir -p $(COMPOSERTMP)
	mv composer.phar $(COMPOSERTMP)/composer


So we have 3 ways of dealing with composer, whereas a single one should be used everywhere. I'll propose a patch for this as I think that's what is creating the issues reported.

maximilianhuber:
In this PR all three cases are consistend and match the first version (AFAIK). Plus you have the option to implement packaging without breaking changes using `$PATH`.

An alternative would be to introduce a new environmental variable which can be used to define the position of composer. Would it be an option to call make with
```
COMPOSER=/tmp/composer/composer make composer_install
```
and within the makefile use something like ${COMPOSER:-composer}` ??

maximilianhuber:
The (untested) changes in b84af50 might allow you to run the makefile with
```
make COMPOSER=/tmp/composer/composer composer_install
```
is that ok?

bcornec:
No the calls to download composer are all different in that they all place the resulting file in different places, creating a mess when you need to call the tool afterwards through a single call !

The problem is splitting download from usage, not the Variable.
Look at #739 to see what I mean.

maximilianhuber:
But the download is splitted. The makefiles do never download composer. For that one either calls `fo-instaldepts` or installs it by hand.

In your example of the mess are still two of three variants introduced by your movement of composer to `/tmp/composer`

maximilianhuber:
@bcornec With 3735e45 the idea suggested in the comment above is now working and one is able to call make in the following way (after downloading composer.phar to `/tmp/composer/composer.phar`)
```bash
make COMPOSER_PHAR=/tmp/composer/composer phpvendors
#or
make COMPOSER_PHAR=/tmp/composer/composer install
```

This allows you to download composer at an arbitrary timepoint and in an arbitrary way, without introducing braking changes to other parts of the code.

##

**note:** I was not able to find the code in `pbconf/fossology/rpm/fossology.spec` which takes care of the download of composer.

maximilianhuber:
one might want to replace the line https://github.com/fossology/fossology/pull/738/commits/76ca5c25235ba8d8a099b0211e3dd4b41edae61c#diff-2f7bc9d6d5478304df37e9900a6e32f2R35 (and other lines which download composer) by an script

maximilianhuber:
How to install composer programmatically is described here: https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md

maximilianhuber:
This PR now contains a script to install composer 
- of an specified version
- from an trustworthy source (i.e. github, with specified commit hash)

maximilianhuber:
@mcjaeger This PR is now rebased to current master

mcjaeger:
actually it seems like the fo-apache.conf is not copied anymore, but that is another issue as this seems to be also in master

PR #739, opened 2017-02-01
Improve composer handling

bcornec:
Download composer into a fossology subdir (tmp)
setup COMPOSER_HOME to avoid creation of a .composer dir at install
Fix Docker and Vagrant to use make composer_install

This should help fixing #731, #732 in a better way than #738 IMHO. Please check and comment


src/Makefile
```diff
@@ -86,7 +86,7 @@ composer_download:
 composer_install:
 	@echo "current dir: '$${PWD}'"
 	$(INSTALL_DATA) composer.json composer.lock $(DESTDIR)$(MODDIR)
-	(cd $(DESTDIR)$(MODDIR); $(COMPOSERTMP)/composer install -q --no-dev)
+	(cd $(DESTDIR)$(MODDIR); export COMPOSER_HOME=/dev/null ; $(COMPOSERTMP)/composer install -q --no-dev)
```
maximilianhuber: I think that this will not work, since `TOP=..` and thus this line evaluates to 
```
(cd some/other/folder; export ....; ../tmp/composer/composer install -q --no-dev)
```
which does not work.

Setting `COMPOSER_HOME` is a good idea.

Dockerfile
```diff
@@ -32,8 +32,7 @@ RUN set -x \
  && /fossology/install/scripts/php-conf-fix.sh --overwrite \
  && /fossology/utils/fo-installdeps -e -y \
  && $_cleanup
-RUN curl -sS https://getcomposer.org/installer | php && \
-    mv composer.phar /usr/local/bin/composer
+RUN make composer_install
```
maximilianhuber: The script `fo-installdeps` already handles the installation of composer, i.e. writes it to `/usr/local/bin/composer`.  (see https://github.com/fossology/fossology/blob/4ed3875bcc096bc093a921613772597a06be76c8/utils/fo-installdeps#L261)

Vagrantfile
```diff
@@ -30,8 +30,7 @@ sudo apt-get update -qq -y
 
 sudo apt-get install -qq curl php5 git libspreadsheet-writeexcel-perl libdbd-sqlite3-perl
 
-curl -sS https://getcomposer.org/installer | php
-sudo mv composer.phar /usr/bin/composer
+make composer_install
```
maximilianhuber: Here you mean `make composer_download`

Dockerfile
maximilianhuber: here you mean `make composer_download`

src/Makefile
bcornec: Do we have a variable expressing the full absolute path name of the installation tree ?

maximilianhuber: Not as far as I know

Dockerfile
bcornec: Wel, so we should remove it from fo-installdeps to let it only in the Makefile (or create a separate script).

Why do we need that: because for packaging, we don't need to invoke fo-installdeps, because deps are handled by packages. So we just need to handle what is missing, which is composer. So either through the Makefile or a specific script. But that one needs to be invoked from all other places instead of duplicating.

src/Makefile
bcornec: Ok, using readlink -f will fix this...

Vagrantfile
bcornec: Oops right !

bcornec:
With regards to documentation, where should I make modifications ?

bcornec:
I've made an updated version of the patch which takes in account your remarks, and also fixes additional points which were missing in the first version. Except the doc, that should look better.

bcornec:
I think I've now also fixed the doc, so please review again.

Dockerfile
```diff
@@ -32,8 +32,7 @@ RUN set -x \
  && /fossology/install/scripts/php-conf-fix.sh --overwrite \
  && /fossology/utils/fo-installdeps -e -y \
  && $_cleanup
-RUN curl -sS https://getcomposer.org/installer | php && \
-    mv composer.phar /usr/local/bin/composer
+RUN make composer_download && make composer_install
```
maximilianhuber: At this point in the script are the makefiles not yet added to the context (will be done two lines below) thus this fails.

##

One should be able to download all dependencies (e.g. composer) before adding the source (for rebuild speed reasons / one should not need internet for the rebuild if only source is changed)

Dockerfile
```diff
@@ -32,11 +32,10 @@ RUN set -x \
  && /fossology/install/scripts/php-conf-fix.sh --overwrite \
  && /fossology/utils/fo-installdeps -e -y \
  && $_cleanup
-RUN curl -sS https://getcomposer.org/installer | php && \
-    mv composer.phar /usr/local/bin/composer
 
 ADD . .
 RUN chmod +x /fossology/docker-entrypoint.sh
+RUN make composer_download && make composer_install
```
maximilianhuber: This change now forces one to download composer every time one change a single line in the source tree and builds the image. (e.g. in the development loop)

I do not like this change, all dependencies should be loaded before adding the source.

src/testing/jenkins/Fossology_Source_Install.xml
```diff
@@ -100,19 +100,20 @@ else
...
+git config --global https.proxy &quot;http://xxx:8088&quot;;
 git clone --depth 1 https://github.com/fossology/fossology.git fossology;
 pwd;
-export http_proxy=http://web-proxy.cce.hp.com:8088/
+export http_proxy=http://xxx:8088/
 cd ./fossology/;
 ./utils/fo-installdeps -e -y;
-curl -sS --proxy http://web-proxy.cce.hp.com:8088 https://getcomposer.org/installer | php &amp;&amp; mv composer.phar /usr/local/bin/composer;
+make composer_download
+make composer_install
```
maximilianhuber: The `composer_install` target is also called when `make install` is called, i.e. two lines below (see: https://github.com/fossology/fossology/blob/ae754b26e849b4c9068ed6f4aa354310165aec0c/src/Makefile#L53)

Vagrantfile
```diff
@@ -30,8 +30,8 @@ sudo apt-get update -qq -y
 
 sudo apt-get install -qq curl php5 git libspreadsheet-writeexcel-perl libdbd-sqlite3-perl
 
-curl -sS https://getcomposer.org/installer | php
-sudo mv composer.phar /usr/bin/composer
+make composer_download
+sudo make composer_install
```
maximilianhuber: The `composer_install` target is also called later when `sudo make install` is called (see: https://github.com/fossology/fossology/blob/ae754b26e849b4c9068ed6f4aa354310165aec0c/src/Makefile#L53)

Dockerfile
maximilianhuber: The `composer_install` target is also called later when `make install` is called (see: https://github.com/fossology/fossology/blob/ae754b26e849b4c9068ed6f4aa354310165aec0c/src/Makefile#L53)

maximilianhuber:
You might need to change the following line in `src/Makefile`:
```makefile
install: all $(INSTALLDIRS) composer_install
```
to split the composer call from the install

Dockerfile
bcornec: Well, I agree with you !! That's why at first, when we starting talking about that, my recommendation was to say: 
integrate composer in your tree. Create a script to update the version if needed. But put it under git, and at least all these stuff disappear, and the update of composer is managed separately, fully outside of build/install.

This is tru because this is not provided by default by distributions on which you install and provide FOSSology. So yes, *my* way of doing it would be to create a small script to downlaod (or update) composer, and put it under your $TOP/bin dire, and then call it from there, knowing it will be there. WDYT ? I can cook a patch for this instead.

bcornec:
Hello Maximilian

As per the discussion upper, here is a new version which proposes to integrate composer in tree, and provide a script to mange its update. Vagrant and Docker, and Makefile adapted accordingly. Le tme know if you think we are in the right direction !

maximilianhuber:
AFAIK adding composer to our git tree is no option, since we do not want to redistribute composer.


@mcjaeger what do you think?

maximilianhuber:
Are we sure that running `curl -sS https://getcomposer.org/installer | php` on different machines in different environments always creates the same `composer.phar`? Or does this script apply some customization?

mcjaeger:
I also do not like it. But, we really need to get the packaging out ASAP. composer.phar is distributed as MIT and I will update the license info then for the release. So, I vote yes, let's pull this in.

mcjaeger:
About docuemntation:

1) this github project wiki
2) wiki fossology org

Bu-huuuut: @maximilianhuber @bcornec, is it working  it works for you? tested the vagrant file on projejct root level on my system show this in `/var/log/apache/error.log` when opening fossology, (which returns a blank page):

> [Thu Feb 02 17:38:19.329712 2017] [:error] [pid 27004] [client 10.0.2.2:56342] PHP Fatal error:  require_once(): Failed opening required '/usr/local/share/fossology/vendor/autoload.php' (include_path='.:/usr/share/php:/usr/share/pear') in /usr/local/share/fossology/lib/php/common.php on line 25

looking into the vagrant output at installing I have the following line

> ==> default: current dir: '/vagrant/src'
> ==> default: /usr/bin/install -D -m 644 composer.json composer.lock /usr/local/share/fossology
> ==> default: (cd /usr/local/share/fossology; export COMPOSER_HOME=/dev/null ; ../utils/composer install -q --no-dev)
> ==> default: /bin/sh: 1: 
> ==> default: ../utils/composer: not found
> ==> default: make[1]: 
> ==> default: *** [composer_install] Error 127
> ==> default: make[1]: Leaving directory `/vagrant/src'
> ==> default: make: 
> ==> default: *** [install-src] Error 2
> ==> default: *** Running postinstall for everything ***

Vagrantfile
```diff
@@ -30,9 +30,6 @@ sudo apt-get update -qq -y
 
 sudo apt-get install -qq curl php5 git libspreadsheet-writeexcel-perl libdbd-sqlite3-perl
 
-curl -sS https://getcomposer.org/installer | php
```
mcjaeger: it seem like that this removal in fact results in a non working installation now. even in the checked out branch I have the composer file there (1,8MB), but it is not found by the vagrant provisioning.

mcjaeger:
should `pbconf/Vagrantfile` be changed with 

> # for pkg in "fossology-composer" "fossology"; do
>     pb -t -p fossology sbx2build   fossology
>     pb -p    fossology build2prep  fossology
> #   pb -t -p fossology sbx2pkg     fossology
>     pb -t -p fossology sbx2pkg2ins fossology

maximilianhuber:
I think vagrant provisioning is failing, since the proposed fix via `readlink -f` is missing (see [comment above](https://github.com/fossology/fossology/pull/739#discussion_r98951446)).

The `readlink -f` solution is tricky, since one can call the make command from a different folder.

mcjaeger:
for me executing `docker build .` leads to an error:

> (cd /usr/local/share/fossology; export COMPOSER_HOME=/dev/null ; ../utils/composer install -q --no-dev)
> /bin/sh: 1: ../utils/composer: not found
> Makefile:82: recipe for target 'composer_install' failed
> make[1]: Leaving directory '/fossology/src'
> make[1]: *** [composer_install] Error 127
> Makefile:45: recipe for target 'install-src' failed
> make: *** [install-src] Error 2

please clarify, `readlink -f` should be applied where?

maximilianhuber:
The problem is that you change the folder via `cd /usr/local/share/fossology` and the call `../utils/composer install -q --no-dev` which is **relative to the makefile in which these lines are placed**.

You can use `readlink -f` to make an path absolute **relative to the current location of the caller**.

That you can call the make command from different places can be seen by the fact that the following commands do the same (on vagrant):
```bash
cd /vagrant && make install # readlink -f ../utils/composer is here /utils/composer
# and
cd /vagrant/src && make install # readlink -f ../utils/composer is here /vagrant/utils/composer (as expected)
```

maximilianhuber:
Let me mention the alternative #738 again, which solves the packaging problem (AFAIK) without touching the other parts (and thus is not breaking anything).

maximilianhuber:
Another (not really beautiful) solution might be, to also install composer to the target folder (e.g. `/usr/local/share/fossology`) in a way that keeps its relative position. (That would be an ugly hack)

mcjaeger:
I have create a wiki page for documentation here:

https://github.com/fossology/fossology/wiki/Notepad-Building-Packages

src/testing/jenkins/Fossology_Source_Install.xml
```diff
@@ -128,7 +129,7 @@ cat /etc/issue |grep -iE &apos;ubuntu 12.04&apos;;
 if [ $? -eq 0 ] ; then
   cp ../../composer.json /srv/fossology/test4packageinstall;
   cd /srv/fossology/test4packageinstall;
-  composer install;
+  $mypwd/utils/composer install;
```
maximilianhuber: (unimportant comment:)
Here one should call the makefile?

Is someone actually using this file?

bcornec:
In #738, composer is in /tmp,and wee conclude that wasn't acceptable.
The big difference is that by having composer.phar in your tree, you do not rely at all on what other people do at the other end. You control which version you deliver and use. It's up to you to use the new script to download a new version, only when it's needed and on the platform you want.

You have fully decoupled the management of that from the build phase (with or without packages).
I think this is needed, and of course change the way it was done before.

So the fact we're converging is a good sign that we're solving probably all the issues each of us was looking at.

maximilianhuber:
in #738 the placement of composer is completely customizable. I just have used `/tmp/...` for your packaging since it was convenient. But #738 does not force you to place the binary there. You could either rely on PATH or specify it exatly by yourself. 

>  You control which version you deliver and use.

The script `utils/fo-update-composer` always updates to latest and thus takes the control away.?

bcornec:
The script is not called during build ! And that should NOT be called during build. A reprocueable build can not rely on deps you do not control precisely (other packages are ok, a randomly downloaded file is not IMO). It's only called manually when you want to update your version of composer (and because you have a good reason for that) or in the Vagrantfile/Dockerfile, if you want to automate its installation (which is now optional in fact so could also be removed).

maximilianhuber:
> ... other packages are ok ...

This PR will prevent the usage of the composer package provided by some distributions. There are composer packages [e.g. for Ubuntu](http://packages.ubuntu.com/yakkety/composer). 

##

Can you please answer the following question:
- does the `curl ... | php` command, which installs composer, act the same on every system? are the artifacts equal on the binary level?

src/Makefile
```diff
@@ -76,17 +75,12 @@ $(CLEANDIRS):
...
-	curl -sS https://getcomposer.org/installer | php
-	mkdir -p $(COMPOSERTMP)
-	mv composer.phar $(COMPOSERTMP)/composer
+	$(COMPOSERPATH)/composer install -q
 
 composer_install:
 	@echo "current dir: '$${PWD}'"
 	$(INSTALL_DATA) composer.json composer.lock $(DESTDIR)$(MODDIR)
-	(cd $(DESTDIR)$(MODDIR); $(COMPOSERTMP)/composer install -q --no-dev)
+	(cd $(DESTDIR)$(MODDIR); export COMPOSER_HOME=/dev/null ; $(COMPOSERPATH)/composer install -q --no-dev)
```
maximilianhuber: This still does not work if the `COMPOSERPATH` if it is not absolute (same in #738)

bcornec:
That's why my patch provides an absolute path for it indeed.

You cannot trust an environment to be safe IMO. It's way better to give full path names.

maximilianhuber:
We already trust the environment enough to use `php`, `wget`, `curl`, the package manager, ... without full path. I do not think that `composer` is the problem.

##

> You cannot trust an environment to be safe IMO. It's way better to give full path names.

But that will cause unexpected problems in the **from source** installation. The relative path works for `make phpvendors` but not for `make composer_install`. Maybe rename `COMPOSERPATH` to make clear that it has to be absolute?

bcornec:
wrt composer packages, that would be the best approach indeed. So depending on the distro we could put a requirement on the composer package provided when it exists and deal with the lack of it when it doesn't, either with a file based approach as here or by providing our own package to solve the issue.

I can't really comment on the similarity of the result of the cul | php command on different distros. I'd tend to say it may well produce different results as the local php invokation will lead to different version of php, and thus the result will probably be different. But I'm not a PHP expert at all, so using just my good sense to guess here.

bcornec:
If you want I'll rename it COMPOSERABSPATH if that's better for you. And also fix the conflict.

mcjaeger:
please help: the vagrant file on project root does not execute properly, it shows the error:
```
==> default: writing VERSION file for BUILD
==> default: make -C ./src composer_download
==> default: make[1]: Entering directory `/vagrant/src'
==> default: make[1]: *** 
==> default: No rule to make target `composer_download'
==> default: .  Stop.
==> default: make[1]: Leaving directory `/vagrant/src'
==> default: make: *** [composer_download] Error 2
==> default: make -C ./src composer_install
==> default: make[1]: Entering directory `/vagrant/src'
==> default: current dir: '/vagrant/src'
==> default: /usr/bin/install -D -m 644 composer.json composer.lock /usr/local/share/fossology
==> default: /usr/bin/install: 
==> default: target ‘/usr/local/share/fossology’ is not a directory
==> default: make[1]: 
==> default: *** [composer_install] Error 1
==> default: make[1]: Leaving directory `/vagrant/src'
==> default: make: *** [composer_install] Error 2
```
indeed, looking into file https://github.com/fossology/fossology/blob/dev/composer3/src/Makefile , `composer_download` is not there:
I think calling `composer_download` must be removed from the vagrant file on project root in line 50, if I am not mistaken?

mcjaeger:
as maximilian wrote `You have also made the composer path configurable, but you have no fallback if the environmental variable is not set.` ... I was coming across this now too. maybe it is also ok to set this if this was not set, because I did:
```
sudo utils/fo-update-composer /home/fossy/composer
export COMPOSERABSPATH=/home/fossy/ 
sudo make install
```
-> does not work because environment variables are not pulled when sudoing, maybe you guys do that all the time, but I needed some  time until I got this with `sudo -E make install`.

bcornec:
Right. I'm updating the patch.

mcjaeger:
source install works apart from missing documentation in the wiki, but I will update that once we have merged this.

mcjaeger:
(offtopic) somehow the apache config from `/usr/local/etc/fossology/conf/src-install-apache-example.conf` is not copied anymore. this is a new bug?

mcjaeger:
about the latest commit: I was actually mentioning ` composer_download ` not `composer_install`? is that on purpose?

maximilianhuber:
Your last commit states that it **put[s] back the composer_install one**, but the diff only shows removals. And it does not touch the lines changed in the previous commit https://github.com/fossology/fossology/pull/739/commits/31ff0c4b8d5c4aa4dd0978d6e821e198f3173e62 and thus can not revert it.

Are you testing your changes before pushing?

maximilianhuber:
You introduce breaking changes to the source installation but you have not adopted the files under `/pbconf/` to cope with the changes. I think that **this pr is breaking the packaging**.

Have you tested packaging with this PR?

maximilianhuber:
@bcornec said in https://github.com/fossology/fossology/pull/739#issuecomment-277944091:
> I can't really comment on the similarity of the result of the cul | php command on different distros. [...]

I do not think that this is enough evidence.

maximilianhuber:
In this PR projectbuilder still tries to build `fossology-composer` and thus **breaks packaging**.
```
[...]
==> default: Dependency Installed:
[...]                                          
==> default: Complete!
==> default: ++ systemctl start docker
==> default: Running provisioner: shell...
    default: Running: inline script
==> default: ++ cat
==> default: ++ for pkg in '"fossology-composer"' '"fossology"'
==> default: ++ pb -t -p fossology sbx2build fossology-composer
==> default: Project fossology is not Project-Builder compliant. at /usr/share/perl5/vendor_perl/ProjectBuilder/VCS.pm line 729, <GIT> line 4.
==> default: ERROR: Inconsistency detected:
==> default:        * /home/vagrant/prj/fossology/pbconf (PBCONFDIR) refers to git@github.com:fossology/fossology.git but
==> default:        * /home/vagrant/.pbrc refers to https://github.com/fossology/fossology.git
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

mcjaeger:
@maximilianhuber I have also noticed that the vagrant file in `pbcobnf/` is broken. I took this pull request as a basis to work on commonly to see if we get that working jointly between us: src install, vagrant file, docker. Once the vagrant file in project root is working there is no deal to add that change to the vagrant file in `pbconf/`

maximilianhuber:
@bcornec as an alternative, you might want to have a look at https://github.com/fossology/fossology/pull/738/files#diff-57bd0ecebf88871e52ba1baabeb22f2bR60 of #738. There is a script which installs composer in a specified version and downloads it from a trustworthy source (github with specified commit hash).

Therefore #738 should now satisfy
- no root is needed to install composer
- composer can be installed to and used from arbitrary places
- consistency, i.e. all composer installations (should) use `utils/install_composer.sh`

without
- breaking changes
- an 1.8MB blob in the source tree

##

More information on the programmatically installation of composer can be found at https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md

mcjaeger:
currently, there seems to be an issue with project builder in general. I tried to make a `pbconf/Vagrantfile`:
```
==> default: ++ pb -p fossology build2prep fossology
==> default: Project: fossology
==> default: Action: build2prep
==> default: INFO: ------ Starting to prepare build environement ------
==> default: rpm -q --quiet epel-release-latest-7.noarch... 
==> default: NOT OK but non blocking
==> default: Adding package to setup repository... 
==> default: ERROR running command (sudo rpm -Uvh /tmp/pb.ZjtBZu55qV/epel-release-latest-7.noarch.rpm) with cwd=/home/vagrant, pid=24202 at /usr/share/perl5/vendor_perl/ProjectBuilder/Base.pm line 178.
==> default: 	ProjectBuilder::Base::pb_system('sudo rpm -Uvh /tmp/pb.ZjtBZu55qV/epel-release-latest-7.noarch...', 'Adding package to setup repository') called at /usr/share/perl5/vendor_perl/ProjectBuilder/Distribution.pm line 597
==> default: 	ProjectBuilder::Distribution::pb_distro_setuprepo_gen('HASH(0x35ecfb0)', '/home/vagrant/prj/fossology/pbdelivery/pbrc', 'addrepo') called at /usr/share/perl5/vendor_perl/ProjectBuilder/Distribution.pm line 531
==> default: 	ProjectBuilder::Distribution::pb_distro_setuprepo('HASH(0x35ecfb0)') called at /usr/bin/pb line 1620
==> default: 	main::pb_build2prep() called at /usr/bin/pb line 859
==> default: ERROR: child (sudo rpm -Uvh /tmp/pb.ZjtBZu55qV/epel-release-latest-7.noarch.rpm) cwd=/home/vagrant exited with value 1
==> default: error: open of /tmp/pb.ZjtBZu55qV/epel-release-latest-7.noarch.rpm failed: No such file or directory
```
I have no clue as
* I think it has worked yesterday
* I am seeing no project builder change.  (version from Jan 3rd)
* this error is in project builder
* logging into the vagrant machine indeed does not show the created tmp dir there...

maximilianhuber:
This should be solved by https://github.com/fossology/fossology/pull/738. Maybe one can close this one until its functionality is especially needed. (ping @mcjaeger @bcornec )

mcjaeger:
@bcornec maybe close this and create a new PR with the new branch: https://github.com/fossology/fossology/tree/dev/pbvagrantcent

mcjaeger:
@bcornec we can close this right?

bcornec:
I'm not particularly happy with composer (having a non-reproduceable build is always a concern IMO) but for now I don't have a better proposal (except packaging composer !)

PR #749, opened 2017-02-22
fix(scripts) : update timezone info to php.ini

ag4ums:


install/scripts/php-conf-fix.sh
```diff
@@ -6,7 +6,24 @@
 #
 echo 'Automated php.ini configuration adjustments'
 echo
-phpIni=/etc/php5/apache2/php.ini
+if [ -f /etc/debian_version ]
```
maximilianhuber: At the moment there is no default case, i.e. `phpIni` might be empty.

I would recommend to update the code as follows:
```bash
if [ -f /etc/redhat-release ]; then
    phpIni=/etc/php.ini
    TIMEZONE=`readlink /etc/localtime | sed 's%/usr/share/zoneinfo/%%'`
else
    phpIni=/etc/php5/apache2/php.ini
    TIMEZONE=`cat /etc/timezone`
fi
if [ -z $TIMEZONE ]; then
    TIMEZONE="America/Denver"
fi
```

install/scripts/php-conf-fix.sh
```diff
@@ -6,7 +6,24 @@
...
+    phpIni=/etc/php5/apache2/php.ini
+    TIMEZONE=`cat /etc/timezone`
+    if [ -z $TIMEZONE ]
+    then
+        TIMEZONE="America/Denver"
+    fi
+elif [ -f /etc/redhat-release ]
+then
+    phpIni=/etc/php.ini
+    TIMEZONE=`readlink /etc/localtime | sed 's/.*\usr\/share\/zoneinfo\///'`
```
maximilianhuber: There is missing the backslash in front of `usr`

Even simpler: you can replace the delimiter by `%` and write the command without escaping as 
```
    ... | sed 's%/usr/share/zoneinfo/%%'`
```

maximilianhuber: readlink is a command from `coreutils` which might not be installed everywhere. I think this is OK here.

install/scripts/php-conf-fix.sh
```diff
@@ -22,6 +39,8 @@ then
     sed -i 's/^\(post_max_size\s*=\s*\).*$/\1701M/' php.ini
     echo 'Setting max upload filesize to 700M'
     sed -i 's/^\(upload_max_filesize\s*=\s*\).*$/\1700M/' php.ini
+    echo Setting timezone to $TIMEZONE
+    sed -i "s@.*date.timezone =.*@date.timezone = $TIMEZONE@" php.ini
```
maximilianhuber: please use the same delimeters as above to be consistent. It might also be good to use single quotes to prevent globbing. I.e. use
```bash
    sed -i 's/.*date.timezone =.*/date.timezone = '"$TIMEZONE"/' php.ini
```

install/scripts/php-conf-fix.sh
ag4ums: but then what will be the location for phpIni in default case?

maximilianhuber: I would say the default case is the debain case which might match most other distros (as shown in the code snipped above).

install/scripts/php-conf-fix.sh
ag4ums: Is this working? seems it cannot escape the / in the path in $TIMEZONE

maximilianhuber: Yes, you might be right.

Maybe change the delimiter from `@` to `%` to be consistent with the line 11 (or change line 11).

mcjaeger:
hm, tested, unfortunately, the `php-conf-fix.sh`  is not copied to the package installation (what I expected). I think I was confused by this file, which actually does not call the script neither, but applied a part of it:

https://github.com/fossology/fossology/blob/master/debian/fossology-web.postinst

:( will try to fix the php installation and update

maximilianhuber:
@mcjaeger this already solves the problem (when called manually). The problem that this is not done via the package installation is a separate problem.

PR #751, opened 2017-02-28
Fix target name for stanalone nomos

AMDmi3:


mcjaeger:
Hi, thank you very much for this pull request!!! just one minor thing, could I please ask you for a boiler plate thing if possible?

could you please change the commit message according to contribution guidelines https://github.com/fossology/fossology/blob/master/CONTRIBUTING.md
very simple:
change `Fix target name for stanalone nomos` to `fix(nomos): target name for stanalone nomos compilation`
it should be just an amend commit with a subsequent rebase and force push. If that does not work for you for technical reasons, just write an e-mail to me.

AMDmi3:
Done

Issue #723, opened 2016-12-06
Standalone Docker container fails due to missing priveledges


maximilianhuber:
In the `docker-entrypoint.sh` script is the line https://github.com/fossology/fossology/commit/c8b4b5485064a7681ea6681f1c8c2c8bf24c5279#diff-60713fb19f6c50a2473835df63ff17f9R52 which starts the postgresql-db for the standalone container.

Since this script is due to recent changes called (#720) with an unprivileged user, which is not allowed to start the db service, it fails.

--

this bug was discovered by Manfred Wilner and posted on the dev-mailing-list:
> I followed direction for docker on Ubuntu and was able to pull down the image but:
> ```
> Digest: sha256:5103804e8decb33f8bf1ac13dea7757648e3b752800a915784170f8af4eceb7a
> Status: Downloaded newer image for fossology/fossology:latest
> *****************************************************
> WARNING: No database host was set and therefore the
> internal database without persistency will be used.
> THIS IS NOT RECOMENDED FOR PRODUCTIVE USE!
> *****************************************************
> chmod: changing permissions of '/var/run/postgresql': Operation not permitted
> ```

maximilianhuber:
Maybe the following changes might help:
```diff
diff --git Dockerfile Dockerfile
index a30564c..4e71dbe 100644
--- Dockerfile
+++ Dockerfile
@@ -82,6 +82,5 @@ EXPOSE 8081
 
 ################################################################################
 VOLUME /srv/fossology/repository/
-USER fossy
 
 ENTRYPOINT ["/fossology/docker-entrypoint.sh"]
diff --git install/docker-compose.base.yml install/docker-compose.base.yml
index b96e015..f97ebd6 100644
--- install/docker-compose.base.yml
+++ install/docker-compose.base.yml
@@ -24,3 +24,4 @@ services:
       - FOSSOLOGY_DB_HOST=fossology-db
       - FOSSOLOGY_DB_NAME=fossology
       - FOSSOLOGY_DB_USER=fossy
+    user: fossy
```

bufferoverflow:
the problem is fo-postinstall.sh (needed for initialize the database) need root to start.
possible solutions:
* Skip root validation if there is only database initialization
* do the rest of initializations while building docker image (as root)

florianwittmann:
I have the same problem and also getting the
`chmod: changing permissions of '/var/run/postgresql': Operation not permitted`
with the standalone docker image without external postgres.

[...]

PR #758, opened 2017-03-08
fix(docker): change Dockerfile, docker run command

max-wittig:
remove 'USER fossy' from Dockerfile, /etc/init.d/postgresql cannot be started without being root

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,6 +82,5 @@ EXPOSE 8081
 
 ################################################################################
 VOLUME /srv/fossology/repository/
-USER fossy
 
 ENTRYPOINT ["/fossology/docker-entrypoint.sh"]
--- a/install/docker-compose.base.yml
+++ b/install/docker-compose.base.yml
@@ -24,3 +24,4 @@ services:
       - FOSSOLOGY_DB_HOST=fossology-db
       - FOSSOLOGY_DB_NAME=fossology
       - FOSSOLOGY_DB_USER=fossy
+    user: fossy


maximilianhuber:
This changes makes the containers run privileged again.

Maybe adding `user: fossy` to the `docker-compose.yml` might be a good replacement to the removed line. (see https://github.com/docker/compose/issues/1532)

bufferoverflow:
I just tested that change and I think we should merge.

maximilianhuber:
Tested for both, singleton- and compose-setup?

mcjaeger:
if review and test is OK, we should merge it, so we have a working docker single container that works for demo purposes when we tag 3.1

pguttmann:
"so we have a working docker single container that works for demo purposes when we tag 3.1" => +1

maximilianhuber:
This PR is related to issue #723

maximilianhuber:
Can you please add the following changes to the file `install/docker-compose.base.yml`:
```diff
diff --git a/install/docker-compose.base.yml b/install/docker-compose.base.yml
index b96e015..f97ebd6 100644
--- a/install/docker-compose.base.yml
+++ b/install/docker-compose.base.yml
@@ -24,3 +24,4 @@ services:
       - FOSSOLOGY_DB_HOST=fossology-db
       - FOSSOLOGY_DB_NAME=fossology
       - FOSSOLOGY_DB_USER=fossy
+    user: fossy
```

I have tested the compose setup with this changes and I would also suggest to merge it.

max-wittig:
updated the changes to `install/docker-compose.base.yml` proposed by @maximilianhuber

PR #759, opened 2017-03-09
chore(travis-ci): add docker test

max-wittig:


maximilianhuber:
- After you have started docker you could also use `curl` to check whether fossology is running
- you could also try the same with the docker-compose setup

.travis.yml
```diff
@@ -31,9 +31,14 @@ install:
...
  - sudo chown $(whoami) /var/local/cache/fossology
 
+sudo: required
+
+services:
+  - docker
+
 before_script:
  - cd src/
- - composer install --dev
```
maximilianhuber: Does all the php-tests run and work without the ` --dev` argument?

max-wittig: They should. --dev is deprecated. Dev packages are automatically installed, when `composer install` is run

max-wittig: I re-added the `--dev` option. The last test always fails without error log in Travis CI, even though it is completely unrelated to the changes I made

bufferoverflow:
travis says that used phpunit is only for php 7 ... see https://travis-ci.org/fossology/fossology/jobs/209661405#L835

maximilianhuber:
ping @steffen-weber

src/testing/docker/travis_docker_test.sh
```diff
@@ -3,4 +3,4 @@ docker build -t fossology/test .
 container_id=`docker run -d fossology/test`
 ip=`docker inspect --format '{{ .NetworkSettings.IPAddress }}' $container_id | awk "NR==1{print;exit}"`
 docker ps | grep -q fossology/test
-curl -L -I http://$ip:8081/repo | grep -q "200 OK"
\ No newline at end of file
+curl -L -I http://127.0.0.1:8081/repo/ | grep -q "200 OK"
```
maximilianhuber: Please check whether the page really appears. E.g. grep for some text on the homepage.

If the composer installation fails Apache still delivers an empty page successfully.

src/testing/docker/travis_docker_test.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+docker build -t fossology/test . 
+container_id=`docker run -d fossology/test`
+ip=`docker inspect --format '{{ .NetworkSettings.IPAddress }}' $container_id | awk "NR==1{print;exit}"`
+docker ps | grep -q fossology/test
+curl -L -I http://127.0.0.1:8081/repo/ | grep -q "200 OK"
```
maximilianhuber: Please check whether the page really appears. E.g. grep for some text on the homepage.

If the composer installation fails Apache still delivers an empty page successfully.

maximilianhuber:
I think you will need to squash the commits.

src/testing/docker/travis_docker_test.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+docker build -t fossology/test . 
+container_id=`docker run -p 8081:80 -d fossology/test`
+ip=`docker inspect --format '{{ .NetworkSettings.IPAddress }}' $container_id | awk "NR==1{print;exit}"`
+docker ps | grep -q fossology/test
+curl -L -I http://$ip:8081/repo/ | grep -q "200 OK"
```
maximilianhuber: You should also grep in the full response for some content expected on the fossology page. If the composer installation fails, apache still serves an empty page with response code 200

install/docker-compose.base.yml
```diff
@@ -24,3 +24,4 @@ services:
       - FOSSOLOGY_DB_HOST=fossology-db
       - FOSSOLOGY_DB_NAME=fossology
       - FOSSOLOGY_DB_USER=fossy
+    user: fossy
```
maximilianhuber: Why is this visible in the diff here? These changes are already merged. Maybe you should rebase on master or cleanup your branch.

max-wittig: it's not working yet. The only way to test travis-ci is to commit. No review necessary yet.
I will always cleanup, before this needs to be reviewed again.

src/testing/docker/travis_docker_test.sh
max-wittig: Apache serves 404, because /repo doesn't exist normally

install/docker-compose.base.yml
maximilianhuber: not if everything except composer is successful. Then are the basic files placed and the folders created but composer does not create the classloader and all dependencies.

src/testing/docker/travis_docker_test.sh
maximilianhuber: not if everything except composer is successful. Then are the basic files placed and the folders created but composer does not create the classloader and all dependencies.

PR #760, opened 2017-03-10
fix(unpacking): fix unpacking of mime-type application/java-archive

### LINKED ISSUE #711: Fossology not unpacking certain JAR files
When uploading artefacts that contain third party JARs Fossology is treating some of these JARs as binaries and is not unpacking them. This results in Fossology being unable to report on the licenses contained within them.

After investigating I believe this is due to the mime types of the JARs. Those with a mime type of 'application/zip' work fine and Fossology unpacks these and can analyse their contents, but those with a mime type of 'application/java-archive' do not appear to get unpacked and do not get analysed (mime type is checked with the command 'file --mime-type <JAR filename>').

I suspect this is due to this file https://github.com/fossology/fossology/blob/master/src/ununpack/agent/ununpack_globals.h only having a command for the mime type 'application/jar' defined and no command defined for the 'application/java-archive' mime type.

As an example, the following JARs do not get unpacked within Fossology when contained within the artefact that is uploaded:
http://central.maven.org/maven2/log4j/log4j/1.2.16/log4j-1.2.16.jar
http://central.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.8.4/jackson-databind-2.8.4.jar
http://central.maven.org/maven2/org/scala-lang/scala-library/2.11.8/scala-library-2.11.8.jar
http://central.maven.org/maven2/org/codehaus/groovy/groovy-all/2.4.4/groovy-all-2.4.4.jar

max-wittig:
Fixes #711


mcjaeger:
who tested it besides max?

bufferoverflow:
@mcjaeger I did so. Sorry for not using the review feature of GitHub, will do next time.

PR #761, opened 2017-03-10
fix(user-creation): email needs to be unique and required

### LINKED ISSUE #664: Many users can be created using same email
Tested :  7c7e3d3 

Create user in Admin -> Users -> Add 

With username "**user1**" and email "**user1@xcompany.com**"

now create another

With username "**user2**" and email "**user1@xcompany.com**"

Though user email is optional. i think email should be unique.

max-wittig:
this change requires email adresses, like usernames to be unique

Fixes #664

### DIFF
--- a/src/www/ui/user-add.php
+++ b/src/www/ui/user-add.php
@@ -85,12 +85,28 @@ class user_add extends FO_Plugin {
       $text = _("Passwords did not match. Not added.");
       return ($text);
     }
+    
+    if(empty($Email))
+    {
+      $text = _("Email must be specified. Not added.");
+      return ($text);
+    }
+
     /* Make sure email looks valid */
     $Check = preg_replace("/[^a-zA-Z0-9@_.+-]/", "", $Email);
+
     if ($Check != $Email) {
       $text = _("Invalid email address.  Not added.");
       return ($text);
     }
+    
+    /* Make sure email is unique */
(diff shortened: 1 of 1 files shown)


src/www/ui/user-add.php
```diff
@@ -85,12 +85,28 @@ function Add() {
...
     $Check = preg_replace("/[^a-zA-Z0-9@_.+-]/", "", $Email);
+
     if ($Check != $Email) {
       $text = _("Invalid email address.  Not added.");
       return ($text);
     }
+    
+    /* Make sure email is unique */
+    $email_count = $this->dbManager->getSingleRow("SELECT * FROM users WHERE user_email = '".$Email."';");
+    if ($email_count > 1) {
```
maximilianhuber: The return value of `getSingleRow` is an array. I think there is a missing `sizeof`

bufferoverflow: please use `select count(*) ...`

max-wittig: That's what I wanted to use in the first place, but it didn't work somehow. Gonna try that again.

maximilianhuber: You might have to unpack the singleton array containing the resulting integer.

shaheemazmalmmd:
i saw that in most of the cases people use email as "username".

**consider the following case:**

if the username of one user is "user1@example.com" and email is "user@example.com"

and the same user adds another user with credentials vice versa like

username : "user@example.com" and email : "user1@example.com" 

i think both username and email should be checked for unique.

max-wittig:
username is already validated and required to be unique

PR #764, opened 2017-03-13
fix(email-validation): apply correct email validation

max-wittig:
before nearly any string would count as correct email


src/www/ui/user-add.php
```diff
@@ -93,9 +93,8 @@ function Add() {
     }
 
     /* Make sure email looks valid */
-    $Check = preg_replace("/[^a-zA-Z0-9@_.+-]/", "", $Email);
-
-    if ($Check != $Email) {
+    if (!filter_var($Email, FILTER_VALIDATE_EMAIL)
```
fogninid: `'` is a valid character for emails: this introduces a SQL vulnerability when $Email is concatenated to build a SQL statement.

Replace string concatenation with a parametrized prepared statement, like is done a few lines later.

max-wittig:
Will fix. Thanks.

max-wittig:
potential SQL injection was fixed. Could you review again @fogninid ?

PR #767, opened 2017-03-14
fix(organize): sort admin content list alphabetically

max-wittig:


maximilianhuber:
same as #768. Maybe move it to a library?

max-wittig:
I don't know, if that's possible because of the template twig file.

max-wittig:
Moved method to tools.js as requested @maximilianhuber

max-wittig:
could you review again @maximilianhuber

PR #768, opened 2017-03-14
fix(upload): sort upload to reuse select alphabetically

### LINKED ISSUE #710: Upload from file > Reuse: Sort file list alphabetically
In the Upload > From File interface, the user can opt to reuse clearing results from another file. Under point 7, the user can select this file from a list of files displayed in a small window.

Currently, this selection method has has the following problems:
- The file list is not sorted in any recognizable way. I would prefer alphabetical sorting.
- The window is very small, so the user has to scroll a lot.

Ideally, the user would just start typing the file name, and with each character, a list of file starting with the corresponding string would be displayed. But alphabetical sorting would be a start.

max-wittig:
Fixes #710

### DIFF
--- a/src/reuser/ui/template/agent_reuser.js.twig
+++ b/src/reuser/ui/template/agent_reuser.js.twig
@@ -13,6 +13,7 @@ $('#{{ reuseFolderSelectorName }}').change(function () {
           option.value = key;
           packageForReuse.append(option);
         });
+        sortList('#{{ uploadToReuseSelectorName }} option');
       })
       .fail(failed);
 });


maximilianhuber:
same as #767. Maybe move it to a library?

max-wittig:
Moved method to tools.js as requested @maximilianhuber

max-wittig:
Moved method to tools.js as requested @maximilianhuber

maximilianhuber:
merge with #767? This PR does not work without it and is just a single line

max-wittig:
merged with PR #767

PR #769, opened 2017-03-17
fix(ecc): resolved error in apache-errorlog display while handling ecc

ag4ums:
resolve error in error-log for browsing ecc


max-wittig:
Could you elaborate how to reproduce the error-log, while browsing ECC? I'm not getting any errors in `/var/log/apache2/error.log`

shaheemazmalmmd:
Install Master and then

-> go to single file view of ECC for any file where finding exists.
-> view multiple files with ECC findings using page navigator.
-> now open  /var/log/apache2/error.log you should see some warnings.

Now install contrib/fix-ecc-error-log

follow the same procedure as above. and check the log

Thank You

PR #774, opened 2017-03-20
feat(organize): allow moving/copying of multiple files/folders at the same time

max-wittig:
this feature is a big time-saver for organizing files

### DIFF
--- a/src/www/ui/template/admin_content_move.html.twig
+++ b/src/www/ui/template/admin_content_move.html.twig
@@ -22,8 +22,8 @@
       <form name="formy" method="post">
         <input type="hidden" name="fromFolder" id="fromFolder" value="-1"/>
 
-        <label for="foldercontent">{{ 'Select the upload or folder you wish to move'|trans }}:<br/>
-          <select name="foldercontent" id="foldercontent" size="20">
+        <label for="foldercontent">{{ 'Select uploads or folders you wish to move'|trans }}:<br/>
+          <select name="foldercontent[]" id="foldercontent" multiple="multiple" size="20">
           </select>
         </label>
         <br/>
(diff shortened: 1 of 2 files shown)


max-wittig:
Could somebody review this PR?

src/www/ui/template/admin_content_move.html.twig
```diff
@@ -22,8 +22,8 @@
       <form name="formy" method="post">
         <input type="hidden" name="fromFolder" id="fromFolder" value="-1"/>
 
-        <label for="foldercontent">{{ 'Select the upload or folder you wish to move'|trans }}:<br/>
-          <select name="foldercontent" id="foldercontent" size="20">
+        <label for="foldercontent">{{ 'Select uploads or folders you wish to move'|trans }}:<br/>
+          <select name="foldercontent[]" id="foldercontent" multiple="multiple" size="20">
```
maximilianhuber: What do the square brackets mean?

max-wittig: Multiselect. Otherwise it doesn't work. Try it without. 
I think it's a PHP specific thing

max-wittig: Found a stackoverflow post, but nothing specifiy as to why it is needed.
http://stackoverflow.com/questions/30819489/select-multiple-not-working

maximilianhuber: Yes

> They address a limitation of PHP, which doesn't generate an array automatically if multiple values with the same name are submitted, [...]

max-wittig: so this functionality isn't working as it appears to work?
Multiple select is also possible, but there are no square brackets[]. Is it even sending all selected files?
That's another issue, but maybe a potenial bug. 
image

PR #775, opened 2017-03-21
Dev/select2 upload reuse

max-wittig:
- add select2 javascript libary, which enables the use of a search box on the File --> Upload File Page 
- Update jQuery and jQuery UI to the latest version
  - I've done some manual UI testing and everything seems to work as before with the updated libs
  - more testing may still be necessary by other contributors, who are more familiar with the extended 
    features of FOSSology

## Current implementation:

reuse-before

-----------------------------------------------

## New implementation:
reuse-after2
reuse-after


src/www/ui/template/include/foot.html.twig
```diff
@@ -4,6 +4,6 @@
    are permitted in any medium without royalty provided the copyright notice and this notice are preserved.
    This file is offered as-is, without any warranty.
 #}
-<script src="scripts/jquery-1.11.1.min.js" type="text/javascript"></script>
+<script src="scripts/jquery-3.2.0.min.js" type="text/javascript"></script>
```
maximilianhuber: Have you tested that this change is compatible with all our functionality?

This might break something or change the behavior of some code. Please also check that the browser console does not log any new errors or exceptions.

max-wittig: I opened everything, did many common actions I know and checked the browser console all the time. But I'm not yet familiar with all the functionality fossology has to offer. That's why more people should test this change as described above.

maximilianhuber: Sorry. I have not read the description.

shaheemazmalmmd:
Tool tip UI seems to be broken after the changes. please fix 

tooltipnotvisible

max-wittig:
Thanks for testing. What browser are you using?

shaheemazmalmmd:
I have found this in both Mozilla and Chrome browsers.

max-wittig:
Issue is fixed. I forgot to actually also update the jQuery UI css file.

mcjaeger:
@max-wittig  could you pls. where are the gif files from? it is part of jquery right?

max-wittig:
the png files are part of jQuery UI

maximilianhuber:
@shaheemazmalmmd have you tested this PR in depth?

max-wittig:
I'm gonna run the UI tests with the updated jQuery

mcjaeger:
@maximilianhuber do you consider your comments as review?

mcjaeger:
I tested it with some uploads, I noticed a lag for letting new uploads appear, might be also my slow computer -> OK.

maximilianhuber:
@mcjaeger sorry for my late response. My comment were no complete review but the added code looks good.

PR #781, opened 2017-03-30
fix(advice-licenses): fix public notes not being saved, when creating/editing a license

max-wittig:
Before in the advice_license mod, public notes were not saved at all. 
image


mcjaeger:
ehem, description please?

max-wittig:
description added

PR #787, opened 2017-03-31
WIP: chore(testing): add E2E selenium tests

max-wittig:
### Selenium Tests with Node and Protractor
Tests found already a few issues.
3 of which I already fixed in these PR: 

- PR #780
- PR #781
- PR #782

#### 10 sec section of the 150 seconds test
9_74iq

PR #788, opened 2017-04-03
fix(folder-deletion): don't delete duplicate files in other folder, when folder is deleted

max-wittig:
To reproduce issue:
1. Create a new Folder with Name "TestFolder"
2. Upload file (in this example test.zip)
3. Copy test.zip into TestFolder -> test.zip should now exist in the Software Repository and in the folder TestFolder
4. Delete TestFolder
Expected Behavior: test.zip still exists in the Software Repository, because it was only copied from there in TestFolder before
Actual Behavior: test.zip is gone from the Software Repository
This commit should fix this issue. Testing from more contributors is still necessary

This issue was found using the E2E Tests in PR #787

### DIFF
--- a/src/delagent/agent/util.c
+++ b/src/delagent/agent/util.c
@@ -559,9 +559,9 @@ int listFoldersRecurse (long Parent, int Depth, long Row, int DelFlag, int user_
     return 1;
   }
 
-  /* Find all folders with this parent and recurse */
+  /* Find all folders with this parent and recurse, but don't show uploads, if they also exist in other directories */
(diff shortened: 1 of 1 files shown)


max-wittig:
Already fixed internally

max-wittig:
Hey @shaheemazmalmmd I saw that you solved this issue another way internally. What would you suggest would be the best way to get the fix for this issue upstream? Either with your changes or these changes.

shaheemazmalmmd:
Please remove warnings 

warningdelagent

max-wittig:
could you also compare your fix from the NG version? Maybe it solves this issue better?

shaheemazmalmmd:
this commit is included in https://github.com/fossology/fossology/pull/830

PR #791, opened 2017-04-06
fix(docker): revert docker and docker-compose to older, working state

max-wittig:
As coordinated with @bufferoverflow and @mcjaeger 
This also enables the use of E2E selenium tests with Travis CI later on...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 # FOSSology Dockerfile
 # Copyright Siemens AG 2016, fabio.huser@siemens.com
-# Copyright TNG Technology Consulting GmbH 2016, maximilian.huber@tngtech.com
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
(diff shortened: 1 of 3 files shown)


mcjaeger:
let's keep ninka testing otherwise we need to remove the ninca wrapper code from the repo as well.
@ag4ums has developed a fix for that
 
>   # TO resolve ninka build dependency
>   - git clone --branch release-1.1104 https://github.com/dagolden/IO-CaptureOutput.git
>   - cd IO-CaptureOutput && perl Makefile.PL && make && make test && make install && cd ..
>   - rm -rf IO-CaptureOutput

let's use this.

mcjaeger:
@maximilianhuber  please let us know in case you would like to have some time review or work on this.

maximilianhuber:
@mcjaeger yes, that would be great. I will try to fix the bug this weekend to prevent the rollback of the docker setup.

bufferoverflow:
The big question here is , when shall the release happen. @max-wittig has some time to make this ready today and add the ninka fix, mission complete.

We can always add the other stuff again if @maximilianhuber is successful at the weekend.

bufferoverflow:
Have this ready and integrate #787 is much more important from my perspective.

mcjaeger:
integrate #787 into docker and travis-ci? or only travis-ci?

if maximilian asks for few days it is just fair to wait.

maximilianhuber:
Please have a look at the PR https://github.com/fossology/fossology/pull/793 which fixes the problem by reverting the root Dockerfile (the same way as this PR does). But I do not revert the compose setup, instead I move the complex Dockerfile to `install/docker-compose.Dockerfile`. Thus we still have a setup which runs as non privileged user.

maximilianhuber:
The commit 2d8474c is not related to docker. Please move it to its own PR.

maximilianhuber:
This breaks the tests. Please fix the following
```
RROR: Can't create '/usr/local/man/man3'
Do not have write permissions on '/usr/local/man/man3'
```
(after moving it to its own PR)

max-wittig:
How to fix that?

maximilianhuber:
you might need to add `sudo` to be able to do the `make install`

max-wittig:
replaced by PR #793

PR #770, opened 2017-03-17
chore(travis): add docker test

max-wittig:
fossology build into container and run

selenium E2E tests are coming soon

PR #793, opened 2017-04-07
fix(docker): use a simpler Dockerfile for standalone build

maximilianhuber:
The modified Dockerfile did not work with the split-up docker-compose
setup. This moves the "complicated" dockerfile to
`install/docker-compose.Dockerfile` and places the old Dockerfile at the
root level.

TODO: merge these dockerfiles again

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 # FOSSology Dockerfile
 # Copyright Siemens AG 2016, fabio.huser@siemens.com
-# Copyright TNG Technology Consulting GmbH 2016, maximilian.huber@tngtech.com
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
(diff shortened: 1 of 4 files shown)


maximilianhuber:
This is a potential replacement for #791

max-wittig:
The standalone docker image doesn't work
<pre>Starting FOSSology job scheduler: FATAL scheduler.c.542: could not find group "fossy"
FATAL set_usr_grp() aborting due to error: Success </pre>

maximilianhuber:
I am checking it. I thought that the file `/Dockerfile` should be exactly the one used in #791

maximilianhuber:
Yes, you are right, one also has to split the entrypoint script.
Will be fixed in a few seconds

maximilianhuber:
should be fixed

max-wittig:
I'm gonna test the standalone docker image again.

Dockerfile
```diff
@@ -10,77 +9,32 @@
...
-        --scheduler-only
+RUN /fossology/install/scripts/install-spdx-tools.sh
 
-RUN set -x \
- && mkdir -p /var/log/fossology \
- && chown -R fossy:fossy /var/log/fossology \
- && chgrp -R fossy /usr/local/etc/fossology/ \
- && chmod -R g+wr /usr/local/etc/fossology/ \
- && chown fossy:fossy /usr/local/etc/fossology/Db.conf
+RUN /fossology/install/scripts/install-ninka.sh
```
max-wittig: Don't include ninka in the docker container, as it doesn't work currently.

maximilianhuber: That is an independent problem and is not related to the docker setup.

max-wittig: It kinda is, because then docker is installing a broken dependency into FOSSology. If you then try to upload a file and check the ninka checkbox, it will fail 100%

maximilianhuber: This problem will be solved by https://github.com/fossology/fossology/pull/799

max-wittig: #798 fixes just travis, but combined with your PR #799 it should work.

maximilianhuber: have missed that travis was not using the `install-ninka.sh` script. this is now fixed in #799

maximilianhuber:
@max-wittig @mcjaeger @bufferoverflow : here are all tests green and the pr is approved. I think this one can be merged and then can @max-wittig rebase #770.

PR #796, opened 2017-04-10
This PR is just created to test travis together with the simpler docker setup

maximilianhuber:


maximilianhuber:
This PR should not be created here, sorry.

PR #803, opened 2017-04-18
fix(spdx2): add files with no license found to generated output format

shaheemazmalmmd:


mcjaeger:
please provide a description!

mcjaeger:
I have installed the branch on http://52.26.97.143/repo/ (login testuser/test) seems to work, but requires further testing

src/spdx2/agent/spdx2.php
```diff
@@ -436,8 +436,12 @@ protected function addScannerResults(&$filesWithLicenses, ItemTreeBounds $itemTr
     {
       $reportedLicenseId = $this->licenseMap->getProjectedId($row['rf_fk']);
       $shortName = $this->licenseMap->getProjectedShortname($reportedLicenseId);
-      if ($shortName != 'No_license_found' && $shortName != 'Void') {
-        $filesWithLicenses[$row['uploadtree_pk']]['scanner'][] = $shortName;
+      if ($shortName != 'Void') {
+        if($shortName != 'No_license_found'){
+          $filesWithLicenses[$row['uploadtree_pk']]['scanner'][] = $shortName;
+        }else{
+          $filesWithLicenses[$row['uploadtree_pk']]['scanner'][] = "";
```
maximilianhuber: If the license secanners do not find any licenses, it might correct to return `"NONE"`

src/spdx2/agent/spdx2utils.php
```diff
@@ -63,7 +63,11 @@ static public function addPrefixOnDemand($license, $spdxValidityChecker = null)
       }
       else
       {
-        return self::$prefix . $license;
+        if(!empty($license)){
+          return self::$prefix . $license;
+        }else{
+          return "NOASSERTION";
```
maximilianhuber: I think it would be more consistent to call the function with `addPrefixOnDemand("NOASSERTION")` instead of an empty string and add
```
    if($license === "NOASSERTION" || $license === "NONE"){
        return $license;
    }
```
to the top of `addPrefixOnDemand`.

maximilianhuber: But you can still leave this here as default for an empty string

src/spdx2/agent/spdx2.php
maximilianhuber: @mcjaeger @kestewart : is it correct to state `LicenseInfoInFile: NONE` if the scanners were not able to find any license? or should we act defensively and say `NOASSERTION`?

kestewart: @maximilianhuber 
- if the scanners aren't able to find any license then `LicenseInfoInFile: NONE` is accurate.
It is ok to say `NOASSERTION`, but the preference would be to provide the information that no information was found in the file by using `NONE`.  see comment on issue

mcjaeger: I like the `NOASSERTION` more because it represents the same rationale as with FOSSology and its "no license known": actually there might be a license, but the scanner did not find it. As with `NONE` it sounds like there is really nothing there.

src/spdx2/agent/spdx2utils.php
mcjaeger: but is there a chance that `addPrefixOnDemand(...` will add a `LicenseREf-` prefix to `NOASSERTION` then?

maximilianhuber: Not if you add the suggested `if($license === "NOASSERTION"...` to the top of the function `addPrefixOnDemand`

mcjaeger: Ah, Ok got it. reading is an advantage ...

mcjaeger:
retested with two packages, checking numbers of entries -> all OK.

PR #805, opened 2017-04-18
chore(docs): updating readme and changelog, consolidation of license information

mcjaeger:
updated readme, screenshots and changelogs, removed license file and added content to the readme file, but added notice file instead (also knows as README_OSS in some organisations), also adding spdx tag value file. checked with markdown editor.


mcjaeger:
actually, it seems like github is automatically proposing users for reviewing...

PR #809, opened 2017-05-01
fix(licenseref): changing shortname of 3DFX license to 'Glide'

### LINKED ISSUE #744: Invalid 3DFX license info in SPDX:RDF file
When scanning a file containing the 3DFX Glide license text (https://spdx.org/licenses/Glide.html), both Nomos and Monk correctly label the license as 3DFX. This can be seen in `?mod=view-license`. When an SPDX:RDF file is generated, the following is written to the file:

`<spdx:licenseInfoInFile rdf:resource="http://spdx.org/licenses/3DFX"/>`

This resource is invalid - it should be `https://spdx.org/licenses/Glide`. This causes problems when analyzing the file with spdx tools, as it does not recognize the license. Manually editing the RDF file to use the correct resource link fixes the problem.

Tested on: Version: [3.1.0rc2], Branch: [master], Commit: [#3b080b]

timurphy:
Fixes #744. SPDX:RDF output now contains the correct license ref:
`<spdx:licenseInfoInFile rdf:resource="http://spdx.org/licenses/Glide"/>`


max-wittig:
Code looks good and seems to work, but maybe you could also change it in the `Bulk Scan` and `User Decision` Section of the License Browser, while you're at it. That would make it consistent throughout the whole project.
image
image

timurphy:
Hi @max-wittig, this is showing up as Glide on my side:

max-wittig:
That is weird. Tried using another browser and clearing the caches. Still Glide doesn't seem to be in the list for bulk scan and user decision. 3DFX still is though.
I just ran `sudo make install` and also tried running the fo-postinstall script again. Did I forget something?

image

timurphy:
Are you running this as a fresh install? I'm assuming `install/db/licenseref.sql` is run as part of the initial setup and (also assuming) bulk scan and user decision pull these tags from the database.

If you run a `SELECT rf_shortname FROM license_ref WHERE rf_pk=199`, does it give you 3DFX or Glide?

shaheemazmalmmd:
Changes will create duplicate license for existing fossology instances. 

I think the update query fails.

3dfxandglide

mcjaeger:
actually, if we cannot upgrade existing instances, this is a no go then.

timurphy:
I think I've found the problem. `fossinit.php` assumes `rf_shortname` is the identifying key for a license, so changing this value results in a duplicate record being created: https://github.com/fossology/fossology/blob/34cc66a0508e5f1855f15df6339db0c074ee916f/install/fossinit.php#L413

This entire section ignores `rf_pk`, including when inserting new licenses. This is the true primary key and should be used as such.

It will take a couple of weeks for me to test this properly. Watch this space.

timurphy:
The latest commit resolves the licenseref updating issue. I've done a bit of refactoring here as there were a few issues with the current implementation:

1. Use `rf_pk` as the primary key instead of `rf_shortname`, which was partially to blame for the duplication issue mentioned above.
2. Update queries were using hard-coded field names, with some fields missing (e.g. `rf_pk`). This meant that some fields were not updated. Changed to dynamically generate the query based on the actual fields in the table.
3. Update queries were open to SQL injection vulnerabilities. Changed to use parameterised queries.
4. DbManager did not allow `queryOnce` to take a parameter list. Added this option.

I've run a number of tests using both blocks of modified code (update and insertion of licenses), and this appears to be working well. I would welcome some extra testing though.

Note that the changes to the Sqlite `query` method could not be tested, as I can't see how to configure FOSSology to use this DBMS. If someone could test this or point me in the right direction, that would be appreciated.

mcjaeger:
I tested it stand alone / fresh install -> OK, but I think we need to test an update of an existing instance.

steffen-weber:
I suggest testing with an >=30d old existing instance, i.e., before rf_pk=571, rf_name='RSA-Cryptoki' was inserted. Then add a new license that probably get automatically rf_pk=571. Now updating the instance and check if the new license still exists.

timurphy:
@steffen-weber this will cause the newly-created license to be overwritten, as it will have the same `rf_pk` as `RSA-Cryptoki`. Which is not good.

I'm not sure if this is fixable given the current design. Although `rf_pk` is set in `licenseref.sql`, it's not actually used as the license identifier - `rf_shortname` seems to have this role. This means that `rf_pk` may be different across installations for the same license, and also means `rf_shortname` cannot reliably be modified by `fossinit.php`. I can't think of a reliable way of fixing this without breaking existing installations.

steffen-weber:
Overwriting user's licenses sounds bad. How about this strategy?
* insert Glide as duplicated license as your first commit did via initLicenseRefTable()
* update the 3DFX text, name
* delete newly inserted Glide license
* update 3DFX shortname to Glide
* increase `$sysconfig['Release']`

The `$sysconfig['Release']` in `fossinit.php` should be useful to determine if migration step is required.

timurphy:
That could work. Do we have a `patches.php` script or similar which we could put this in? Adding this logic to `fossinit.php` could make that script messy in the future.

mcjaeger:
@timurphy no problem with keeping this PR open, but let us know if you feel like you are going ahead with this or not. I think changing the SPDX conformant Id to glide is an important fix.

timurphy:
@mcjaeger sorry, I haven't had time to make the changes. I'll try to get a PR in for this in the next few weeks.

shaheemazmalmmd:
@timurphy : Can you please rebase this branch with master (merge conflicts)

timurphy:
I've made the changes and tested on a fresh install, upgrade on system having rf_shortname '3DFX' and upgrade on system having rf_shortname 'Glide'. I'm not familiar with the way `$sysconfig['Release']` is used so please review and let me know if I've done that wrong.

steffen-weber:
code looks good to me

PR #811, opened 2017-05-01
fix(bulk-scan): don't schedule bulk scan, if no license is selected by user

### LINKED ISSUE #641: Add warning when bulk scanning with no license selected (for addition or removal)
When doing the bulk scans it may happen that users do a bulk scan and forget to select the license, because the see the license in the pop up menue already.

A warning would be required to tell the user that no license was actually selected (by pressing plus or minus) and the bulk scan would be useless.

In the screenshot: Apache-2.0 is selected an a users perception could be that everything is all right to kick-off the bulk scan. But it is not, because the table "action"/"license" has not entry.

036 oss fossyng 20160223 screenshot bulk ui improvement idea warning

max-wittig:
the user is shown an alert dialog, to prevent useless bulk scans
If there is a better way to inform the user about this error, please let me know

Fixes #641

### DIFF
--- a/src/www/ui/scripts/change-license-common.js
+++ b/src/www/ui/scripts/change-license-common.js
@@ -109,10 +109,36 @@ function scheduledDeciderError (responseobject, resultEntity) {
   resultEntity.show();
 }
 
+function isUserError(bulkActions, refText) {
+    var errorText = "";
+    if(bulkActions.length < 1) {
+        errorText += "No licenses to bulk scan selected\n";
+    }
+
+    if(refText.trim().split(" ").length < 2) {
+        errorText += "Reference text needs to be at least 2 words long"
+    }
+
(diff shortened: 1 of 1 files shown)


max-wittig:
bulk reference text is now also required to contain at least 2 words. Thanks @shaheemazmalmmd

DineshDevarajan:
I need to use bulk reference text as "MODULE_LICENSE("GPL");" which is a single word and as this feature has made the reference text to at least 2 words I am unable to run the scan as it results in an error "Reference text needs to be at least 2 words long". 

I understand when the length is 0 it should not run. So Would like to know why the reference text is restricted for length=1.

krutztq:
created a new ticket: https://github.com/fossology/fossology/issues/1654

Issue #754, opened 2017-03-03
Nomos regexp error


AMDmi3:
There seem to be problem with specific nomos regexp:

```
Compile failed, regex #706
regex = "program is linked with (|and uses) microsoft distributable code"
FATAL nomos_regex.c.49: regcomp failure: empty (sub)expression
```

mcjaeger:
thanks for pointing out, in order to cover it could you please help and provide us with more details when you have encountered this error?

AMDmi3:
FreeBSD 11.0 amd64, nomos standalone build, pcre 8.4. The problem is reproducible when running nomos on a text file with single line:

```
program is linked with microsoft distributable code
```

I think the problem is obvious from regexp definition. There also seem to be a double space problem (e.g. the rule matches ```with<space>and<space>uses<space>microsoft``` and ```with<space><space>microsoft```, but not ```with<space>microsoft```)

The problems is fixed for me by

```
-%STR%  "program is linked with (|and uses) microsoft distributable code"
+%STR%  "program is linked with( | and uses )microsoft distributable code"
```

mcjaeger:
actually, for the examples above it is fixed:

https://github.com/fossology/fossology/pull/817

but apparently another similar case has been found:

https://github.com/fossology/fossology/issues/822

AMDmi3:
> actually, for the examples above it is fixed

It isn't. After the mentioned change there still is an empty subexpression.

max-wittig:
but didn't you open #822 for the same issue? Or are these different issues?

AMDmi3:
The same as I've originally submitted, but not the same that was fixed here in 0b8e58b.

max-wittig:
I get that. But is it correct to close one if the issues you've created for this problem?

AMDmi3:
Yes, let's leave this one closed and #822 active, with additional examples.

PR #817, opened 2017-05-03
fix(nomos): issue #754, regexp error

maximilianhuber:
This PR fixes issue #754. The solution was already suggested in #754 and on the fossology-devel mailinglist by 

as suggested in #754 and by dirkx@... on the mailing list


shaheemazmalmmd:
Code looks good. works as expected

PR #818, opened 2017-05-05
chore(backup): add s3 backup script

max-wittig:
Wiki page: [S3-Backup](https://github.com/fossology/fossology/wiki/S3-Backup)

### DIFF
--- a/utils/backup/S3/fo-backup-s3
+++ b/utils/backup/S3/fo-backup-s3
@@ -1,0 +1,50 @@
+#!/bin/bash
+# Copyright (C) Siemens AG 2017
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# version 2 as published by the Free Software Foundation.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source ${DIR}/before
+set -x
+
+if [[ ${1} = "-l" ]]; then
+  LOGGING=${2}
+  if [[ -z "${LOGGING}" ]]; then
+    echo "Please specify a file, where to save the log file to"
+    exit 1
+  else
+    exec &> >(tee -a ${LOGGING})
+  fi
+fi
+
+${DIR}/install
+
+echo Starting backup at `date`
+echo Creating diff of repository at `date`
+find /srv/fossology/repository/ -type f -exec du -h {} \; | sort -n | ${DIR}/venv/bin/aws s3 cp --sse AES256 - ${AWS_DIFF_LOCATION}
+
+echo Start database dump at `date`
+#### postgresql needs to be restarted for the dump to work
+/etc/init.d/postgresql restart
+#### these steps are split, because of permissions
+su - postgres -c 'pg_dumpall' > ${TMP_FILE}
+cp ${TMP_FILE} /srv/fossology/database.txt
+echo Done database dump. Starting Upload at `date`
+
+#### sync repository
+echo "Uploading and starting backup to S3..."
+tar -zcf - /srv/fossology/ | ${DIR}/venv/bin/aws s3 cp --expected-size 200000000000 --sse AES256 - ${AWS_BACKUP_LOCATION}
+
+echo "Uploaded backup as ${FILENAME} to the S3 Bucket ${BUCKET}"
+echo Backup script done at `date`
+${DIR}/after
+rm ${TMP_FILE}
--- a/utils/backup/S3/fo-restore-s3
+++ b/utils/backup/S3/fo-restore-s3
@@ -1,0 +1,81 @@
+#!/bin/bash
+# Copyright (C) Siemens AG 2017
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# version 2 as published by the Free Software Foundation.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source ${DIR}/before
+set -x
+
+if [[ ${1} = "-l" ]]; then
+  LOGGING=${2}
+  if [[ -z "${LOGGING}" ]]; then
+    echo "Please specify a file, where to save the log file to"
+    exit 1
+  else
+    exec &> >(tee -a ${LOGGING})
+  fi
+fi
+
+echo Starting restore at `date`
+
+#### stop fossology scheduler
+echo "Stopping fossology scheduler..."
+/etc/init.d/fossology stop
+
+#### stop apache2
+echo "Stopping apache2..."
+/etc/init.d/apache2 stop
+
+echo "Start removing old data at `date`. Starting in 10 seconds..."
+sleep 10
+${DIR}/../../fo-cleanold --delete-repository --delete-database
+
+#### sync full tar
+echo Done removing old data. Starting download+extracting of "${FILENAME}" at `date`
+${DIR}/venv/bin/aws ${ENDPOINT} s3 cp --expected-size 200000000000 --sse AES256 "${AWS_BACKUP_LOCATION}" - | tar --warning=no-timestamp -xzf - -C /
+
+echo Checking diff of repository at `date`
+AWS_TMP_DIFF_LOCATION=`mktemp`
+CURRENT_TMP_DIFF_LOCATION=`mktemp`
+
+#### Copy diff from S3 to tmp
+${DIR}/venv/bin/aws s3 cp --sse AES256 "${AWS_DIFF_LOCATION}" "${AWS_TMP_DIFF_LOCATION}"
+find /srv/fossology/repository/ -type f -exec du -h {} \; | sort -n > "${CURRENT_TMP_DIFF_LOCATION}"
+echo 'Comparing diff of repository...'
+
+if cmp -s "${CURRENT_TMP_DIFF_LOCATION}" "${AWS_TMP_DIFF_LOCATION}"
+then
+  echo "Repository diff check successful. Continuing..."
+else
+  echo Diffs are different!!!
+  echo Continuing anyway...!
+fi
+
+echo Done download of full archive. Start restoring database at `date`
+mv /srv/fossology/database.txt "${TMP_FILE}"
+chmod 777 "${TMP_FILE}"
+
+/etc/init.d/postgresql restart
+
+if [[ -n "${LOGGING}" ]]; then
+  echo --------- Start database log ---------
+  su - postgres -c "psql -f ${TMP_FILE}"
+  echo --------- End database log ---------
+else
+  su - postgres -c "psql -f ${TMP_FILE}" > /dev/null 2>&1
+fi
+
+echo Done restoring database. Restore script done at `date`
+${DIR}/after
+rm ${TMP_FILE}
+rm ${CURRENT_TMP_DIFF_LOCATION}
+rm ${AWS_TMP_DIFF_LOCATION}
--- a/utils/backup/S3/after
+++ b/utils/backup/S3/after
@@ -1,0 +1,16 @@
+#!/bin/bash
+# Copyright Siemens AG 2017
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+#
+
+#### start apache2
+echo "Starting apache2..."
+/etc/init.d/apache2 start
+
+#### start fossology scheduler
+echo "Starting fossology scheduler..."
+/etc/init.d/fossology start
--- a/utils/backup/S3/before
+++ b/utils/backup/S3/before
@@ -1,0 +1,53 @@
+#!/bin/bash
+# Copyright Siemens AG 2017
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+#
+
+TMP_FILE=`mktemp`
+
+if [[ $(id -u) -ne 0 ]]; then
+ echo "Please run as root"
+ exit 1
+fi
+
+if [[ -z "${BUCKET}" ]]; then
+ echo "Please set the BUCKET variable to the name of your S3 Bucket"
+ exit 1
+fi
+
+### Filename to restore from. Hostname is used, if none set
+if [[ -z "${FILENAME}" ]]; then
+ FILENAME=${HOSTNAME}.tar
+elif [[ "${FILENAME}" != *tar ]]; then
+ FILENAME=${FILENAME}.tar
+fi
+
+### for debugging e.g. local AWS S3 Bucket
+if [[ -n "${ENDPOINT}" ]]; then
+  ENDPOINT="--"${ENDPOINT}
+else
+  ENDPOINT=""
+fi
+
+${DIR}/install
+
+#### check for AWS config file
+if [[ ! -f ~/.aws/config ]]
+then
+	${DIR}/venv/bin/aws configure
+fi
+
+AWS_BACKUP_LOCATION="s3://${BUCKET}/${FILENAME}"
+AWS_DIFF_LOCATION="${AWS_BACKUP_LOCATION}-repo-diff.txt"
+
+#### stop fossology scheduler
+echo "Stopping fossology scheduler..."
+/etc/init.d/fossology stop
+
+#### stop apache2
+echo "Stopping apache2..."
+/etc/init.d/apache2 stop
--- a/utils/backup/S3/install
+++ b/utils/backup/S3/install
@@ -1,0 +1,18 @@
+#!/bin/bash
+# Copyright Siemens AG 2017
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+#
+# Description: cleans installs requirements for the backup and restore
+
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+#### install dependencies
+apt-get install -y python3-pip python3.4-venv libyaml-dev
+python3 -m venv venv
+${DIR}/venv/bin/pip3 install awscli


max-wittig:
Successfully tested on dev-fossology

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+if [[ ${1} = "-l" ]]; then
+ LOGGING=${2}
+ if [ -z ${LOGGING} ]; then
+  echo "Log dir not set."
+  exit 1
+ fi
+fi
+
+if [ -z ${BUCKET} ]; then
+ echo "Bucket env variable needs to be set"
```
maximilianhuber: Please mention the correct name of the missing variable in the echo output

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+
+if [[ $(id -u) -ne 0 ]]; then
+ echo "Please run as root"
+ exit 1
+fi
+
+if [[ ${1} = "-l" ]]; then
+ LOGGING=${2}
+ if [ -z ${LOGGING} ]; then
+  echo "Log dir not set."
```
maximilianhuber: Please mention the correct name of the missing variable in the echo output

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+if [[ $(id -u) -ne 0 ]]; then
+ echo "Please run as root"
+ exit 1
+fi
+
+if [[ ${1} = "-l" ]]; then
+ LOGGING=${2}
+ if [ -z ${LOGGING} ]; then
```
maximilianhuber: it would be better to write `[ -z "${LOGGING}" ];` which is best practice (AFAIK).

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+# modify it under the terms of the GNU General Public License
+# version 2 as published by the Free Software Foundation.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
```
maximilianhuber: missing `set -x` in this script.

utils/backup/S3/clean
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Copyright Siemens AG 2017
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+#
+# Description: cleans tmp files
+
```
maximilianhuber: missing (but potentially unnecessary) `set -x` in this script.

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+./install
+
+#### check for AWS config file
+if [ ! -f ~/.aws/config ]
+then
+	venv/bin/aws configure
+fi
+
+if [ -n "${LOGGING}" ]; then echo Starting backup at `date` >> ${LOGGING}; fi
```
maximilianhuber: pleas use a consistent style of formatting for if statements

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+# version 2 as published by the Free Software Foundation.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+if [[ $(id -u) -ne 0 ]]; then
```
maximilianhuber: please be consistent with your uses of `[` and `[[`

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+echo "Stopping apache2..."
+/etc/init.d/apache2 stop
+
+#### copy database to /srv/fossology
+echo "Backing up database..."
+/etc/init.d/postgresql restart
+
+if [ -n "${LOGGING}" ]; then echo Creating diff of repository at `date` >> ${LOGGING}; fi
+echo "Creating repo diff..."
+ls -RlahGg  --time-style='+' /srv/fossology/repository/ | cut -d ' ' -f 2-5 | venv/bin/aws s3 cp --sse AES256 - s3://${BUCKET}/${FILENAME}-repo-diff.txt
```
maximilianhuber: it is best practice to add quotes around  `s3://...diff.txt` to prevent problems with spaces in the inserted variables.

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+if [[ $(id -u) -ne 0 ]]; then
+ echo "Please run as root"
+ exit 1
+fi
+
+if [[ ${1} = "-l" ]]; then
```
maximilianhuber: please add a help function which describes how the command should be called and what parameters are expected

utils/backup/S3/fo-backup-s3
maximilianhuber: You are just logging the headings of the steps, this is not very helpful.

maybe consider to use some trick like:
```bash
if [ -n "${LOGGING}" ]; then
  exec &> >(tee -a $LOGGING)
fi
```
and then just `echo "Starting backup at $(date)"`

utils/backup/S3/fo-restore-s3
```diff
@@ -0,0 +1,108 @@
...
+venv/bin/aws ${ENDPOINT} --debug s3 cp --expected-size 200000000000 --sse AES256 s3://${BUCKET}/${FILENAME} - | tar --warning=no-timestamp -xzf - -C /
+
+if [ -n "${LOGGING}" ]; then echo Checking diff of repository at `date` >> ${LOGGING}; fi
+DIFF=`venv/bin/aws s3 cp --sse AES256 s3://${BUCKET}/${FILENAME}-repo-diff.txt -`
+CURRENT_SUM=`ls -RlahGg  --time-style='+' /srv/fossology/repository/ | cut -d ' ' -f 2-5`
+echo 'Comparing diff of repository...'
+
+if [ "${DIFF}" != "${CURRENT_SUM}" ]; then
+  echo Diff are different!!!
+  echo Diff are different!!! >> ${LOGGING}
```
maximilianhuber: This does not work if `$LOGGING` is unset

utils/backup/S3/fo-restore-s3
```diff
@@ -0,0 +1,108 @@
...
+# modify it under the terms of the GNU General Public License
+# version 2 as published by the Free Software Foundation.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
```
maximilianhuber: the `set -x` is missing

maximilianhuber: some of the comments for `fo-backup-s3` are also applicable here, maybe merge the scripts to reduce duplicated code?

utils/backup/S3/fo-backup-s3
maximilianhuber: As far as i see do the scripts fail if they are called from another directory, e.g. via `utils/backup/S3/fo-backup-s3`

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+
+#### sync repository
+echo "Uploading and starting backup to S3..."
+tar -zcvf - /srv/fossology/ | venv/bin/aws --debug s3 cp --expected-size 200000000000 --sse AES256 - s3://${BUCKET}/${FILENAME}
+
+if [ -n "${LOGGING}" ]; then echo Done upload. Cleanup at `date` >> ${LOGGING}; fi
+
+#### start apache2
+echo "Starting apache2..."
+/etc/init.d/apache2 start
```
maximilianhuber: You should only start the servers if they were running before starting this script

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+echo "Backing up database..."
+/etc/init.d/postgresql restart
+
+if [ -n "${LOGGING}" ]; then echo Creating diff of repository at `date` >> ${LOGGING}; fi
+echo "Creating repo diff..."
+ls -RlahGg  --time-style='+' /srv/fossology/repository/ | cut -d ' ' -f 2-5 | venv/bin/aws s3 cp --sse AES256 - s3://${BUCKET}/${FILENAME}-repo-diff.txt
+
+if [ -n "${LOGGING}" ]; then echo Start database dump at `date` >> ${LOGGING}; fi
+#### these steps are split, because of permissions
+su postgres -c 'pg_dumpall' > /tmp/database.txt
```
maximilianhuber: It is best practice to use a tempfile generated with `mktemp`

maximilianhuber:
It might be a good idea to extract the **backup** part from the **send to s3** part. This would allow the usage of multiple backends (e.g. scp, sftp, local directory) and might improve testability.

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,92 @@
...
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+if [[ $(id -u) -ne 0 ]]; then
+ echo "Please run as root"
+ exit 1
+fi
+
+if [[ ${1} = "-l" ]]; then
+ LOGGING=${2}
```
maximilianhuber: It might be better to call this variable `LOG` or `LOGFILE`

utils/backup/S3/fo-backup-s3
maximilianhuber: The correct message might be `echo "Log file not set."`

maximilianhuber:
Please do not create wiki pages before the corresponding features are merged.

utils/backup/S3/fo-backup-s3
max-wittig: I though that I should only use [[, when I need it, as it provides additional functionality. But not sure...

maximilianhuber: I think that `[[` have no disadvantages other than loosing the compability with `sh` (which is at most places just a configured `bash`).

I think here would also be a `[` enough.

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,43 @@
...
+source ${DIR}/before
+set -x
+${DIR}/install
+
+echo Starting backup at `date`
+
+
+
+echo Creating diff of repository at `date`
+ls -RlahGg --time-style='+' /srv/fossology/repository/ | cut -d ' ' -f 2-5 | ${DIR}/venv/bin/aws s3 cp --sse AES256 - "s3://${BUCKET}/${FILENAME}-repo-diff.txt"
```
maximilianhuber: Why is this recursive? This `cut` splits folders with spaces in their names in arbitrary places.
Why do you keep the inode-count?

Please add variables for magic strings like `"s3://${BUCKET}/${FILENAME}-repo-diff.txt"` which are used in multiple places.

utils/backup/S3/fo-restore-s3
```diff
@@ -0,0 +1,63 @@
...
+${DIR}/../../fo-cleanold --delete-repository --delete-database
+
+#### sync full tar
+echo Done removing old data. Starting download+extracting of ${FILENAME} at `date`
+${DIR}/venv/bin/aws ${ENDPOINT} --debug s3 cp --expected-size 200000000000 --sse AES256 s3://${BUCKET}/${FILENAME} - | tar --warning=no-timestamp -xzf - -C /
+
+echo Checking diff of repository at `date`
+DIFF=`${DIR}/venv/bin/aws s3 cp --sse AES256 s3://${BUCKET}/${FILENAME}-repo-diff.txt -`
+CURRENT_DIFF=`ls -RlahGg --time-style='+' /srv/fossology/repository/ | cut -d ' ' -f 2-5`
```
maximilianhuber: You are calling `ls -RlahGg --time-style='+' /srv/fossology/repository/` at multiple places, please move them to an function.

utils/backup/S3/fo-restore-s3
```diff
@@ -0,0 +1,63 @@
...
+echo "Stopping fossology scheduler..."
+/etc/init.d/fossology stop
+
+#### stop apache2
+echo "Stopping apache2..."
+/etc/init.d/apache2 stop
+
+echo Start removing old data at `date`
+
+${DIR}/../../fo-cleanold --delete-repository --delete-database
```
maximilianhuber: You might need to ensure that the database is running.
This might delete productive data, you should display a warning and give the option to exit.

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,43 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source ${DIR}/before
```
maximilianhuber: You are sourcing the `before` before setting `set -x`. You can't be sure that `$FILENAME` is filled and that apache is stopped

utils/backup/S3/fo-restore-s3
```diff
@@ -0,0 +1,63 @@
...
+cp /srv/fossology/database.txt ${TMP_FILE}
+
+/etc/init.d/postgresql restart
+
+if [[ -n "${LOGGING}" ]]; then
+  echo --------- Start database log ---------
+  su postgres -c "psql -f ${TMP_FILE}"
+  echo --------- End database log ---------
+else
+  su postgres -c "psql -f ${TMP_FILE}" > /dev/null 2>&1
```
maximilianhuber: why do you also suppress `STDERR`?

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,43 @@
...
+
+
+echo Creating diff of repository at `date`
+ls -RlahGg --time-style='+' /srv/fossology/repository/ | cut -d ' ' -f 2-5 | ${DIR}/venv/bin/aws s3 cp --sse AES256 - "s3://${BUCKET}/${FILENAME}-repo-diff.txt"
+
+echo Start database dump at `date`
+#### postgresql needs to be restarted for the dump to work
+/etc/init.d/postgresql restart
+#### these steps are split, because of permissions
+TMP_FILE=`mktemp`
```
maximilianhuber: already done in `before`

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,43 @@
...
+echo Starting backup at `date`
+
+
+
+echo Creating diff of repository at `date`
+ls -RlahGg --time-style='+' /srv/fossology/repository/ | cut -d ' ' -f 2-5 | ${DIR}/venv/bin/aws s3 cp --sse AES256 - "s3://${BUCKET}/${FILENAME}-repo-diff.txt"
+
+echo Start database dump at `date`
+#### postgresql needs to be restarted for the dump to work
+/etc/init.d/postgresql restart
```
max-wittig: because it needs to be restarted for the dump to work. I don't think start would work

utils/backup/S3/fo-restore-s3
max-wittig: because there are many errors that come up, because tables already exist etc...

utils/backup/S3/fo-backup-s3
```diff
@@ -0,0 +1,43 @@
...
+#### these steps are split, because of permissions
+TMP_FILE=`mktemp`
+su postgres -c 'pg_dumpall' > ${TMP_FILE}
+cp ${TMP_FILE} /srv/fossology/database.txt
+echo Done database dump. Starting Upload at `date`
+
+#### sync repository
+echo "Uploading and starting backup to S3..."
+tar -zcvf - /srv/fossology/ | ${DIR}/venv/bin/aws --debug s3 cp --expected-size 200000000000 --sse AES256 - "s3://${BUCKET}/${FILENAME}"
```
max-wittig: Yes, but maybe the default value will change overtime and then it wouldn't work anymore. What do you think?

utils/backup/S3/fo-backup-s3
max-wittig: `splits folders with spaces in their names in arbitrary places`
No it just splits the column. There is a tab in there, not spaces.

maximilianhuber: Since you are listing recursively, the `s -RlahGg --time-style='+'` also lists the names of the corresponding folders as the first "word" in a line (for sub directories). If they contain spaces you are splitting them and keeping a part of them.

max-wittig: Do you happen to know a way to remove the inode-count, besides cut?

maximilianhuber: # example:
```
$  tree ./                                                                                                                                                                                                                               /tmp/test
./
├── Folder with splaces
│   └── file
└── otherFile
```
and
```
$ ls -RlahGg --time-style='+' . | cut -d ' ' -f 2-5                                                                                                                                                                                     /tmp/test
.:
20K
 3 4,0K 
16  12K 
 2 4,0K 
 1  

with splaces:
8,0K
2 4,0K  .
3 4,0K  ..
1  
```

## Note:
here you also see that the `ls -RlahGg --time-style='+'` contains on my system spaces and not tabs => the filenames are not listed.

max-wittig: you're right. Also we could also just use tree for the diff, right? Only filesize would be gone then

maximilianhuber: If you want a list of all files together with their sizes use something like `find . -type f -exec du -h {} \;`

utils/backup/S3/fo-backup-s3
maximilianhuber: I would not know why this should be necessary (except for reloading the config)

utils/backup/S3/fo-backup-s3
max-wittig: Thank you!

utils/backup/S3/fo-backup-s3
maximilianhuber: or maybe this forces the DB to write everything out...

utils/backup/S3/fo-restore-s3
```diff
@@ -0,0 +1,80 @@
...
+source ${DIR}/before
+set -x
+
+if [[ ${1} = "-l" ]]; then
+ LOGGING=${2}
+ if [[ -z "${LOGGING}" ]]; then
+  echo "Please specify a file, where to save the log file to"
+  exit 1
+ fi
+ else
```
maxhbr: indentation is confusing here

utils/backup/S3/fo-restore-s3
```diff
@@ -0,0 +1,80 @@
...
+  exec &> >(tee -a ${LOGGING})
```
maxhbr: you only call this command if no argument `"-l"` was passed. That looks wrong

maxhbr: same in the other file

mcjaeger:
@max-wittig how about writing the status of this pull request as of now?

max-wittig:
The Pull Request is working, but it's not possible to run the restore and backup script at the same time.
There is no safety measure to ensure that a full backup is downloaded.

PR #823, opened 2017-05-12
Two GUI improvements: comment column with line breaks and yellow flags for "To Be Discussed" items

alpianon:
If an item needs to be discussed with other members of the team:

1. it is likely that the one who set the decision type to "To be discussed" wants to write a (possibly) long comment to explain why she/he thinks that item needs to be discussed, but long comments are displayed without line breaks, so the corresponding column may become too large to be easily read;
2. members of the team need to quickly identify the items that other members marked as "to be discussed", while fossology marks them with red flag, as if they have not been examined by anyone.

So:

1. I modified comment column style so that it has line breaks
2. I added a yellow flag to mark items that require discussion


alpianon:
I mean the comments displayed when you "open" the single file
image

Before my modification I saw this instead:
image

max-wittig:
That's still not an ideal solution, when text get's too long. But maybe better than before?
What do you think @shaheemazmalmmd @mcjaeger @ag4ums 
image

alpianon:
I agree, not ideal, but still better than before.

alpianon:
Uhm, changing comments on old commits is a bit too complicated and messy (https://help.github.com/articles/changing-a-commit-message/), maybe it would be better if I delete my first fork, refork the project and commit again my changes with the "right" comments, what do you think?

max-wittig:
it's not complicated. You might want to look into [rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)
<pre>
git rebase -i HEAD~3
git push -f origin master
</pre>

Also for the future. The best way to contribute is to use a new branch on your fork and propose a PR from there. 

**Never commit directly to master**

alpianon:
I think I have done right now... I also added to my file those (unneeded) whitespaces which were in the original files so that the diff is much clearer to read)

src/www/ui/template/browse_file.js.twig
```diff
@@ -31,10 +31,14 @@ function createDirlistTable() {
         {"sTitle":"Edited Results","sClass":"left","bSortable":false},
         {"sTitle":"Clearing Status","sClass":"clearingStatus center","bSortable":false,"bSearchable":false,"sWidth":"5%","mRender":function ( data, type, full ) {
             if(type!='display') return data;
-            if (data==='green'){
-              return '<img alt="done" src="images/green.png" class="icon-small"/>';
+            switch (data){
+              case 'green':
+                return '<img alt="done" src="images/green.png" class="icon-small"/>';
+              case 'yellow':
+                return '<img alt="done" src="images/yellow.png" class="icon-small"/>';
```
mcjaeger: I wondered how about changing this to not done, as the yellow this is " to be discussed " indicating that someone should come back to this and clarify.

mcjaeger:
looks also good to me, just mentioning that the alternate text displayed in stead of the image could be corrected to `not done` (or `not quite done`, har har).

mcjaeger:
@max-wittig I was wondering about your never-to-master comment - may could could be more verbose and explain why you suggest this. - maybe you intended to suggest that he might add other features in future and should have a clean master ready? Please help here to shed some light here and explain also why you suggest to never commit directly to master in his fork?

alpianon:
> looks also good to me, just mentioning that the alternate text displayed in stead of the image could be corrected to not done (or not quite done, har har).

I did not think about that, I agree (why not simply "to be discussed"?)

max-wittig:
There are several reasons for not committing directly to master, even on your fork
1. Good practice, keeps master "clean" and your fork on the same state. No one is allowed to directly 
    push to the original master, so you also shouldn't commit on your master

2. Committing directly to your fork makes trying out your pull request more complicated
    Normally you can just follow this guide: [Checking out pull requests locally](https://help.github.com/articles/checking-out-pull-requests-locally/)
    and run this command
    <pre>git fetch origin pull/ID/head:BRANCHNAME</pre>
    This guide doesn't work anymore, if you commit on your master
3. Let's say you commit on your master branch and the original upstream project doesn't accept your 
    changes. Are you gonna rebase your master with a force push? Normally you can just delete the 
    branch, where you submitted the PR. Master stays untouched
4. Let's pretend you want to work on multiple Issues at the same time and submit multiple Pull 
    Requests. That's not possible, because every change requires another Pull Request from another 
    branch of yours. But now your master isn't clean anymore. How are you gonna branch of master 
    now?

If I got anything wrong, let me know. That's just my understanding, why using another branch is best practice :smiley: 

@mcjaeger @alpianon

PR #830, opened 2017-05-19
fix(delagent): delete child folder by parent id

shaheemazmalmmd:
**Delete child folder from command line.** 

Deletion of child folder use to delete all the reference folders as there was no link to its parent.  now use "-p" to indicate the child folders parent.

Also included https://github.com/fossology/fossology/pull/788 after removing a run-time warning

To reproduce issue:

Create a new Folder with Name "TestFolder"
Upload file (in this example test.zip)
Copy test.zip into TestFolder -> test.zip should now exist in the Software Repository and in the folder TestFolder
Delete TestFolder
Expected Behavior: test.zip still exists in the Software Repository, because it was only copied from there in TestFolder before
Actual Behavior: test.zip is gone from the Software Repository
This commit should fix this issue. Testing from more contributors is still necessary

### DIFF
--- a/src/delagent/agent/delagent.c
+++ b/src/delagent/agent/delagent.c
@@ -2,7 +2,7 @@
  delagent: Remove an upload from the DB and repository
 
  Copyright (C) 2007-2013 Hewlett-Packard Development Company, L.P.
- Copyright (C) 2015-2016 Siemens AG
+ Copyright (C) 2015-2017 Siemens AG
 
  This program is free software; you can redistribute it and/or
  modify it under the terms of the GNU General Public License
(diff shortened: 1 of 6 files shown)


max-wittig:
How should the behavior of delagent be, after this change? What should I expect?

max-wittig:
This does work, but I noticed another issue, very closly related to this one. I thought that this PR would fix this too, but unfortunatly it doesn't. 

**To reproduce:**
1. Upload a file into the Software Repository
2. Create a Folder and call it for example `AAA`
3. Copy the file into `AAA`
4. Delete the file in the Folder `AAA`

**Expected behavior:**
File is gone in Folder `AAA`, but still exists in the Software Repository

**Actual Behavior:**
File is gone in the Folder `AAA` as well as in the Software Repository

### Should we fix this in a separate PR or do you want to patch it in this one, because the issue is so closly related?

mcjaeger:
(from discussion)

**Cases**
1.  File is not copied (no copies exist anymore), user selects the only occurrence for a file -> deletion from software repo is OK, in addition to removal from folder.
2. File is copied at least once, user select one out of at least two copies for deletion -> remove entry from folder is obviously OK. question is to delete the upload entirely from software repo or not.

PR #832, opened 2017-05-22
feat(license_administration): add sorting, update datatable

max-wittig:
### Results
* Faster loading times
* Sortable table

---

image

## New additions
*  Search on each column
*  Default sorting by short name

---

screenshot-2018-5-23 license administration


mcjaeger:
suggestion from discussion: let's keep the license text 3 first lines view: it is fact used for checking quickly if the licenses need beautyfying because of comment characters. Maybe it could be relaced with the URL so there are no space issues, also the cell could be with flex width then.

max-wittig:
changed the behavior again, so that the license textfield is included again. This does impact performance

src/www/ui/template/admin_license-upload_form.html.twig
```diff
@@ -4,12 +4,13 @@
...
    This file is offered as-is, without any warranty.
 #}
+
 <form name="Updatefm" action="{{ actionUri }}" method="post">
   <input type="hidden" name="req_marydone" value="{{ req_marydone }}"/>
   <input type="hidden" name="req_shortname" value="{{ req_shortname }}"/>
   <input type="hidden" name="rf_pk" value="{{ rf_pk }}"/>
   <input type="hidden" name="{% if rf_pk %}updateit{% else %}addit{% endif %}" value="true"/>
-  <table>
+  <table id="dataTableTest">
```
shaheemazmalmmd: id="dataTableTest" can be removed as it was not used anywhere

shaheemazmalmmd:
After testing i observed that the license table is getting loaded for two times. plain table and datatable.
i think it would be good to apply a class to existing table functionality then loading it in datatable. because the time taken to load data in datatable is more compared to normal table. May be because to total no of records is 400.

max-wittig:
It isn't loaded two times. It is printed once by PHP and then the dataTables plugin applies to this table.
It's not really possible to further improve the performance with this setup, as the text column contains just too much text to be parsed by the DOM in a timely matter. If you want to further improve the loading times, that only way would be to stop showing the text column on page load

maximilianhuber:
It might be an alternative to add the table data via an large JS object. Parsing the JS JSON syntax might be faster then parsing an inline html table to extract the data from there.

src/www/ui/scripts/tools.js
```diff
@@ -83,3 +83,8 @@ function sortList(selector)
         $(o).text(arr[i].t);
     });
 }
+
+function popUpLicenseText(licenseID, title) {
```
maximilianhuber: You are adding this function here but it is never called in your added code

src/www/ui/admin-license-file.php
```diff
@@ -57,6 +57,7 @@ function RegisterMenus()
   public function Output()
   {
     $V = ""; // menu_to_1html(menu_find($this->Name, $MenuDepth),0);
+    $V .= '<script src="scripts/jquery-3.2.0.min.js"></script>';
```
maximilianhuber: You are importing jquery here (see first screenshot) altough it gets also imported in the footer (see second screenshot).

##### first screenshot:

##### second screenshot:

src/www/ui/admin-license-file.php
```diff
@@ -103,11 +104,26 @@ public function Output()
...
+      $V .= '<script src="scripts/tools.js"></script>';
+      $V .= $this->getDataTableScript();
+    }
     return $V;
   }
 
+  function getDataTableScript()
+  {
+    return <<<EOT
+    <script src="scripts/jquery.dataTables.min.js" type="text/javascript"></script>
```
maximilianhuber: It looks like that `css/jquery.dataTables.css` is loaded by default but `scripts/jquery.dataTables.min.js` is added on demand? That looks inconsistend (but is not a problem of this PR).

maximilianhuber:
maybe pagination would also help to reduce the rendering time of the table (would need the changes from [the comment above](https://github.com/fossology/fossology/pull/832#issuecomment-314687604))

maximilianhuber:
Please update or remove outdated screenshots.

src/www/ui/admin-license-file.php
max-wittig: I don't see another solution. JQuery is loaded too late. Even if I try to add 
<pre>
<script src="scripts/jquery.dataTables.min.js"></script>
<script src="scripts/tools.js"></script>
</pre>
to the `foot.html.twig` tools.js would be loaded too late.

maximilianhuber:
There is no CSS styling for the paging-numbers:

src/www/ui/admin-license-file.php
```diff
@@ -202,64 +205,52 @@ function LicenseList($namestr, $filter)
...
+        "<textarea readonly rows='3' cols='40'>$vetext</textarea>",
+        "$row[rf_url]"
+      );
+
+      $tableBodyData[] = $row;
     }
+
+    $ob .= "
+    <script>
+     var licenseTable = '" . urlencode(json_encode(($tableBodyData))) . "';
```
maximilianhuber: There are to many braces.

maximilianhuber: I think that base64-encoding might be faster and simpler encoding.

src/www/ui/admin-license-file.php
maximilianhuber: I think for this problem there is the usual [`$(document).ready()`-pattern](https://learn.jquery.com/using-jquery-core/document-ready/). Just call your function to populate the table after everything is loaded.

src/www/ui/admin-license-file.php
maximilianhuber: Is the encoding necessary? I think `"[...] var licenseTable = ".json_encode($tableBodyData)."; [...]"` should also work and should be escaped correctly

max-wittig: It doesn't work without encoding, but you're right. Maybe base64 is faster

max-wittig: The $(document).ready()-pattern isn't working in that case. Also none of the other later loading functions work

maximilianhuber: But we use Datatables also on other pages. Do they all request another jQuery?

maximilianhuber: Works perfectly fine without the encoding and parsing:
```diff
diff --git a/src/www/ui/admin-license-file.php b/src/www/ui/admin-license-file.php
index 8d999101b..d92b2ad7f 100644
--- a/src/www/ui/admin-license-file.php
+++ b/src/www/ui/admin-license-file.php
@@ -244,7 +244,7 @@ class admin_license_file extends FO_Plugin

     $ob .= "
     <script>
-     var licenseTable = '" . urlencode(json_encode(($tableBodyData))) . "';
+     var licenseTable = " . json_encode($tableBodyData) . ";
      insertLicenseTable(licenseTable);
     </script>";
     pg_free_result($result);
diff --git a/src/www/ui/scripts/tools.js b/src/www/ui/scripts/tools.js
index 9d35a2799..fdc2b7be3 100644
--- a/src/www/ui/scripts/tools.js
+++ b/src/www/ui/scripts/tools.js
@@ -85,9 +85,8 @@ function sortList(selector)
 }

 function insertLicenseTable(licenseTableJSON) {
-    var data = JSON.parse(urldecode(licenseTableJSON));
     $('#adminLicenseTable').DataTable({
-      data: data,
+      data: licenseTableJSON,
       deferRender: true,
       responsive: true,
       scroller: true
```

maxhbr:
@max-wittig please implement or solve the requested changes

src/www/ui/admin-license-file.php
max-wittig: this doesn't work for me in Google Chrome

maxhbr: can you paste here a screenshot of the error in the dev-log?

max-wittig:
also included the jQuery DataTables script with $vars['scripts'] is ignored for this page

maxhbr:
Have you had a look into the templates why that variable is not used?

max-wittig:
I don't think that this page has a template. The only template that is loaded is for `admin_license-upload_form.html.twig`, but this isn't contained on this page

ChristophNiehoff:
Made some modifications to the existing PR:
* Removed the base64-encoding. Now the JS objects are written directly using the twig-template. This allows to escape the license texts in a simple way.
* When creating the datatable, only the first 120 characters (corresponding roughly to the first three lines) of the license texts are shown. The complete license text is loaded only on demand through a mouseover-event of the textboxes. This reduces the traffic from ~1.5MB to ~35kB.  On my local setup this has a quite good performance, but it should definitely be test in a realistic environment.
* There was a error in the SQL query for getting the license data for a given "License family name"
* jquery is loaded only once now.

maxhbr:
Instead of amending the commit from @max-wittig, id might be better to add your own additional commit.

src/www/ui/template/admin_license_file.js.twig
```diff
@@ -0,0 +1,118 @@
...
+//
+
+{% for entry in data %}
+result.push({
+    "edit":               [ "{{ entry['rf_pk'] }}", "{{ entry['marydone'] }}", "{{ entry['rf_shortname'] }}"],
+    "marydone":           "{{ entry['marydone'] }}",
+    "rf_pk":              "{{ entry['rf_pk'] }}",
+    "rf_spdx_compatible": "{{ entry['rf_spdx_compatible'] }}",
+    "rf_shortname":       "{{ entry['rf_shortname'] }}",
+    "rf_fullname":        "{{ entry['rf_fullname']|escape('js') }}",
```
maxhbr: You should escape every value, not only some of them.

ChristophNiehoff: But these are only IDs and Booleans. Do you think they can possibly contain dangerous characters?

maxhbr: shortname is a string...

ChristophNiehoff:
Split up into two separate commits and implemented the change requested by @maxhbr

src/www/ui/template/admin_license_file.js.twig
```diff
@@ -0,0 +1,118 @@
...
+        if( $(this).data('textloaded')==false ) {
+            setLicenseText($(this).data('licenseid'), $(this).attr('id'))
+            $(this).data('textloaded',true);
+        }
+
+    });
+});
+
+function setLicenseText(licenseID, textareaID) {
+    var result = $.ajax({
```
maxhbr: maybe remove `var result = `

src/www/ui/template/admin_license_file.js.twig
```diff
@@ -0,0 +1,118 @@
...
+
+{% for entry in data %}
+result.push({
+    "edit":               [ "{{ entry['rf_pk'] }}", "{{ entry['marydone'] }}", "{{ entry['rf_shortname']|escape('js') }}"],
+    "marydone":           "{{ entry['marydone'] }}",
+    "rf_pk":              "{{ entry['rf_pk'] }}",
+    "rf_spdx_compatible": "{{ entry['rf_spdx_compatible'] }}",
+    "rf_shortname":       "{{ entry['rf_shortname']|escape('js') }}",
+    "rf_fullname":        "{{ entry['rf_fullname']|escape('js') }}",
+    "rf_text":            [ "{{ entry['rf_pk'] }}", "{{ entry['rf_text'][:120]|escape('js') }}" ],
```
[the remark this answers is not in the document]
ChristophNiehoff: The whole procedure of loading the full license texts via ajax was designed in such a way that the user does not recognize its existence. So, I would prefer not to use ellipsis.

ChristophNiehoff:
@maxhbr Implemented your requested changes.

mcjaeger:
some issue with this PR, this new table is not really that it loads slower (apparently 50% at first page impression), but one cannot search all the licenses anymore. For example, one would like to look for a particular license id named `OPAL`. 

Problem is that Web browser search is only working with full table viewed, however, the table in this PR shows at max 100 entries. Also the setting is not sticky and having 10 entries at every reload is painful too.

GMishx:
Added new changes:
* Search by column
* Default sorting by short name

PR #835, opened 2017-05-23
feat(monk): make use of rf_active to detect monk scan for licenses

shaheemazmalmmd:
monk will scan license only when rf_acive is true
changes in license_ref table, Changed rf_active from false to true.
please do postinstall to test.


max-wittig:
What does rf_active mean? If I understand this correctly this commit will improve performance by reducing scans of monk?

shaheemazmalmmd:
rf_active refers to the status of license in license administration.

activestatuscapture

In this PR, monk scanner scans for the licenses with status active(yes/true).

mcjaeger:
we have added this feature to pull off licenses from bulk scanning that are part of the server database. You need to keep them in the database for maintaining old results, however, for monk scanning they do not contribute to an optimal result.

Example BSD variant "France Telecom"

max-wittig:
I'm just getting a 500 Error, when opening /repo
The /fo-postinstall script did ran
Maybe it conflicts with my existing uploads?

shaheemazmalmmd:
Hi, i have tested this branch with existing uploads it works as expected. can you explain me how did you install this changes ??

max-wittig:
It tried it again and it seems to work now. Weird.
Just testing the UI and checked the code. Looks good so far, but I don't know how to check what monk scans in detail

mcjaeger:
some notes on the testing:
* just take some package with a full license text, i.e. an Apache 2.0 licensed package with license text on project root folder (commons-*, etc.)
* BEFORE deactivate in License Browser the Apache 2.0 license
* Upload and scan with Monk: Monk shall not find the Apache 2.0 license text.

shaheemazmalmmd:
After seeing the error message i think that fossology is not able to get "businessrules.obligationmap" which is there in master but not in this branch. just do make clean and make install from root folder then  run fo-postinstall. i hope it should work

max-wittig:
Monk doesn't start for me on your branch. I don't know why :/

max-wittig:
Some questions:
If you understood this correctly

1. To disable an existing license you have to create a new license in `Admin-->License-Admin-->Add License` and set the concluded license to the license you want to disable?
2. How to remove user created licenses again?
3. Let's say you re-enable a license, but ran monk through the upload as this license was set to not active --> you can't schedule monk again on this upload

PR #837, opened 2017-05-29
WIP: feat(api): add fossology REST api

max-wittig:
## FOSSology REST OpenAPI specification:
can be viewed in rendered form on http://editor2.swagger.io/

**Implemented**:

GET                /repo/api/v1/uploads/{id}
DELETE          /repo/api/v1/uploads/{id}
GET               /repo/api/v1/search/
GET               /repo/api/v1/users/
GET               /repo/api/v1/users/{id}
DELETE         /repo/api/v1/users/{id}
GET              /repo/api/v1/jobs/


pguttmann:
It's great to see some thoughts/progress on this front :)
I'd like two suggest a couple of things:
1. Under the Job section, perhaps a "Status" report capability could be included?
2. Break "Licenses" out from Organize and place in a new section Reports along with other report types such as copyright, SPDX etc...

max-wittig:
Hey thanks for the input.
1. Isn't there already a status report capability for the jobs section or am I missing something?
``` yaml
/jobs:
    get:
      tags:
      - Job
      summary: Gets all jobs
      description: Returns all jobs with their status
      responses:
        200:
          description: OK
          schema:
            type: array
            items:
              $ref: '#/definitions/Job'

Job:
    type: object
    properties:
      job_id:
        type: number
        format: int
      status:
        type: string
      filename:
        type: string
      eta:
        type: string
```

2.  That's why I used /organize/licenses modelled after the fossology menu, but you're right the whole reports section is still missing

image

pguttmann:
Oops - I guess I was missing something, not you. Sorry about that - I missed the clarification: "description: Returns all jobs with their status"

max-wittig:
No problem. Also feel free to edit the checklist above

mcjaeger:
two questions: 

1) will the api documentation index file be generated? or can it be generated with a target? I would like to avoid checking that in?

2) I am not sure the way the API is best organised along the menue structure of  menu structure is maybe not optimal from a REST perspective thinking resource oriented.

3) Why not starting small and do a single use case like uploading an upload and retrieving a report like readme oss? I think the large coverage of functions poses implication on how to design the resource paths that may be necessary on the first attempt.

src/www/ui/api/helper/DbHelper.php
```diff
@@ -0,0 +1,157 @@
+<?php
```
mcjaeger: please consider file headers for all new files:
https://github.com/fossology/fossology/wiki/Coding-Style#default-license-and-file-headers

TheInvoker:
I'm trying the wget command but it just returns the html of the page. It does show what I want which is the license information. Is there some configuration I need to do to get it working?

max-wittig:
for now it should work, if you insert this into `/etc/apache2/sites-enabled/000-default.conf`
<pre>
**RewriteEngine on
RewriteCond %{REQUEST_FILENAME} !-d
RewriteCond %{REQUEST_FILENAME} !-f
RewriteRule   "^/repo/api/(. *)"  "/usr/local/share/fossology/www/ui/api/index.php" [L]**
</pre>

mcjaeger:
@max-wittig please let us know if you consider continue working on this branch and when.

mcjaeger:
closing this for now, could be reopnened in case

PR #838, opened 2017-05-30
feat(reuse): search all folders

max-wittig:
This PR enabled the search of all uploads to reuse. No need to specify a specific folder anymore, but it's still possible to do so.

#### Please check the performance implications of this pull request with a larger amount of folders and uploads

image
<hr />

image

### DIFF
--- a/src/lib/php/Dao/FolderDao.php
+++ b/src/lib/php/Dao/FolderDao.php
@@ -202,8 +202,24 @@ GROUP BY group_fk
     $this->dbManager->freeResult($res);
     return $results;
   }
-  
-  
+
+  public function getAllFolderIds()
+  {
+    $statementName = __METHOD__;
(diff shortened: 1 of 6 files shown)


mcjaeger:
good point: the switch with the checkbox switched between
* search for a specific folder
* or just all folders

src/lib/php/Dao/FolderDao.php
```diff
@@ -202,8 +202,24 @@ public function countFolderUploads($parentId, $userGroupMap)
...
+    $allIds = array();
+    for($i=0; $i < sizeof($results); $i++)
+    {
+      array_push($allIds, intval($results[$i]['folder_pk']));
+    }
+
+    return $allIds;
+  }
+
   public function getFolderChildUploads($parentId, $trustGroupId)
```
ag4ums: please provide test-cases for getAllFolderIds()

Issue #826, opened 2017-05-17
Some cleared licenses are reported incorrectly in SPDX exports.


swinslow:
Issue:
When both (A) an archive contains multiple identical copies of a file, and (B) that file has been identified and cleared using the "Apply decision to all future occurrences of this file" option, then an SPDX export will report incorrect concluded license results for those files.

Steps to reproduce:
1) Upload an archive that contains multiple copies of the same file.  A sample archive is attached as [example.tar.gz](https://github.com/fossology/fossology/files/1007983/example.tar.gz); this archive contains three copies of the same file.
2) In the upload settings, select Monk and Nomos scanning options.
3) After the scan completes, manually go into the license browser and identify one of the files as "BSD".  Select the option "Apply decision to all future occurrences of this file" before hitting submit.
4) The other two identical files should also be cleared and marked as "BSD" in the license browser after this.
5) Go back to the "Browse" tab, and for the example.tar.gz upload, choose "-- select action --" => "Export SPDX tag:value".
6) After the SPDX tag:value report is generated, look at the LicenseConcluded tags for the three files in the archive.  

Expected result:
In the SPDX report, each of the three files should have the tag "LicenseConcluded: LicenseRef-BSD"

Actual result:
In the SPDX report, one of the three files has the tag "LicenseConcluded: LicenseRef-BSD AND LicenseRef-BSD AND LicenseRef-BSD".  The other two files have the tag "LicenseConcluded: NOASSERTION".

A copy of the SPDX report is here: [SPDX2TV_example.tar.gz_1495029013.txt](https://github.com/fossology/fossology/files/1008007/SPDX2TV_example.tar.gz_1495029013.txt).

swinslow:
I've confirmed the same behavior is occurring with DEP5 exported reports.  See 
[DEP5_example.tar.gz_1495030615.txt](https://github.com/fossology/fossology/files/1008021/DEP5_example.tar.gz_1495030615.txt):

    . . .
    Files: example/file2.txt
    License: BSD AND BSD AND BSD
    
    Files: example/file1.txt
           example/file3.txt
    License: NoLicenseConcluded
    Comment: scanners found: NOASSERTION
    . . .

mcjaeger:
I think that DEP5 and SPDX use the same implementation for querying the database so the second observation should be OK (otherwise a fix in one corner maybe forgotten at the other generation format)

swinslow:
[...]

PR #839, opened 2017-05-30
WIP: fix(spdx2): fix incorrect license results in SPDX export

swinslow:
This PR is a first cut at addressing issue #826, for the incorrect SPDX reports of licenses that are cleared using the “Apply decision to all future occurrences of this file” option.

The edits to /src/spdx2/agent/spdx2.php are intended to fix the incorrect NOASSERTION results by generating the LicenseConcluded result using a similar process as the createFileDataRow() function in src/www/ui/async/AjaxExplorer.php.

I’m marking this PR as a WIP for two reasons:
1. This revision doesn’t separately walk through the reportinfo columns in the clearing_event table, so it may not be pulling in all of the expected data for the SPDX report.
2. I’m quite new to each of Fossology, PHP and Github, so I apologize in advance if anything here is obviously incorrect or submitted in the wrong manner!

### DIFF
--- a/src/spdx2/agent/spdx2.php
+++ b/src/spdx2/agent/spdx2.php
@@ -20,6 +20,7 @@ namespace Fossology\SpdxTwo;
 
 use Fossology\Lib\Agent\Agent;
 use Fossology\Lib\BusinessRules\LicenseMap;
+use Fossology\Lib\BusinessRules\ClearingDecisionFilter;
 use Fossology\Lib\Dao\ClearingDao;
 use Fossology\Lib\Dao\CopyrightDao;
 use Fossology\Lib\Dao\TreeDao;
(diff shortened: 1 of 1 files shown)


mcjaeger:
> I’m quite new to each of Fossology, PHP and Github, so I apologize in advance if anything here is obviously incorrect or submitted in the wrong manner!

no problem, I think the way you have started to work on your fork is pretty safe. Great to see that you contribute for this. actually I am not using that "apply all future occurrences" selection, so we may have missed that obviously.

On the remark dep5 from #826 : I think it was to expect because it shares some login with the spdx generation.

swinslow:
One comment on this PR: 
The way that it's currently implemented, for files that would previously be output as "LicenseConcluded: NOASSERTION", those files will now be output as "LicenseConcluded: NONE".
I am not yet familiar enough with Fossology's internals to know the most correct way to address this (though I'm taking a look at it). But I did want to highlight the issue so that anyone using this PR in its current form isn't caught off-guard.

swinslow:
I've identified another problem with this PR, as currently implemented:

In the exported SPDX file, with this patch in place, Fossology no longer correctly reproduces the license texts for LicenseRef-* licenses from clearing results. (It reproduces the texts for licenses detected by scanners, but not from clearings.)

I have a fix for this, but I don't yet have a clean fix for the issue I mentioned in my comment last week (above). In light of that, I'm going to close this PR for the time being. I'll submit a new PR after I've addressed the other issue as well.

Issue #786, opened 2017-03-31
nomos corrections?


rlintu:
Hi,
I have a lot of corrections to nomos which I haven't had time to contribute. I understand that
it is not a good practice to contribute large chunk of code, but I'm afraid that I do not have
time to split the code into smaller, more manageable, pieces. 

I can make a pull request with all changes it that is OK. 
Otherwise I can try to categorize the changes into logical units (GPL corrections, Freeware corrections
etc.) but that may take time.

PR #842, opened 2017-06-06
feat(new-agent): license scanner for SPDX-License-Identifier

ag4ums:
This new license scanner identifies `SPDX-License-Identifier:   <license-name>` license information from  from files.


src/ojo/Makefile
```diff
@@ -0,0 +1,50 @@
+# Copyright Siemens AG 2017-2018
```
mcjaeger: no !!! :-D

rlintu:
Do we need a new agent for this? nomos can be enhanced to detect those license identifiers, actually I've done that for few licenses (and some other corrections https://github.com/fossology/fossology/issues/786)

rlintu:
Sorry, wrong button pressed ...

mcjaeger:
yes we do, because it should enable auto conclusion when the SPDX license identifier is matched. Opposed to Nomos regex matching we could do:

* Ojo finds a SPDX-id-based license statement
* Nomos does not find anything else

-> auto conclude and done.

mcjaeger:
some projects to test with:

https://github.com/ARM-software/arm-trusted-firmware
https://github.com/ARM-software/CMSIS_5
https://github.com/zephyrproject-rtos/zephyr

more here:
https://wiki.spdx.org/view/Business_Team/Adoption#Use_of_SPDX_Short_Identifiers

mcjaeger:
for the decider, to conclude:
* ojo alone should be OK, no other agent finds something
* if other agents find something different, then no decision

mcjaeger:
Just came across this: how about adding a few words to the ojo license analysis in the upload form just as it is done for monk and nomos? I think users will not understand what ojo does otherwise.

mcjaeger:
I am missing the decider part?

mcjaeger:
I think Ojo is missing in the license list (just see nomos monk and ninka there)

src/ojo/agent/ojos.cc
```diff
@@ -0,0 +1,162 @@
...
+ * @brief Entry point for ojo agent
+ * @page ojo OJO Agent
+ * @tableofcontents
+ *
+ * OJO agent uses regular expressions to find out SPDX License identifiers
+ * from a file.
+ *
+ * The agent runs in multi-threaded mode and creates a new thread for
+ * every pfile for faster processing.
+ *
```
mcjaeger: nice! command line capability!

mcjaeger:
this is going to be cool:

Screen Shot 2019-08-01 at 20 34 17

and this:

Screen Shot 2019-08-01 at 20 34 44

src/ojo/agent/OjosDatabaseHandler.cc
```diff
@@ -0,0 +1,354 @@
...
+unsigned long OjosDatabaseHandler::saveLicenseToDatabase(
+    OjoDatabaseEntry &entry) const
+{
+  QueryResult queryResult = dbManager.execPrepared(
+    fo_dbManager_PrepareStamement(dbManager.getStruct_dbManager(),
+      "ojoInsertLicense",
+      "INSERT INTO license_file"
+      "(rf_fk, agent_fk, pfile_fk, rf_match_pct)"
+      " VALUES($1,$2,$3,100) RETURNING fl_pk",
+      long, long, long),
```
ag4ums: 100 is not require as Ojo don't find any percentage

GMishx: Removed, please check.

debian/control
```diff
@@ -236,6 +236,18 @@ Description: architecture for analyzing software, ninka
  .
  This package contains the ninka wrapper agent programs and their resources.
 
+Package: fossology-ojo
+Architecture: any
+Depends: fossology-common, ${shlibs:Depends}, ${misc:Depends}
+Description: architecture for analyzing software, ninka
+ The FOSSology project is a web based framework that allows you to
```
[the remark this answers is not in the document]
GMishx: Fixed, thanks for pointing out.

PR #844, opened 2017-06-07
fix(obligations): reintegrate lost changes

deveaud-m:
Use correct copyright year for last changes.
Set filename while exporting the obligations CSV.
Allow obligations with no licenses associated.

### DIFF
--- a/src/lib/php/Application/ObligationCsvExport.php
+++ b/src/lib/php/Application/ObligationCsvExport.php
@@ -1,6 +1,6 @@
 <?php
 /*
-Copyright (C) 2015, Siemens AG
+Copyright (C) 2017, Siemens AG
 
 This program is free software; you can redistribute it and/or
 modify it under the terms of the GNU General Public License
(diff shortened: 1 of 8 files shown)


shaheemazmalmmd:
Changes looks good. needs to be tested once

mcjaeger:
(from discussion) let do next steps in a saparate effort:

shaheemazmalmmd:
Updating text from CSV file leads to duplicate obligation. 

steps to reproduce: 
1) Create obligations for any two licenses using add obligation.
2) Export obligations in CSV.
3) Now edit obligation text that is "full text" for both the licenses.
4) Import the same using import functionality.
5) Now check the obligations "edited" obligations appeared twice.

deveaud-m:
@shaheemazmalmmd Thanks for testing and reporting the issue. 

That's right, the current implementation allows two obligations with the same `topic` to coexist in the DB only if the obligation texts differ. This might be needed to associate one identical obligation topic with different licenses and texts. 

From my point of view the feature is implemented correctly but we can still discuss if obligation topics should be unique in the DB.

mcjaeger:
just checked with the "users": title could be the same. At some point we could have an id displayed somewhere, but from my side I think it could be Ok for now.

deveaud-m:
Thanks @mcjaeger for the clarification.
@shaheemazmalmmd could you please merge those changes?

I'm currently working on #845 in a separate branch.

PR #854, opened 2017-06-19
feat(dashboard): add PHP info table

max-wittig:
## Provides PHP version information and loaded modules in admin dashboard
image


src/www/ui/admin-dashboard.php
```diff
@@ -159,7 +196,6 @@ function DatabaseMetrics()
     $connection_count = $row['val'];
 
     /**** Active connection count ****/
-    $current_query = (strcmp($this->pgVersion['server'], "9.2") >= 0) ? "state" : "current_query";
```
mcjaeger: granted that this is an unused variable. But how about adding this feature later at some time, because the current query info could be useful for debugging as well.

max-wittig: This functionality is already implemented in line 222 (Didn't see that either)
image

src/www/ui/admin-dashboard.php
```diff
@@ -176,7 +212,7 @@ function DatabaseMetrics()
    */
   function DatabaseQueries()
   {
-    $V = "<table border=1>\n";
+    $V = "<table border=1 style='width: 100%;'>\n";
```
mcjaeger: pls. consider removing "inline styling"

src/www/ui/admin-dashboard.php
```diff
@@ -308,25 +352,43 @@ public function Output() {
     $this->pgVersion = pg_version($PG_CONN);
     
     $V="";
-    $V .= "<table border=0 width='100%'><tr>\n";
+    $V .= "<table style='width: 100%;' border=0>\n";
```
maximilianhuber: Please do not use inline styling. Please use CSS instead of table specific styling options.

mcjaeger:
@max-wittig if it is OK, let's merge this and not add the versions of the modules, could be added later as well.

shaheemazmalmmd:
@max-wittig : please rebase this branch with master (merge conflicts)

PR #857, opened 2017-06-21
fix(debian9): add compatibility with debian9

max-wittig:
Maybe also works for ubuntu 16.04+, but not tested!

* Why tests were touched: [Only variables should be passed by reference](http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.error-handling.strict)
* Updated default PHP Unit as Fossology doesn't support PHP 5.4 anyway
* Removed PHP 5.4 from travis-ci for the same reason


max-wittig:
Some issues exist with permissions on debian 8 with this PR

src/testing/docker/default-docker-test.sh
```diff
@@ -13,10 +13,10 @@ docker create -p 8081:8081 --name fossology-test --net fossology-testnet --ip 17
 docker start fossology-test
 
 #### is container running
-docker ps | grep -q fossology-test
+docker ps | grep fossology-test
 
 #### fossology needs up to 15 seconds to startup
-sleep 15
+sleep 30
```
maximilianhuber: please update or remove the comment above

src/testing/docker/default-docker-test.sh
```diff
@@ -13,10 +13,10 @@ docker create -p 8081:8081 --name fossology-test --net fossology-testnet --ip 17
...
 
 #### is fossology reachable? --> check title
-curl -L -s http://172.18.0.22/repo/ | grep -q "<title>Getting Started with FOSSology</title>"
+curl -v -L -s http://172.18.0.22 | grep -q "<title>Apache2 Debian Default Page: It works</title>"
```
maximilianhuber: This test is no longer checking that fossology is running or responding! Please add also the old test back in.

utils/fo-installdeps
```diff
@@ -110,13 +110,13 @@ if [ $BUILDTIME ]; then
...
               wheezy) apt-get $YesOpt install $YesOpt postgresql-server-dev-9.1;;
-              sid) apt-get  $YesOpt install $YesOpt postgresql-server-dev-8.4;;
               lucid|natty) apt-get $YesOpt install postgresql-server-dev-8.4;;
               isadora) apt-get $YesOpt install postgresql-server-dev-8.4;;
               precise|oneiric|quantal|raring|saucy) apt-get $YesOpt install postgresql-server-dev-9.1;;
               trusty) apt-get $YesOpt install postgresql-server-dev-9.3;;
               jessie) apt-get $YesOpt install postgresql-server-dev-9.4;;
-              xenial) apt-get $YesOpt install postgresql-server-dev-9.5;;
+              xenial|yakkety) apt-get $YesOpt install postgresql-server-dev-9.5;;
+              stretch|buster|sid|zesty|artful) apt-get $YesOpt install postgresql-server-dev-9.6;;
```
max-wittig: So it's okay, if I understand that correctly?

maximilianhuber: it will break soon and is the same problem as in https://github.com/fossology/fossology/pull/864. Maybe it is a good idea to remove the `sid` option but I am not sure.
=> It is OK until someone finds a better solution.

src/lib/php/Test/TestLiteDb.php
```diff
@@ -63,7 +63,8 @@ function __construct($dbFileName = null)
     
     $sqlite3Connection = new SQLite3($this->dbFileName, SQLITE3_OPEN_READWRITE | SQLITE3_OPEN_CREATE);
 
-    $container->get('db.manager')->setDriver(new SqliteE($sqlite3Connection));
+    $liteDb = new SqliteE($sqlite3Connection);
+    $container->get('db.manager')->setDriver($liteDb);
```
max-wittig: because `Only variables should be passed by reference` starting with PHP 7.0

utils/fo-installdeps
bill-auger: IMHO it is reasonable to remove 'sid' support - as a bleeding edge unstable 'rolling' distro, much like arch, users are expected to "know what they are doing" - and for the same reason arch support was deemed to be not a priority, it is probably an extremely small use-case that anyone is running a server on 'sid' - the docker image is still there for those few

but on that note the 'buster' distro is just as volatile trailing only a few weeks behind 'sid' so they should stay or go together - but the main thing is that this is not going to "break" until v9.7 comes out - which will not exactly be broken - it is just adding yet another version to the stack as you see has been done several times already - and regardless of debian, this would still be the case for the next bi-annual release of ubuntu that includes v9.7

practically speaking, ubuntu is actually the more troublesome target than debian because it is guaranteed to requires a routine update to these install files every 6 months - so to be consistent with this decision, note that dropping support for 'buster' and 'sid' and supporting only the LTS releases of debian is the equivalent to only supporting ubuntu LTS every 2 years and not supporting any of the interim versions as has been done all along

bill-auger:
@maximilianhuber -

to be clear, the point i raised was that ubuntu support is actually more work to maintain than debian because it is guaranteed to requires a routine update to these install files every 6 months - even if no dependencies change, a new ubuntu entry would need to be added every 6 months - that would not be the case for debian - unless dependencies change, a new debian entry would not need to be added until the next debian LTS version in 2-3 years - if every bi-annual ubuntu version is to be supported, then keeping 'buster' and 'sid' will be exactly zero additional work because whatever dependency versions ubuntu requires they take from debian testing (buster)

utils/fo-installdeps
maximilianhuber: > to be clear, the point i raised was that ubuntu support is actually more work to maintain than debian because it is guaranteed to requires a routine update to these install files every 6 months - even if no dependencies change, a new ubuntu entry would need to be added every 6 months - that would not be the case for debian - unless dependencies change, a new debian entry would not need to be added until the next debian LTS version in 2-3 years - if every ubuntu version is to be supported, then keeping 'buster' and 'sid' will be exactly zero additional work because whatever dependency versions ubuntu requires, they take from what is current in debian testing (currently buster)

But Ubuntu will at least not break in the future. Unless someone of the developers uses buster or sid, we will not recognize when the dependencies break on these systems.

And AFAIK we can not test the dependencies with travis (unless we set up docker files for that)?

bill-auger: wel as i said it will not "break" - it will just require a new version of the PG libs - and that WILL happen - and it will happen in buster and ubuntu at the same time (within a few months)

you could use debootstrap (or several other tools) for that - but yea it's probably more trouble than it's worth - to be consistent then i would suggest to drop support for the non-LTS versions of ubuntu because it is no easier to support those on travis right?

also as i mention on my original PR - the entire hassle of version numbers could be avoided by requiring the metapackages like:
```
postgresql-server-dev-all
instead of
postgresql-server-dev-9.5
or
postgresql-server-dev-9.6

postgresql php php-pgsql php-cli php-curl php-xml
instead of
postgresql-9.5 libapache2-mod-php5 php5 php5-pgsql php5-cli php5-curl
or
postgresql-9.6 libapache2-mod-php7.0 php7.0 php7.0-pgsql php7.0-cli php7.0-curl php7.0-xml;;
```

just some ideas to lessen the maintenance load - fwiw - thi sis a nice feature but very few upstreams manage a script like this - they usually just post the dependencies in the README and let it to the user and a ./configure script to prepare the environment

maximilianhuber: Using metapackages leads to other problems. We explicitly use specific php and psql versions since we do not want to support all of them.

In my opinion, if one tries to install fossology on a not supported system, one should fail early in the install-deps-step instead of silently later, when some features in the application break.

+1 for the idea with the `./configure` step for verifying dependencies.

max-wittig:
(from the discussion): add agent tests for PHP 7.0 and 7.1

utils/fo-installdeps
```diff
@@ -178,23 +178,23 @@ if [ $RUNTIME ]; then
...
               trusty)
-                 apt-get $YesOpt install postgresql-9.3;;
+                 apt-get $YesOpt install postgresql-9.3 libapache2-mod-php5 php5 php5-pgsql php5-cli php5-curl;;
               jessie)
-                 apt-get $YesOpt install postgresql-9.4;;
-              xenial)
-                 apt-get $YesOpt install postgresql-9.5;;
+                 apt-get $YesOpt install postgresql-9.4 libapache2-mod-php5 php5 php5-pgsql php5-cli php5-curl;;
+              xenial|yakkety)
+                 apt-get $YesOpt install postgresql-9.5 libapache2-mod-php5 php5 php5-pgsql php5-cli php5-curl;;
```
keestux: Drop the '5' from php packages?

bill-auger: keestux - to be clear, when you say drop support for php5, what you are really suggesting is to drop support for: squeeze, wheezy, lucid, natty, isadora, precise, oneiric, quantal, raring, saucy, trusty, jessie, xenial, yakkety

if you are only suggesting to drop the number '5' and require the meta-packages php, php-pgsql, php-cli, php-curl, etc. instead of php,5 php5-pgsql, php5-cli, php5-curl - that has already been suggested in this PR

keestux: Nevermind. I just got confused by Github which showed commit a4af5cc, but which is not present in the fresh clone that I made.

mcjaeger:
tested with debian 9 (net install) -> OK

tested with ubuntu 16.04.03 ->

1. composer fails, requires to be installed
```
sudo apt-get install php-mbstring
sudo apt-get install php-curl
```
to get the make install running ....

bill-auger:
@mcjaeger -

did you solve that error? i noticed it is exactly the same error as i got compiling on parabola

https://github.com/fossology/fossology/pull/864

PR #859, opened 2017-06-26
feat(install): provide easy install script

max-wittig:
This uses the instruction in the fossology wiki
https://github.com/fossology/fossology/wiki/Install-from-Source

### DIFF
--- a/install/easy-install
+++ b/install/easy-install
@@ -1,0 +1,29 @@
+#!/bin/bash
+# FOSSology Easy Install script
+# Copyright Siemens AG 2017
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+#
+# Description: Easy install script for fossology. Goes through the steps on the Fossology wiki
+set -e
+cd "$(dirname $(readlink -f "${BASH_SOURCE[0]}"))"/..
+sudo utils/fo-cleanold
+make clean
+sudo utils/fo-installdeps
+make
+sudo make install
+sudo /usr/local/lib/fossology/fo-postinstall
+sudo cp install/src-install-apache-example.conf /etc/apache2/sites-available/fossology.conf
+sudo ln -s /etc/apache2/sites-available/fossology.conf /etc/apache2/sites-enabled/fossology.conf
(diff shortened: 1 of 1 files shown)


install/easy-install
```diff
@@ -0,0 +1,28 @@
...
+# Description: Easy install script for fossology. Goes through the steps on the Fossology wiki
+set -e
+cd "$(dirname $(readlink -f "${BASH_SOURCE[0]}"))"/..
+sudo utils/fo-cleanold
+make clean
+sudo utils/fo-installdeps
+make
+sudo make install && sudo /usr/local/lib/fossology/fo-postinstall
+sudo cp install/src-install-apache-example.conf /etc/apache2/sites-available/fossology.conf
+sudo cp install/src-install-apache-example.conf /etc/apache2/sites-enabled/fossology.conf
```
maximilianhuber: Please use `a2ensite`, which creates the corresponding symlink

install/easy-install
```diff
@@ -0,0 +1,28 @@
...
+# without any warranty.
+#
+# Description: Easy install script for fossology. Goes through the steps on the Fossology wiki
+set -e
+cd "$(dirname $(readlink -f "${BASH_SOURCE[0]}"))"/..
+sudo utils/fo-cleanold
+make clean
+sudo utils/fo-installdeps
+make
+sudo make install && sudo /usr/local/lib/fossology/fo-postinstall
```
maximilianhuber: here is the `&&` unnecessary, just use a newline

install/easy-install
max-wittig: I don't know if that's the right way, because a2ensite only works on debian based systems.

maximilianhuber: Then please symlink via
```
ln -s /etc/apache2/sites-available/fossology.conf /etc/apache2/sites-enabled/fossology.conf
```
That is the expected way of placing the config in `sites-enabled`.

This will only work on some systems (e.g. the debian based ones), other systems (e.g. RHEL) might use `/etc/httpd/conf.d/` instead of `/etc/apache2/sites-enabled/`.

max-wittig:
(followup from discussion): actually it's okay how it is. No user-data is deleted and the offical wiki also references fo-cleanold without parameters as first [step](https://github.com/fossology/fossology/wiki/Install-from-Source).

PR #867, opened 2017-06-28
feat(obligations): extend database and obligation management with new data:

deveaud-m:
* ob_type: type of obligation {Obligation, Restriction, Risk or Right}
* ob_classification: the level of attention the obligation could become (optional)
* ob_modifications: does this obligation concerns modified source code (optional)
* ob_comment: from the obligation admin (optional)

Additionally: add the ability to delete an obligation and throws a confirm dialogue box.
CSV import and export functions are up-to-date with the new schema.

### DIFF
--- a/src/lib/php/Application/ObligationCsvExport.php
+++ b/src/lib/php/Application/ObligationCsvExport.php
@@ -52,7 +52,7 @@ class ObligationCsvExport {
   public function createCsv($ob=0)
   {
     $csvarray = array();
-    $sql = "SELECT ob_pk, ob_topic,ob_text
+    $sql = "SELECT ob_pk,ob_type,ob_topic,ob_text,ob_classification,ob_modifications,ob_comment
             FROM obligation_ref;";
     if ($ob>0)
     {
@@ -61,8 +61,10 @@ class ObligationCsvExport {
       $row = $this->dbManager->getSingleRow($sql,$stmt);
       $vars = $row ? array( $row ) : array();
       $liclist = $this->obligationMap->getLicenseList($ob);
+      $candidatelist = $this->obligationMap->getLicenseList($ob, True);
       array_shift($vars);
       array_push($vars,$liclist);
+      array_push($vars,$candidatelist);
       $csvarray = $vars;
     }
     else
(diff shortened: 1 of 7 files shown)


deveaud-m:
This is the new UI look:
addobligation

deveaud-m:
@max-wittig Thanks a lot for taking the time to test the changes.

The two branches `master` and `improved-obligations` are not compatible. The DB schema in the new branch has been extended and new data field were added according to the more precise requirements I've got from the users. If you fill up the table `obligation_ref` using the master branch, fields which are mandatory on the new branch are simply missing.

I'm not aware of the "official" process regarding bringing changes in the DB and ensuring compatibility with previous branches. Maybe @mcjaeger can help us merge this new feature more properly.

maximilianhuber:
@deveaud-m your formatter uses wrong settings, we indent with two spaces instead of 4

maximilianhuber:
Your commit 1 touches `+170 −43` lines, 2 touches `+283 −113` lines and the reformatting commit does`+778 −844`. Thus the reformatting changes clearly more code than you have touched and thus makes all `git blame` information in the corresponding files unusable.

Please only format exactly the lines you have touched.

deveaud-m:
@maximilianhuber Thanks for the review. I deleted the commit leading to way more syntax errors than expected, installed `phpcs` and patched the code according to FOSSology's coding guidelines. Now the CI jobs have all passed.

deveaud-m:
@max-wittig Thanks for the review.

Regarding the colors: you're right, I will restrict the color drop-down menu to one single choice. 

Regarding the associated licenses: I changed the DB schema for the associated licenses from `obligation_map` to `obligation_license_map` and `obligation_candidate_map`. This breaks the compatibility with the former data model. To enable a smooth transition between the previous version and this one, the list of obligations from the actual DB should be exported once and re-imported in the new CSV header. Once the data model is stable, this one-time step shouldn't be required anymore.

maximilianhuber:
> To enable a smooth transition between the previous version and this one, the list of obligations from the actual DB should be exported once and re-imported in the new CSV header.

This can not be done via a migration script in the DB?

PR #868, opened 2017-06-29
feat(login): add openid connect login

max-wittig:
[Wiki entry](https://github.com/fossology/fossology/wiki/OpenID-Connect)

### DIFF
--- a/src/composer.json
+++ b/src/composer.json
@@ -15,7 +15,8 @@
         "symfony/http-foundation": "2.6.*",
         "symfony/dependency-injection": "2.6.*",
         "twig/twig": "1.*",
-        "twig/extensions": "1.*"
+        "twig/extensions": "1.*",
+        "jumbojett/openid-connect-php": "0.1.*"
     },
 
     "require-dev": {
--- a/src/www/ui/core-auth.php
+++ b/src/www/ui/core-auth.php
@@ -194,31 +194,95 @@ class core_auth extends FO_Plugin
     return (@$_SERVER['REMOTE_ADDR']);
   }
 
-  /**
-   * \brief This is only called when the user logs out.
-   */
   public function Output()
   {
     $userName = GetParm("username", PARM_TEXT);
     $password = GetParm("password", PARM_TEXT);
     $referrer = GetParm("HTTP_REFERER", PARM_TEXT);
+    $openidProviderLogin = GetParm("openidProviderLogin", PARM_TEXT);
+
+    global $SysConf;
+    $openIdConnect = $SysConf["OPENID_CONNECT"];
+    $provider_name = $openIdConnect["provider_name"];
+    $provider_url = $openIdConnect["provider_url"];
+    $client_id = $openIdConnect["client_id"];
+    $client_secret = $openIdConnect["client_secret"];
+    $redirect_url = $openIdConnect["redirect_url"];
+    $claim = $openIdConnect["claim"];
+
+    $isOpenIDLogin = !empty($_GET["code"]) || !empty($openidProviderLogin);
+    if($isOpenIDLogin)
+    {
+
+      $client = new OpenIDConnectClient($provider_url, $client_id, $client_secret);
+      $client->setRedirectURL($redirect_url);
+      $client->authenticate();
+      try
+      {
+        $userName = $client->requestUserInfo($claim);
+      }
+      catch(Exception $e)
+      {
+        return false;
+      }
+
+      $validLogin = $this->checkOpenIDUser($userName);
+
+      if($validLogin)
+      {
+        $referrer = "?mod=home";
+      }
+      else
+      {
+        /**
+         * redirect the user to the auth page to signal
+         * login failure
+         */
+        $referrer = "?mod=auth";
+      }
+    }
+    else
+    {
+      $validLogin = $this->checkUsernameAndPassword($userName, $password);
+
+      $this->vars['loginFailure'] = !empty($userName) || !empty($password);
+      if (!empty($userName) && $userName != 'Default User')
+      {
+        $this->vars['userName'] = $userName;
+      }
+    }
+
     if (empty($referrer))
     {
       $referrer = GetArrayVal('HTTP_REFERER', $_SERVER);
     }
-    $referrerQuery = parse_url($referrer,PHP_URL_QUERY);
-    if($referrerQuery) {
+
+    /**
+     * redirects the user to the auth page, incase of failed OpenId Login or failed
+     * normal login
+     */
+    if($validLogin || $isOpenIDLogin)
+    {
+      return new RedirectResponse($referrer);
+    }
+
+    $referrerQuery = parse_url($referrer, PHP_URL_QUERY);
+    if ($referrerQuery)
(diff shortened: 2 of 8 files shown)


mcjaeger:
I think we need to see what happened with the old OAuth PR code

maximilianhuber:
I have a theoretical question:

if there is some user with username **user** in the system which does not use the openid login, i.e. has no account on the auth provider, is then the following attack possible:
create an account with username **user** on the auth-provider and log in with your new defined credentials.

As far as I understand the code (after rough inspection) you only check
- [whether the passed credentials are valid by asking the auth provider](https://github.com/fossology/fossology/pull/868/files#diff-d75185ba1f976f7b22e254541e49e3a7R219) and
- [whether the username is in the system](https://github.com/fossology/fossology/pull/868/files#diff-d75185ba1f976f7b22e254541e49e3a7R323)

Can you please tell me where I am wrong and why this simple attack does not work?

src/www/ui/page/HomePage.php
```diff
@@ -21,6 +21,7 @@
 use Fossology\Lib\Plugin\DefaultPlugin;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use OpenIDConnectClient;
```
maximilianhuber: this statement looks displaced.

src/composer.json
```diff
@@ -15,7 +15,8 @@
         "symfony/http-foundation": "2.6.*",
         "symfony/dependency-injection": "2.6.*",
         "twig/twig": "1.*",
-        "twig/extensions": "1.*"
+        "twig/extensions": "1.*",
+        "jumbojett/openid-connect-php": "0.1.*"
```
maximilianhuber: On github is already version `0.3` released

maximilianhuber: This is licensed under Apache2.0 which is incompatible with our GPLv2. That might be a problem for distribution! (ping @mcjaeger)

src/www/ui/core-auth.php
```diff
@@ -194,53 +194,109 @@ function getIP()
...
+    $client_secret = $openIdConnect["client_secret"];
+    $redirect_url = $openIdConnect["redirect_url"];
+    $claim = $openIdConnect["claim"];
+
+    $isOpenIDLogin = !empty($_GET["code"]) || !empty($openidProviderLogin);
+    if($isOpenIDLogin)
     {
-      $referrer = GetArrayVal('HTTP_REFERER', $_SERVER);
+
+      $client = new OpenIDConnectClient($provider_url, $client_id, $client_secret);
```
maximilianhuber: there is no import corresponding to the new dependency? Please add an explicit import (if applicable).

max-wittig:
<pre>
create an account with username user on the auth-provider and log in with your new defined credentials
</pre>
Yes that would be possible, but you should trust the OpenID Provider you write in the config file. It's the same with the current Oauth Implementation

maximilianhuber:
But if you add Github (which is supported AFAIK) as auth-provider, everyone can create accounts with arbitrary usernames?

max-wittig:
No. The accountname has to already be in the fossology database. No accounts can be created using this implementation

maximilianhuber:
Yes, you can't create a user in fossology. I talk about the other direction.

If you deploy a default fossology installation (with the user fossy), add Github as OpenID Provider and I own the account **fossy** on Github, I am able to log into your instance, right?

maximilianhuber:
#### To clearify my point:
> ```
> create an account with username user on the auth-provider and log in with your new defined credentials
> ```
> Yes that would be possible, but you should trust the OpenID Provider you write in the config file. It's the same with the current Oauth Implementation

If everyone is able to **choose an arbitrary username**, you also have to trust everyone, not only the Provider.

max-wittig:
If you would do that, yes. The administrator needs to edit a config file and add the OpenID provider that is trusted.
**But you're right. This needs clarification in the wiki**

src/composer.json
max-wittig: If I understand this correctly, we already have Apache-2.0 in the Fossology project as a dependency
https://github.com/fossology/fossology/blob/master/NOTICES#L4068

mcjaeger:
how about asking the JumboJett person for relicensing the software as GPL-2.0?

src/composer.json
mcjaeger: this is apparently test files, but building functionality upon Apache-2.0 libs is maybe more difficult. How about asking the author if reliense is possible?

maximilianhuber:
Potential alternative library: https://pear.php.net/package/OpenID/

max-wittig:
(from the discussion): use basic auth, as there doesn't seem to be a good PHP OpenID Connect Library alternative right now...

NicolasToussaint:
Is the OpenID feature totally abandoned ?
If so, we should update the wiki page [OpenID-Connect](https://github.com/fossology/fossology/wiki/OpenID-Connect)

bhcopeland:
This feature looks abandoned. The code doesn't seem to be in the project, so shouldn't the wiki be updated?

maxhbr:
The content of the mentioned wiki entry was:

```markdown
# To use Fossology with a OpenID provider

Make sure, that you trust the OpenID Provider. In most cases it should be controlled by your authority, because otherwise users would be able to login with any already existing username in the Fossology database. 

***

## 1. Get the credentials from your OpenID Connect Identify Provider(e.g. Gitlab, ..)

Example Gitlab:
![image](https://user-images.githubusercontent.com/6639323/27686561-8f4862d4-5cd3-11e7-9788-47c07f0b14ee.png)
![image](https://user-images.githubusercontent.com/6639323/27686594-b58a952a-5cd3-11e7-85f0-4eaca5bccd5b.png)

## 2. Insert these credentials into your `/etc/local/etc/fossology/fossology.conf` file

[Here](https://docs.gitlab.com/ee/integration/openid_connect_provider.html) you can find the claims, that are available for Gitlab. The claim is in your fossology instance then used as username. The user has to exist already in Fossology to successfully authentificate. For Fossology it's important that you redirect to `?mod=auth`, otherwise it won't work
<pre>
[OPENID_CONNECT]
provider_url = gitlab.com
provider_name = Gitlab
client_id = cbea8587ae802c65a95d7abe3c9db97dc83000048edc7daad82124d77dde82b3
client_secret = 5edb5f38b5f9998c1278c4d5d78ffee35642936d9e3acc4cc5d30a437d3856fc
redirect_url = "http://fossology.example.com/repo?mod=auth"
claim = username
</pre>

## 3. On the Login page, the user should now see a "Login with PROVIDER_NAME" Button
![image](https://user-images.githubusercontent.com/6639323/27686980-5b2a9164-5cd5-11e7-8393-53e719de46d4.png)

## 4. There conditions are required for OpenID Connect to work
* The data of your OpenID Connect Provider was correctly inserted into the fossology.conf file
* The OpenID connect provider accepts the authentification of the user
* The fossology username is the same as the claim that was returned from the OpenID provider
```

maxhbr:
@bhcopeland thank you for pointing out.

PR #869, opened 2017-06-30
feat(copyright): split tables, separate tables for copyright and emai…

shaheemazmalmmd:
This pull request will split the copyright table.

copyrights will be stored in "copyright" table.

email/url/author will be stored in "author" table.

### DIFF
--- a/install/db/dbmigrate_copyright-author.php
+++ b/install/db/dbmigrate_copyright-author.php
@@ -1,0 +1,42 @@
+<?php
+/***********************************************************
+ Copyright (C) 2017 Siemens AG
+ 
+ This program is free software; you can redistribute it and/or
+ modify it under the terms of the GNU General Public License
+ version 2 as published by the Free Software Foundation.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License along
+ with this program; if not, write to the Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ ***********************************************************/
+
+
+/**
+ * \brief reinsert the old email/url/author to new author table
+ * and delete the same from copyright table
+ */
+
+  $countAuthorColumns = $dbManager->getSingleRow("SELECT count(*) FROM copyright WHERE type <> 'statement' AND content IS NOT NULL;",array(),'getCountAuthorColumns');
+
+  if($countAuthorColumns['count'] > 0){
+
+    echo "Insert email/url/author data from copyright to author table... \nIt takes some time depending number of columns... \n";
+
+    $dbManager->queryOnce("INSERT INTO author (ct_pk, agent_fk, pfile_fk, content, hash, type, copy_startbyte, copy_endbyte, is_enabled) 
+                           SELECT ct_pk, agent_fk, pfile_fk, content, hash, type, copy_startbyte, copy_endbyte, is_enabled 
+                           FROM copyright WHERE type <> 'statement' AND content IS NOT NULL;");
+    $countInsertedAuthorColumns = $dbManager->getSingleRow("SELECT count(*) FROM author au INNER JOIN copyright co 
+                           ON au.pfile_fk = co.pfile_fk WHERE au.ct_pk = co.ct_pk AND au.content = co.content;",array(),'getCountInsertedAuthorColumns');      
+    if($countAuthorColumns['count'] == $countInsertedAuthorColumns['count']){
+      echo "Deleting the email/url/author data from copyright table...\n";
+      $dbManager->queryOnce("DELETE FROM copyright WHERE type <> 'statement' AND content IS NOT NULL;");
+    }else{
+      echo "Something went wrong please execute the postinstall again...\n";
+    }
+  }
--- a/src/copyright/agent/database.cc
+++ b/src/copyright/agent/database.cc
@@ -300,15 +300,22 @@ bool CopyrightDatabaseHandler::insertNoResultInDatabase(long int agentId, long i
 
 bool CopyrightDatabaseHandler::insertInDatabase(DatabaseEntry& entry) const
 {
+  std::string tableName = IDENTITY;
+ 
+  if("author" == entry.type ||
+     "email" == entry.type ||
+     "url" == entry.type){
+    tableName = "author";
+   }
+
   return dbManager.execPrepared(
     fo_dbManager_PrepareStamement(
       dbManager.getStruct_dbManager(),
-      "insertInDatabase",
-      "INSERT INTO "
-      IDENTITY
-      "(agent_fk, pfile_fk, content, hash, type, copy_startbyte, copy_endbyte)"
-        " VALUES($1,$2,$3,md5($3),$4,$5,$6)",
-      long, long, char*, char*, int, int
+      ("insertInDatabaseFor" + tableName).c_str(),
+      ("INSERT INTO "+ tableName +
+      "(agent_fk, pfile_fk, content, hash, type, copy_startbyte, copy_endbyte)" +
+        " VALUES($1,$2,$3,md5($3),$4,$5,$6)").c_str(),
+        long, long, char*, char*, int, int
     ),
     entry.agent_fk, entry.pfile_fk,
     entry.content.c_str(),
--- a/src/lib/php/Dao/CopyrightDao.php
+++ b/src/lib/php/Dao/CopyrightDao.php
@@ -70,12 +70,18 @@ class CopyrightDao extends Object
     }
 
     $statementName = __METHOD__.$tableName;
-    $sql = "SELECT * FROM $tableName WHERE copy_startbyte IS NOT NULL AND pfile_fk=$1";
     $params = array($pFileId);
-    if($agentId!=0) {
+    $addAgentValue = "";
+    if(!empty($agentId)) {
       $statementName .= '.agentId';
-      $sql .= ' AND agent_fk=$2';
+      $addAgentValue = ' AND agent_fk=$2';
       $params[] = $agentId;
+    }
+    $getHighlightForTableName = "SELECT * FROM $tableName WHERE copy_startbyte IS NOT NULL AND pfile_fk=$1 $addAgentValue";
+    if($tableName != "copyright"){
+      $sql = $getHighlightForTableName;
+    }else{
+      $sql = "$getHighlightForTableName UNION SELECT * FROM author WHERE copy_startbyte IS NOT NULL AND pfile_fk=$1 $addAgentValue";
     }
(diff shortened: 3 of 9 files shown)


max-wittig:
**Seems to work fine functionally**, but I'm getting this error message while running `fo-postinstall`.
Did I miss something? I'm using existing uploads in the repository
Or is it maybe only for new instances?
(I'm using PHP 7.0 with debian 9 and postgresql 9.6.3)
<pre>
Insert email/url/author data from copyright to author table... 
 It takes some time depending number of columns... 
PHP Warning:  pg_query(): Query failed: ERROR:  column "statement" does not exist
LINE 3: ...                     FROM copyright WHERE type <> "statement...
                                                             ^ in /usr/local/share/fossology/lib/php/Db/Driver/Postgres.php on line 81
[2017-07-04 13:40:52] /usr/local/share/fossology/lib/php/libschema.php.CRITICAL: ERROR:  column "statement" does not exist LINE 3: ...                     FROM copyright WHERE type <> "statement...                                                              ^ [] []
PHP Fatal error:  Uncaught Fossology\Lib\Exception: error executing: INSERT INTO author (ct_pk, agent_fk, pfile_fk, content, hash, type, copy_startbyte, copy_endbyte, is_enabled) 
                           SELECT ct_pk, agent_fk, pfile_fk, content, hash, type, copy_startbyte, copy_endbyte, is_enabled 
                           FROM copyright WHERE type <> "statement" AND content IS NOT NULL;

ERROR:  column "statement" does not exist
LINE 3: ...                     FROM copyright WHERE type <> "statement...
                                                             ^ in /usr/local/share/fossology/lib/php/Db/DbManager.php:120
Stack trace:
#0 /usr/local/share/fossology/lib/php/Db/DbManager.php(160): Fossology\Lib\Db\DbManager->checkResult(false, 'INSERT INTO aut...')
#1 /usr/local/lib/fossology/dbmigrate_copyright-author.php(32): Fossology\Lib\Db\DbManager->queryOnce('INSERT INTO aut...')
#2 /usr/local/lib/fossology/fossinit.php(321): require_once('/usr/local/lib/...')
#3 {main}
  thrown in /usr/local/share/fossology/lib/php/Db/DbManager.php on line 120
Database connectivity is good.
</pre>

I could also provide more debug info, if you need it

shaheemazmalmmd:
May be because of double quotes postgres is assuming "statement" as column. will make changes.

install/db/dbmigrate_copyright-author.php
```diff
@@ -0,0 +1,43 @@
...
+ ***********************************************************/
+
+
+/**
+ * \brief reinsert the old email/url/author to new author table
+ * and delete the same from copyright table
+ */
+
+
+  $countAuthorColumns = $dbManager->getSingleRow("SELECT count(*) FROM copyright WHERE type <> 'statement' AND content IS NOT NULL;",array(),__METHOD__);
```
max-wittig: Maybe we should use `!=` instead of `<>` because it clearly reads `not equal`?
I had to look up, what `<>` does and both are supported. What do you think?

maximilianhuber: What is the value of `__METHOD__` at this line of code?

maximilianhuber:
This also migrates correctly in the docker-compose setup:
```
$ docker-compose up --build
[...]
fossology-scheduler_1  | DB schema has been updated for fossology.
fossology-scheduler_1  | Database schema update completed successfully.
fossology-scheduler_1  | Update reference licenses
fossology-scheduler_1  | Insert email/url/author data from copyright to author table... 
fossology-scheduler_1  | It takes some time depending number of columns... 
fossology-scheduler_1  | Deleting the email/url/author data from copyright table...
[...]
```
And the result in the UI was checked.

src/copyright/agent/database.cc
```diff
@@ -300,21 +300,38 @@ bool CopyrightDatabaseHandler::insertNoResultInDatabase(long int agentId, long i
...
-      "(agent_fk, pfile_fk, content, hash, type, copy_startbyte, copy_endbyte)"
-        " VALUES($1,$2,$3,md5($3),$4,$5,$6)",
-      long, long, char*, char*, int, int
-    ),
-    entry.agent_fk, entry.pfile_fk,
-    entry.content.c_str(),
-    entry.type.c_str(),
-    entry.copy_startbyte, entry.copy_endbyte
-  );
+  if(!strcmp(entry.type.c_str(),IDENTITY) || strcmp(entry.type.c_str(),"statement")){
```
shaheemazmalmmd: "Statement" implies to copyrights. and in current database design only copyrights have different name in type. others have same  as IDENTITY.

maximilianhuber: but `IDENTITY` might be equal to `"copyright"`: https://github.com/siemens/fossology/blob/a5a5276ce806f330756957b9713957f7ad3ff49d/src/copyright/agent/identity.hpp#L31

The only other option for `IDENTITY` is `"ecc"` (see `identity.hpp`). This implies that only ecc is placed in another table called ecc and everything else is placed in the table called `author`? At least this is how I understand the code.

maximilianhuber: ### to explain my confusion:

You can rewrite this roughly in the following way (related to the [other comment below](https://github.com/fossology/fossology/pull/869/files#r126643170)):
```cc
if(  !strcmp(entry.type.c_str(),IDENTITY) // [1]
   || strcmp(entry.type.c_str(),"statement") // [2]
  ){
    tablename = "author";
} else { 
    tablename = IDENTITY;
}
// [...]
```

then
- the statement [1] says that "if you have to handle some entry whose type does not match your identity, then write the result to the `author`-table" and
- the statement [2] says something like "if you have to handle some entry of type `statement` then write it to the `author`-table"

In the case when `IDENTITY="copyright"` this means that you write everything to the `author` table, since either the type is never equal to `copyright` (i.e. is `statement` or something else for non-copyrights).

##

This result has not been verified in the database.

shaheemazmalmmd: >> The only other option for IDENTITY is "ecc" (see identity.hpp). This implies that only ecc is placed in another table called ecc and everything else is placed in the table called author? At least this is how I understand the code.

Here i'm comparing the IDENTITY with "entry.type.c_str()" which doesn't contain "copyright". More over there are almost 4 types "ecc, statement, email, url, author" that's why it would be better to compare it with IDENTITY.

shaheemazmalmmd: >> In the case when IDENTITY="copyright" this means that you write everything to the author table, since either the type is never equal to copyright (i.e. is statement or something else for non-copyrights).

Not everything to author table. Only "Author/Email/URL" data, copyright statements gets inserted to copyright table. and ECC findings to ecc table.

maximilianhuber: But copyrights have the type `"statement"` which makes `strcmp(entry.type.c_str(),"statement")` true and thus writes them to the `author` table.

maximilianhuber: A clean (and potentially correct) implementation might be:
```cc
if(strcmp(entry.type.c_str(),"author") || strcmp(entry.type.c_str(),"email") || strcmp(entry.type.c_str(),"url")){
    tablename = "author";
} else { 
    tablename = IDENTITY;
}
[...]
```

maximilianhuber: My confusion is now reduced by the following insight: **`strcmp(str1, str2)` evaluates to `false`** if `str1` is equal to `str2`.

I still think that the condition in the if is not very readable and should be refactored.

maximilianhuber:
This pull request breakes ECC.

I have tried it with a tar.gz only containing the following file
```
this is a Test for Laser Cryptography Crypto
Elliptic Curve
```
and then I am no longer able to open the ECC tab for that upload (other uploads with no ECC findings still show the empty ECC table.
It fails with the following error in the postgres log:
```
ERROR:  bind message supplies 2 parameters, but prepared statement "Fossology\Lib\Dao\CopyrightDao::getHighlightsecc.agentId" requires 1
STATEMENT:  SELECT * FROM ecc WHERE copy_startbyte IS NOT NULL AND pfile_fk=$1
```

The root cause for this problem is the broken if statement in the file `database.cc` which is already under heavy discussion. It is writing `ecc` statements to the `authors` table since `IDENTITY==ecc`:
```
fossology=# select * from author;
 ct_pk | agent_fk | pfile_fk |                         content                          |               hash               |  type  | copy_startbyte | copy_endbyte | is_enabled 
-------+----------+----------+----------------------------------------------------------+----------------------------------+--------+----------------+--------------+------------
[...]
    98 |        5 |       11 | MAINTAINERCLEANFILES                                     | 28e706ab0c4d709383ca1a4788ac3ef1 | author |           6958 |         6978 | t
    99 |        5 |       11 | maintainers to use                                       | 1df0d8923c31fad13b97f475a07c4e0f | author |           7645 |         7663 | t
   100 |       13 |       33 | Laser Cryptography Crypto                                | ee6d4424674af847e6fa62def91fd4b5 | ecc    |             22 |           47 | t
   102 |       13 |       33 | Laser Cryptography Crypto                                | ee6d4424674af847e6fa62def91fd4b5 | ecc    |             22 |           47 | t
(65 rows)
```

src/copyright/agent/database.cc
maximilianhuber: Currently the if statement does not work for `ecc` since if `entry.type.c_str() == "ecc" == IDENTITY` the statment
```
!strcmp(entry.type.c_str(),IDENTITY) || strcmp(entry.type.c_str(),"statement"
```
evaluates to true and writes `ecc` information to the `author` table.

src/lib/php/Dao/CopyrightDao.php
```diff
@@ -70,13 +70,19 @@ public function getHighlights($uploadTreeId, $tableName="copyright", $agentId=0,
...
     $params = array($pFileId);
-    if($agentId!=0) {
+    $addAgentValue = "";
+    if(!empty($agentId)) {
       $statementName .= '.agentId';
-      $sql .= ' AND agent_fk=$2';
+      $addAgentValue = ' AND agent_fk=$2';
       $params[] = $agentId;
     }
+    $getHighlightForTableName = "SELECT * FROM $tableName WHERE copy_startbyte IS NOT NULL AND pfile_fk=$1$addAgentValue";
```
maximilianhuber: minor inconsistency:
here you write `$1$addAgentValue`
below you write `$1 $addAgentValue`

maximilianhuber:
Deployed and tested: works fine.

maximilianhuber:
It looks like your commit message is to long (longer then 72 chars) which fails the conventional-changelog test:


@max-wittig can we modify the changelog test so that it is less strict
@mcjaeger i would count this as **all tests passed**

max-wittig:
I thought that was fine, as 72 chars is that is recommended for most projects to be descriptive, short commit message
See: [Source1](https://www.reddit.com/r/git/comments/20ko8g/why_do_a_lot_of_developers_apply_a_72character/) [Source2](https://github.com/torvalds/linux/pull/17#issuecomment-5661185)

I can change the restrictions or this commit message may be changed. Just let me know

shaheemazmalmmd:
I will change the commit message for now. lets discuss this restrictions later :)

PR #871, opened 2017-07-06
feat(organize): allow searching for folders to copy/move to

max-wittig:
Uses select2, which is already part of Fossology
image


src/www/ui/template/components/select-folder.html.twig
```diff
@@ -7,7 +7,7 @@
 {% if selected is not defined %}
   {% set selected = -1 %}
 {% endif %}
-<select name="{{ name }}"{% if id is defined %} id="{{ id }}"{% endif %}>
+<select style="width: 100%" name="{{ name }}"{% if id is defined %} id="{{ id }}"{% endif %}>
```
maximilianhuber: Inline styling...

PR #873, opened 2017-07-07
feat(delete): add select2 to folder select

max-wittig:
image
Uses existing select2 libary

### DIFF
--- a/src/delagent/ui/admin-upload-delete.php
+++ b/src/delagent/ui/admin-upload-delete.php
@@ -161,7 +161,9 @@ class admin_upload_delete extends FO_Plugin
     $V.= "<ol>\n";
     $text = _("Select the folder containing the file to delete: ");
     $V.= "<li>$text";
-    $V.= "<select name='folder' ";
+    $V.= "<br />";
+    $V.= "<select id='folder_select' name='folder' ";
+
     $V.= "onLoad='Uploads_Get((\"" . Traceback_uri() . "?mod=upload_options&folder=-1' ";
     $V.= "onChange='Uploads_Get(\"" . Traceback_uri() . "?mod=upload_options&folder=\" + this.value)'>\n";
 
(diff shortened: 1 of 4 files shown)


mcjaeger:
pls. consider changing title and commit message, maybe...

src/delagent/ui/admin-upload-delete.php
```diff
@@ -190,6 +193,14 @@ public function Output()
     $text = _("Delete");
     $V.= "<input type='submit' value='$text'>\n";
     $V.= "</form>\n";
+
+    $V.= <<<EOT
```
maximilianhuber: why do you use heredoc syntax here, and [at other places](https://github.com/siemens/fossology/blob/0159a0f38590e5d8739e401619da039b9b9d6102/src/www/ui/admin-dashboard.php#L129-L165) the usual multiline string syntax?

src/delagent/ui/admin-upload-delete.php
```diff
@@ -161,13 +161,16 @@ public function Output()
     $V.= "<ol>\n";
     $text = _("Select the folder containing the file to delete: ");
     $V.= "<li>$text";
-    $V.= "<select name='folder' ";
+    $V.= "<br />";
+    $V.= "<div>";
```
maximilianhuber: Why is this additional `div` necessary?

src/delagent/ui/admin-upload-delete.php
max-wittig: I choose multiline string syntax, if I need to use a PHP variable e.g. here:
<pre>phpversion($currentExtensionName); </pre>
https://github.com/siemens/fossology/blob/0159a0f38590e5d8739e401619da039b9b9d6102/src/www/ui/admin-dashboard.php#L129-L165

Otherwise I use EOT, because it does look better imho and you can be sure that there is no PHP code executed in the EOT.

Is that bad practice?

maximilianhuber: The php statements in heredocs get also executed and replaced. If you want to prevent this use the nowdoc syntax, i.e.
```
    $V.= <<<'EOT'
[...]
```

mcjaeger:
needs to be tested!

PR #874, opened 2017-07-07
feat(delete): allow deletion of multiple uploads

max-wittig:

image

### DIFF
--- a/src/delagent/ui/admin-upload-delete.php
+++ b/src/delagent/ui/admin-upload-delete.php
@@ -1,7 +1,7 @@
 <?php
 /***********************************************************
  Copyright (C) 2008-2013 Hewlett-Packard Development Company, L.P.
- Copyright (C) 2015-2016 Siemens AG
+ Copyright (C) 2015-2017 Siemens AG
 
  This program is free software; you can redistribute it and/or
  modify it under the terms of the GNU General Public License
@@ -19,12 +19,12 @@
 
 use Fossology\Lib\Auth\Auth;
 use Fossology\Lib\Dao\UploadDao;
-
+use \delagent\ui\DeleteMessages;
+use \delagent\ui\DeleteResponse;
 /**
  * \file admin_upload_delete.php
  * \brief delete a upload
  */
-
 define("TITLE_admin_upload_delete", _("Delete Uploaded File"));
 
 /**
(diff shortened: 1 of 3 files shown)


mcjaeger:
it would be interested for testing to test deleting a copy of the upload and the original at the same time

max-wittig:
Actually it works fine with copies of uploads, but while testing I noticed that the multi-select actually didn't work when changing folders. Fixed it now.

src/delagent/ui/admin-upload-delete.php
```diff
@@ -114,27 +115,79 @@ function Delete($uploadpk, $Depends = NULL)
...
+      }
+
+      if(in_array(DeleteMessages::NO_PERMISSION, $errorMessages))
+      {
+        $displayMessage .= "No permission to delete " .
+          array_count_values($errorMessages)[DeleteMessages::NO_PERMISSION]. " uploads<br/>";
+      }
+
+      $displayMessage .= _("Deletion of ") .
+        (sizeof($uploadpks)-sizeof($errorMessages)) . " projects queued";
```
maximilianhuber: here you use `gettext` inconsistently. You add it for the first half of the sentence but not the second.

src/delagent/ui/admin-upload-delete.php
```diff
@@ -114,27 +115,79 @@ function Delete($uploadpk, $Depends = NULL)
...
+    {
+      $V .= DisplayMessage($deleteResponse->getDeleteMessageString().$deleteResponse->getAdditionalMessage());
+    }
+    else
+    {
+      $displayMessage = "";
+
+      if(in_array(DeleteMessages::SCHEDULING_FAILED, $errorMessages))
+      {
+        $displayMessage .= "<br/>Scheduling failed for " .
```
maximilianhuber: your output only add the leading `<br/>` only in this case, in the other cases it is missing. This might be an inconsistency

src/delagent/ui/admin-upload-delete.php
```diff
@@ -114,27 +115,79 @@ function Delete($uploadpk, $Depends = NULL)
...
+        $errorMessages[] = $deleteResponse;
+      }
+    }
+
+    /**
+     * Output normal message, if only one upload
+     */
+    if(sizeof($uploadpks) < 2)
+    {
+      $V .= DisplayMessage($deleteResponse->getDeleteMessageString().$deleteResponse->getAdditionalMessage());
```
maximilianhuber: if `sizeof($uploadpks) == 0` then `$deleteResponse==""`. I do not know what the behaviour of `""->getDeleteMessageString()` is.

src/delagent/ui/admin-upload-delete.php
```diff
@@ -114,27 +115,79 @@ function Delete($uploadpk, $Depends = NULL)
...
   /**
+   * @param $uploadpks
+   * @brief starts deletion and handles error messages
+   * @return string
+   */
+  function initDeletion($uploadpks)
+  {
+    $V = "";
+    $errorMessages = [];
+    $deleteResponse = "";
```
maximilianhuber: initialized as empty string, but later used as type `DeleteResponse`

src/delagent/ui/admin-upload-delete.php
```diff
@@ -114,27 +115,79 @@ function Delete($uploadpk, $Depends = NULL)
...
+    for($i=0; $i < sizeof($uploadpks); $i++)
+    {
+      $deleteResponse = $this->TryToDelete($uploadpks[$i]);
```
maximilianhuber: loosing the explicit cast to `int` makes this a security problem, especially with the unclean usage in `admin_upload_delete->Delete` (please also fix the `Delete` function).

src/delagent/ui/admin-upload-delete.php
max-wittig: The breaks are alright in this case from a visual standpoint. It's done so that it only breaks the warning messages., not the info message. That's why you got breaks at the end of the other cases

src/delagent/ui/admin-upload-delete.php
```diff
@@ -114,27 +115,84 @@ function Delete($uploadpk, $Depends = NULL)
...
+    {
+      $deleteResponse = $this->TryToDelete(intval($uploadpks[$i]));
+
+      if($deleteResponse->getDeleteMessageCode() != DeleteMessages::SUCCESS)
+      {
+        $errorMessages[] = $deleteResponse;
+      }
+    }
+
+    if(sizeof($uploadpks) <= 0)
```
maximilianhuber: It would be cleaner to move that to the top of the function. That is where one expects it.

src/delagent/ui/admin-upload-delete.php
```diff
@@ -114,27 +115,84 @@ function Delete($uploadpk, $Depends = NULL)
...
+    {
+      return DisplayMessage("No uploads selected");
+    }
+
+    /**
+     * Output normal message, if only one upload
```
maximilianhuber: This comment is necessary.

PR #877, opened 2017-07-14
fix(license): remove junk characters from LGPL-2.1 license text

shaheemazmalmmd:


mcjaeger:
pls. help a little more and .... could you pls. specify what "good" means. Like you have tested it or did you look at the code?

max-wittig:
Sorry @mcjaeger I just looked at the code changes before. Now I also did some tests with the fossology instance build from this PR.
Tested by scanning a file containing the LGPL-2.1 License. Nomos and Monk are recognizing LGPL-2.1 correctly. Also the entry in `Admin > License Admin > Select License` looks good.

PR #887, opened 2017-08-07
Feature: JSON output

daniellandau:
Optional JSON output in some of the CLI scanners.


daniellandau:
@maximilianhuber, now the tests pass.

maxhbr:
Can you tell me which compiler versions (in the travis test) are supported?

maxhbr:
It would be great if you could add tests. This would also help us to understand the feature.

daniellandau:
Everything in nomos and monk is supported for all compiler versions. In copyright, gcc-4.8 and gcc-4.6 are not supported.

The idea is to output valid JSON for further analysis in other tools from the scanners. I don't have time at the moment to work on tests.

maxhbr:
@mcjaeger: Note: this PR adds a new dependency as file into our code base (14000 LOC)

steffen-weber:
@daniellandau it should be possible to create and include /install/scripts/install-json.sh that pulls https://github.com/nlohmann/json/blob/develop/src/json.hpp to ensure that Fossology uses the latest version

daniellandau:
I would imagine fetching an unverified update from the internet probably doesn't bode well with for example Debian with their reproducible builds. Wouldn't it in any case be better to update the version of json.hpp purposefully as needed, with a git log of when updates happen?

steffen-weber:
Good point for not pulling the develop version, but fetching https://github.com/nlohmann/json/releases/download/v2.1.1/json.hpp allows to view tests and history of json.hpp and avoids duplication.

daniellandau:
Ok. The just added commit fetches `json.hpp` on the fly. I think vendoring it would be a better idea, though.

PR #889, opened 2017-08-10
fix(travis): do not build multiple times

maxhbr:
Currently travis does the `make`-step to often. For example when the dockerized setup is tested and the setup on the host can be ignored.

This PR refactors the `.travis.yml` file to improve build times and to prevent tests from failing due to unrelated stuff.

### The mapping between the tests is as follows:
- The tests **.1** to **.3** do no longer do `make`
- The test for PHP 7.1 was added as `allowed-failues` but not as real matrix entry, this was changed => test-.17
- The docker tests now check whether the scheduler is running and responding

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@
 # build FOSSology on Travis CI - https://travis-ci.org/
 
 language: php
+dist: trusty
 php:
   - '5.6'
 addons:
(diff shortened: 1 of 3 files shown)


maximilianhuber:
@max-wittig This now also tests PHP7.{0,1} as allowed failures.

PR #891, opened 2017-08-16
feat(candidate): add delete feature to candidate licenses

shaheemazmalmmd:
* Delete option added for the candidate license.
* Delete only works when candidate license not in use.
* If the candidate license is in use then show the file paths where this license is used.

### DIFF
--- a/src/www/ui/page/AdminLicenseCandidate.php
+++ b/src/www/ui/page/AdminLicenseCandidate.php
@@ -1,6 +1,6 @@
 <?php
 /***********************************************************
- * Copyright (C) 2014-2015 Siemens AG
+ * Copyright (C) 2014-2017 Siemens AG
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
@@ -21,9 +21,11 @@ namespace Fossology\UI\Page;
 use Fossology\Lib\Auth\Auth;
 use Fossology\Lib\BusinessRules\LicenseMap;
 use Fossology\Lib\Dao\LicenseDao;
+use Fossology\Lib\Dao\TreeDao;
 use Fossology\Lib\Db\DbManager;
 use Fossology\Lib\Plugin\DefaultPlugin;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Response;
 
 class AdminLicenseCandidate extends DefaultPlugin
@@ -36,7 +38,10 @@ class AdminLicenseCandidate extends DefaultPlugin
   private $highlightRenderer;
   /** @var TextRenderer */
   private $textRenderer;
+  /** @var LicenseDao */
   private $licenseDao;
+  /** @var TreeDao */
+  private $treeDao;
 
 
   function __construct()
@@ -52,6 +57,7 @@ class AdminLicenseCandidate extends DefaultPlugin
     $this->highlightRenderer = $this->getObject('view.highlight_renderer');
     $this->textRenderer = $this->getObject('view.text_renderer');
     $this->licenseDao = $this->getObject('dao.license');
+    $this->treeDao = $this->getObject('dao.tree');
   }
 
   /**
@@ -61,6 +67,7 @@ class AdminLicenseCandidate extends DefaultPlugin
    */
   protected function handle(Request $request)
   {
+
     $rf = intval($request->get('rf'));
     if ($rf<1)
     {
@@ -137,6 +144,9 @@ class AdminLicenseCandidate extends DefaultPlugin
         }
         $vars['message'] = 'Sorry, this feature is not ready yet.';
         break;
+      case 'deletecandidate':
+        return $this->doDeleteCandidate($rf);
+        break;
     }
       
     return $this->render('admin_license_candidate-merge.html.twig', $this->mergeWithDefault($vars));
@@ -152,14 +162,17 @@ class AdminLicenseCandidate extends DefaultPlugin
     $dbManager->prepare($stmt = __METHOD__, $sql);
     $res = $dbManager->execute($stmt);
     $aaData = array();
+    $delete = "";
     while ($row = $dbManager->fetchArray($res))
     {
       $link = Traceback_uri() . '?mod=' . self::NAME . '&rf=' . $row['rf_pk'];
       $edit = '<a href="' . $link . '"><img border="0" src="images/button_edit.png"></a>';
+      $delete = '<img border="0" id="deletecandidate'.$row['rf_pk'].'" onClick="deleteCandidate('.$row['rf_pk'].')" src="images/icons/close_16.png">';
+
       $aaData[] = array($edit, htmlentities($row['rf_shortname']),
           htmlentities($row['rf_fullname']),
           '<div style="overflow-y:scroll;max-height:150px;margin:0;">' . nl2br(htmlentities($row['rf_text'])) . '</div>',
-          htmlentities($row['group_name'])
+          htmlentities($row['group_name']),$delete
           );
     }
     $dbManager->freeResult($res);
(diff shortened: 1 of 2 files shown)


src/www/ui/page/AdminLicenseCandidate.php
```diff
@@ -262,6 +279,32 @@ private function mergeCandidate($candidate, $suggest, $vars)
...
+                               ."rf_fk=$1 AND removed=TRUE);");
+    $result = $dbManager->execute($stmt, array($rfPk));
+    $dataFetch = $dbManager->fetchAll($result);
+    $dbManager->freeResult($result);
+    if(empty($dataFetch)){
+      $dbManager->getSingleRow('DELETE FROM license_candidate WHERE rf_pk=$1', array($rfPk), __METHOD__.".delete");
+      return new Response('true', Response::HTTP_OK, array('Content-type'=>'text/plain'));
+    }else{
+      $treeDao = $this->getObject('dao.tree');
+      $Path  = "Cannot remove license, because it is in use at \n\n";
```
maxhbr: The variable name `$Path` is confusing here, and please start variables with lowercase.

src/www/ui/page/AdminLicenseCandidate.php
```diff
@@ -262,6 +279,32 @@ private function mergeCandidate($candidate, $suggest, $vars)
...
+      foreach($dataFetch as $cnt => $uploadTreeFk){
+        $Path .= $cnt+1;
+        $Path .= ") ".$treeDao->getFullPath($uploadTreeFk['uploadtree_fk'], 'uploadtree')."\n";
```
maxhbr: please use a real html `<ol>`-list instead of `1) .... 2)....`.

maxhbr: Or write JSON and render it in the frontend

src/www/ui/page/AdminLicenseCandidate.php
```diff
@@ -61,6 +67,14 @@ function __construct()
    */
   protected function handle(Request $request)
   {
+    $action = $request->get('action');
```
maxhbr: later there is a switch-case on `$request->get('do')` which determines the action. Maybe it would be good to use this field also here.

src/www/ui/page/AdminLicenseCandidate.php
```diff
@@ -61,6 +67,14 @@ function __construct()
...
+    $deleteRfPk = intval($request->get('deleteRfPk'));
```
maxhbr: Same as above, maybe use the already used filed `'rf'` instead of the new `'deleteRfPk'`

src/www/ui/page/AdminLicenseCandidate.php
```diff
@@ -262,6 +279,32 @@ private function mergeCandidate($candidate, $suggest, $vars)
     return true;
   }
 
+  protected function doDeleteCandidate($rfPk)
+  {
+    $dbManager = $this->getObject('db.manager');
+    $stmt = __METHOD__.".$rfPk";
```
maxhbr: `$rfPk` is a argument used in the statement and should not be encoded in the statement name. The name should only depend on the query not the instances of its arguments.

src/www/ui/page/AdminLicenseCandidate.php
maxhbr: ~~This completely ignores permissions? You might be able to see names of uploads which you have no permissions to see.~~ is only usable for admins.

src/www/ui/page/AdminLicenseCandidate.php
```diff
@@ -262,6 +279,32 @@ private function mergeCandidate($candidate, $suggest, $vars)
...
+    $dbManager->prepare($stmt, "SELECT DISTINCT(uploadtree_fk) FROM clearing_event "
```
maxhbr: it is simpler to read if you use multiline strings and formatting:
```php
   $dbManager->prepare($stmt,
        "SELECT DISTINCT(uploadtree_fk)
           FROM clearing_event 
          WHERE rf_fk=$1
            AND removed=false
            AND uploadtree_fk NOT IN (SELECT DISTINCT(uploadtree_fk)
                                        FROM clearing_event
                                       WHERE rf_fk=$1
                                         AND removed=true)");
```
which can (AFAIK) be simplified to 

```php
   $dbManager->prepare($stmt,
        "SELECT DISTINCT(uploadtree_fk)
           FROM clearing_event 
          WHERE rf_fk=$1
            AND removed=false");
```

maxhbr: This only finds usages as clearing decisions, but not as scanner findings.

src/www/ui/template/admin_license_candidate.html.twig
```diff
@@ -50,7 +51,25 @@
...
+          $.ajax({
+            type: 'POST',
+            dataType: 'text',
+            url: '?mod=admin_license_candidate&action=deletecandidate',
+            data: { deleteRfPk : rfPk },
+            success: function(dataOnSuccess) {
+              if(dataOnSuccess == "true"){
+                $("#deletecandidate" + rfPk).closest("tr").hide();
+              }else{
+                alert(dataOnSuccess);
```
maxhbr: we use `jquery-ui` which would allow you to use a more sophisticated alert dialog, like `$.alert(...)`.

src/www/ui/template/admin_license_candidate.html.twig
```diff
@@ -50,7 +51,25 @@
     return otable;
   }
     
-    
+    function deleteCandidate(rfPk) {
+        if (confirm("Remove this license from candidate license list?")) {
```
maxhbr: Same as below here, it is better to use `$.confirm(...)`

src/www/ui/page/AdminLicenseCandidate.php
shaheemazmalmmd: Javascript alert box will not render any kind of html tags. So plan text is used here

src/www/ui/page/AdminLicenseCandidate.php
shaheemazmalmmd: Candidate licenses not included in any scanner findings. they are used once scanning gets finished.

src/www/ui/page/AdminLicenseCandidate.php
maxhbr: Below I suggested to use the jquery-ui alert and confirm boxes, which support html tags.

src/www/ui/page/AdminLicenseCandidate.php
maxhbr: the `reportImport` PR will change this and create candidates as scanner findings.

maxhbr:
@shaheemazmalmmd please fix the formatting issue in `AdminLicenseCandidate.php`:

src/www/ui/page/AdminLicenseCandidate.php
mcjaeger: I think we need this then with the report agent

src/www/ui/page/AdminLicenseCandidate.php
```diff
@@ -262,6 +282,32 @@ private function mergeCandidate($candidate, $suggest, $vars)
     return true;
   }
 
+  protected function doDeleteCandidate($rfPk)
+  {
+    $dbManager = $this->getObject('db.manager');
+    $stmt = __METHOD__.".getUploadtreeFkForUsedCandidates";
+    $dbManager->prepare($stmt, "SELECT DISTINCT(uploadtree_fk) FROM clearing_event "
```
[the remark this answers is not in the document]
shaheemazmalmmd: Sorry i'm not able to understand you. i think the query is already formatted in multi line and looks simple(not a big query). What more changes you want me to do here ??

query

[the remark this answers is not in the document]
shaheemazmalmmd: Query seems to be correct as every time fossology inserts a new decision into clearing_event table (not update existing). so all old decisions still exists in table.  

Lets suppose user adds a license and removes the same. 

Then in removed column in CE table will have ( false, true) for the same license id (rf_pk). So this needed to be checked in both cases

src/www/ui/page/AdminLicenseCandidate.php
```diff
@@ -262,6 +275,35 @@ private function mergeCandidate($candidate, $suggest, $vars)
     return true;
   }
 
+  protected function doDeleteCandidate($rfPk)
+  {
+    $dbManager = $this->getObject('db.manager');
+    $stmt = __METHOD__.".getUploadtreeFkForUsedCandidates";
+    $dbManager->prepare($stmt, "SELECT DISTINCT(uploadtree_fk)
```
maxhbr: please update this statement as discussed

mcjaeger:
rested after last commit and checked with @maxhbr work as expected, opened an issue #916 because maybe it is Ok for first like that.

NesrinCeylan:
Is there any option to delete the licenses which where added before?

PR #893, opened 2017-08-18
fix(php): resolve pass-by-reference errors

timurphy:
Seeing these in the Travis build logs (job 16):

```
Only variables should be passed by reference
/home/travis/build/fossology/fossology/src/lib/php/Test/TestPgDb.php:62
/home/travis/build/fossology/fossology/src/lib/php/Test/TestLiteDb.php:66
```

This causes 181 errors to occur in this job. This fix resolves all errors so that the job succeeds again.

EDIT: this is for the PHP7 build job. Only test files are changed so no production functionality is affected.


maxhbr:
This fix is already contained in https://github.com/fossology/fossology/pull/857

PR #897, opened 2017-09-07
chore(changelog): update to commitlint

max-wittig:
This also fixes the current build problems in travis ci


maxhbr:
Travis uses the following code to check out the dev-branch:


Can you reproduce the commitlint error localy by using the same commands?

The workaround currently looks very ugly and I think it can be simplified.

PR #906, opened 2017-09-12
feat(copyright): also show deactivated copyrights/urls/authors/... in the UI

maxhbr:
This PR adds a new table with deactivated statements to all views of Copyright, the result looks like:

and

To make the `Email/URL/Author` more readable this adds tabs which look like:

(there is a slight css problem, i.e. the `Undo` is black and not blue. That will be fixed by https://github.com/fossology/fossology/pull/904)

### DIFF
--- a/src/copyright/ui/CopyrightView.php
+++ b/src/copyright/ui/CopyrightView.php
@@ -68,7 +68,7 @@ class CopyrightView extends Xpview
 
     $vars = array('statement'=>$tableVars,
         'content' => "$output\n",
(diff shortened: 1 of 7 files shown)


maxhbr:
@mcjaeger This PR contains now the discussed changes to the headers of the tables.

mcjaeger:
so, I tested it: activtation, deactivation in aggregated and single file works and can be seen in the generated readme oss output. But ... adding text findings manuallz is not supported by this PR, right?

PR #907, opened 2017-09-13
feat(treeView): add operation to make multiple files irrelevant

shaheemazmalmmd:


src/www/ui/change-license-processPost.php
```diff
@@ -95,10 +95,20 @@ function Output()
...
-      /** @var ItemTreeBounds */
-      $itemTreeBounds = $this->uploadDao->getItemTreeBounds($itemId,$itemTableName);
-      $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
+      if(!is_array($itemId)) {
+        $itemTableName = $this->uploadDao->getUploadtreeTableName($itemId);
+        /** @var ItemTreeBounds */
+        $itemTreeBounds = $this->uploadDao->getItemTreeBounds($itemId,$itemTableName);
+        $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
+      }else{
+        $uploadTreeId = "";
```
maxhbr: This looks unnecessary

src/www/ui/change-license-processPost.php
```diff
@@ -95,10 +95,20 @@ function Output()
...
+        foreach($itemId as $uploadTreeId){
+          $itemTableName = $this->uploadDao->getUploadtreeTableName($uploadTreeId);
+          /** @var ItemTreeBounds */
+          $itemTreeBounds = $this->uploadDao->getItemTreeBounds($uploadTreeId,$itemTableName);
+          $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
```
maxhbr: You only keep the last `$errMsg`?

src/www/ui/change-license-processPost.php
```diff
@@ -95,10 +95,20 @@ function Output()
...
-      $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
+      if(!is_array($itemId)) {
+        $itemTableName = $this->uploadDao->getUploadtreeTableName($itemId);
+        /** @var ItemTreeBounds */
+        $itemTreeBounds = $this->uploadDao->getItemTreeBounds($itemId,$itemTableName);
+        $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
+      }else{
+        $uploadTreeId = "";
+        foreach($itemId as $uploadTreeId){
+          $itemTableName = $this->uploadDao->getUploadtreeTableName($uploadTreeId);
```
maxhbr: You could move the following three lines to a function, since they are duplicated to the other if-case above

src/www/ui/template/browse_file.js.twig
```diff
@@ -60,12 +60,25 @@ function createDirlistTable() {
...
-  $("#dirlist_length select").css('text-align','right');  
+  $("#dirlist_length select").css('text-align','right');
+
+  var i = 0;
+  $('#markIrrelevant').click(function () {
+    var currentArray = [];
+    $("input:checkbox[id=selectedForIrrelevant]:checked").each(function() {
+      getCurrentVal = $(this).val();
+      if(jQuery.inArray(getCurrentVal, currentArray) < 0 && getCurrentVal != ""){
+        currentArray[i++] = getCurrentVal;
```
maxhbr: why not `currentArray.push(getCurrentVal)`? (this makes the variable `i` unnecessary)

src/www/ui/change-license-processPost.php
```diff
@@ -95,20 +95,31 @@ function Output()
...
-      $itemTreeBounds = $this->uploadDao->getItemTreeBounds($itemId,$itemTableName);
-      $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
-      if (empty($errMsg))
-      {
-        return new JsonResponse(array('result'=>'success'));
+      if(!is_array($itemId)) {
+        $this->doMarkIrrelevant($itemId, $groupId, $userId);
+      }else{
+        foreach($itemId as $uploadTreeId){
+          $this->doMarkIrrelevant($uploadTreeId, $groupId, $userId);
```
maxhbr: Here you throw away the error message

src/www/ui/change-license-processPost.php
```diff
@@ -95,20 +95,31 @@ function Output()
...
     {
-      $itemTableName = $this->uploadDao->getUploadtreeTableName($itemId);
-      /** @var ItemTreeBounds */
-      $itemTreeBounds = $this->uploadDao->getItemTreeBounds($itemId,$itemTableName);
-      $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
-      if (empty($errMsg))
-      {
-        return new JsonResponse(array('result'=>'success'));
+      if(!is_array($itemId)) {
+        $this->doMarkIrrelevant($itemId, $groupId, $userId);
```
maxhbr: Here you throw away the error message

src/www/ui/change-license-processPost.php
```diff
@@ -95,20 +95,31 @@ function Output()
...
+  function doMarkIrrelevant($itemId, $groupId, $userId)
+  {
+    $itemTableName = $this->uploadDao->getUploadtreeTableName($itemId);
+    /** @var ItemTreeBounds */
+    $itemTreeBounds = $this->uploadDao->getItemTreeBounds($itemId,$itemTableName);
+    $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
+    if(!empty($errMsg)){
+      return $this->errorJson($errMsg,$errMsg);
+    }
+    return $errMsg;
```
maxhbr: At this point `$errMsg` is always empty. Why not return something useful or a default value?

maxhbr: Previously here was `return new JsonResponse(array('result'=>'success'));` used

src/www/ui/change-license-processPost.php
shaheemazmalmmd: @this point there wont be any errors as it returns directly from this doMarkIrrelevant function.

shaheemazmalmmd: https://github.com/siemens/fossology/blob/fcab67c6ea5b6c6caf44aaf8b1d97c24a094365e/src/www/ui/change-license-processPost.php#L174

maxhbr: I do not understand what you mean.

Your function `doMarkIrrelevant` returns something related to the error message, but that is dropped and ignored both times when you call it.

src/www/ui/change-license-processPost.php
shaheemazmalmmd: You are right $errMsg will be empty. but there is no additional case handled here so couldn't return a default value. or shall i do just "return;" ??

maxhbr: maybe returning `""` would be better.

shaheemazmalmmd: sure ok.

src/www/ui/change-license-processPost.php
```diff
@@ -95,20 +95,34 @@ function Output()
...
-      $itemTreeBounds = $this->uploadDao->getItemTreeBounds($itemId,$itemTableName);
-      $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
-      if (empty($errMsg))
-      {
-        return new JsonResponse(array('result'=>'success'));
+      if(!is_array($itemId)) {
+        $responseMsg = $this->doMarkIrrelevant($itemId, $groupId, $userId);
+      }else{
+        foreach($itemId as $uploadTreeId){
+          $responseMsg = $this->doMarkIrrelevant($uploadTreeId, $groupId, $userId);
```
maxhbr: new error messages still overwrite old ones. Is this intended?

src/www/ui/change-license-processPost.php
```diff
@@ -95,20 +95,34 @@ function Output()
...
     if(!empty($decisionMark))
     {
-      $itemTableName = $this->uploadDao->getUploadtreeTableName($itemId);
-      /** @var ItemTreeBounds */
-      $itemTreeBounds = $this->uploadDao->getItemTreeBounds($itemId,$itemTableName);
-      $errMsg = $this->clearingDao->markDirectoryAsIrrelevant($itemTreeBounds,$groupId,$userId);
-      if (empty($errMsg))
-      {
-        return new JsonResponse(array('result'=>'success'));
+      if(!is_array($itemId)) {
```
mcjaeger: please try to make it like https://github.com/fossology/fossology/wiki/Coding-Style

maxhbr: Please adhere our codingstyle: https://github.com/fossology/fossology/wiki/Coding-Style

src/www/ui/template/browse.html.twig
```diff
@@ -54,7 +54,7 @@
               </th>
               <th><input type="checkbox" id="openCBoxFilter"/>{{ 'open'|trans }}</th>
               <th></th>
-              <th></th>
+              <th><button type="button" id="markIrrelevant" >MarkAsIrrelevant</button></th>
```
maxhbr: Why not `Mark as irrelevant` with spaces?

shaheemazmalmmd: Because of spacing issue in action td, it increases the size of button as well as td.

shaheemazmalmmd: sizeofbutton

mcjaeger:
I am not sure this feature works this way. I think users want to have all rows shown selected at one click. The use case is to search for `Makefile` or `pom.xml` using the filter function and then select those as irrelevant with a single click.
Other point is that FOSSology has differnt location with multi selection and this could be resembled, for example at the File browser view:
screen shot 2017-10-01 at 23 59 00
or in the copyright view.

PR #910, opened 2017-09-25
feat(copyright): allow to have multiple copyright decisions

maxhbr:
**Note:** the migration script is currently broken. It does not register correctly whether it was already called.

## 

related to #906 and #882


maxhbr:
This PR currently contains problems:
- prepared statement naming has problems and `Exception: Existing Statement mismatch` is thrown
- reports are not yet updated to handle multiple decisions
- migration script does not determine correctly whether something has to be migrated

mcjaeger:
tested and works in general. Also readme oss renders the multiple statements (tested also with two packages. however, it does not seem to arrive in the SPDX. not know if that is normal or unintended. At the same time, for this file, the same copyright is rendered three times.

screen shot 2017-10-03 at 22 33 39

shaheemazmalmmd:
**Copyright Not Exists In ReadmeOSS**

Add the copyright findings for any file with no copyrights found. Then Generate ReadMe

maxhbr:
@shaheemazmalmmd I think that this is also true for the current master.

PR #911, opened 2017-09-26
feat(treeView): add remove option for irrelevant decisions

shaheemazmalmmd:
.... using the edit functionality

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -756,41 +756,82 @@ INSERT INTO clearing_decision (
    * @param int $groupId
    * @param int $userId
    */
-  protected function markDirectoryAsIrrelevantIfScannerDetected(ItemTreeBounds $itemTreeBounds,$groupId,$userId)
+  public function deleteIrrelevantDecisionsFromDirectory(ItemTreeBounds $itemTreeBounds,$groupId,$userId)
   {
-    $params = array($itemTreeBounds->getLeft(), $itemTreeBounds->getRight(), $userId, $groupId, DecisionTypes::IRRELEVANT, DecisionScopes::ITEM);
-    $options = array(UploadTreeProxy::OPT_SKIP_THESE=>'noLicense',
-                     UploadTreeProxy::OPT_ITEM_FILTER=>' AND (lft BETWEEN $1 AND $2)',
-                     UploadTreeProxy::OPT_GROUP_ID=>'$4');
-    $uploadTreeProxy = new UploadTreeProxy($itemTreeBounds->getUploadId(), $options, $itemTreeBounds->getUploadTreeTableName());
-    $statementName = __METHOD__ ;
-    $sql = $uploadTreeProxy->asCte()
-        .' INSERT INTO clearing_decision (uploadtree_fk,pfile_fk,user_fk,group_fk,decision_type,scope) 
-          SELECT uploadtree_pk itemid,pfile_fk pfile_id, $3, $4, $5, $6 FROM UploadTreeView';
-    $this->dbManager->prepare($statementName, $sql);
-    $res = $this->dbManager->execute($statementName,$params);
-    $this->dbManager->freeResult($res);
-  }  
-    
+    $this->markDirectoryAsIrrelevantIfScannerDetected($itemTreeBounds, $groupId, $userId, true);
+    $this->markDirectoryAsIrrelevantIfUserEdited($itemTreeBounds, $groupId, $userId, true);
+  }
+  
   /**
    * @param ItemTreeBounds $itemTreeBounds
    * @param int $groupId
    * @param int $userId
    */
-  protected function markDirectoryAsIrrelevantIfUserEdited(ItemTreeBounds $itemTreeBounds,$groupId,$userId) 
+  protected function markDirectoryAsIrrelevantIfScannerDetected(ItemTreeBounds $itemTreeBounds,$groupId,$userId,$removeDecision=false)
+  {
+    $statementName = __METHOD__ ;
+    $params = array($itemTreeBounds->getLeft(), $itemTreeBounds->getRight());
+    $params[] = $groupId;
+    $a = count($params);
+    $options = array(UploadTreeProxy::OPT_SKIP_THESE=>'noLicense',
+                     UploadTreeProxy::OPT_ITEM_FILTER=>' AND (lft BETWEEN $1 AND $2)',
+                     UploadTreeProxy::OPT_GROUP_ID=>'$'.$a.'');
+    $uploadTreeProxy = new UploadTreeProxy($itemTreeBounds->getUploadId(), $options, $itemTreeBounds->getUploadTreeTableName());
+    if(!$removeDecision) {
+      $params[] = $userId;
+      $params[] = DecisionTypes::IRRELEVANT;
+      $params[] = DecisionScopes::ITEM;
+      $sql = $uploadTreeProxy->asCte()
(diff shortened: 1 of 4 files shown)


src/www/ui/scripts/change-license-browse.js
```diff
@@ -98,3 +98,22 @@ function markDecisions(uploadTreeIdForMultiple) {
...
 }
+
+function deleteMarkedDecisions() {
+  var data = {
+    "uploadTreeId": $('#uploadTreeId').val(),
+    "decisionMark": 'deleteIrrelevant'
+  };
+  resultEntity = $('bulkIdResult');
+    var txt;
+    var pleaseConfirm = confirm("You are about to delete all irrelevant decisions. Please confrim!");
```
mcjaeger: please fix typo

src/lib/php/Dao/ClearingDao.php
```diff
@@ -756,41 +756,73 @@ public function markDirectoryAsIrrelevant(ItemTreeBounds $itemTreeBounds,$groupI
...
-        .' INSERT INTO clearing_decision (uploadtree_fk,pfile_fk,user_fk,group_fk,decision_type,scope) 
-          SELECT uploadtree_pk itemid,pfile_fk pfile_id, $3, $4, $5, $6 FROM UploadTreeView';
+    if(!$removeDecision){
+      $params[] = $userId;
+      $params[] = DecisionTypes::IRRELEVANT;
+      $params[] = DecisionScopes::ITEM;
+      $sql = $uploadTreeProxy->asCte()
+          .' INSERT INTO clearing_decision (uploadtree_fk,pfile_fk,user_fk,group_fk,decision_type,scope) 
+            SELECT uploadtree_pk itemid,pfile_fk pfile_id, $'.($a+1).', $'.$a.', $'.($a+2).', $'.($a+3).' FROM UploadTreeView
+	    WHERE NOT EXISTS (SELECT uploadtree_fk FROM clearing_decision WHERE decision_type=$5 and uploadtree_fk=UploadTreeView.uploadtree_pk)';
```
ChristophNiehoff: There is still a `$5` hardcoded which should be replaced by `$'.($a+2).'`

src/lib/php/Dao/ClearingDao.php
```diff
@@ -756,41 +756,73 @@ public function markDirectoryAsIrrelevant(ItemTreeBounds $itemTreeBounds,$groupI
...
+    }else{
+      $params[] = DecisionTypes::IRRELEVANT;
+      $sql = $uploadTreeProxy->asCte()
+          .' DELETE FROM clearing_decision WHERE clearing_decision_pk IN (SELECT clearing_decision_pk FROM clearing_decision WHERE 
+           decision_type = $'.($a+1).' AND uploadtree_fk IN (SELECT uploadtree_pk FROM UploadTreeView) ORDER BY clearing_decision.date_added DESC)';
```
ChristophNiehoff: Is the `ORDER BY` statement really necessary?

src/lib/php/Dao/ClearingDao.php
```diff
@@ -756,41 +756,73 @@ public function markDirectoryAsIrrelevant(ItemTreeBounds $itemTreeBounds,$groupI
...
-          SELECT itemid,pfile_id, $'.$a.', $'.($a+1).', $'.($a+2).', $'.($a+3).' FROM allDecs ad WHERE type_id!=$'.($a+2));
+    if(!$removeDecision){
+      $params[] = $userId; 
+      $a = count($params);
+      $params[] = $groupId;
+      $params[] = DecisionTypes::IRRELEVANT;
+      $params[] = DecisionScopes::ITEM;
+      $this->dbManager->prepare($statementName, $decisionsCte
+          .' INSERT INTO clearing_decision (uploadtree_fk,pfile_fk,user_fk,group_fk,decision_type,scope) 
+            SELECT itemid,pfile_id, $'.$a.', $'.($a+1).', $'.($a+2).', $'.($a+3).' FROM allDecs ad WHERE type_id IS NULL');
```
ChristophNiehoff: I don't really understand why you changed `WHERE type_id != IRRELEVANT` into `WHERE type_id IS NULL`.  Could you please explain?

src/lib/php/Dao/ClearingDao.php
```diff
@@ -756,41 +756,73 @@ public function markDirectoryAsIrrelevant(ItemTreeBounds $itemTreeBounds,$groupI
...
+    }else{
+      $params[] = DecisionTypes::IRRELEVANT;
+      $this->dbManager->prepare($statementName, $decisionsCte
+          .' DELETE FROM clearing_decision WHERE clearing_decision_pk IN (SELECT clearing_decision_pk FROM clearing_decision WHERE
```
ChristophNiehoff: Correct me if I'm wrong, but isn't this the same query as the one used in `markDirectoryAsIrrelevantIfUserEdited` ? 
Does this mean that in `deleteIrrelevantDecisionsFromDirectory(ItemTreeBounds` you are calling two functions that make the same query?  
Wouldn't it be easier to just have one function for the `deleteIrrelevantDecisions` case?

src/lib/php/Dao/ClearingDao.php
```diff
@@ -756,41 +756,73 @@ public function markDirectoryAsIrrelevant(ItemTreeBounds $itemTreeBounds,$groupI
...
+      $params[] = DecisionTypes::IRRELEVANT;
+      $params[] = DecisionScopes::ITEM;
+      $this->dbManager->prepare($statementName, $decisionsCte
+          .' INSERT INTO clearing_decision (uploadtree_fk,pfile_fk,user_fk,group_fk,decision_type,scope) 
+            SELECT itemid,pfile_id, $'.$a.', $'.($a+1).', $'.($a+2).', $'.($a+3).' FROM allDecs ad WHERE type_id IS NULL');
+    }else{
+      $params[] = DecisionTypes::IRRELEVANT;
+      $this->dbManager->prepare($statementName, $decisionsCte
+          .' DELETE FROM clearing_decision WHERE clearing_decision_pk IN (SELECT clearing_decision_pk FROM clearing_decision WHERE 
+           decision_type = $'.count($params).' AND uploadtree_fk IN (SELECT itemid FROM allDecs) ORDER BY clearing_decision.date_added DESC)');
```
ChristophNiehoff: Maybe you could define a variable `$a = count($params)`. Then this query would look more consistent with the other queries...

src/lib/php/Dao/ClearingDao.php
shaheemazmalmmd: Yeah, its required to get the recent record.

src/lib/php/Dao/ClearingDao.php
shaheemazmalmmd: There are two functions calls inside ``deleteIrrelevantDecisionsFromDirectory``. Because of two different decisions types.

Scanner detected ``markDirectoryAsIrrelevantIfScannerDetected``
And 
User Edited ``markDirectoryAsIrrelevantIfUserEdited``

We use different tables for this. that is why its required to use two functions

src/lib/php/Dao/ClearingDao.php
ChristophNiehoff: Sorry, but I don't really get it.  Most probably, I have an error in my thinking, but doesn't `$sql` essentially look like
`[CTE] DELETE FROM table WHERE field in ( *something* ORDER BY date DESC) `  ?
Why does the ordering make a difference here?

mcjaeger:
@ChristophNiehoff are the changhes applied OK here?

mcjaeger:
works for me with various back and forth setting irrelevant using different measures

PR #923, opened 2017-10-24
feat(unifiedreport): add upload history url to title table

shaheemazmalmmd:
add groupname next to username
correct warnings in obligation

componentlinkandgroup

### DIFF
--- a/src/lib/php/Dao/UserDao.php
+++ b/src/lib/php/Dao/UserDao.php
@@ -1,6 +1,6 @@
 <?php
 /***********************************************************
- * Copyright (C) 2014-2015 Siemens AG
+ * Copyright (C) 2014-2017 Siemens AG
  * Author: J. Najjar, S. Weber, A. Wührl
  *
  * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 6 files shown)


src/unifiedreport/ui/agent-foreport.php
```diff
@@ -49,7 +49,7 @@ protected function handle(Request $request)
     $userId = Auth::getUserId();
     $jobId = JobAddJob($userId, $groupId, $upload->getFilename(), $uploadId);
     $error = "";
-    $jobQueueId = $reportGenAgent->AgentAdd($jobId, $uploadId, $error, array());
+    $jobQueueId = $reportGenAgent->AgentAdd($jobId, $uploadId, $error, array(), tracebackTotalUri());
```
ChristophNiehoff: I am a little bit confused why you are storing the URI in the `jq_cmd_args`. At first sight, it is not clear to me, why the URI should be a "command argument".  Or is `jq_cmd_args` intended to be a general field for remarks about a job?  Anyways, its it really necessary to store this (redundant) information in the database? Is there a reason why you don't simply call `tracebackTotalUri()` in `UnifiedReport`?

shaheemazmalmmd: As the report gets generated in agent side i have to pass it through command line args. 

tracebackTotalUri() only works for UI not CLI.

src/lib/php/Dao/UserDao.php
```diff
@@ -363,10 +363,10 @@ public function addGroup($groupName)
     return $groupNowExists['group_pk'];
   }
 
-  public function addGroupMembership($groupId, $userId)
+  public function addGroupMembership($groupId, $userId, $groupPerm=1)
```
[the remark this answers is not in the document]
shaheemazmalmmd: Its a refactor. but removing it.

PR #924, opened 2017-10-26
feat(report): report assessment summary checkbox selection

ag4ums:


src/unifiedreport/agent/reportStatic.php
```diff
@@ -113,6 +113,18 @@ function clearingProtocolChangeLogTable(Section $section)
     $section->addTextBreak();
   }
 
+  function checkValue($cell, $value, $text)
```
ChristophNiehoff: I think the name of this method is confusing. My first association was something like "testValue"...
Maybe a name like "doCheckValue" or "addCheckbox" would be better...

src/unifiedreport/agent/reportStatic.php
```diff
@@ -160,30 +171,39 @@ function assessmentSummaryTable(Section $section, $otherStatement)
     $table->addCell($cellFirstLen)->addText(htmlspecialchars(" "), $leftColStyle, "pStyle");
     $table->addCell($cellLen)->addText(htmlspecialchars(" "), $rightColStyleBlue, "pStyle");
 
+    $nocriticalfiles = "no critical files found, source code and binaries can be used as is";
```
ChristophNiehoff: For all these strings a whitespace should be added at the beginning. Otherwise the checkboxes touch the text.

mcjaeger:
@ChristophNiehoff  I am Ok testing it without looking at MS word (As I do not have word on my computer as well)

ChristophNiehoff:
Unfortunatelly, the output doesn't look very good in LibreOffice.

LibreOffice Writer (Version: 5.3.1.2)
screenshot from 2017-10-30 11-47-32

However, in Microsoft Word (Microsoft Office 365 ProPlus, Version 15.0.4937.1000) it looks as expected:
image

ChristophNiehoff:
The code itself looks good regarding my change requests. However, I am not yet setting the "approved" label for my review because of the issues mentioned in my previous comment.

mcjaeger:
I am testing with LibreOffice too on macosx and did not see it, I will retest.

mcjaeger:
created an issue here: https://github.com/fossology/fossology/issues/933

Issue #866, opened 2017-06-27
License clearing results from clearingDao->getFileClearingsFolder are incorrect


swinslow:
Issue:
When both (A) an upload contains multiple identical copies of a file, and (B) that file has been identified and cleared using the "Apply decision to all future occurrences of this file" option, then the results array returned from getFileClearingsFolder() in /src/lib/php/Dao/ClearingDao.php is incorrect. It contains duplicate clearing results for the file that was marked with global scope, and lacks clearing results for other copies of the same file in that upload.

This appears to be causing SPDX exports to report incorrect concluded license results for those files (see issue #826). It may also be causing errors in other functions that rely on this function.

Steps to reproduce:

1) Upload an archive that contains multiple copies of the same file. A sample archive is attached as [newtest.tar.gz](https://github.com/fossology/fossology/files/1105714/newtest.tar.gz); this archive contains three copies of the same file.
2) In the upload settings, select Monk and Nomos scanning options.
3) After the scan completes, manually go into the license browser and identify mitfile1.txt as "MIT". Select the option "Apply decision to all future occurrences of this file" before hitting submit.
4) After this, the other two identical files (mitfile2.txt and mitfile3.txt) should now also show up as cleared and should be marked as "MIT" in the license browser. Each of these files has a unique uploadTreeId but shares the same pfileId.
5) The function getFileClearingsFolder() in /src/lib/php/Dao/ClearingDao.php will now return incorrect results.

To confirm this, I dumped the return value from a call to getFileClearingsFolder() to a temporary text file. 
I added the following temporary code to /src/spdx2/agent/spdx2.php, in the function getFilesWithLicensesFromClearings(), just after the call to getFileClearingsFolder():
```php
$tempf = fopen("/tmp/getFileClearingsFolder-report.txt", "w");
$decisions_arr = print_r($clearingDecisions, true);
fwrite($tempf, $decisions_arr);
fclose($tempf);
```

A copy of this report is attached as [getFileClearingsFolder-report.txt](https://github.com/fossology/fossology/files/1105715/getFileClearingsFolder-report.txt).

Expected result:
getFileClearingsFolder() should return an array with a separate ClearingDecision object for each of mitfile1.txt, mitfile2.txt, and mitfile3.txt. 

Actual result:
[...]

PR #925, opened 2017-10-30
fix(lib): decision for future occurrence of files

ag4ums:
On successful Testing, this fix should close issue #866 and #826.


shaheemazmalmmd:
After deciding the licenses are getting separated and a empty column found in clearing history.

Please find that attached screenshot. 

historycrashed

PR #950, opened 2017-11-28
fix(bulk): separate td for each image and add width for select

shaheemazmalmmd:
**To Reproduce:**

-> Add a new license or candidate license to fossology database with min 50 characters for shortname.
-> Go to single file view click on bulk recognition.
-> Select license will be out of the box.

screen shot 2017-12-20 at 11 47 41


mcjaeger:
after fix: 
screen shot 2017-12-20 at 12 56 27

PR #956, opened 2017-12-07
fix(ui): checkbox param call more adaptable with php 5.4

ag4ums:


src/www/ui/ui-view-info.php
```diff
@@ -677,6 +677,26 @@ function ShowReportInfo($Upload)
...
 
+  /**
+    * @param array $checkBoxListParams
+    * @return $cbSelectionList
+   */
+
+  protected function getCheckBoxSelectionList($checkBoxListParams)
+  {
+    foreach($checkBoxListParams as $checkBoxListParam){
+      if(empty(GetParm($checkBoxListParam, PARM_STRING))){
```
mcjaeger: should be like that:
GetParm($checkBoxListParam, PARM_STRING) != ''

PR #960, opened 2017-12-13
fix(lib): container.php access from cache

ag4ums:
saw the error from the log that the container.php was not accessible and found out that is some installation modes the assigned rights were not sufficient.

Unfortunately the conditions are unclear, because it happened at some time on "some instance". 

It seems to be connected with the generation of the SPDX.


mcjaeger:
maybe we could test it with 027 as well?

PR #962, opened 2017-12-20
add user findings of copyright and ecc from files with non-agent finding

shaheemazmalmmd:
Point is that when manually adding a finding by the user in the single file view Web UI on a file where the agent did not find anything ("text finding"), then this finding is not put into the reporting.

These are rare cases, as the usual way is to just correct scanner findings. However, manually added things should be there too.

### DIFF
--- a/src/lib/php/Dao/CopyrightDao.php
+++ b/src/lib/php/Dao/CopyrightDao.php
@@ -148,6 +148,100 @@ class CopyrightDao extends Object
     $this->dbManager->execute(__METHOD__, array($decisionId, $pfileId));
   }
 
+  /**
+   * @param $tableName
+   * @param $uploadTreeTableName
+   * @param $uploadId
+   * @param $type
+   * @param $extrawhere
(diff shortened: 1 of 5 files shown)


src/lib/php/Dao/CopyrightDao.php
```diff
@@ -148,6 +148,120 @@ public function undoDecision($tableName,$pfileId, $decisionId)
...
+          $scannerEntries[$key]['textfinding'] = $value['textfinding'];
+          $scannerEntries[$key]['comments'] = $value['comments'];
+          $scannerEntries[$key]['description'] = $value['description'];
+        }
+      }
+
+      return $scannerEntries;
+    } else {
+      $editedEntries = $this->getEditedEntries($tableNameDecision, $uploadTreeTableName, $uploadId, $decisionType);
+//      print_r($editedEntries);exit;
```
ag4ums: please remove the debugging statement.

Issue #963, opened 2017-12-26
SPDX tag-value exporter fills in NONE for FileCopyrightText (should be NOASSERTION)


swinslow:
**Issue**:
In FOSSology 3.1.0, when the copyright analysis agent has not been run for an upload, and the SPDX tag-value exporter agent is run, the resulting SPDX file lists "FileCopyrightText: NONE" for each file in the upload.

According to the SPDX specification (see https://spdx.org/spdx-specification-21-web-version#h.206ipza), NONE is meant as a statement that the file contains no copyright information. This should likely be NOASSERTION instead, which signals that the reporter is making no statement about whether copyright info is present in the file. The SPDX RDF exporter correctly reports this as NOASSERTION.

This issue was originally reported on the fossology-devel mailing list by Tran Van Truong on 2017-12-25.

**Steps to reproduce**:

1. In FOSSology 3.1.0, upload an archive but do not run the copyright analysis agent on upload.
2. Export the archive in SPDX RDF and SPDX tag-value formats.
3. Compare the "spdx:copyrightText" field in RDF with the "FileCopyrightText" field in tag-value.

**Expected result**: both should be NOASSERTION because no copyright scan has been performed
**Actual result**: the RDF field contains a URI reference to the NOASSERTION value; the tag-value field contains the text "NONE".

PR #964, opened 2017-12-26
fix(spdx): make SPDX-rdf and SPDX-tv templates consistent

maxhbr:
This changes the SPDX-tv template as suggested in #963.


shaheemazmalmmd:
Code looks good 
noassertion

PR #970, opened 2018-01-16
CSV import from obligations: improve checks while associating Licenses

deveaud-m:
Importing obligations from a CSV file with licenses not available
in the DB leads to a `not-null constrains` Postgres error.
With those changes, the license association first checks for the
availability of the license's shortname before associating the
license with the obligation.

### DIFF
--- a/src/lib/php/Application/ObligationCsvImport.php
+++ b/src/lib/php/Application/ObligationCsvImport.php
@@ -131,7 +131,7 @@ class ObligationCsvImport {
     return ($row === false) ? false : $row['ob_pk'];
   }
 
-  private function compareLicList ($exists, $listFromCsv, $candidate, $row)
+  private function compareLicList($exists, $listFromCsv, $candidate, $row)
   {
     $getList = $this->obligationMap->getLicenseList($exists, $candidate);
     $listFromDb = $this->reArrangeString($getList);
@@ -164,6 +164,9 @@ class ObligationCsvImport {
     /* @var $dbManager DbManager */
     $dbManager = $this->dbManager;
     $exists = $this->getKeyFromTopicAndText($row);
+    $associatedLicenses = "";
+    $candidateLicenses = "";
+    $listFromCsv = "";
     $msg = "";
     if ($exists !== false)
     {
@@ -182,7 +185,7 @@ class ObligationCsvImport {
         $msg .=" No Changes in CandidateLicense";
       }
       else {
-        $this->clearListFromDb($exists, $listFromCsv, true);
+        $this->clearListFromDb($exists, $listFromCsv);
         if(!empty($row['candidatenames'])) {
           $associatedLicenses = $this->AssociateWithLicenses($row['candidatenames'], $exists, True);
         }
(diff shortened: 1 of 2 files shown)


ag4ums:
no error now in csv import.
->But csv import is not updating, if obligation table field/s are modified.
-> import updated CSV from one server to another server may be necessary to keep the sever in Sync. 
Test steps followed  : 
*1. "Admin->ObligationAdmin->CSV Export" (from server 1)
*2. "Admin->ObligationAdmin->CSV Import" (in server 2)
*3. Obligation are correctly uploaded in server 2.
*5. "Admin->Select Obligation" Select and obligation to modify (in server 1)
*4. Modify(add/delete) Associated Licenses, change Classification field values for that obligation topic. and select update (in server 1).
*5. repeat step 1. and step 2 .
*6 changed fields(Associated Licenses and Classification) are not getting updated.

src/lib/php/Application/ObligationCsvImport.php
```diff
@@ -181,34 +183,56 @@ private function handleCsvObligation($row)
...
         $this->obligationMap->associateLicenseWithObligation($obPk, $licId, $candidate);
         if ($associatedLicenses == "")
         {
           $associatedLicenses = "$license";
+        } else {
+          $associatedLicenses .= ";$license";
         }
-        else
+      } else {
+        if ($message == "")
```
mcjaeger: @deveaud-m just a question: so in both cases the message is set to the same string, but one is appending and another is replacing (if the string is empty), concatenation could work on the empty string as well. Just asking if we could change this into a straight concatenation removing the conditional?

src/lib/php/Application/ObligationCsvImport.php
```diff
@@ -181,34 +183,56 @@ private function handleCsvObligation($row)
...
         }
       }
+    }
 
-      return $associatedLicenses;
+    if (!empty($associatedLicenses))
+    {
+      if ($message == "")
+      {
+        $message = "$associatedLicenses were associated.\n";
```
mcjaeger: same as above...

src/lib/php/Application/ObligationCsvImport.php
deveaud-m: @mcjaeger Sure, go ahead!

src/lib/php/Application/ObligationCsvImport.php
```diff
@@ -154,20 +154,22 @@ private function handleCsvObligation($row)
     $new = $dbManager->fetchArray($resi);
     $dbManager->freeResult($resi);
 
+    $associatedLicenses = "";
+    $candidateLicenses = "";
     if (!empty($row['licnames'])) {
       $associatedLicenses = $this->AssociateWithLicenses($row['licnames'], $new['ob_pk']);
+    } else {
```
steffen-weber: the else case is not required since `$associatedLicenses` is initialized with that string in line 157

shaheemazmalmmd:
Works as expected

deveaud-m:
Great job @ag4ums and @shaheemazmalmmd! Thanks for taking over this PR.

PR #975, opened 2018-01-23
feat(deleteUpload): unlink upload in case of multiple upload not delete

shaheemazmalmmd:
- Upload a new package "upload-A" in Software Repository
- Create a new folder "Test"
- Make a copy of upload A in Test
- Delete upload A in software repository
- The copied upload should not get deleted

### DIFF
--- a/src/delagent/ui/Page/AdminUploadDelete.php
+++ b/src/delagent/ui/Page/AdminUploadDelete.php
@@ -20,6 +20,7 @@ namespace Fossology\DelAgent\UI\Page;
 
 use Fossology\Lib\Auth\Auth;
 use Fossology\Lib\Dao\UploadDao;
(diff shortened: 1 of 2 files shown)


ag4ums:
tested with the below scenario: 
* create folders under Software Repository(SR) "test1" and "test2"
* upload package "upload1" under SR>test1
* copy SR>test1 under SR>test2
* delete "upload1" from Software Repository->test1
* the package "upload1" is also getting deleted from SR>test2>test1.
@shaheemazmalmmd @mcjaeger Is this fix is expected to cover the above scenario or this is expected behavior.

ag4ums:
Will raise another ticket to address the above scenario,
For this pull req :
Code looks good, tested and working fine.

PR #978, opened 2018-01-30
feat(license): add acknowledgements to license clearing

shaheemazmalmmd:
- upload a package and go to single file clearing.
- you should find new acknowledgement TD.
- add new acknowledgement to the file and make it identified.
- Generate READMEOSS and Unified Report.
- check if the acknowledgements exists

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -1,6 +1,6 @@
 <?php
 /*
-Copyright (C) 2014-2017, Siemens AG
+Copyright (C) 2014-2018, Siemens AG
 Author: Johannes Najjar
 
 This program is free software; you can redistribute it and/or
@@ -218,7 +218,8 @@ class ClearingDao extends Object
               ce.removed AS removed,
               ce.type_fk AS event_type_id,
               ce.reportinfo AS reportinfo,
-              ce.comment AS comment
+              ce.comment AS comment,
+              ce.acknowledgement AS acknowledgement
             FROM decision
             LEFT JOIN users ON decision.user_id = users.user_pk
             LEFT JOIN clearing_decision_event cde ON cde.clearing_decision_fk = decision.id
@@ -252,6 +253,7 @@ class ClearingDao extends Object
       $eventGroupId = $row['event_group_id'];
       $comment = $row['comment'];
       $reportInfo = $row['reportinfo'];
+      $acknowledgement = $row['acknowledgement'];
 
       if ($clearingId !== $previousClearingId && $itemId !== $previousItemId)
       {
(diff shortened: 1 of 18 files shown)


src/lib/php/Data/Clearing/ClearingEvent.php
```diff
@@ -46,6 +46,10 @@ class ClearingEvent extends Object implements LicenseClearing
    * @param int $groupId
    * @param int $eventType
    * @param ClearingLicense $clearingLicense
+   * @param boolean $removed
+   * @param string $reportinfo
+   * @param string $acknowledgement
+   * @param string $comment
```
ag4ums: $removed, $reportinfo, $acknowledgement, $comment are these require

src/lib/php/Dao/ClearingDao.php
```diff
@@ -852,17 +866,28 @@ public function getMainLicenseIds($uploadId, $groupId)
...
     $this->dbManager->insertTableRow('upload_clearing_license',
             array('upload_fk'=>$uploadId,'group_fk'=>$groupId,'rf_fk'=>$licenseId));
   }
-  
+
+  /**
+   * @param uploadId
+   * @param int $groupId
+   * @param int $licenseId
+   */
```
ag4ums: please also add return param

src/readmeoss/agent/readmeoss.php
```diff
@@ -120,27 +131,61 @@ private function writeReport($contents, $uploadId)
...
+   * @param $extract
+   * @param $break
+   */ 
+  private function createReadMeOSSFormat($addSeparator, $dataForReadME, $extract='text', $break)
+  {
+    $outData = "";
+    foreach($dataForReadME as $statements){
+     $outData .= $statements[$extract] . $break;
+     if(!empty($addSeparator)){
+       $outData .= $addSeparator . $break;
```
ag4ums: please correct the alignment

mcjaeger:
@ag4ums would you mind to clarify if the review comments are covered.

Also, I suspect that the same error shown as in https://github.com/fossology/fossology/pull/980 has also the same cause.

ag4ums:
now this is fixed with the pull request #1063

NicolasToussaint:
Do you think that the Acknowledgement information could also be exported to the SPDX reports ?

It seems to me that the ["File Contributor"](https://spdx.org/spdx_specification_2_0_html#h.sqyw64) entry in the SPDX 2.0 Specification would be a good candidate, what do you think ?

mcjaeger:
I think the issue is here at the spdx spec: https://github.com/spdx/spdx-spec/issues/28

I am not sure if FileContributor is the intended option here. Maybe as an interim solution it could be done. However, I am not sure we should build FOSSology (and SW360) to have it in FileContributor then if there is a new field planned. @kestewart would you have an update on this?

NicolasToussaint:
Good spot, though I'm a bit confused on how these fields should be used.
About the AttributionText field you refer to:
> It can be considered as a combination of information found in properties like FileCopyrightText, LicenseInfoInFile, FileContributor, but it might not simply the sum of these values.

Anyhow, I guess I would be fine with a temporary export to the FileContributor.

mcjaeger:
Hm, I see the point that this field is "partially matching" maybe it could be good to have the acknowledgement there for the first. It think it depends, how likely the spec is updated within the next three months, for example ...

PR #979, opened 2018-01-31
chore(debian) : added support for ubuntu 16.04 and php7.0

ag4ums:
debian packages to support php7.0
and support for ubuntu 16.04


mcjaeger:
OK, fun, but this disables debian 8 support, right?

ag4ums:
corrected and now working with debian(8/9) and ubuntu(14.04/ 16.04).

PR #995, opened 2018-03-06
fix(browse): ajax browse required login

### LINKED ISSUE #994: allow anonymous user to browse (mod=browse)
Is it possible ? I can't find my way to the setting that would allow that. I'm using the latest Docker release.

My goal is to allow the default "Software repository" folder to be readable for everyone, even non logged-in users...

Instead I get the following:

screenshot-2018-3-5 browse

There is this "Processing..." notice that is freezed
In the web dev pane I see:

```
12:28:20,219 default DEBUG executing
'FossologyLibDaoUserDao::getUserChoices' took 0.461ms (1 queries, avg 0.461ms)
repo:365:1
12:28:20,994 Request Failed: error, Internal Server Error
tools.js:60:3
12:28:20,996 default DEBUG DI container setup (cached: yes) (0.371ms)
```

However it works when logged in (as `fossy`)...

In addition, it doesn't seem related to the docker setup as stated in https://lists.linuxfoundation.org/pipermail/fossology-devel/2018-March/002953.html

maxhbr:
The browse plugin `ui-browse.php` does not require login to be used, but the
ajax backend `AjaxBrowse.php` required it.

The problem was caused due to the migration of the parent class `FO_Plugin` to
`DefaultPlugin` in the `AjaxBrowse.php`. In `FO_Plugin` is by default no login
required, this logic is flipped in the newer implementation `DefaultPlugin`.

closes #994

### DIFF
--- a/src/www/ui/async/AjaxBrowse.php
+++ b/src/www/ui/async/AjaxBrowse.php
@@ -54,9 +54,9 @@ class AjaxBrowse extends DefaultPlugin
   function __construct()
   {
     parent::__construct(self::NAME, array(
-        self::PERMISSION => Auth::PERM_READ
-      ));
-        
+      self::REQUIRES_LOGIN => false,
+      self::PERMISSION => Auth::PERM_READ
+    ));
     global $container;
     $this->uploadDao = $container->get('dao.upload');
     $this->userDao = $container->get('dao.user');


mhow2:
I cloned and changed to the said branch the used the shipped docker-compose in `./install` to generate a new docker image. Seems to work on my side. On this screenshot I tweaked the permissions so the uploads appear as anonymous user.
screenshot-2018-3-7 browse

PR #1000, opened 2018-03-13
fix(delagent): change query which deletes all files with same pfile

shaheemazmalmmd:
**REQUIREMENTS**
* Two same packages with different versions.

**HOW TO TEST**
* On a clean instance 
* Install master and upload package1 (I used zlib-1.2.7)
* On the same instance upload package2 (I used zlib-1.2.8)
* Delete package 1
* Check Browse-view both the packages are getting deleted.
* Now install contrib/delAgentOrphanRecordsUploadTree and repeat the tests.


mcjaeger:
tested, seems to work

PR #1001, opened 2018-03-13
fix(reuser): remove warnings and errors with testcases for reuse

shaheemazmalmmd:
Fixed failing testcases and warnings(phpunit) in reuser for debian OS. 

**ERRORS**
`PHP Notice:  Array to string conversion in /builds/fossology/fossologyng/src/lib/php/Db/Driver/Postgres.php on line 72
dropdb: database removal failed: ERROR:  database "reusersched" is being accessed by other users
DETAIL:  There is 1 other session using the database.

1) Fossology\Reuser\Test\SchedulerTest::testReuserRealScanWithARepoClearingEnhanced
pg_execute(): Query failed: ERROR:  invalid input syntax for integer: "Array"
`

### DIFF
--- a/src/composer.json
+++ b/src/composer.json
@@ -35,7 +35,6 @@
             "Fossology\\Agent\\Copyright\\UI\\": "copyright/ui",
             "Fossology\\DeciderJob\\UI\\": "deciderjob/ui",
             "Fossology\\Decider\\UI\\": "decider/ui",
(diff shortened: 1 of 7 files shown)


mcjaeger:
@shaheemazmalmmd mayybe you could consider some little desc like were warnings just hidden or what was changed.

Issue #985, opened 2018-02-08
Search: wrong number of files matching the search criteria (>100)


dwojtasiuk:
Hello,

The count in Search stops at 100, which is the max number of records per page. In case there are more than a hundred files matching the search criteria, you have to go to the last page of search results to see how many files are actually there.
![image](https://user-images.githubusercontent.com/36267903/35976763-689410ca-0ce1-11e8-8ba2-1228d8ec3f00.png)

Version: [3.2.0rc1], Commit: [#dc43fd]

amanjain97:
@daniellandau @mcjaeger Apart from this, why there is an indentation in the search results (201 and 202). Is that an issue with UI or it is relevant to the application?

amanjain97:
@mcjaeger Please review the PR #1004 and close the issue.

GMishx:
Working fine now. Thanks @amanjain97 
![file-counter](https://user-images.githubusercontent.com/18077542/42694343-ba97a0d6-86cf-11e8-8975-7af3b62a5d54.png)

PR #1004, opened 2018-03-15
fix(search.php): Fix the number of files matching the search criteria

amanjain97:
This fix issue #985 

screen shot 2018-03-15 at 12 40 52 pm

It displays the total number of files that match the search criteria 
Also, the message about the UI so that users are not confused why indentation is there in the search results.

### DIFF
--- a/src/www/ui/search.php
+++ b/src/www/ui/search.php
@@ -22,7 +22,7 @@ use Fossology\Lib\Dao\UploadDao;
 
 class search extends FO_Plugin
 {
-  protected $MaxPerPage  = 100;  /* maximum number of result items per page */
+  protected $MaxPerPage  = 10;  /* maximum number of result items per page */
   /** @var UploadDao */
   private $uploadDao;
   
@@ -76,14 +76,14 @@ class search extends FO_Plugin
    * \param $SizeMin  Minimum file size, -1 if unused
    * \param $SizeMax  Maximum file size, -1 if unused
    * \param $searchtype "containers", "directory" or "allfiles"
-   * \return array of uploadtree recs.  Each record contains uploadtree_pk, parent, 
-   *         upload_fk, pfile_fk, ufile_mode, and ufile_name
+   * \return array of uploadtree recs and total uploadtree recs count. Each record contains uploadtree_pk, parent, upload_fk, pfile_fk, ufile_mode, and ufile_name
    */
   function GetResults($Item, $Filename, $tag, $Page, $SizeMin, $SizeMax, $searchtype, $License, $Copyright)
   {
     global $PG_CONN;
     $UploadtreeRecs = array();  // uploadtree record array to return
     $totalUploadtreeRecs = array();  // total uploadtree record array to return
+    $totalUploadtreeRecsCount = 0;
     $NeedTagfileTable = true;
     $NeedTaguploadtreeTable = true;
 
@@ -102,7 +102,7 @@ class search extends FO_Plugin
  
        /* Check upload permission */
        if (!$this->uploadDao->isAccessible($upload_pk, Auth::getGroupId())) {
-        return array($UploadtreeRecs, $totalUploadtreeRecs);
+        return array($UploadtreeRecs, $totalUploadtreeRecsCount);
       }
     }
 
@@ -131,7 +131,7 @@ class search extends FO_Plugin
       {
         /* tag doesn't match anything, so no results are possible */
         pg_free_result($result);
-        return array($UploadtreeRecs, $totalUploadtreeRecs);
+        return array($UploadtreeRecs, $totalUploadtreeRecsCount);
       }
 
       /* Make a list of the tag_pk's that satisfy the criteria */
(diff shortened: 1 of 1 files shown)


maxhbr:
@amanjain97 please fix your local git setup, so that your git commits are actually related to your github accound which you use to create PRs and to communicate.

amanjain97:
I have configured the Git account using git config. Can you please confirm do I need to make a new pull request for this issue or I will keep this mind from the next Pull request.

maxhbr:
The problem is that you have not configured the email adress in your local git config correctly. It currently states `Aman Jain <amanjain@Amans-MacBook-Pro.local>`. Please update that, ammend the commit, and push with `--force-with-lease` to your feature branch.

maxhbr:
It looks better now, but still not perfect. The committer is updated to your public email adress, but the author is still the old setup. To update it you could use
```
git commit --amend --author="Aman Jain <amanjain5221@gmail.com>"
```

amanjain97:
Can you please check now. Apologies for the inconvenience.

maxhbr:
Yes, the commit metadata looks good now.

amanjain97:
Can we merge it now ?

src/www/ui/search.php
```diff
@@ -101,7 +102,7 @@ function GetResults($Item, $Filename, $tag, $Page, $SizeMin, $SizeMax, $searchty
  
        /* Check upload permission */
        if (!$this->uploadDao->isAccessible($upload_pk, Auth::getGroupId())) {
-        return $UploadtreeRecs;
+        return array($UploadtreeRecs, $totalUploadtreeRecs);
```
maxhbr: `$totalUploadtreeRecs` is always `array()`

src/www/ui/search.php
```diff
@@ -130,7 +131,7 @@ function GetResults($Item, $Filename, $tag, $Page, $SizeMin, $SizeMax, $searchty
       {
         /* tag doesn't match anything, so no results are possible */
         pg_free_result($result);
-        return $UploadtreeRecs;
+        return array($UploadtreeRecs, $totalUploadtreeRecs);
```
maxhbr: `$totalUploadtreeRecs` is always `array()`

src/www/ui/search.php
```diff
@@ -266,6 +267,20 @@ function GetResults($Item, $Filename, $tag, $Page, $SizeMin, $SizeMax, $searchty
 
     $Offset = $Page * $this->MaxPerPage;
     $SQL .= " ORDER BY ufile_name, uploadtree.pfile_fk";
+    $result = pg_query($PG_CONN, $SQL);
```
maxhbr: Since you are only interested in the size of the result of `$SQL` you could tweak the query to directly return the number of rows.

This code also implies that you ask the DB twice, with and without paging. Why then use paging in the first place if one has already fetched all data.

src/www/ui/search.php
```diff
@@ -280,7 +295,7 @@ function GetResults($Item, $Filename, $tag, $Page, $SizeMin, $SizeMax, $searchty
       }
     }
     pg_free_result($result);
-    return($UploadtreeRecs);
+    return array($UploadtreeRecs, $totalUploadtreeRecs);
```
maxhbr: it is enought to just return the size of `$totalUploadtreeRecs`

maxhbr: please also update the function documentation, that the return value is changed

src/www/ui/search.php
maxhbr: This is copied code and should be factored out to a function (in fact there are more elaborate libraries to talk to the DB in the source code)

src/www/ui/search.php
```diff
@@ -411,11 +426,13 @@ function Output()
     if ($CriteriaCount)
     {
       if (empty($Page)) { $Page = 0; }
-      $UploadtreeRecs = $this->GetResults($Item,$Filename,$tag,$Page,$SizeMin,$SizeMax,$searchtype,$License, $Copyright);
+      $result = $this->GetResults($Item,$Filename,$tag,$Page,$SizeMin,$SizeMax,$searchtype,$License, $Copyright);
```
maxhbr: the variable name `$result` is way less descriptive

src/www/ui/search.php
amanjain97: I want to return an array of $UploadtreeRecs and $totalUploadtreeRecs. In case I have missed something, can you please suggest how can I return two variables from a function.

src/www/ui/search.php
amanjain97: Ok. Thanks for the feedback. I will update the patch soon.

amanjain97: Can you please review the approach. 
How about fetching all results without paging in $totalUploadTreeRecs and then just return the $MaxPerPage results according to the page number as $uploadTreeRecs.
``` 
return array($uploadTreeRecs, $totalUploadTreeRecsCount)
``` 
In this way, we will be executing the query only once and no copied code will be there.

src/www/ui/search.php
amanjain97: In case I have missed, are there any naming conventions in Fossology?

mcjaeger:
@maxhbr OK to merge?

PR #1066, opened 2018-04-16
fix(licenseUpdate): update the license parameters with same shortname

ag4ums:
Steps to test.

User unable to update the license fields from License Administration.
Eg : Updating Risk, is giving error.

This PR fixes the above issue, and user should be able to update the require fields except the license shortname(unique).

### DIFF
--- a/src/www/ui/admin-license-file.php
+++ b/src/www/ui/admin-license-file.php
@@ -352,7 +352,21 @@ class admin_license_file extends FO_Plugin
     $check_count = $this->dbManager->getSingleRow($sql,array($rfId,$shortname,$text),__METHOD__.'.countLicensesByNomos');
     return (0 < $check_count['count']);
   }
-  
+
+
+  /** @brief check if shortname is changed */
+  private function isShortNameExists($rfId, $shortname)
+  {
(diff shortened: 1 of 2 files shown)


GMishx:
Please consider disabling the shortname filed while updating the license.
Since it is enabled, users might try to change its value (which is not allowed) and the operation will fail. This might be frustrating.

src/www/ui/admin-license-file.php
```diff
@@ -352,7 +352,21 @@ private function isShortnameBlocked($rfId,$shortname,$text)
...
   }
-  
+
+
+  /** @brief check if shortname is changed */
+  private function isShortNameExists($rfId, $shortname)
+  {
+    $sql = "SELECT rf_shortname FROM license_ref WHERE rf_pk=$1";
+    $row = $this->dbManager->getSingleRow($sql,array($rfId),__METHOD__.'.DoNotChnageShortName');
+    if ($row['rf_shortname'] === $shortname) {
```
GMishx: Please consider case insensitive comparison as right-hand value is coming from the UI form.

PR #1067, opened 2018-04-16
fix(import-csv): syntax error in importing license-csv

ag4ums:
import licenses in csv format throws error.

image

### DIFF
--- a/src/www/ui/template/admin_license_from_csv.html.twig
+++ b/src/www/ui/template/admin_license_from_csv.html.twig
@@ -8,14 +8,13 @@
 
 {% block content %}
   <p>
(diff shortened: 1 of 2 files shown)


src/www/ui/template/admin_license_from_csv.html.twig
```diff
@@ -8,14 +8,11 @@
 
 {% block content %}
   <p>
-    {{ 'This option permits uploading a single file (which may be iso, tar, rpm, jar, zip, bz2, msi, cab, etc.) from
-        your computer to FOSSology.'| trans }}
-    {{ 'Your FOSSology server has imposed a maximum upload file size of'| trans }}
```
GMishx: Won't it be good to show the maximum upload file size of PHP to user?

PR #1072, opened 2018-04-21
fix(author): Fix multiple entries in author table

GMishx:
## How To Test
1.  Get a package with at least one file which contains only author information (email/author/URL) and no copyright statement.
2.  Install master and upload the package.
3.  Upload the same package again (multiple times).
4.  Check the Email/URL/Author tab of the uploaded package. The count of entries in the table will be the multiple of actual occurrence and the number of times the package is uploaded.

## After fix
1.  Install the fix and run `sudo install/fo-postinstall`.
2.  Repeat the steps above.
3.  The problem will be fixed.

### Other fixes
1.  Decouple copyright and author table's index sequence.

Sample test package: [30-seconds-of-code](https://github.com/Chalarangelo/30-seconds-of-code)

### DIFF
--- a/src/copyright/agent/copyright.cc
+++ b/src/copyright/agent/copyright.cc
@@ -131,7 +131,7 @@ int main(int argc, char** argv)
     {
       int uploadId = atoi(fo_scheduler_current());
 
-      if (uploadId == 0) continue;
+      if (uploadId <= 0) continue;
 
       int arsId = writeARS(state, 0, uploadId, 0, dbManager);
 
--- a/src/www/ui/core-schema.dat
+++ b/src/www/ui/core-schema.dat
@@ -432,8 +432,8 @@
 
 
   $Schema["TABLE"]["author"]["ct_pk"]["DESC"] = "";
-  $Schema["TABLE"]["author"]["ct_pk"]["ADD"] = "ALTER TABLE \"author\" ADD COLUMN \"ct_pk\" int8 DEFAULT nextval('copyright_ct_pk_seq'::regclass)";
-  $Schema["TABLE"]["author"]["ct_pk"]["ALTER"] = "ALTER TABLE \"author\" ALTER COLUMN \"ct_pk\" SET NOT NULL, ALTER COLUMN \"ct_pk\" SET DEFAULT nextval('copyright_ct_pk_seq'::regclass)";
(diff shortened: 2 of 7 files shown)


GMishx:
Please test it on existing instances as well since the query required to clean table may take pretty long time.

mcjaeger:
retested also on instance with data ..., confirmed, works

src/www/ui/core-schema.dat
```diff
@@ -1938,7 +1944,7 @@
   $Schema["INDEX"]["clearing_event"]["clearing_event_uploadtree_fk_idx"] = "CREATE INDEX clearing_event_uploadtree_fk_idx ON clearing_event USING btree (uploadtree_fk);";
   $Schema["INDEX"]["clearing_event"]["clearing_event_uploadtree_group_fk_idx"] = "CREATE INDEX clearing_event_uploadtree_group_fk_idx ON clearing_event USING btree (uploadtree_fk, group_fk, date_added);";
 
-  $Schema["INDEX"]["copyright"]["agent_fk_idx"] = "CREATE INDEX agent_fk_idx ON copyright USING btree (agent_fk);";
+  $Schema["INDEX"]["copyright"]["copyright_agent_fk_idx"] = "CREATE INDEX copyright_agent_fk_idx ON copyright USING btree (agent_fk);";
```
shaheemazmalmmd: Please consider droping index the old index if you are cheating a new index with different name.

src/copyright/agent/copyright.cc
```diff
@@ -131,7 +131,7 @@ int main(int argc, char** argv)
     {
       int uploadId = atoi(fo_scheduler_current());
 
-      if (uploadId == 0) continue;
+      if (uploadId <= 0) continue;
```
shaheemazmalmmd: Can you please tell me why this change was made ? in which case uploadId  goes to a negative value ?

src/www/ui/core-schema.dat
GMishx: It will be taken care here https://github.com/fossology/fossology/blob/master/src/lib/php/libschema.php#L372 by dropIndexes()

src/copyright/agent/copyright.cc
GMishx: The call here can pass negative values
https://github.com/fossology/fossology/blob/master/src/scheduler/agent/agent.c#L242

PR #1094, opened 2018-05-08
doc(screenshots): add wrongly deleted screenshots back into the source code

maxhbr:
They were accidentally with some cleanup commit


GMishx:
Won't it be good to provide new screenshots? These are from 2015's build and UI shown in `examples/Concludeb.jpg` has changed since.

maxhbr:
I also think that updating the screenshots is a good idea, but not the scope of this PR.

PR #1096, opened 2018-05-10
fix(user): Update user's current group while removing from group

### LINKED ISSUE #1003: Problem with group selection for a user
**How to reproduce:**

1.  Add a user to a group (say 'TestGroup1').
2.  Login with the user's account and change to newly assigned group (TestGroup1) and logout.
3.  Remove the user from the same group (TestGroup1).
4.  Login again with the user's account.

No information will be displayed and the Ajax query get struck as the user have no permissions in the group (TestGroup1). Since the user is a part of only group now, he/she can not change the group to solve the issue.

2018-03-15 09_52_05-browse

**Expected result:**

If the user is removed from the group, it should also be reflected in the selected group when the user logs in again.

GMishx:
**Fixes**
1.  If the current user's group is deleted, update the PHP Session.
2.  If a user is removed from it's selected group, select a new valid group.

**How to reproduce:**
1.  Add a user ('user1') to a group (say 'TestGroup1').
2.  Login with the user's ('user1') account and change to newly assigned group (TestGroup1) and logout.
3.  Remove the user from the same group (TestGroup1) from another account.
4.  Login again with the user's ('user1') account, the ajax call will stall.
---
1.  Create a new group ('TestGroup1').
2.  Change current group to 'TestGroup1'.
3.  Goto Admin >> Groups >> Delete Group.
4.  Delete 'TestGroup1'.
5.  Browse another page, the ajax call will stall.

Closes #1003.

### DIFF
--- a/src/lib/php/Dao/UserDao.php
+++ b/src/lib/php/Dao/UserDao.php
@@ -23,24 +23,30 @@ use Fossology\Lib\Auth\Auth;
 use Fossology\Lib\Db\DbManager;
 use Fossology\Lib\Util\Object;
 use Monolog\Logger;
+use Symfony\Component\HttpFoundation\Session\Session;
 
 class UserDao extends Object
 {
   const USER = 0;
   const ADMIN = 1;
   const ADVISOR = 2;
-  
+
   const SUPER_USER = 'fossy';
 
   /* @var DbManager */
   private $dbManager;
   /* @var Logger */
   private $logger;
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd:
I'm still able to reproduce the issue. Please have a look

GMishx:
Can you please check again @shaheemazmalmmd ?

PR #1102, opened 2018-05-18
fix(schema): check with the table property for current scheme

shaheemazmalmmd:
**HOW TO TEST**
- Run install and fo-postinstall on any existing fossology instance.
- Check for the execution time.

### DIFF
--- a/src/lib/php/libschema.php
+++ b/src/lib/php/libschema.php
@@ -1,7 +1,7 @@
 <?php
 /*
  Copyright (C) 2008-2014 Hewlett-Packard Development Company, L.P.
- Copyright (C) 2014-2015, Siemens AG
+ Copyright (C) 2014-2015, 2018 Siemens AG
 
  This library is free software; you can redistribute it and/or
  modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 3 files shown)


src/lib/php/libschema.php
```diff
@@ -383,7 +383,8 @@ function dropIndexes()
       }
       foreach ($IndexInfo as $name => $sql)
       {
-        if (empty($name) || $this->schema['INDEX'][$table][$name] == $sql)
+        if (empty($name) || $this->schema['INDEX'][$table][$name] == $sql ||
+             (str_replace($table." ", "public.".$table." ", $this->schema['INDEX'][$table][$name]) == $sql))
```
GMishx: Won't it be good to use str_replace at line no. 889 inside `addIndexes()` so that the call will be reduced to only once?

src/lib/php/libschema.php
```diff
@@ -415,7 +416,8 @@ function applyIndexes()
       }
       foreach ($indexInfo as $name => $sql)
       {
-        if (empty($name) || $this->currSchema['INDEX'][$table][$name] == $sql)
+        if (empty($name) || $this->currSchema['INDEX'][$table][$name] == $sql ||
+             ($this->currSchema['INDEX'][$table][$name] == str_replace($table." ", "public.".$table." ", $sql)))
```
GMishx: This call can also be removed with the previous change.

src/lib/php/libschema.php
shaheemazmalmmd: No. Only PSQL 9.5 or above  needs replace with "Public".  i think most of the people are still using older versions. so both the checks are required.

GMishx: What I am saying is that instead of replacing the string during the check, if we replace it at the time of assignment, we can save one `str_replace` call.

PR #1104, opened 2018-05-21
feat(readmeoss): add license shortname above the license text

shaheemazmalmmd:
**TO TEST**
- Upload any component and do some license clearing.
- Generate the READMEOSS and check if the license shortname exists above text.


GMishx:
Can we remove hard-coded value to `true` while calling `createReadMeOSSFormat()` in all the three calls?

GMishx:
Tested, working as expected.

PR #1113, opened 2018-05-29
fix(email): Prevent scheduler crashes

GMishx:
## Fixes
1. Check if SMTP variable is set, then only send email otherwise print a notice in log.
2. Separate function to get command (readability, traceability).
3. Check email client's return code.
4. Safe clear for PQresult (SafePQclear) which checks
    1. Result is not NULL
    2. Clean it
    3. Set result to NULL to prevent double-clear


GMishx:
Done the changes to update rows which already exists in DB.

PR #1126, opened 2018-06-22
fix(sysconfig): Change structure of values

GMishx:
Changed the structure of columns and values from plain string to array to ease modification.

### DIFF
--- a/src/lib/php/common-sysconfig.php
+++ b/src/lib/php/common-sysconfig.php
@@ -70,21 +70,21 @@ function ConfigInit($sysconfdir, &$SysConf)
   global $PG_CONN;
 
   /*************  Parse VERSION *******************/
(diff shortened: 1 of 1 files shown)


src/lib/php/common-sysconfig.php
```diff
@@ -195,56 +195,51 @@ function Populate_sysconfig()
 {
   global $PG_CONN;
 
-  $Columns = "variablename, conf_value, ui_label, vartype, group_name, group_order, description, validation_function, option_value";
+  $Columns = array("variablename", "conf_value", "ui_label", "vartype", "group_name",
```
ag4ums: change the variable name according to https://github.com/fossology/fossology/wiki/Coding-Style#coding-guidelines-for-php

Issue #1007, opened 2018-03-15
Building Fossology 3.2.0 fails on Fedora 27


msahyoun777:
While building Fossology 3.2.0 on Fedora 27 using the easy-install command, I get compile errors in pkagent.c

The relevant warnings and errors follow. I can upload the entire build output if necessary:

_```
make -s -C pkgagent
make[2]: Entering directory '/home/foss/Source/fossology-3.2.0/src/pkgagent'
writing VERSION file for pkgagent
make[3]: Entering directory '/home/foss/Source/fossology-3.2.0/src/pkgagent/agent'
In file included from pkgagent.h:31:0,
                 from pkgagent.c:31:
/usr/include/rpm/rpmlib.h:20:2: warning: #warning RPM 4.4.x compatibility layer has been removed in RPM >= 4.14 [-Wcpp]
 #warning RPM 4.4.x compatibility layer has been removed in RPM >= 4.14
  ^~~~~~~
pkgagent.c: In function 'ReadHeaderInfo':
pkgagent.c:384:18: warning: implicit declaration of function 'tagName'; did you mean 'ttyname'? [-Wimplicit-function-declaration]
     strcat( fmt, tagName(tag[i]));
                  ^~~~~~~
                  ttyname
pkgagent.c:384:18: warning: passing argument 2 of 'strcat' makes pointer from integer without a cast [-Wint-conversion]
In file included from pkgagent.h:23:0,
                 from pkgagent.c:31:
/usr/include/string.h:129:14: note: expected 'const char * restrict' but argument is of type 'int'
 extern char *strcat (char *__restrict __dest, const char *__restrict __src)
              ^~~~~~
pkgagent.c:387:14: warning: implicit declaration of function 'headerSprintf'; did you mean 'headerWrite'? [-Wimplicit-function-declaration]
     msgstr = headerSprintf(header, fmt, rpmTagTable, rpmHeaderFormats, &errstr);
              ^~~~~~~~~~~~~
              headerWrite
pkgagent.c:387:41: error: 'rpmTagTable' undeclared (first use in this function); did you mean 'rpmTagType'?
     msgstr = headerSprintf(header, fmt, rpmTagTable, rpmHeaderFormats, &errstr);
                                         ^~~~~~~~~~~
                                         rpmTagType
pkgagent.c:387:41: note: each undeclared identifier is reported only once for each function it appears in
[...]

PR #1116, opened 2018-06-07
feat(fossology): Support for Bionic Beaver

GMishx:
## Fixes
1.  Change dependency from heirloom-mailx to heirloom-mailx | s-nail (alias).
1.  Nomos: change sprintf statements to handle longer strings without unexpected results.
1.  Added dependencies for bionic
1.  Change Object class to ObjectType
1.  Added PHP 7.2 tests in Travis as allow failure (can be made necessary)

## Updates
1.  Updated composer dependencies.

## Drop support
**(Already dropped by #1153)**
Dropped support for following distros as they are no longer supported by their producer/developer.

<table>
<thead>
<tr>
<th>Code name</th>
<th>Distro</th>
<th>Version</th>
<th>Link</th>
</tr>
</thead>
<tbody>
<tr>
<td>Lenny</td>
<td>Debian</td>
<td>5</td>
<td><a href="https://wiki.debian.org/DebianLenny" rel="nofollow">https://wiki.debian.org/DebianLenny</a></td>
</tr>
<tr>
<td>Squeeze</td>
<td>Debian</td>
<td>6</td>
<td><a href="https://www.debian.org/releases/squeeze/" rel="nofollow">https://www.debian.org/releases/squeeze/</a></td>
</tr>
<tr>
<td>Hardy</td>
<td>Ubuntu</td>
<td>8.04 </td><td rowspan="9"><a href="https://wiki.ubuntu.com/Releases" rel="nofollow">https://wiki.ubuntu.com/Releases</a></td>
</tr>
<tr>
<td>Lucid</td>
<td>Ubuntu</td>
<td>10.04</td>
</tr>
<tr>
<td>Natty</td>
<td>Ubuntu</td>
<td>11.04</td>
</tr>
<tr>
<td>Oneiric</td>
<td>Ubuntu</td>
<td>11.10</td>
</tr>
<tr>
<td>Precise</td>
<td>Ubuntu</td>
<td>12.04</td>
</tr>
<tr>
<td>Quantal</td>
<td>Ubuntu</td>
<td>12.10</td>
</tr>
<tr>
<td>Raring</td>
<td>Ubuntu</td>
<td>13.04</td>
</tr>
<tr>
<td>Saucy</td>
<td>Ubuntu</td>
<td>13.10</td>
</tr>
<tr>
<td>Yakkety</td>
<td>Ubuntu</td>
<td>16.10</td>
</tr>
<tr>
<td>Isadora</td>
<td>Linux Mint</td>
<td>9</td>
[...]

PR #1139, opened 2018-07-11
fix(pkgagent): Added support for RPM >= 4.14

robertvolkmann:
Fix for #1007 . Building fails on Fedora 27 because RPM >= 4.14 has removed the RPM 4.4.x API compatibility.

## Changes

Used the following defines from the removed rpmlegacy.h:
```
#define headerSprintf(_h, _fmt, _tbltags, _exts, _emsg) \
	headerFormat((_h), (_fmt), (_emsg))
#define	tagName		rpmTagGetName
```
edit: Dropped support for RPM 4.4.x and RHEL/CentOS 5.

## How to test

Compile on Fedora 27.

## Question

Should we remove the legacy RPM 4.4.x stuff? RHEL/CentOS 5 are the remaining users of RPM 4.4.x but RHEL is in the Extended Life-cycle Support and CentOS 5 is now EOL.

### DIFF
--- a/src/pkgagent/agent/Makefile
+++ b/src/pkgagent/agent/Makefile
@@ -26,8 +26,6 @@ LDFLAGS_LOCAL = $(FO_LDFLAGS) -lpq -lrpm -lrpmio
 
 EXE = pkgagent
 HDRS = pkgagent.h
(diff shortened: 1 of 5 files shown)


GMishx:
Thank you @robertvolkmann for the contribution made.

Unfortunately, the changes suggested by you are already in my pull request #1116. You can pull the branch and please verify if the changes are working on your environment.

PR #1156, opened 2018-08-08
fix(wget_agent): Ignore test_proxy_ftp, because it is flaky on travis.

robertvolkmann:
## Description

The test test_proxy_ftp is flaky, because there are some issues with ftp and required sudo on travis. Since the essential part of the test is already commented out, we should ignore it until there exists a solution for travis.

### Changes

fix(wget_agent): Ignore test_proxy_ftp, because it is flaky on travis.
fix(wget_agent): Clean up the database only if it is necessary.


maxhbr:
The travis run here fails due to rate limitation on github, triggered by some unrelated test which talks to github. Code is good and the PR can be merged.

PR #1167, opened 2018-08-17
chore(travis): Simplified travis.yml

robertvolkmann:
* Removed unused dependencies.
* Removed caching for apt, because travis does not recommend it.
* Refactored matrix build config.
* Use yaml referencing to simplify compiler installation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,133 +6,122 @@
 language: php
 dist: trusty
 php:
-  - '7.0'
+  - 7.0
+  - 5.6
+  - 7.1
 addons:
- - postgresql: "9.3"
- - apt:
-    sources:
-    - ubuntu-toolchain-r-test
-    packages:
-    - gcc-4.8
-    - g++-4.8
-    - gcc-5
-    - g++-5
-    - gcc-6
-    - g++-6
+  apt:
+    packages: &default_packages
+      - bzip2
+      - cabextract
+      - cpio
+      - genisoimage
+      - libboost-program-options-dev
+      - libboost-regex-dev
+      - libcppunit-dev
+      - libcunit1-dev
+      - libdbd-sqlite3-perl
+      - libglib2.0-dev
+      - liblocal-lib-perl
+      - libmagic-dev
+      - librpm-dev
+      - libspreadsheet-writeexcel-perl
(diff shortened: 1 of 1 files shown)


maxhbr:
@robertvolkmann please rebase (due to https://github.com/fossology/fossology/pull/1173)

GMishx:
This PR really helps in maintaining the YAML. Out off curiosity, I found a difference in the travis runs. Can you explain the behavior why the job sequence is altered (the PHPUnit test cases came at top)?

Your PR:
pr

Master:
master

robertvolkmann:
Perhaps because the matrix expansions runs before any include.

.travis.yml
```diff
@@ -6,133 +6,122 @@
...
 env:
   global:
-   - PHPTESTSUITE=''
-   - CHECKBEFORE='cppcheck -q -isrc/nomos/agent_tests/testdata/NomosTestfiles/ -isrc/testing/dataFiles/ --suppress=*:src/copyright/agent/json.hpp src/'
-   - MAKETARGETS='all test'
+    - PHPTESTSUITE=''
+    - CHECKBEFORE='cppcheck -q -isrc/nomos/agent_tests/testdata/NomosTestfiles/ -isrc/testing/dataFiles/ --suppress=*:src/copyright/agent/json.hpp src/'
+    - MAKETARGETS='all test'
+  matrix:
+    - PHPTESTSUITE='Fossology PhpUnit Test Suite' CHECKBEFORE='' MAKETARGETS='build-lib VERSIONFILE build-cli'
```
GMishx: There are two `matrix` elements here. Can we merge the same in the `matrix` following `after_script`?

robertvolkmann: No, we cannot if we want the matrix expansion. The `matrix` element in `env` is necessary since we use global environment variables.
>Sometimes you may want to use environment variables that are global to the matrix, i.e. they’re inserted into each matrix row. That may include keys, tokens, URIs or other data that is needed for every build. In such cases, instead of manually adding such keys to each env line in matrix, you can use global and matrix keys to differentiate between those two cases.

See: https://docs.travis-ci.com/user/environment-variables/#global-variables

PR #1169, opened 2018-08-17
feat(docker): Implemented multi-staged build.

robertvolkmann:
Reduced the compressed size of the docker image from [378 MB](https://hub.docker.com/r/fossology/fossology/tags/) to [269 MB](https://hub.docker.com/r/robertvolkmann/fossology/tags/).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,14 +9,18 @@
 #
 # Description: Docker container image recipe
 
-FROM debian:jessie
+FROM debian:jessie as builder
 
 LABEL maintainer="Fossology <fossology@fossology.org>"
 
 WORKDIR /fossology
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get update \
- && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends lsb-release sudo \
+ && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+      git \
+      lsb-release \
+      php5-cli \
+      sudo \
  && rm -rf /var/lib/apt/lists/*
 
 COPY ./utils/fo-installdeps ./utils/fo-installdeps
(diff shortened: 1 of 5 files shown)


GMishx:
Unfortunately, I am unable to run Copyright (and sub programs) on the image created by your PR.

Please check the attached screenshots.

`fossology/fossology:latest`:
2018-08-23 11_38_17-ubuntu running - oracle vm virtualbox _ 2

`robertvolkmann/fossology:multi-stage`:
2018-08-23 11_38_40-ubuntu running - oracle vm virtualbox _ 2

GMishx:
@robertvolkmann, can you please update the image in docker hub as I have an old docker version which does not support multistage builds?

robertvolkmann:
@GMishx [Docker image updated.](https://hub.docker.com/r/robertvolkmann/fossology/tags/)

Dockerfile
```diff
@@ -27,30 +31,59 @@ COPY ./src/pkgagent/mod_deps ./src/pkgagent/
...
+COPY --from=builder /fossology/dependencies-for-runtime /fossology
+
+WORKDIR /fossology
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get update \
+ && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+      curl \
+      lsb-release \
+      sudo \
+      libboost-regex1.55.0 libboost-program-options1.55.0 \
```
GMishx: Can be removed after merging #1177

robertvolkmann: Done.

Dockerfile
```diff
@@ -27,30 +31,58 @@ COPY ./src/pkgagent/mod_deps ./src/pkgagent/
...
+ && rm -rf /var/lib/apt/lists/*
 
 COPY . .
 
-RUN make install && make clean
+RUN /fossology/utils/install_composer.sh
 
-# the database is filled in the entrypoint
-RUN /usr/local/lib/fossology/fo-postinstall --agent --common --scheduler-only --web-only --no-running-database
+RUN make && make install && make clean
```
GMishx: One suggestion, this statement can be reduced to
`RUN make install clean`

robertvolkmann: The split in three commands ensures that the installation runs only after a successful compilation and so on. This is necessary since we recursive call multiple Makefiles.

GMishx: The `install` recipe already have the dependency on `all` recipe (which is called when simply calling `make`). So `install` will only run if compilation (i.e. `all`) is completed successfully. And then you can chain the `clean` recipe in the same command.

robertvolkmann: Implemented.

PR #1092, opened 2018-05-08
fix(testCases): Fix scheduler and ununpack test cases

GMishx:
**Changes:**

1.  Can directly call `make test` from top directory to test every agent.
2.  Prevent printing VERSION file command.
3.  Change LLVM repository from precise to trusty.
4.  Changes suggested by clang-7 (C++17):
>  Removed call to move for copy elision.
>  copyright_error
>  wget_error

**How to test:**

1.  Make directory `/srv/fossologyTestRepo` and give permission to current user:
```
sudo mkdir -p /srv/fossologyTestRepo
sudo chown $(whoami) /srv/fossologyTestRepo
```
2.  Create `fossologytest` and `fossy` user in PostgreSQL and save their passwords in `~/.pgpass`.
```
psql -c "CREATE USER fossy WITH CREATEDB LOGIN PASSWORD 'fossy';" -U postgres
psql -c "CREATE USER fossologytest WITH CREATEDB LOGIN PASSWORD 'fossologytest';" -U postgres
echo "localhost:*:*:fossy:fossy" >> ~/.pgpass
echo "localhost:*:*:fossologytest:fossologytest" >> ~/.pgpass
chmod 0600 ~/.pgpass
```
3.  Install composer dependencies
```
cd src && composer install
```
4.  Run `make all test`
5.  Install cppcheck `sudo apt-get install cppcheck`.
6.  Run `cppcheck -q -isrc/nomos/agent_tests/testdata/NomosTestfiles/ -isrc/testing/dataFiles/ --suppress=*:src/copyright/agent/json.hpp src/`

PR #1170, opened 2018-08-17
Revert "fix(testCases): Fix scheduler and ununpack test cases"

maxhbr:
The pullrequest fossology/fossology#1092 introduces multiple problems (see https://github.com/fossology/fossology/pull/1092#commitcomment-30133056 and  https://github.com/fossology/fossology/pull/1092#issuecomment-413890496)

Reverts fossology/fossology#1092

@robertvolkmann thanks for pointing the issues out.


maxhbr:
Summaries of issues mentioned in the conversation of fossology/fossology#1092:

This PR
- elongates the execution times of from 2h30 - 3h30 to 4h45 - 5h30
- makes `make test` run the script `utils/prepare-test` which
  - **requires `sudo`**
  - calls `apt-get` to install dependencies
    - which is an **unexpected** side effect
    - which **only works on ubuntu / debain**
  - calls `chown $(whoami) /var/local/cache/fossology` and **breaks the current fossology instance**
  - creates the user `fossy` with password `fossy` in the db
    - this might change the password of `fossy` back to default which is a **potentially security risk** and **might break the current fossology instance**
  - creates a testuser in the current postgres which might be in productive use
  - writes two new lines to `~/.pgpass` on every run
- reactivates the blinking ftp test

GMishx:
Instead of reverting the PR, what I suggest is
1.  To remove the `utils/prepare-test` from the `test` target. Let the user decide to run it or not.
2.  Deactivate the FTP test (again).
3.  Reduce the redundant PHP-GCC test cases.

By removing the redundant PHP-GCC test cases, we can bring down the test matrix from 22 tests to **11**.

GMishx:
Okay, agree. Will add the warning message as well.

GMishx:
Please check #1173

maxhbr:
obsolete due to #1173

PR #1172, opened 2018-08-19
fix(scheduler): make init script wait for postgresql on startup

### LINKED ISSUE #1171: Scheduler fails on boot if database is not available
### Description

When using the Debian package version of Fossology 3.3.0, the Fossology scheduler tries to run on system boot. However, it fails because the PostgreSQL database is also booting but is not yet available.

This may occur in other situations that use the Fossology init script, beyond the Debian package.

#### How to reproduce

1. Install Fossology using the Debian package for 3.3.0
2. Reboot the system
3. The Fossology scheduler will try to run and will fail, leaving the following message in `/var/log/fossology/fossology.log`:

```
2018-08-19 11:01:07 scheduler [705] :: FATAL database.c.669: Unable to connect to the database: "ERROR: Unable to connect to the database
2018-08-19 11:01:07 scheduler [705] ::    Connection string: 'dbname=************** host=************** user=************** password=**************'
2018-08-19 11:01:07 scheduler [705] ::    Connection status: '1'
2018-08-19 11:01:07 scheduler [705] ::   Check: /usr/local/etc/fossology/Db.conf
2018-08-19 11:01:07 scheduler [705] :: "
2018-08-19 11:01:07 scheduler [705] :: FATAL errno is: Success
```

### Versions

* Last commit id on master: 9fb9c7 (according to header within Fossology interface)
```
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.5 (stretch)
Release:	9.5
Codename:	stretch
```

### Fossology logs

See snippet pasted above

### Fix

One fix appears to be as simple as telling the init script to wait until postgresql is available. I'm going to submit this as a PR.

I think this would fix it for situations like the Debian package (where the scheduler and PostgreSQL are on the same system). However, this solution may not work for people who are using the init script in a context where the database is remote.

swinslow:
Signed-off-by: Steve Winslow <swinslow@gmail.com>

## Description

Fixes #1171 

This makes the init script for the scheduler agent wait for PostgreSQL to start up.

### Changes

In the init script, 'postgresql' was added to the Required-Start: and Required-Stop: header lines.

## How to test

Please note that I have tested this in production, when using the Debian packages for 3.3.0, by manually making this change to `/etc/init.d/fossology`. I've confirmed that this appears to fix the recurring problem in #1171 where the scheduler would fail to start, due to the unavailability of the PostgreSQL database during boot-up.

I have not rebuilt the packages with the change made in this PR, so someone who is familiar with the packages may of course want to do so.

### DIFF
--- a/src/scheduler/agent/defconf/init.d/fossology.in
+++ b/src/scheduler/agent/defconf/init.d/fossology.in
@@ -8,8 +8,8 @@
 #
 ### BEGIN INIT INFO
 # Provides:          {$PROJECT}
-# Required-Start:    $network $local_fs $remote_fs $syslog $named
-# Required-Stop:     $network $local_fs $remote_fs $syslog $named
+# Required-Start:    $network $local_fs $remote_fs $syslog $named postgresql
+# Required-Stop:     $network $local_fs $remote_fs $syslog $named postgresql
 # Default-Start:     2 3 4 5
 # Default-Stop:      0 1 6
 # Short-Description: FOSSology scheduler daemon


GMishx:
Thanks for the fix. But what if I am running an instance with an external DB and don't have PostgreSQL installed on FOSSology instance?

maxhbr:
@GMishx good point

swinslow:
Yes -- I mentioned the point @GMishx raised at the bottom of the issue report in #1171. Apologies for not mentioning it here also.

I think the fix is helpful at a minimum for people who are using the Debian packaged version (may also be the case for the Vagrant install, though I haven't tried it). In that use case, someone is expecting to use the PostgreSQL server that is automatically installed, and may not understand why the agents are not running on their system.

Perhaps there should be separate init scripts for the packaged install, vs. for people who are experienced and familiar enough with Fossology setup to be able to configure a PostgreSQL database on a separate instance?

maxhbr:
> Perhaps there should be separate init scripts for the packaged install, vs. for people who are experienced and familiar enough with Fossology setup to be able to configure a PostgreSQL database on a separate instance?

I would support that.

maxhbr:
@swinslow @GMishx due to the problems with systems, which do not control the postgres themself, I think it would be better to revert this PR and reopen the issue. We need a solution which works in both cases.

PR #1173, opened 2018-08-20
Fix/test recipie prepare

GMishx:
## Description

Fix suggested by #1170

### Changes

- .travis.yml
    - Remove redundant GCC test cases with different flavours of PHP as PHP test cases are covered under PHPUnit section.
    - This results in reduction of test cases from 22 => 10.
- Makefile
    - Remove preparetest target from test target.
    - User need to run `./utils/prepare-test -afty` once to setup test environment.
        - See `./utils/prepare-test -h` for more.
- wget
    - Disable the failing FTP test.
    - Thanks @robertvolkmann and @maxhbr for pointing the issues out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,9 +66,6 @@ $(CLEANDIRS):
 phpvendors:
 	$(MAKE) -C $(FOSRCDIR) phpvendors
 
-preparetest: ./utils/prepare-test
-	./utils/prepare-test -y -a
-
 # release stuff
 tar: dist-testing
 dist-testing:
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ phpvendors:
 	$(MAKE) -C $(FOSRCDIR) phpvendors
 
 preparetest: ./utils/prepare-test
-	./utils/prepare-test
+	./utils/prepare-test -y -a
 
 # release stuff
 tar: dist-testing
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ $(UNINSTALLDIRS):
 	$(MAKE) -C $(@:uninstall-%=%) uninstall
 
 # test depends on everything being built first
-test: all preparetest $(TESTDIRS)
+test: all $(TESTDIRS)
 $(TESTDIRS):
 	$(MAKE) -C $(@:test-%=%) test
 
--- a/utils/prepare-test
+++ b/utils/prepare-test
@@ -23,14 +23,90 @@
 # canonical location for such info, then fix it there first and also
 # update this file and the INSTALL document.
 
-sudo mkdir -p /var/local/cache/fossology
-sudo chown $(whoami) /var/local/cache/fossology
-sudo mkdir -p /srv/fossologyTestRepo
-sudo chown $(whoami) /srv/fossologyTestRepo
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -67,7 +67,7 @@ phpvendors:
 	$(MAKE) -C $(FOSRCDIR) phpvendors
 
 preparetest: ./utils/prepare-test
```
maxhbr: `make preparetest` is still breaking the system without warning. I would suggest to remove that target and call `./utils/prepare-test -y -a` directly in the travis script

utils/prepare-test
```diff
@@ -23,14 +23,90 @@
...
+
+EOF
+
+while getopts ":ayh" o; do
+  case "${o}" in
+    a)
+      WarnAccept='Y'
+      ;;
+    y)
+      YesOpt='-y'
```
maxhbr: this variable is not used for apt get

utils/prepare-test
```diff
@@ -23,14 +23,90 @@
...
+if echo ${WarnAccept} | grep -q -e "^[yYnN]$" ; then
+  answer="${WarnAccept}"
+else
+  read -r -n 1 -p "Are you sure run this script?(Y/n) " answer
+  if [[ -z ${answer} ]]; then
+    answer="Y"
+  fi
+fi
+
+echo ""
```
maxhbr: this is equivalent to just
```
echo
```

utils/prepare-test
```diff
@@ -23,14 +23,90 @@
...
+      show_help; exit;;
+    *)
+      show_help; exit 1;;
+   esac
+done
+
+if echo ${WarnAccept} | grep -q -e "^[yYnN]$" ; then
+  answer="${WarnAccept}"
+else
+  read -r -n 1 -p "Are you sure run this script?(Y/n) " answer
```
maxhbr: yes should not be the default in a "if you run this script your productive instance may die"-script

utils/prepare-test
```diff
@@ -23,14 +23,90 @@
...
+      YesOpt='-y'
+      ;;
+    h)
+      show_help; exit;;
+    *)
+      show_help; exit 1;;
+   esac
+done
+
+if echo ${WarnAccept} | grep -q -e "^[yYnN]$" ; then
```
maxhbr: Why not 
```bash

￼     a)
￼       answer='Y'
￼       ;;
```

and then
```bash
if [[ -z ${answer} ]]; then
  read -r -n 1 -p "Are you sure run this script?(Y/n) " answer
￼   if [[ -z ${answer} ]]; then
￼     answer="Y"
￼   fi
￼ fi
```

maxhbr:
now most tests fail with `psql: FATAL:  database "fossology" does not exist`...

utils/prepare-test
```diff
@@ -23,14 +23,95 @@
...
+  if [[ -z ${WarnAccept} ]]; then
+    WarnAccept="N"
+  fi
+fi
+
+echo
+
+if echo ${WarnAccept} | grep -q -e "^[yY]$" ; then
+  sudo mkdir -p /var/local/cache/fossology
+  sudo chown "$(whoami)" /var/local/cache/fossology
```
maxhbr: I still do not like this line. This should kill the running fossology instance in most cases (e.g. if whoami != fossy)

PR #1178, opened 2018-08-27
chore(travis): Removed global environment variables usage.

robertvolkmann:
## Changes

* Moved syntax check and static code analysis in separate steps.
* Merged docker tests into one job to build the image only once.
* Removed not working phpunit update.
* Removed already installed packages.
* Split make all test into two steps.

## How to test

Compare the logs of travis-ci.

PR #1183, opened 2018-08-27
feat(composer): Updated development dependencies.

robertvolkmann:
## Description

The `composer.lock` is out-of-sync to `composer.json`. Someone has missed to update the `composer.lock` after changing `composer.json`.

### Changes

* Generated a recent `composer.lock`.
* Updated the development dependencies to the latest versions with support for php 5.6.
* Switched to Mockery::pattern() since Mockery 1.0 removed implicit regex matching.
* Upgraded bootstrapping of Hamcrest

## How to test

* Run tests or check logs on travis-ci.
* Test fossology.

### DIFF
--- a/src/composer.json
+++ b/src/composer.json
@@ -20,12 +20,12 @@
         "phpoffice/phpword" : "v0.13.*"
     },
     "require-dev" : {
-        "mockery/mockery" : ">=0.9.1",
-        "hamcrest/hamcrest-php" : "1.1.*",
-        "squizlabs/php_codesniffer" : "1.*",
-        "sebastian/phpcpd" : "2.*",
-        "php-coveralls/php-coveralls" : "1.0.*",
-        "phpunit/phpunit" : "~5"
+        "hamcrest/hamcrest-php" : "^2",
+        "mockery/mockery" : "^1",
+        "php-coveralls/php-coveralls" : "^2",
+        "phpunit/phpunit" : "^5.7",
+        "sebastian/phpcpd" : "^3",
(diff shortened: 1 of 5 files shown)


GMishx:
Consider removing the PHP environment dependency from `composer.json` to get the PHP version from the environment and updating the dependencies accordingly.

It tried it once, there were some dependency issues with PHPUnit.

It will be a good thing to do and will also fix the no-op installs of PHPUnit in travis (removed in #1178 )

robertvolkmann:
@GMishx From my understanding, we should maintain the version lock to php 5.6 in the platform section as long as it is our minimal supported php version. Otherwise we will accidentally create a `composer.lock` that is not compatible with php 5.6.

In the meantime we should add a travis job to test the compatibility with phpunit 6. Maybe something like:
```
if [[ ${TRAVIS_PHP_VERSION:0:3} != "5.6" ]]; then
  rm -f src/composer.lock
  sed -i 's/5\.6/7.0/' src/composer.json
  composer require --dev --no-update --working-dir=src phpunit/phpunit ^6
fi
```

PS: Exists already a plan to drop support for php 5.6?

robertvolkmann:
- [PHPUnit 6 is compatibly with PHP 7.0, PHP 7.1, PHP 7.2](https://phpunit.de/). Actually we don't need to support PHPUnit 7 for PHP 7.1+.
- [Trusty supports only PHP 5.5](https://packages.ubuntu.com/trusty/php5).
- After the release of Debian Buster we should drop official support for Jessie and PHP 5.6.

PR #1187, opened 2018-08-28
fix(easy-install): only use the SUDO command if necessary

NicolasToussaint:

## Description

The install script forcibly uses the `sudo` command ; however, this fails when the user running the script is already `root` (when I install Fossology inside a Chroot).
The proposed changes simply use the `sudo` command when the current user is not `root`

### Changes

I consider that the `sudo` command is not needed if the user is allowed to `touch /` ... simple way to test `root` permissions.

## How to test

1. Run script as root user
2. Run script as non-root user

### DIFF
--- a/install/easy-install
+++ b/install/easy-install
@@ -8,19 +8,20 @@
 # without any warranty.
 #
 # Description: Easy install script for fossology. Goes through the steps on the Fossology wiki
+touch / 2>/dev/null || SUDO=sudo
 set -e
 cd "$(dirname $(readlink -f "${BASH_SOURCE[0]}"))"/..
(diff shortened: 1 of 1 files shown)


robertvolkmann:
Instead of touching root, you should check via `id --user` that the user ID is equal to zero. But I don't understand the issue. On my system I can run `sudo` as root and everything works. Have you tried to install `sudo` in your Chroot?

NicolasToussaint:
Oh, my root user can now use the `sudo` command - thought I didn't install it manually.
I should have checked the root of the problem instead of crafting this hack :-/
I guess the PR can be ignored then.

PR #1190, opened 2018-08-28
feat(test): Bypass API rate limit of github.

robertvolkmann:
## Description

[Sometimes](https://travis-ci.org/fossology/fossology/jobs/421499631) the PHP tests failed because they hit the [API rate limit of GitHub](https://developer.github.com/v3/#rate-limiting).

### Changes

Replaced the API call in the tests with a test double.

## How to test

* Run the `RepositoryApiTest`.
* Check that the Aboutpage still works.


src/lib/php/Application/RepositoryApi.php
```diff
@@ -16,42 +16,95 @@
...
+
+class CurlRequestServer
+{
+  public function create($url)
+  {
+    return new CurlRequest($url);
+  }
+}
+
+class RepositoryApi
```
maxhbr: We usually do not have multiple classes in one file

robertvolkmann: Done.

src/lib/php/Application/test/RepositoryApiTest.php
```diff
@@ -16,22 +16,57 @@
...
-  public function testGetCommitsOfLastDays() {
-    $result = $this->repositoryApi->getCommitsOfLastDays(60);
-    assertThat($result, is(not(emptyArray())));
-    assertThat($result[0], hasKey('sha'));
+
+  public function testGetCommitsOfLastDays()
+  {
+    $mockCurlRequestServer = \Mockery::mock('CurlRequestServer');
+    $mockCurlRequestServer->shouldReceive('create')->once()
+      ->with('https://api.github.com/repos/fossology/fossology/commits?since=2018-06-28T12:00:00Z')
```
maxhbr: Where does the date `2018-06-28T12:00:00Z` come from?

robertvolkmann: From the above `time`. The return value represent the data as unix timestamp and is the test double for the php `time` function.

PR #1199, opened 2018-09-04
Improve json hpp install

NicolasToussaint:
## Description

I have several difficulties when installing Fossology from sources:
- sometimes my network is unstable, and the whole install process fails because the download of the json.hpp file fails
- Quite often, the machines I install Fossology on require a proxy to access the internet.

### Changes

- Check first that file final file is not already available (sith correct checksum)
- Added the Retry and Timeout options to curl, to make the installation process more robust
- Let the user configure a proxy by sourcing a new configuration file: `/etc/fossology/fossology-proxy.conf`

## How to test

Install Fossology from source, through a proxy

### DIFF
--- a/install/scripts/install-json-hpp.sh
+++ b/install/scripts/install-json-hpp.sh
@@ -1,9 +1,5 @@
 #!/bin/bash
 
-# Source fossology config file, for Proxy settings
-conf_file=/etc/fossology.conf
-[ -r $conf_file ] && . $conf_file
-
 cd "$(dirname "$0")/../.."
 FINAL_FILE=./src/copyright/agent/json.hpp
 FILE_SHASUM="fbdfec4b4cf63b3b565d09f87e6c3c183bdd45c5be1864d3fcb338f6f02c1733"
--- a/install/scripts/install-json-hpp.sh
+++ b/install/scripts/install-json-hpp.sh
@@ -19,6 +19,10 @@ _check_file() {
 # Check if file is already available
 _check_file "$SHASUM" "$FINAL_FILE" && exit 0
 
+# Proxy Setup: If needed, configure the proxy variables in the appropriate config file
+proxy_file=/etc/fossology/fossology-proxy.conf
+[ -r $proxy_file ] && . $proxy_file
+
 TMP=$(mktemp)
 curl -s $CURL_OPION -o "$TMP" -L $FILe_URL
 SUCCESS=$?
--- a/install/scripts/install-json-hpp.sh
+++ b/install/scripts/install-json-hpp.sh
@@ -2,7 +2,7 @@
 
 cd "$(dirname "$0")/../.."
 FINAL_FILE=./src/copyright/agent/json.hpp
-FILE_SHASUM="faa2321beb1aa7416d035e7417fcfa59692ac3d8c202728f9bcc302e2d558f57"
+FILE_SHASUM="fbdfec4b4cf63b3b565d09f87e6c3c183bdd45c5be1864d3fcb338f6f02c1733"
 FILE_URL="https://github.com/nlohmann/json/releases/download/v2.1.1/json.hpp"
 CURL_OPTIONS="--max-time 10 --retry 4"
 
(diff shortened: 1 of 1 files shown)


install/scripts/install-json-hpp.sh
```diff
@@ -2,20 +2,42 @@
 
 cd "$(dirname "$0")/../.."
 FINAL_FILE=./src/copyright/agent/json.hpp
-SHASUM=fbdfec4b4cf63b3b565d09f87e6c3c183bdd45c5be1864d3fcb338f6f02c1733
-TMP=$(mktemp)
-CHECKSUMFILE=$(mktemp)
+FILE_SHASUM="faa2321beb1aa7416d035e7417fcfa59692ac3d8c202728f9bcc302e2d558f57"
+FILE_URL="https://github.com/nlohmann/json/releases/download/v2.1.1/json.hpp"
```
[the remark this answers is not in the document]
NicolasToussaint: Oh, thanks, will fix that.

install/scripts/install-json-hpp.sh
```diff
@@ -2,20 +2,42 @@
...
+}
+
+# Check if file is already available
+_check_file "$SHASUM" "$FINAL_FILE" && exit 0
 
-curl -s -o "$TMP" -L https://github.com/nlohmann/json/releases/download/v3.1.2/json.hpp
-echo "$SHASUM $TMP" > "$CHECKSUMFILE"
-sha256sum --quiet -c "$CHECKSUMFILE"
+# Proxy Setup: If needed, configure the proxy variables in the appropriate config file
+proxy_file=/etc/fossology/fossology-proxy.conf
```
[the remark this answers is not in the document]
NicolasToussaint: Thanks, didn't realize I will reuse that config entry

install/scripts/install-json-hpp.sh
```diff
@@ -1,21 +1,43 @@
...
+conf_file=/etc/fossology.conf
+[ -r $conf_file ] && . $conf_file
+
 cd "$(dirname "$0")/../.."
 FINAL_FILE=./src/copyright/agent/json.hpp
-SHASUM=fbdfec4b4cf63b3b565d09f87e6c3c183bdd45c5be1864d3fcb338f6f02c1733
-TMP=$(mktemp)
-CHECKSUMFILE=$(mktemp)
+FILE_SHASUM="fbdfec4b4cf63b3b565d09f87e6c3c183bdd45c5be1864d3fcb338f6f02c1733"
+FILE_URL="https://github.com/nlohmann/json/releases/download/v2.1.1/json.hpp"
```
GMishx: Please check the version as well. (v2.1.1 > v3.1.2)

install/scripts/install-json-hpp.sh
```diff
@@ -1,21 +1,43 @@
 #!/bin/bash
 
+# Source fossology config file, for Proxy settings
+conf_file=/etc/fossology.conf
```
GMishx: It is not a good idea to hard code path here.
If you are installing FOSSology from Debian packages, then this path is correct.
If you install FOSSology from source, the path is more like `/usr/local/etc/fossology/fossology.conf`

install/scripts/install-json-hpp.sh
NicolasToussaint: Stupid me, URL is also fixed.

install/scripts/install-json-hpp.sh
GMishx: You can try the process target here:
https://github.com/fossology/fossology/blob/master/Makefile.process#L16
And make something like:
https://github.com/fossology/fossology/blob/master/install/defconf/fossology.conf.in#L48

The variables which you can look into are:
https://github.com/fossology/fossology/blob/master/Makefile.conf#L40

NicolasToussaint: Many thanks for the explanations and pointers - I have removed the proxy part of the PR altogether, I will have to work on this separately later.

ag4ums:
the fix is here #1226  so closing this.

PR #1210, opened 2018-09-19
fix(schema): Add missing constraints

### LINKED ISSUE #274: Missing constraint in license_file
---

Author Name: **Bob Gobeille** 
Original Redmine Issue: 6753, http://www.fossology.org/issues/6753
Original Date: 2014/03/18
Original Assignee: Bob Gobeille

---

the license_file table needs a foreign key constraint on rf_fk to insure that any given rf_fk refers to a real license_ref record.

GMishx:
## Description

Add missing constraints, indexes and removed some old tables.

Added new migration file to handle the new constraints.

Also update the default release value in `sysconfig` table to `3.3.0`.

### New constraints:

1. `bucket_container_buck`
1. `bucket_file_bucket_fk_fkey`
1. `bucket_file_pfile_fk_fkey`
1. `clearing_decision_event_event_fk_fkey`
1. `clearing_decision_event_decision_fk_fkey`
1. `clearing_event_job_fk_fkey`
1. `copyright_decision_pfile_fk_fkey`
1. `ecc_agent_fk_fkey`
1. `ecc_pfile_fk_fkey`
1. `ecc_decision_pfile_fk_fkey`
1. `foldercontents_parent_fk_fkey`
1. `group_user_member_user_group_ukey`
1. `highlight_keyword_pfile_fk_fkey`
1. `keyword_agent_fk_fkey`
1. `keyword_pfile_fk_fkey`
1. `keyword_decision_pfile_fk_fkey`
1. `license_file_rf_fk_fkey`
1. `license_ref_bulk_upload_fk_fkey`
1. `obligation_ref_pkey`
1. `obligation_ref_md5_ukey`
1. `pkg_deb_req_pkg_fk_fkey`
1. `pkg_rpm_req_pkg_fk_fkey`
1. `report_cache_report_cache_uploadfk_fkey`
1. `report_info_pkey`
1. `report_info_upload_fk_fkey`
1. `reportgen_upload_fk_fkey`
1. `reportgen_job_fk_fkey`
1. `upload_pfile_fk_fkey`
1. `upload_clearing_license_upload_fk_fkey`
1. `upload_clearing_license_rf_fk_fkey`

### New indexes

1. `upload_clearing_license_upload_group_idx`

### Removed tables

1. `agent_runstatus`
1. `attachments`
1. `runstat_up_agent` (view)

Please check [core-schema.dat](src/www/ui/core-schema.dat) for the constraint definition.

Closes #274


install/fossinit.php
```diff
@@ -166,6 +166,11 @@ function explainUsage()
   require_once("$LIBEXECDIR/resequence_author_table.php"); // If table exists, clean up for Schema
 }
 
+if($isUpdating && substr(str_replace(".", "", $sysconfig['Release']), 0, 3) <= 330)
+{
+  require_once("$LIBEXECDIR/dbmigrate_3.3-3.4.php");
+  Migrate_33_34($dbManager, $Verbose);
```
maxhbr: Shouldn't the `Migrate_33_34` call be later in the script, e.g. after the older migration scripts? otherwise one would have to call the script multiple times?

install/fossinit.php
```diff
@@ -346,7 +351,7 @@ function explainUsage()
 {
   require_once("$LIBEXECDIR/dbmigrate_multiple_copyright_decisions.php");
 
-  $sysconfig['Release'] = "3.1.0";
+  $sysconfig['Release'] = "3.3.0";
```
maxhbr: Doesn't that cause problems with current installations, which have `3.1.0` written in the DB as current datamodel version?

I would find better to not change this line but instead have the following additional lines:
```php
// fix release-version datamodel-version missmatch
if($isUpdating && (empty($sysconfig['Release']) || $sysconfig['Release'] == '3.1.0')) {
    $sysconfig['Release'] = "3.3.0";
}
```

install/fossinit.php
```diff
@@ -166,6 +166,11 @@ function explainUsage()
   require_once("$LIBEXECDIR/resequence_author_table.php"); // If table exists, clean up for Schema
 }
 
+if($isUpdating && substr(str_replace(".", "", $sysconfig['Release']), 0, 3) <= 330)
```
maxhbr: -  Why is this better than
  ```php
  if($isUpdating && (empty($sysconfig['Release']) || $sysconfig['Release'] == '3.3.0'))
  ```

- would your code throw a NPE if `$sysconfig['Release'] === null` ?

install/fossinit.php
maxhbr: But this means, that if you migrate from 3.0.0 -> 3.4.0 you have to call the script multiple times?

install/fossinit.php
maxhbr: oh, I have overlooked the `<=`.

install/fossinit.php
maxhbr: fixed by: https://github.com/fossology/fossology/pull/1210#discussion_r220063278

PR #1198, opened 2018-09-04
feat(api): add fossology REST api

GMishx:
Continuation of #837

## FOSSology REST OpenAPI specification:
Can be viewed in rendered form on http://editor2.swagger.io/

**Implemented**:

| Request | API |
| --- | --- |
| GET       | /repo/api/v1/auth/ |
| GET       | /repo/api/v1/uploads/{id} |
| DELETE | /repo/api/v1/uploads/{id} |
| PATCH   | /repo/api/v1/uploads/{id} |
| PUT       | /repo/api/v1/uploads/{id} |
| GET       | /repo/api/v1/uploads/ |
| GET       | /repo/api/v1/search/ |
| GET       | /repo/api/v1/users/ |
| GET       | /repo/api/v1/users/{id} |
| DELETE | /repo/api/v1/users/{id} |
| GET       | /repo/api/v1/jobs/ |
| GET       | /repo/api/v1/jobs/{id} |
| POST     | /repo/api/v1/jobs/ |

The user can get the required authentication header by calling:
`/repo/api/v1/auth/?username=<username>&password=<password>`

### Steps to create HTML from Swagger YAML
1. Install [pretty-swag](https://github.com/twskj/pretty-swag#installation).
1. Change the `host` and `schema` in `src/www/ui/api/documentation/openapi.yaml` to match current implementation.
1. Generate the HTML:
    ```bash
    pretty-swag -i path/to/openapi.yaml -c path/to/pretty-swag.json -o path/to/output.html
    ```
1. Host the HTML

Sample (with host as HTML): https://gmishx.github.io/fossologyrest/output.html

PR #1214, opened 2018-09-28
feat(restapi): New endpoints

GMishx:
## Description

1. Changed the [Silex](https://silex.symfony.com/) framework to [Slim](https://www.slimframework.com/) framework.
2. Add new endpoints

### Changes

New paths:

| Method | Path |
| --- | --- |
| GET | /folders |
| POST | /folders |
| GET | /folders/{id} |
| DELETE | /folders/{id} |
| PATCH | /folders/{id} |
| PUT | /folders/{id} |
| GET | /report |
| GET | /report/{id} |
| POST | /uploads |

Continuation of #1198


max-wittig:
The folders endpoint doesn't seem to work for me
```
127.0.0.1:8081/repo/api/v1/folders/
{
    "code": 404,
    "message": "Resource not found",
    "type": "ERROR"
}
```

image

GMishx:
New endpoints in this PR are for Version 2. So try using
`127.0.0.1:8081/repo/api/v2/folders/`

max-wittig:
@GMishx If version 1 is not released yet, why call it version 2? API versioning works differently from software versioning in most cases. The API version should only be changed in rare cases, if something really breakes. (e.g. Gitlab, Github are using the same version for years, even though they always add new endpoints)

max-wittig:
@GMishx I haven't reviewed the code, yet, but awesome work from a feature perspective :+1:  Looking forward to seeing this in the real world!

More info about url versioning (which we use) can be found here also: 
https://restfulapi.net/versioning/

```txt
APIs only need to be up-versioned when a breaking change is made. Breaking changes include:

    a change in the format of the response data for one or more calls
    a change in the response type (i.e. changing an integer to a float)
    removing any part of the API.

Breaking changes should always result in a change to the major version number 
for an API or content response type.

Non-breaking changes, such as adding new endpoints or new response parameters,
do not require a change to the major version number. 
However, it can be helpful to track the minor versions of APIs when changes
are made to support customers who may be receiving 
cached versions of data or may be experiencing other API issues.
```

src/Makefile
```diff
@@ -84,7 +84,7 @@ phpvendors:
 composer_install:
 	@echo "current dir: '$${PWD}'"
 	$(INSTALL_DATA) composer.json composer.lock $(DESTDIR)$(MODDIR)
-	(cd $(DESTDIR)$(MODDIR); export COMPOSER_HOME=/dev/null; $(COMPOSER_PHAR) install --no-plugins --no-scripts --no-dev)
+	(cd $(DESTDIR)$(MODDIR); $(COMPOSER_PHAR) install --no-plugins --no-scripts --no-dev)
```
maxhbr: I think there was a reason for that export. I think it caused permission problems on some systems / when run with specific user.

GMishx: 1. I removed the export to get caching capability of composer.
1. Anyway, you need to run this target as root user. So this should not cause any permission issues.

maxhbr: There were problems on some systems/users. And if one runs this target for creating a package, it is ran without root privileges. And there is also nothing that would someone prevent from installing as user into a local folder, there are also no root privileges needed. 

This change will break for some users.

GMishx: - I just tried building packages on this branch (using `dpkg-buildpackage -I`) and it went through just fine.
- From the [composer documents](https://getcomposer.org/doc/03-cli.md#composer-home), the `COMPOSER_HOME` by default directs to current user's home directory. And every user have access to their home directory.
- This change will really speed up the source install process.

I can not think of any good reason to revert the change back.

GMishx: Reverting the change here. Will open a separate PR for caching.

PR #1215, opened 2018-09-28
fix(tests): fix PHPCS and phpunit testcases for deciderjob and reuser

shaheemazmalmmd:
**Description**

* Changes are made to correct PHPCS
* add reuser tests to phpunit
* add comments to public tests

**How to test**

install the branch and run PHPCS and phpunit testcases in your local machine.

### DIFF
--- a/src/reuser/agent/Makefile
+++ b/src/reuser/agent/Makefile
@@ -13,14 +13,14 @@ include $(VARS)
 COPY = reuser.php version.php ReuserAgent.php
 WRAP = reuser
 
-all: version.php
+all: version.php reuser
 
 version.php: version-process_php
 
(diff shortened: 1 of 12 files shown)


src/reuser/agent/Makefile
```diff
@@ -13,7 +13,7 @@ include $(VARS)
 COPY = reuser.php version.php ReuserAgent.php
 WRAP = reuser
 
-all: version.php
+all: version.php reuser
```
GMishx: It will be better to add the `reuser` dependency in test recipe instead in `all` as it is only meant for testing (and is throwing [some warnings](https://travis-ci.org/fossology/fossology/jobs/436563120#L3446) as well)

PR #1224, opened 2018-10-16
feat(select2): Use select2 lib for drop-downs

### LINKED ISSUE #1223: Using select 2 library for all large drop-down menus in FOSSology
### Description

Most of the drop-downs in FOSSology grow large which makes it difficult to find required options in them.
Especially in the case of Folders drop-down, the indentation makes it not searchable in Firefox.

It will be better to utilize the search functionality provided by the [select2](https://github.com/select2/select2) library.

##### Not so friendly in Firefox
old

##### Proposed UI
new

GMishx:
## Description

Use select2 library to add search functionality in long drop-downs.

### Changes

Added new CSS class called `ui-render-select2` to call/apply select2 library on the elements with this class.

Closes #1223

### DIFF
--- a/src/delagent/ui/admin-folder-delete.php
+++ b/src/delagent/ui/admin-folder-delete.php
@@ -128,9 +128,9 @@ class admin_folder_delete extends FO_Plugin {
     $V.= "</ul>\n";
     $text = _("Select the folder to delete:  ");
     $V.= "<P>$text\n";
-    $V.= "<select name='folder'>\n";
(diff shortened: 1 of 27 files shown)


shaheemazmalmmd:
* **@GMishx Please add min-width to bulk license as for short license names it will be difficult search other licenses.**

min_width


* **Can you also add select2 in upload with reuse where we select the folder.**

ncapture

GMishx:
@shaheemazmalmmd, changes requested by you are done. Can you please check again?

PR #1226, opened 2018-10-20
feat(copyright): Use package based dependency for json

### LINKED ISSUE #1084: Change Json HPP dependency to package-based
This works then Deb9 based and related systems as the package is there.

For red hat / yum based systems, there needs to be another solution.

GMishx:
## Description

Remove dependency from external file download for json-hpp and shift to packaged based dependency (`libjsoncpp` from Debian based and `jsoncpp` from RedHat based systems).

### Changes

1. Remove download for external json-hpp file.
1. Add `jsoncpp` package dependency.

## How to test

Compile the `copyright` package in Debian and RedHat based systems and check the JSON output.

```bash
$ sudo ./utils/fo-installdeps -ye
$ make copyright
$ cd src/copyright/agent
$ ./copyright -J ./copyright.cc
```

Closes #1084


mcjaeger:
appreciate this commit, but I did some testing on 3.4.0 rc1 already to make sure it is working. I am not so sure about adding this on short final. maybe tests fail because of timing issues of course.

GMishx:
The tests are now fixed. Can you please check again?

mcjaeger:
(from discussion) let#s compare output of copyright with open ssl for external dependency and package dependency.

mcjaeger:
(from discussion:) better to use the c++ lib instead of the c one

GMishx:
Found `jsoncpp` for Fedora 27.

And can be enabled in CentOS 7 using EPEL.
`sudo yum install epel-release`

src/copyright/agent/copyright.cc
```diff
@@ -54,7 +54,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 #include "copyright.hpp"
 
 #ifndef DISABLE_JSON
```
ag4ums: we are installing the lib now so "#ifndef DISABLE_JSON"  can be removed,

src/copyright/mod_deps
```diff
@@ -0,0 +1,117 @@
...
+      case "$CODENAME" in
+        jessie|trusty)
+          apt-get $YesOpt install libjsoncpp0;;
+        stretch|buster|sid|artful|bionic|xenial|cosmic)
+          apt-get $YesOpt install libjsoncpp1;;
+        *) echo "ERROR: Unknown or Unsupported $DISTRO $CODENAME release, please report to the mailing list"; exit 1;;
+      esac;;
+    Fedora)
+      yum $YesOpt install \
+        jsoncpp-devel
```
ag4ums: need to be change to runtime lib

src/copyright/agent/copyright.cc
GMishx: Done

src/copyright/mod_deps
GMishx: Done

GMishx:
Tested on Ubuntu 18.04 and CentOS 7

GMishx:
Compilation failed on Fedora 27 and fixed accordingly.

mcjaeger:
tested again with u1604: works

PR #1228, opened 2018-10-22
fix(core-schema): drop constraint from clearing_event and create migr…

mcjaeger:
Issue that the constrains are preventing to add candidate licenses as clearing result.

### DIFF
--- a/src/www/ui/core-schema.dat
+++ b/src/www/ui/core-schema.dat
@@ -1886,7 +1886,6 @@
   $Schema["CONSTRAINT"]["clearing_decision_event_event_fk_fkey"] = "ALTER TABLE \"clearing_decision_event\" ADD CONSTRAINT \"clearing_decision_event_event_fk_fkey\" FOREIGN KEY (\"clearing_event_fk\") REFERENCES \"clearing_event\" (\"clearing_event_pk\") ON UPDATE NO ACTION ON DELETE CASCADE;";
   $Schema["CONSTRAINT"]["clearing_decision_event_decision_fk_fkey"] = "ALTER TABLE \"clearing_decision_event\" ADD CONSTRAINT \"clearing_decision_event_decision_fk_fkey\" FOREIGN KEY (\"clearing_decision_fk\") REFERENCES \"clearing_decision\" (\"clearing_decision_pk\") ON UPDATE NO ACTION ON DELETE CASCADE;";
   $Schema["CONSTRAINT"]["clearing_event_pkey"] = "ALTER TABLE \"clearing_event\" ADD CONSTRAINT \"clearing_event_pkey\" PRIMARY KEY (\"clearing_event_pk\");";
-  $Schema["CONSTRAINT"]["clearing_event_rf_fk_fkey"] = "ALTER TABLE \"clearing_event\" ADD CONSTRAINT \"clearing_event_rf_fk_fkey\" FOREIGN KEY (\"rf_fk\") REFERENCES \"license_ref\" (\"rf_pk\") ON UPDATE NO ACTION ON DELETE CASCADE;";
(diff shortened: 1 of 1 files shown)


mcjaeger:
first test is good: vagrant based deloyment does not show the error anymore. both for selecting as user decision and bulk scan, candidate licenses can be used again.

install/db/dbmigrate_drop_table_constraints.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+/**
+ * @brief Drop errored constraints.
+ *
+ * @param DbManager $dbManager
+ * @param string $tableName
+ * @param string $constraintName
+ */
+function dropConstraints($dbManager, $tableName, $constraintName)
```
maxhbr: I am confused, that `dropConstraints` is not called in `fossinit.php`

install/db/dbmigrate_drop_table_constraints.php
```diff
@@ -0,0 +1,47 @@
...
+  $sql = "ALTER TABLE $tableName DROP CONSTRAINT IF EXISTS $constraintName";
+  $dbManager->getSingleRow($sql,array(),"dropConstraints".$constraintName);
+}
+
+$listOfConstraints = [
+  ["clearing_event", "clearing_event_rf_fk_fkey"]
+];
+
+foreach ($listOfConstraints as $listConstraintsRow) {
+  $result = $dbManager->getSingleRow("SELECT count(*) FROM information_schema.table_constraints WHERE constraint_name = $1;", array($listConstraintsRow[1]), 'checkconstraintsIfExists');
```
maxhbr: Maybe use the function `DB_ConstraintExists('clearing_event_rf_fk_fkey')` ( I am not actually sure that this function exists, but I think I have seen it somewhere )

install/db/dbmigrate_drop_table_constraints.php
```diff
@@ -0,0 +1,47 @@
...
+ * @param string $constraintName
+ */
+function dropConstraints($dbManager, $tableName, $constraintName)
+{
+  $sql = "ALTER TABLE $tableName DROP CONSTRAINT IF EXISTS $constraintName";
+  $dbManager->getSingleRow($sql,array(),"dropConstraints".$constraintName);
+}
+
+$listOfConstraints = [
+  ["clearing_event", "clearing_event_rf_fk_fkey"]
```
maxhbr: also drop `license_file_rf_fk_fkey` in `license_file`? (has to be verified if it is also wrong)

install/db/dbmigrate_drop_table_constraints.php
```diff
@@ -0,0 +1,47 @@
...
+];
+
+foreach ($listOfConstraints as $listConstraintsRow) {
+  $result = $dbManager->getSingleRow("SELECT count(*) FROM information_schema.table_constraints WHERE constraint_name = $1;", array($listConstraintsRow[1]), 'checkconstraintsIfExists');
+  if($result['count'] > 0) {
```
maxhbr: This will always drop the constraint `clearing_event_rf_fk_fkey` even if it would be corrected to also include candidate licenses. Maybe also check for database version and apply if version is `3.4.0` and increase afterwards.

GMishx:
Do we really need a migration script? As the constraints which are not mentioned in the `core-schema.dat` will be dropped by this function call:

https://github.com/fossology/fossology/blob/master/src/lib/php/libschema.php#L147

https://github.com/fossology/fossology/blob/master/src/lib/php/libschema.php#L337

shaheemazmalmmd:
Yes you are right, Let me test. Will remove the migration scripts if not required.

GMishx:
Can you please check with new changes @maxhbr ?

maxhbr:
I am still not sure about the following:

What happened with an instance, where data was in the database invalidating the constraint and was updated?
- Did the script fail? 
- Did the script drop data?
- has postgres another way of handling this?

GMishx:
The [migration script](https://github.com/fossology/fossology/blob/master/install/db/dbmigrate_3.3-3.4.php) will drop invalid data before applying the new constraints.

maxhbr:
> The migration script will drop invalid data before applying the new constraints.

That is bad, we should recheck every of the new constraints, and merge this ASAP. (ping @mcjaeger )

PR #1232, opened 2018-10-30
test(postgres): Run test cases in PostgreSQL 10

GMishx:
## Description

Included new test cases for PHP 7.0 and 7.2 with PostgreSQL 10.

### Changes

Replaced `createlang` with `CREATE LANGUAGE` query as `createlang` is deprecated in PostgreSQL 9.3 and removed in PostgreSQL 10.

## How to test

Check Travis and with PostgreSQL 10.


ag4ums:
tested with bionic and xenial need to test with trusty and debian 8/9

PR #1246, opened 2018-11-20
docs(libphp): Added doxygen comments for PHP library

GMishx:
## Description

Added doxygen comments for PHP library.

## How to test

Deployed here: https://gmishx.github.io/

### DIFF
--- a/fossology_doxygen.conf
+++ b/fossology_doxygen.conf
@@ -55,7 +55,7 @@ PROJECT_NAME           = "FOSSology"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 3.4.0-rc2
+PROJECT_NUMBER         =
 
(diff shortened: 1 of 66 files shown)


fossology_doxygen.conf
```diff
@@ -55,7 +55,7 @@ PROJECT_NAME           = "FOSSology"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         =
+PROJECT_NUMBER         = 3.4.0-rc2
```
mcjaeger: 3.4.0? or maybe 3.5.0-rc1?

GMishx: I think it will be better not to hard code the version as it has to be changed on every release and travis anyways [gets it from git](https://github.com/fossology/fossology/blob/master/utils/deploy-pages.sh#L36)

PR #1257, opened 2018-12-20
feat(licenseRef): check flag before updating the license text

shaheemazmalmmd:
## Description

Do not update license text if it is already updated from UI.

### Changes

Add flag to license_ref table and check for different cases.

## How to test

* On a clean fossology instance change license text from ui and execute the postinstall scripts.
* Check if the updated license text still exist.


src/www/ui/core-schema.dat
```diff
@@ -944,12 +944,15 @@
   $Schema["TABLE"]["license_ref"]["rf_spdx_compatible"]["ADD"] = "ALTER TABLE \"license_ref\" ADD COLUMN \"rf_spdx_compatible\" bool DEFAULT false";
   $Schema["TABLE"]["license_ref"]["rf_spdx_compatible"]["ALTER"] = "ALTER TABLE \"license_ref\" ALTER COLUMN \"rf_spdx_compatible\" SET NOT NULL, ALTER COLUMN \"rf_spdx_compatible\" SET DEFAULT false";
 
+  $Schema["TABLE"]["license_ref"]["rf_flag"]["DESC"] = "COMMENT ON COLUMN \"license_ref\".\"rf_flag\" IS 'default will be 1 if it gets updated from UI it will be 2 and from source 3'";
```
GMishx: As per the code, nowhere the `rf_flag` is set to `3`. Can you please check?

shaheemazmalmmd: The flag should be set to 3 when you add a new license text from source.

GMishx:
The flag value never seems to be set to `3`.

Steps followed:

1. Deleted a license from existing FOSSology database and installing the branch again.
    - The license is added back but with `rf_flag` = 1.
2. Updated license text from UI then change the license text in `install/db/licenseref.sql` and added a new column `rf_flag` with value as `3` to the insert statement of the license.
    - The license text is replace with the license text in SQL file but the `rf_flag` is still 2.

Suggestions:
1. I think, the statement `$sql .= " rf_flag=3,"` should be added here: https://github.com/fossology/fossology/pull/1257/files#diff-a4ac1a70eb536e248a45185452d44cc2R455
2. The 3 line change https://github.com/fossology/fossology/pull/1257/files#diff-a4ac1a70eb536e248a45185452d44cc2R483 seems like it does nothing as the file is never read again after the replacement of the text.

PR #1258, opened 2018-12-20
fix(delagent): delete existing clearing events using delagent 

### LINKED ISSUE #1057: Delagent refer to tables which don't exist
Delagent is referring to 3 tables which does not exist any more:

>  licterm_name, agent_lic_status, agent_lic_meta

Check here: https://github.com/fossology/fossology/blob/master/src/delagent/agent/util.c#L281

shaheemazmalmmd:
closes #1057

### DIFF
--- a/src/delagent/agent/delagent.c
+++ b/src/delagent/agent/delagent.c
@@ -1,8 +1,6 @@
 /********************************************************
- delagent: Remove an upload from the DB and repository
-
  Copyright (C) 2007-2013 Hewlett-Packard Development Company, L.P.
- Copyright (C) 2015-2017 Siemens AG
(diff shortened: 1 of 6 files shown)


GMishx:
During testing, while I am trying to delete a folder from Organize >> Folders >> Delete Folder , I get an empty page. While checking the Apache error log, I got the following message:

`[Fri Feb 01 16:27:49.318259 2019] [php7:error] [pid 20480] [client 127.0.0.1:51774] PHP Recoverable fatal error:  Object of class Fossology\\Lib\\Db\\ModernDbManager could not be converted to string in /usr/local/share/fossology/lib/php/Db/Driver/Postgres.php on line 72, referer: http://localhost/repo/?mod=admin_folder_delete`

PR #1278, opened 2019-01-29
feat(maintagent): add feature to delete orphaned records from database

shaheemazmalmmd:
## Description

Functionality to delete all the orphaned records from fossology database

### Changes

Removed Orphan records from uploadtree, clearing_event, clearing_decision, clearing_decision_event, obligation_map, obligation_candidate_maps table.

## How to test

* Upload a package Do clearing for all the files.
* Login to fossology database and delete the uploaded package from upload table.
* Login to fossology UI and add a candidate license and normal license.
* Add this licenses to obligations and delete the added licenses.
* Run maintagent and check if the all the orphaned records deleted from the above tables

### DIFF
--- a/src/maintagent/agent/process.c
+++ b/src/maintagent/agent/process.c
@@ -323,9 +323,10 @@ FUNCTION void NormalizeUploadPriorities()
   fo_scheduler_heart(1);  // Tell the scheduler that we are alive and update item count
   return;  // success
 }
+
 /**
  * @brief reindex of all indexes in fossology database
-  * @returns void but writes status to stdout
+ * @returns void but writes status to stdout
  */
 FUNCTION void reIndexAllTables()
 {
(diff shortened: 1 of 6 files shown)


src/maintagent/agent/process.c
```diff
@@ -351,3 +352,95 @@ FUNCTION void reIndexAllTables()
...
+  char *clearingDecisionEventOrphaned;
+  char *clearingEventOrphaned;
+  char *obligationMapOrphaned;
+  char *obligationCandidateMapOrphaned;
+  long startTime, endTime;
+
+  char *sql1 = " DELETE FROM uploadtree "
+                 " WHERE uploadtree_pk IN (SELECT DISTINCT(uploadtree_pk) "
+               " FROM uploadtree UT LEFT OUTER JOIN upload U ON U.upload_pk = UT.upload_fk "
+                 " WHERE U.upload_pk IS NULL); ";
```
GMishx: This query can be further optimized like this:

```sql
DELETE FROM uploadtree AS UT
WHERE NOT EXISTS (
  SELECT 1
  FROM upload AS U
  WHERE UT.upload_fk = U.upload_pk
);
```

And same goes for every query below.

I got a Execution time boost (calculated using `EXPLAIN ANALYZE`) from 211,327ms to 32,318ms.

src/maintagent/agent/process.c
```diff
@@ -24,81 +24,102 @@
 #include "maintagent.h"
 
 /**********  Globals  *************/
-extern PGconn    *pgConn;        ///< database connection
+PGconn* pgConn = NULL;        // the connection to Database
```
GMishx: Can you please retain the `///<` format for comment string? It is used by Doxygen to create documentation comments for global variables.

src/maintagent/agent/process.c
```diff
@@ -24,81 +24,102 @@
...
 
+/**
+ * \brief simple wrapper which includes PQexec and fo_checkPQcommand
+ * \param desc description for the SQL command, else NULL
```
GMishx: Please check the parameter name.

src/maintagent/agent/process.c
```diff
@@ -24,81 +24,102 @@
...
+  result = PQexec(pgConn, SQL);
+  if (fo_checkPQcommand(pgConn, result, SQL, file, line)) {
+    exitNow(exitNumber);
+  }
+  return result;
+}
+
+/**
+ * \brief Execute SQL query and create the result
+ * \and clear the result.
```
GMishx: Please remove the `\` in front of and.

src/maintagent/agent/process.c
```diff
@@ -24,81 +24,102 @@
...
+ * \param file source file name
+ * \param line source line number
+ * \return PQexec query result
+ */
+PGresult * PQexecCheck(int exitNumber, char *SQL, char *file, const int line)
+{
+  PGresult *result;
+  result = PQexec(pgConn, SQL);
+  if (fo_checkPQcommand(pgConn, result, SQL, file, line)) {
+    exitNow(exitNumber);
```
GMishx: Wouldn't it be better to clear the result before calling `exitNow()`? This can lead to memory leak.

src/maintagent/agent/usage.c
```diff
@@ -42,6 +42,7 @@ FUNCTION void Usage(char *Name)
   printf("  -T   :: Remove orphaned temp tables.\n");
   printf("  -U   :: Process expired uploads (slow).\n");
   printf("  -Z   :: Remove orphaned files from the repository (slow).\n");
+  printf("  -E   :: Remove orphaned rows from database.\n");
```
GMishx: Can we have a `(slow)` notice like other options? Because this flag executes several large queries.

shaheemazmalmmd: I'm still wondering that it should be kept in the slow list or not. because after we fixed delagent it will not find orphaned files to delete 🤔

src/maintagent/agent/process.c
shaheemazmalmmd: it will only enter the if statement it the rows are empty.

PR #1290, opened 2019-02-27
feat(rest): Use bearer auth instead of basic

### LINKED ISSUE #1289: REST API should not use basic authentication
### Description

The FOSSology REST API currently uses only basic authentication which is not secure as the headers contains plain passwords.

API should use some other authentication method like `Bearer` to prevent transmitting passwords in plain text.

### Versions

* Last commit id on master: 0a1cc18cfe6db40f3645d7337fc0bc8a4131002c

GMishx:
## Description

Use JWT based bearer authentication for REST API instead of insecure basic authentication.

### Changes

1. Deprecated `/auth` resource to remove basic authentication.
1. Added new resource `/tokens` to allow users to create new bearer tokens.
1. Use JWT for bearer authentication.
1. Max life of token can be modified from `sysconfig` table (Admin > Customize) (defaults to 30 days).
1. Allow user to create tokens with "read only" and "read/write" access (referred as scope of a token).
1. Read tokens can be used only on **GET** methods. Other methods require write access.

## How JWT will work

1. Whenever a user creates a request to get a new token, a new JWT token is generated in following manner:
    1. Create a payload with `iss`, `aud`, `exp`, `nbf`, `jti` and `scope`.
    1. `jti` is calculated as token's pk from DB and user's pk from DB joined with a `.` and base64 encoded.
    1. Create a random and secure secret key (40 character length) using PHP's [`openssl_random_pseudo_bytes()`](https://secure.php.net/manual/en/function.openssl-random-pseudo-bytes.php) and store it in the DB against the user pk.
    1. Calculate the JWT's digest using `HS256` with this secret key.
1. Whenever a user requests a resource, the JWT is validated using the secret key stored in DB.
    - This method will give us surety that the token was actually generated by FOSSology and is not modified since.
    - The user info is derived from the `jti` payload of the JWT.
1. User also has option to manage the JWT from the UI.
    1. User can create a new token from Admin >> Users >> Edit User Account.
    1. User can revoke a token from the same page.
    1. User can also retrieve their old (non expired) tokens from the same page.

## How to test

Do normal REST testing using the JWT tokens.

Closes #1289 

TODO:
- [x] Create a job in maintenance agent to clean inactive tokens from DB.
    Created issue to work on (#1303)
- [x] Improve function names.
- [x] Move JS from `user-edit` twig to a separate js file.
 - [ ] Update Wiki

### DIFF
--- a/src/www/ui/api/Helper/RestHelper.php
+++ b/src/www/ui/api/Helper/RestHelper.php
@@ -29,6 +29,7 @@ use Fossology\Lib\Dao\FolderDao;
 use Fossology\Lib\Dao\UserDao;
 use Fossology\UI\Api\Models\Info;
 use Fossology\UI\Api\Models\InfoType;
+use Fossology\Lib\Plugin\Plugin;
 
 /**
  * @class RestHelper
@@ -36,6 +37,21 @@ use Fossology\UI\Api\Models\InfoType;
  */
 class RestHelper
 {
+  /**
+   * @var array VALID_SCOPES
+   * Valid scopes for REST authentication tokens.
+   */
+  const VALID_SCOPES = ["read", "write"];
+  /**
+   * @var array SCOPE_DB_MAP
+   * Maps a user readable scope to DB value.
+   */
+  const SCOPE_DB_MAP = ["read" => "r", "write" => "w"];
+  /**
+   * @var int TOKEN_KEY_LENGTH
+   * Length of the token secret key.
+   */
+  const TOKEN_KEY_LENGTH = 40;
   /**
    * @var UploadDao $uploadDao
    * Upload DAO object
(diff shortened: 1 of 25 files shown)


src/www/ui/api/Helper/RestHelper.php
```diff
@@ -196,4 +211,61 @@ public function copyUpload($uploadId, $newFolderId, $isCopy)
...
+   * @param string $tokenScope  The scope of the token requested.
+   * @return boolean|Fossology::UI::Api::Models::Info True if all parameters
+   *         are ok, else an info.
+   */
+  public function validateTokenRequest($tokenExpire, $tokenName, $tokenScope)
+  {
+    $requestValid = true;
+    if (strtotime($tokenExpire) < strtotime("tomorrow") ||
+      ! preg_match("/^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/",
+        $tokenExpire) || strtotime($tokenExpire) > strtotime("+30 days")) {
```
ag4ums: configurable token expiry will be better.
may be from admin -> customize section

mcjaeger:
how about putting the maint agent task into a separate issue?

GMishx:
> how about putting the maint agent task into a separate issue?

Created separate issue #1303 to work on later

mcjaeger:
OK then we would just need review + test

ag4ums:
@GMishx please rebase the branch with master

PR #1293, opened 2019-03-05
feat(licenseref): convert licenseref from sql to json format

### LINKED ISSUE #1286: Move license text from license.ref to JSON (or similar)
### Description

Currently, adding new license texts is done by adding them into the SQL insertion code. This is difficult to modify.

Rather there should be input files for licenses and then, other tools can parse easier or reference to easier.

The main point is how to care for the common ids as they are also already in installed databases.

There could be two uses:

* other projects can use this file more easily than parsing licenseref.sql
* maybe adding new texts to the fossology project / agents will be easier

In the repo there is a file (or mutiple files) for the collection of license texts.

#### Screenshots

n.a. but the current file is here: 

https://github.com/fossology/fossology/blob/master/install/db/licenseref.sql

shaheemazmalmmd:
## Description

Change licenseref.sql to licenseref.json.

### Changes

1. parse JSON file and insert into database

## How to test

1. For a existing fossology instance pull `contrib/featLicenseRefSQLtoJSON` and run the `postinstall` scripts.
2. Check database for all the entries.
3. Upload a file and run monk for sample data.
4. Delete database and run a fresh install.
5. Check database for all the entries.

Closes #1286


GMishx:
After performing some testing, the fix provided by #1257 is not working any more.

maxhbr:
This PR still needs a new code review / test due to force push

install/fossinit.php
```diff
@@ -449,10 +474,11 @@ function initLicenseRefTable($Verbose)
       $rf_flag_check = $row_check['rf_flag'];
 
       $sql = "UPDATE license_ref set ";
-      if($rf_flag_check == 1 ||  $rf_flag == 1) {
-        if ($rf_text_check != $rf_text && !empty($rf_text) && !(stristr($rf_text, 'License by Nomos')))  $sql .= " rf_text='$rf_text', rf_flag='1',";
-      } else {
+      if( ($rf_flag_check == 1 &&  $rf_flag == 1) ||
```
GMishx: If the `$rf_flag_check = 1` and `$rf_flag = 1`, then the license is text is not updated manually by the user. So in this case the `rf_text` should be updated.

I think there should be **only one** condition in the `if` to check `if($rf_flag_check == 2 &&  $rf_flag == 1)` (license text was updated by UI).

shaheemazmalmmd:
@GMishx : done with the requested changes. please check

PR #1305, opened 2019-03-14
fix(scheduler): Respect max run limit of agents

GMishx:
## Description

Currently, the max limit set by agents in their **<agent>.conf** was read and set by scheduler but never checked before scheduling the agent.

This PR checks the same before scheduling the new job/agent.

### Changes

1. Added one new parameter (`run_count`) to the `meta_agent_t` structure to keep the count of running instances of the agent.
1. This count is increased when the agent is scheduled/resumed and decreased when the agent is killed/paused/finished.
1. Check this `run_count` against the agent's `max_run` before scheduling the job.

## How to test

1. Reduce the **max** of an agent by editing it's **.conf**
1. Stop the scheduler and schedule the agent from the UI (this is to give buffer for scheduling enough jobs before scheduler picks them up).
1. Start the scheduler and check the running processes.

### DIFF
--- a/src/scheduler/agent/agent.c
+++ b/src/scheduler/agent/agent.c
@@ -778,6 +778,7 @@ meta_agent_t* meta_agent_init(char* name, char* cmd, int max, int spc)
   strcpy(ma->raw_cmd, cmd);
   strcat(ma->raw_cmd, " --scheduler_start");
   ma->max_run = max;
+  ma->run_count = 0;
   ma->special = spc;
   ma->version = NULL;
   ma->valid = TRUE;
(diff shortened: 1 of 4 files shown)


GMishx:
I also tested the branch on my Ubuntu 18.04 system with 4 cores. I got the following conclusions:

I scheduled 3 monkbulk jobs after killing the scheduler:

image

Updated the max limit for monkbulk in `/usr/local/etc/fossology/mods-enabled/monkbulk/monkbulk.conf` to `1`.

And I restarted the scheduler:

First monk: 
image

Second monk: 
image

---

Maybe the monk jobs you are seeing are just the threads created by omp.

PR #1312, opened 2019-03-23
fix(copyright): fix php notices in copyright hist view

shaheemazmalmmd:
## Description

handle bad requests and remove PHP notices 

### Changes

add check for count of parameters before list them into variables

### DIFF
--- a/src/copyright/ui/ajax-copyright-hist.php
+++ b/src/copyright/ui/ajax-copyright-hist.php
@@ -100,9 +100,18 @@ class CopyrightHistogramProcessPost extends FO_Plugin
     else if($action=="update" || $action=="delete" || $action=="undo")
     {
       $id = GetParm("id", PARM_STRING);
-      list($upload, $item, $hash, $type) = explode(",", $id);
+      $getEachID = array_filter(explode(",", trim($id, ',')), function($var) {
+          return $var !== "";
+      });
+      if(count($getEachID) == 4) {
+        list($upload, $item, $hash, $type) = $getEachID;
+      } else {
+        return new Response('bad request while '.$action,
+                Response::HTTP_BAD_REQUEST,
+                array('Content-type'=>'text/plain')
+          );
+      }
     }
-
 
     /* check upload permissions */
     if (!(($action == "getData" || $action == "getDeactivatedData") &&
--- a/src/copyright/ui/template/histTable.js.twig
+++ b/src/copyright/ui/template/histTable.js.twig
@@ -113,8 +113,8 @@ function updateRows{{ table.type }}(upload, item, hash, kind, replaceText) {
         $("#delete{{ table.type }}" + hash).hide();
         updateElement.html("updating...");
       },
-      error: function() { 
-        alert('error while updating a record'); 
+      error: function(errorResponse) {
+        alert(errorResponse.responseText);
       }
   });
 }
@@ -129,8 +129,8 @@ function delete{{ table.type }}(upload, item, hash, kind) {
         $("#delete{{ table.type }}" + hash).hide();
         $("#update{{ table.type }}" + hash).show();
       },
-      error: function() { 
-        alert('error while deleting a record'); 
+      error: function(errorResponse) {
+        alert(errorResponse.responseText);
       }
   });
 }
@@ -146,6 +146,8 @@ function undo{{ table.type }}(upload, item, hash, kind) {
         $("#update{{ table.type }}" + hash).hide();
         $("#deleteBySelect{{ table.type }}" + hash).prop('checked', false);
       },
-      error: function() { alert('error'); }
+      error: function(errorResponse) {
+        alert(errorResponse.responseText);
+      }
   });
 }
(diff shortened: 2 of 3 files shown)


src/copyright/ui/ajax-copyright-hist.php
```diff
@@ -100,7 +100,15 @@ function Output()
     else if($action=="update" || $action=="delete" || $action=="undo")
     {
       $id = GetParm("id", PARM_STRING);
-      list($upload, $item, $hash, $type) = explode(",", $id);
+      $getEachID = explode(",", trim($id, ','));
```
maxhbr: Why do you trim first? `"2,3,,"` is for me a valid representation of `["2","3","",""]`?

shaheemazmalmmd: hi, All the four values are required here, explode will consider empty as array value. so i'm using trim before i explode. So that i will not get any empty value.

Please let me know if there is any better approach here.

maxhbr: assert that no element is equal to `""`. Your approach can not handle `"2,,,4"` which gets parsed to `["2","","","4"]`

maxhbr:
```suggestion
      list($upload, $item, $hash, $type) = explode(",", $id);
      if (empty($upload) || empty($item) || empty($hash) || empty($type))
      {
          return new JsonResponse(array(...)...);
      }
```
list($upload, $item, $hash, $type) = explode(",", $id);

shaheemazmalmmd: The suggested code will lead to the below PHP Notice: 

```
PHP Notice:  Undefined offset: 1 in /usr/share/fossology/copyright/ui/ajax-copyright-hist.php on line 103
PHP Notice:  Undefined offset: 2 in /usr/share/fossology/copyright/ui/ajax-copyright-hist.php on line 103,
PHP Notice:  Undefined offset: 3 in /usr/share/fossology/copyright/ui/ajax-copyright-hist.php on line 103,
```

shaheemazmalmmd: To handle this i can use `array_filter` which will omit the empty values and give me the exact count.

maxhbr: what was your input? I.e. what was the value of `$id`?

maxhbr: might be OK but then extract the technical implementation into a function `verifyThatAllFourValuesArePresent($ids)` (or something similarly descriptive).

shaheemazmalmmd: Sure will do it this way.

shaheemazmalmmd:
@maxhbr : Please check the latest code changes.

src/copyright/ui/ajax-copyright-hist.php
```diff
@@ -100,7 +100,15 @@ function Output()
...
     {
       $id = GetParm("id", PARM_STRING);
-      list($upload, $item, $hash, $type) = explode(",", $id);
+      $getEachID = array_filter(explode(",", trim($id, ',')));
+      if(count($getEachID) == 4) {
+        list($upload, $item, $hash, $type) = $getEachID;
+      } else {
+        return new JsonResponse(array(
+                 "error" => 'bad request'
+              ), JsonResponse::HTTP_INTERNAL_SERVER_ERROR
```
GMishx: Please return `HTTP_BAD_REQUEST` to match the error message.
```suggestion
              ), JsonResponse::HTTP_BAD_REQUEST
```

mcjaeger:
@maxhbr also changes look good to you too? (because your review has status `requested changes`)

src/copyright/ui/ajax-copyright-hist.php
```diff
@@ -100,10 +100,17 @@ function Output()
     else if($action=="update" || $action=="delete" || $action=="undo")
     {
       $id = GetParm("id", PARM_STRING);
-      list($upload, $item, $hash, $type) = explode(",", $id);
+      $getEachID = array_filter(explode(",", trim($id, ',')));
```
maxhbr: `array_filter` will also drop `"0"`, which casts to `false`. (e.g. `array_filter(array("", "0", "1")) == array("2" => "0")`)

shaheemazmalmmd: I think 0  is not accepted for every value in this ($upload, $item, $hash, $type).

EX: When uploadId is 0 it is an invalid input anyway.

maxhbr: please make that explicit. This will make our code base a little bit cleaner and is best practice in PHP.

mcjaeger: I was googleing it for curiosity ... but what would be good programming practice in this case?

maxhbr: The confusing behaviour can be prevented by adding an explicit callback
```suggestion
      $getEachID = array_filter(explode(",", trim($id, ',')), function($var) {
          return $var !== "";
      });
```
or maybe even something like:
```suggestion
      $getEachID = array_filter(explode(",", trim($id, ',')), function($var) {
          return ! is_numeric($var);
      });
```

PR #1314, opened 2019-03-28
fix(ui): Show license findings for folder with single child

GMishx:
## Description

Currently, if an upload contains a folder with only one file in it, the **License Browser** does not show the license findings and decisions.

### Changes

The UI currently changes the file name if such case occurs.

Now other parameters are also updated at the same time using new function `updateTheFindingsAndDecisions` (reusing the old code).

## How to test

1. Install current master branch.
1. Upload a package with a folder containing single file.
1. Check the parent folder using the **License Browser**.
    - The license findings and decisions will be missing.
1. Install the fix and browse the same upload.
    1. It should show the findings.
    1. The flat view should not be effected.
    1. The other features links ([View][Info][Download][Tag][Edit]) on the right should not be effected.

### DIFF
--- a/src/www/ui/async/AjaxExplorer.php
+++ b/src/www/ui/async/AjaxExplorer.php
@@ -80,6 +80,8 @@ class AjaxExplorer extends DefaultPlugin
     $this->clearingDao = $this->getObject('dao.clearing');
     $this->agentDao = $this->getObject('dao.agent');
     $this->clearingFilter = $this->getObject('businessrules.clearing_decision_filter');
+    $this->filesThatShouldStillBeCleared = [];
+    $this->filesToBeCleared = [];
   }
 
   /**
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
@GMishx : i have tested this branch with classic zlib package. But was not successful with corrections you have made. 

Find the screenshot below.

ScreenShotOfExample

GMishx:
The code changes works if a folder contains a file. In such cases, the parent of the file is the actual folder in the `uploadtree`.

But if the parent of the file is a PDF (as in your test case), the parent of the file becomes `artifact.dir` (since PDFs are extracted by ununpack) and thus the code fails as there is no pfile entry for `artifact.dir`.

So, the parent of `package.qpg` is `qnx` folder which exists in the `pfile` and `uploadtree` table. But the parent of `zlib.3` is `artifact.dir` which in turn has parent `zlib.3.pdf` and `artifact.dir` exists only in `uploadtree` and not in `pfile` table.

I will do some change to check if the parent is `artifact.dir` then get the next parent. It should fix the case.

PR #1319, opened 2019-03-31
fix(apache): Add rewrite string to apache conf

GMishx:
## Description

The rewrite string for Debian packaging's apache conf was missing for the REST API.

### Changes

Added the missing rewrite rule for FOSSology's apache conf file.

## How to test

1. Build and install the Debian package.
1. Check if the REST API can be used.


mcjaeger:
interestingsly, it has been applied at
https://github.com/GMishx/fossology/blob/4a3c7cb0125fb9c2765d7efdda9bf4cb7c7ff5e2/Vagrantfile#L55

PR #1316, opened 2019-03-28
feat(unifiedReport): seperate user and scanner findings of copyrights

shaheemazmalmmd:
## Description

* Separated tables for copyright in to two for scanner findings and user findings.
* Update phpword from v0.13.* to v0.16.*
* fix warnings that appears everytime in unified report job log. 

### Changes

Fix warnings, and separate copyright tables.  

## How to test

* Upload a package and do some clearing of license, copyright and ecc.
* Generate unified report, readmeoss and other report formats check for desired output.
* Check the job log of unified report.

## Log fixed
```2019-03-28 15:40:13 unifiedreport [0] :: JOB[185].unifiedreport[22603.localhost]: "PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /usr/local/share/fossology/unifiedreport/agent/unifiedreport.php on line 274"
2019-03-28 15:40:13 unifiedreport [0] :: JOB[185].unifiedreport[22603.localhost]: "PHP Stack trace:"
2019-03-28 15:40:13 unifiedreport [0] :: JOB[185].unifiedreport[22603.localhost]: "PHP   1. {main}() /usr/local/lib/fossology/fo_wrapper:0"
2019-03-28 15:40:13 unifiedreport [0] :: JOB[185].unifiedreport[22603.localhost]: "PHP   2. require() /usr/local/lib/fossology/fo_wrapper:77"
2019-03-28 15:40:13 unifiedreport [0] :: JOB[185].unifiedreport[22603.localhost]: "PHP   3. UnifiedReport->run_scheduler_event_loop() /usr/local/share/fossology/unifiedreport/agent/unifiedreport.php:846"
2019-03-28 15:40:13 unifiedreport [0] :: JOB[185].unifiedreport[22603.localhost]: "PHP   4. UnifiedReport->processUploadId() /usr/local/share/fossology/lib/php/Agent/Agent.php:341"
2019-03-28 15:40:13 unifiedreport [0] :: JOB[185].unifiedreport[22603.localhost]: "PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /usr/local/share/fossology/unifiedreport/agent/unifiedreport.php on line 277"
2019-03-28 15:40:13 unifiedreport [0] :: JOB[185].unifiedreport[22603.localhost]: "PHP Stack trace:"
[...]

PR #1325, opened 2019-04-04
feat(ci): Run travis jobs on Xenial

GMishx:
## Description

Since Trusty Tahr is reaching its EOL, it is advisable to upgrade systems to Xenial Xerus.

The current builds of FOSSology on Travis were on Trusty, so this PR changes the image to Xenial.

### Changes

1. Change the `dist` from `trusty` to `xenial`.
1. Remove the old GCC version from tests.
1. Added new GCC versions to tests.
1. Remove PHP 5.6 tests as PHP 7.0 is the default version shipped with Xenial.

## How to test

Check the travis logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
-# Copyright Siemens AG, 2014-2018
-# SPDX-License-Identifier:	GPL-2.0 LGPL-2.1
+# Copyright Siemens AG, 2014-2019
+# SPDX-License-Identifier: GPL-2.0 LGPL-2.1
 
 # build FOSSology on Travis CI - https://travis-ci.org/
 
 language: php
-dist: trusty
+dist: xenial
 php: 7.0
 cache:
   ccache: true
@@ -22,6 +22,7 @@ addons:
       - genisoimage
       - libboost-program-options-dev
       - libboost-regex-dev
+      - libglib2.0-dev
       - libcppunit-dev
       - libcunit1-dev
       - libdbd-sqlite3-perl
@@ -32,8 +33,11 @@ addons:
       - librpm-dev
       - libspreadsheet-writeexcel-perl
       - libtext-template-perl
-      - php5-cli
-      - php5-pgsql
+      - php-cli
+      - php-pgsql
+      - php-zip
+      - php-xml
+      - php-mbstring
       - poppler-utils
       - p7zip
       - p7zip-full
@@ -73,15 +77,15 @@ jobs:
         - src/testing/docker/test-standalone.sh
 #### C/C++ agent tests ###########################
     - &compiler-tests
-      env: CC=gcc-4.8 CXX=g++-4.8 CFLAGS='-Wall'
+      env: CC=gcc-5 CXX=g++-5 CFLAGS='-Wall'
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
           packages:
             - *default_packages
-            - gcc-4.8
-            - g++-4.8
+            - gcc-5
+            - g++-5
       install:
         - composer install --prefer-dist --working-dir=src
         - ./install/scripts/install-spdx-tools.sh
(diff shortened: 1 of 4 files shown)


mcjaeger:
wow, did not care so much that gcc 4.8 was release six years ago ...: https://gcc.gnu.org/releases.html

.travis.yml
```diff
@@ -127,21 +128,18 @@ jobs:
         # Use default port
         - sudo sed -i 's/port = 5433/port = 5432/' /etc/postgresql/10/main/postgresql.conf
         # Use 9.6 auth config:
-        - sudo cp /etc/postgresql/{9.6,10}/main/pg_hba.conf
+        - sudo cp /etc/postgresql/{9.5,10}/main/pg_hba.conf
         - sudo service postgresql restart
         - *default-before-script
 #### PHPUnit tests ###########################
     - &phpunit-tests
-      php: 5.6
```
maxhbr: We still support debian jessie with PHP 5.6, so we should not drop the test here

GMishx: Missed that point. Thanks for pointing it out.

mcjaeger:
@maxhbr are your requested changes OK? (because it is in there: https://github.com/fossology/fossology/blob/840ba9b8d740ebcaf3827541aa4c9edb9e0e154f/.travis.yml#L152)

.travis.yml
```diff
@@ -1,10 +1,10 @@
-# Copyright Siemens AG, 2014-2018
-# SPDX-License-Identifier:	GPL-2.0 LGPL-2.1
+# Copyright Siemens AG, 2014-2019
+# SPDX-License-Identifier: GPL-2.0 LGPL-2.1
```
maxhbr: This is not a valid SPDX identifier expression. One has to add `AND` or `OR`
```suggestion
# SPDX-License-Identifier: GPL-2.0 OR LGPL-2.1
```

maxhbr:
@mcjaeger : yes, I do no longer request changes. The old PHP version is now again in the travisfile.

.travis.yml
mcjaeger: right, but it appears being an older issue, which could be excluded from this PR ...

mcjaeger:
I am wondering about 
```
55.19s$ phpdbg -qrr src/vendor/bin/phpunit -csrc/phpunit.xml --testsuite="Fossology PhpUnit Test Suite" --colors=always | grep -v 'script>\|c.log'
[PHP Warning:  Declaration of Fossology\Lib\Report\WeirdCharClearedGetter::getCleared($uploadId, $groupId = NULL, $extended = true, $agentcall = NULL) should be compatible with Fossology\Lib\Report\ClearedGetterCommon::getCleared($uploadId, $groupId = NULL, $extended = true, $agentcall = NULL, $isUnifiedReport = false) in /home/travis/build/fossology/fossology/src/lib/php/Report/tests/ClearedGetterCommonTest.php on line 45]
[PHP Stack trace:]
[PHP   1. {main}() /home/travis/build/fossology/fossology/src/vendor/phpunit/phpunit/phpunit:0]
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.
```
in all the php test build targets, such as https://travis-ci.org/fossology/fossology/jobs/525869379

GMishx:
> I am wondering about
> 
> ```
> 55.19s$ phpdbg -qrr src/vendor/bin/phpunit -csrc/phpunit.xml --testsuite="Fossology PhpUnit Test Suite" --colors=always | grep -v 'script>\|c.log'
> [PHP Warning:  Declaration of Fossology\Lib\Report\WeirdCharClearedGetter::getCleared($uploadId, $groupId = NULL, $extended = true, $agentcall = NULL) should be compatible with Fossology\Lib\Report\ClearedGetterCommon::getCleared($uploadId, $groupId = NULL, $extended = true, $agentcall = NULL, $isUnifiedReport = false) in /home/travis/build/fossology/fossology/src/lib/php/Report/tests/ClearedGetterCommonTest.php on line 45]
> [PHP Stack trace:]
> [PHP   1. {main}() /home/travis/build/fossology/fossology/src/vendor/phpunit/phpunit/phpunit:0]
> PHPUnit 6.5.14 by Sebastian Bergmann and contributors.
> ```
> 
> in all the php test build targets, such as https://travis-ci.org/fossology/fossology/jobs/525869379

It was introduced by #1316 and fixed in #1338.

Issue #1306, opened 2019-03-18
Values of two fields in the Obligations and Risks Administration view always set to "default"


mcjaeger:
### Description

when modifying an obligation in the Obligations and Risks Administration view of FOSSology the values for

"Type of obligation" (Opbigation, restriction ....)

and the values for:

"Optional: level of attention this obligation should raise in the clearing process" (green, white, yellow)

seem to be not stored in the DB or the display is not correct.

#### How to reproduce

Describe the bug and list the steps you used when the issue occurred.
When I have e.g. changed the value of "Type of obligation" to "restriction" then click on update and open it after a while again (the same record) then the selection is again "obligation" The same pplies for "Optional: level of attention this obligation should raise in the clearing process"
#### Screenshots

![image](https://user-images.githubusercontent.com/3418977/54528368-17c46f80-497d-11e9-99ff-cc59a113eb69.png)

### Versions

since some months still in master

sandipbhuyan:
I want to pick up this issue

GMishx:
Fixed by #1329

PR #1329, opened 2019-04-09
fix(obligations): fix UI and connected db to dropdown menu

### LINKED ISSUE #1300: Dropdowns in obligation admin UI are to small to display actual text
### Description

In the "Add Obligation"-admin UI are the drop-down elements to small to actually display the current state. They just show `...`.

### Versions

* Last commit id on master: https://github.com/fossology/fossology/commit/69629cc1f45f2cbde7e1318dba8cf96a3ac3de8a
* Operating System: Linux
* Browser: Chromium (works in Firefox)

hastagAB:
Fix #1300  & #1306 

# Description :
The smaller size of dropdown UI (Showing Three-dots earlier) was fixed. 
The database was not connected to the Dropdown menu so it was showing default selection while editing. [fixed]

# Example Screenshot : 
new1

### DIFF
--- a/src/www/ui/admin-obligation-file.php
+++ b/src/www/ui/admin-obligation-file.php
@@ -337,6 +337,9 @@ class admin_obligation_file extends FO_Plugin
     $vars['obligationClassification'] = array("green"=>"green", "white"=>"white", "yellow"=>"yellow", "red"=>"red");
     $vars['obligationTypes'] = array("Obligation"=>"Obligation", "Restriction"=>"Restriction", "Risk"=>"Risk", "Right"=>"Right");
 
+    $vars['ob_type'] = $row['ob_type'];
+    $vars['ob_classification'] = $row['ob_classification'];
+
     // build scripts
     $vars['licenseSelectorName'] = 'licenseSelector[]';
     $vars['licenseSelectorId'] = 'licenseSelectorId';
--- a/src/www/ui/template/admin_obligation-upload_form.html.twig
+++ b/src/www/ui/template/admin_obligation-upload_form.html.twig
@@ -25,7 +25,7 @@
         {% if isReadOnly %}
           ({{ 'read only'|trans }}){% endif %}
       </td>
-      <td align="left">{{ macro.selectsingle('ob_type', obligationTypes, 'ob_type', 'Obligation', 'style="min-width:120px;"', 4) }}
+      <td align="left">{{ macro.selectsingle('ob_type', obligationTypes, 'ob_type', ob_type, 'style="min-width:120px;"', 4) }}
       </td>
     </tr>
     <tr>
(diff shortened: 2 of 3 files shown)


GMishx:
Thanks for the contribution. Please follow the [Contribution guidelines](https://github.com/fossology/fossology/blob/master/CONTRIBUTING.md) while creating a commit.

mcjaeger:
right, the commit message should be like `fix(obligations): fix UI and connected db to dropdown menu` ... it is just the conventional changelog

hastagAB:
Sorry for the Delay! I was in the middle of my exams.
Fixed the commit message according to the contribution guidelines.
Hope this is fine now.

src/www/ui/admin-obligation-file.php
```diff
@@ -262,6 +262,8 @@ function ObligationTopic($topic)
       $ob .= "<td align=center>$associatedLicenses</td>";
       $ob .= "<td align=center>$candidateLicenses</td>";
       $ob .= "</tr>";
+      $ob_type = $row['ob_type'];
+      $ob_classification = $row['ob_classification'];
```
GMishx: Can you please describe the use of these two new variables? They seems to be unused.

hastagAB: The variables are used in src/www/ui/template/admin_obligation-upload_form.html.twig. Line 28 and 45.
These variables are the actual Obligation Type (ob_type) and Obligation Classification (ob_classification) related to that particular obligation fetched from the database

hastagAB: Earlier these two parameters were hardcoded as string values on line 28 and 45 as 'Obligation' and 'green' respectively

GMishx: The values in the `*.twig` templates goes in the form of array keys. If you want to pass some value to the twig file, you should add it to any one of the array near https://github.com/fossology/fossology/blob/master/src/www/ui/admin-obligation-file.php#L365 as it is done in https://github.com/fossology/fossology/blob/master/src/www/ui/admin-obligation-file.php#L338.

GMishx: Hello @hastagAB, are you planning to working on this?

hastagAB: The selectsingle twig function on line 28 and 45 in the `admin_obligation-upload_form.html.twig` file is expecting a selected value associated with that particular Obligation. But in the current version of the code, the function is being passed a static string value. 
In line 28 it is currently 'Obligation' so no matter what the actual value is the dropdown will always show 'Obligation'
In line 45 it is currently 'green' so no matter what the actual value is in the database the dropdown will always show 'green'

I have replaced these two static strings with the corresponding variables defined in `admin-obligation-file.php` which stores the actual value associated with the particular Obligation.

GMishx: I did understood your point, what I was trying to convey is that the twig needs input for placeholders as an index in an array before rendering the twig template. You can check the code lines I've share before.

PR #1330, opened 2019-04-09
chore(fo-installdeps): drop unsupported distros

maxhbr:
Drop distros from `fo-installdeps.sh` which only support php <5.6. They are no longer supported and should not be listed in the script. This especially affects trusty support.

### DIFF
--- a/src/copyright/mod_deps
+++ b/src/copyright/mod_deps
@@ -75,7 +75,7 @@ CODENAME=$(lsb_release --codename --short)
 if [[ $BUILDTIME ]]; then
   echo "*** Installing $DISTRO buildtime dependencies ***";
   case "$DISTRO" in
-    Debian|Ubuntu|LinuxMint)
+    Debian|Ubuntu)
       apt-get $YesOpt install \
         libjsoncpp-dev
       ;;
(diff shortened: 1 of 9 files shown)


maxhbr:
That might be overhasty, we still have travis tests which test PHP 5.6. Maybe we still support that

GMishx:
Travis will be fixed by #1325. But why are we dropping the support for LinuxMint?

maxhbr:
LinuxMint was not actually supported. It was just in the top level pattern for distros, but there was no actual case which would install the required dependencies.

GMishx:
Can you please clean these two files for trusty dependencies as well:
`src/nomos/mod_deps` and `src/copyright/mod_deps` ?

Also, the following files contain `LinuxMint` in them:
- `src/copyright/mod_deps`
- `src/scheduler/mod_deps`
- `src/nomos/mod_deps`
- `src/mimetype/mod_deps`
- `src/wget_agent/mod_deps`
- `src/delagent/mod_deps`
- `src/ununpack/mod_deps`
- `src/pkgagent/mod_deps`

maxhbr:
@GMishx also the child `mod_deps` scripts are now cleaned up

GMishx:
I'll do a quick test on other systems and merge this PR.

PR #1311, opened 2019-03-22
feat(copyright): Show text findings in copyright-hist

GMishx:
## Description

Show the text findings done manually by the user in `copyright-hist` view in the UI.

### Changes

1. Create new functions to create a simple table (without edit capabilities) to show text findings.
1. Create new class to handle ajax call to create such tables.

## How to test

1. Browse some files for a given upload and add copyright text findings for some files.
1. Browse `mod=copyright-hist` for the same upload.
    1. You should get two tabs for **Copyrights** and **Text findings**.
    1. **Text findings** tab should show the findings added above.
1. Deactivating the text findings should work.
1. The link to goto file should take you to proper file.

At the same time, other views (URL, ECC, Keywords, etc.) should not be effected.

PR #1343, opened 2019-05-03
chore(decisions): Store SHA256 of text findings

GMishx:
## Description

This PR adds additional column to copyright/ecc/keyword decision tables to store the hash (MD5) of the text findings. This will help in reducing the duplicates in the tables and also aid in aggregating the entries for PR #1311 .

### Changes

Add new column, index and unique constraints to copyright/ecc/keyword decision tables.

## How to test

Install the branch on old instance and check if the installation goes through without causing any issue.

### DIFF
--- a/install/db/Makefile
+++ b/install/db/Makefile
@@ -22,6 +22,7 @@ install: all
 	$(INSTALL_DATA) dbmigrate_2.0-2.5-pre.php $(DESTDIR)$(LIBEXECDIR)/dbmigrate_2.0-2.5-pre.php
 	$(INSTALL_DATA) dbmigrate_2.5-2.6.php $(DESTDIR)$(LIBEXECDIR)/dbmigrate_2.5-2.6.php
 	$(INSTALL_DATA) dbmigrate_3.3-3.4.php $(DESTDIR)$(LIBEXECDIR)/dbmigrate_3.3-3.4.php
+	$(INSTALL_DATA) dbmigrate_3.5-3.6.php $(DESTDIR)$(LIBEXECDIR)/dbmigrate_3.5-3.6.php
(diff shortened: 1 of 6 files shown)


install/db/dbmigrate_3.4-3.5.php
```diff
@@ -0,0 +1,144 @@
...
+  return intval($dbManager->getSingleRow($sql, [],
+    __METHOD__ . ".$tableName")['total']);
+}
+
+/**
+ * Migration from FOSSology 3.4.0 to 3.5.0
+ * @param DbManager $dbManager
+ * @param boolean $force Set true to force run the script.
+ */
+function Migrate_34_35($dbManager, $force = false)
```
mcjaeger: naming change

install/db/dbmigrate_3.4-3.5.php
```diff
@@ -0,0 +1,144 @@
...
+  RETURNING 1
+)
+SELECT count(*) AS total FROM updates;
+";
+  return intval($dbManager->getSingleRow($sql, [],
+    __METHOD__ . ".$tableName")['total']);
+}
+
+/**
+ * Migration from FOSSology 3.4.0 to 3.5.0
```
mcjaeger: ... and comment change

install/db/dbmigrate_3.4-3.5.php
GMishx: Done

install/db/dbmigrate_3.4-3.5.php
GMishx: Done

mcjaeger:
(from discussion) to avoid discussion in future, just not use the md5 ... and maybe put the Ui part also in this PR

GMishx:
> (from discussion) to avoid discussion in future, just not use the md5 ... and maybe put the Ui part also in this PR

I have pushed the new changes with hash algorithm as SHA256 and code handling the UI part.

PR #1350, opened 2019-05-14
feat(license): Provide predefined license comments

GMishx:
## Description

This feature will allow admins to add some predefined license comments which can be used by users during clearing with the help of a simple drop down in License Comment window.

Admins can navigate to **Admin >> License Admin >> Standard Comments**.

~Currently admins can add **only 7** comments at max to keep the interface simple.~

### Changes

1. Add new table to hold predefined license comments.
1. Allow users to select these comments with the help of a drop down menu.

## How to test

1. Insert some comments from the **Admin Standard License Comments** page.
1. Clear some files and add the comments.
    1. Try adding comments from the decision table.
    1. Try adding comments from the buik table.
1. Add the comments added from the first step should be visible.
1. Added comments should appear in the reports.

**P.S.:** Whenever a change is done in the comments, the user who changed the comment and timestamp is stored as well. It will help in auditing but is never shown in the UI or returned by any API.

### DIFF
--- a/src/www/ui/core-schema.dat
+++ b/src/www/ui/core-schema.dat
@@ -1015,6 +1015,31 @@
   $Schema["TABLE"]["license_set_bulk"]["acknowledgement"]["ALTER"] = "ALTER TABLE \"license_set_bulk\" ALTER COLUMN \"acknowledgement\" DROP NOT NULL";
 
 
+  $Schema["TABLE"]["license_std_comment"]["lsc_pk"]["DESC"] = "COMMENT ON COLUMN \"license_std_comment\".\"lsc_pk\" IS 'Comment id'";
+  $Schema["TABLE"]["license_std_comment"]["lsc_pk"]["ADD"] = "ALTER TABLE \"license_std_comment\" ADD COLUMN \"lsc_pk\" integer NOT NULL DEFAULT nextval('license_std_comment_lsc_pk_seq'::regclass)";
+  $Schema["TABLE"]["license_std_comment"]["lsc_pk"]["ALTER"] = "ALTER TABLE \"license_std_comment\" ALTER COLUMN \"lsc_pk\" SET NOT NULL, ALTER COLUMN \"lsc_pk\" SET DEFAULT nextval('license_std_comment_lsc_pk_seq'::regclass)";
+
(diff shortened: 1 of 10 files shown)


src/www/ui/core-schema.dat
```diff
@@ -1986,6 +2014,7 @@
   $Schema["CONSTRAINT"]["license_file_pfile_fk_fkey"] = "ALTER TABLE \"license_file\" ADD CONSTRAINT \"license_file_pfile_fk_fkey\" FOREIGN KEY (\"pfile_fk\") REFERENCES \"pfile\" (\"pfile_pk\") ON UPDATE NO ACTION ON DELETE CASCADE;";
   $Schema["CONSTRAINT"]["license_map_pkpk"] = "ALTER TABLE \"license_map\" ADD CONSTRAINT \"license_map_pkpk\" PRIMARY KEY (\"license_map_pk\");";
   $Schema["CONSTRAINT"]["license_map_rf_fkfk"] = "ALTER TABLE \"license_map\" ADD CONSTRAINT \"license_map_rf_fkfk\" FOREIGN KEY (\"rf_fk\") REFERENCES \"license_ref\" (\"rf_pk\") ON UPDATE NO ACTION ON DELETE NO ACTION;";
+  $Schema["CONSTRAINT"]["license_std_comment_user_fk_fkey"] = "ALTER TABLE \"license_std_comment\" ADD CONSTRAINT \"license_std_comment_user_fk_fkey\" FOREIGN KEY (\"user_fk\") REFERENCES \"users\" (\"user_pk\") ON UPDATE SET NULL ON DELETE SET NULL";
```
ag4ums: removing comments while removing the user may not be a good idea,
I think the should be preserved.

ag4ums: Removing comments while removing the user may not be a good idea,
I think the should be preserved.

GMishx: Actually, the foreign key will be set to NULL after removal of user. It will remove the record from the table.

`ON UPDATE SET NULL ON DELETE SET NULL`.

ag4ums: Ok missed it somehow.

PR #1333, opened 2019-04-13
Refactor Travis CI configuration

PeterDaveHello:
## Description

Update Travis CI configuration

### Changes

- Remove all deprecated `sudo:` keys in Travis CI `.travis.yml`
- Disable unnecessary addon to speed up tests
- Leverage yaml anchor for phpunit-tests (so we can easily enable phpdbg to speed up phpunit for php v7.0+)
- Enable ccache to speed up tests
- Enable composer cache to speed up tests
- Enable Fast-Finish to retrieve build result faster
- Fix Coveralls execution path (See previous `Could not open input file: src/vendor/bin/coveralls` error in build logs)
- Fix Coveralls output json file not writable issue
- Run PHPUnit via phpdbg to speed up tests (with a output workaround)

## How to test

Actually, Travis has migrated to the new infra and directly dropped this config, so basically, a passed CI build would be good enough.

It should be easy to find out the build time is shorten, and the Coveralls is working again.

PR #1352, opened 2019-05-14
chore(travis): Set pipefail in Travis CI PHPUnit on PHP 7.0

PeterDaveHello:
## Description

Fix #1333

### Changes

Set shell option pipefail

>  If  set,  the  return value of a pipeline is the value of the last (rightmost)  command  to  exit  with  a non-zero status, or zero if all commands in the pipeline exit successfully.  This  option is disabled by default.

## How to test

Apply the patch on those should fail but didn't.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -162,6 +162,7 @@ jobs:
       php: 7.0
       install: composer update --ignore-platform-reqs --with-dependencies --prefer-dist --working-dir=src phpunit/phpunit
       script:
+        - set -o pipefail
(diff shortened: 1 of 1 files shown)


mcjaeger:
@PeterDaveHello Thank you so much for your contribution, could you pls use conventional changelog messages, like 
`chore(travis): Set pipefail in Travis CI PHPUnit on PHP 7.0`
(for example with changing the commit message with an ammend commit)?

PeterDaveHello:
Sure

PR #1363, opened 2019-05-29
fix(init.d): Implement missing function

### LINKED ISSUE #441: Ubuntu 14.10 init sequence order bug
From Roger at: http://www.fossology.org/boards/3/topics/603

Install fossology On Ubuntu 14.10, It does not auto create init link at rcX.d.
(Example : S20fossology -> ../init.d/fossology on Ubuntu 12.04)
So I create it manual, but use same sequence with postgresql.
After change the init sequence later than postgresql, fossology will auto execute at system boot time.

### LINKED ISSUE #1171: Scheduler fails on boot if database is not available

GMishx:
## Description

This commit adds a weak dependency on postgresql on start.

Fix some warnings be lintian.

### Changes

1. Implement `force-reload` (required function) in `init.d` script.
1. Include default init-functions for Debian.
1. Add lintian override for `script-in-etc-init.d-not-registered-via-update-rc.d` as it is done by `#DEBHELPER#`.
1. Add a weak dependency on postgresql for start.

## How to test

Build the package using `debuild`, check the lintian logs.

Closes #1171
Closes #441

### DIFF
--- a/debian/fossology-scheduler.lintian-overrides
+++ b/debian/fossology-scheduler.lintian-overrides
@@ -1,0 +1,2 @@
+# We use #DEBHELPER# to generate the update-rc and invoke-rc calls
+fossology-scheduler: script-in-etc-init.d-not-registered-via-update-rc.d etc/init.d/fossology
--- a/debian/fossology-scheduler/usr/share/lintian/overrides/fossology-scheduler
+++ b/debian/fossology-scheduler/usr/share/lintian/overrides/fossology-scheduler
@@ -1,0 +1,2 @@
+# We use #DEBHELPER# to generate the update-rc and invoke-rc calls
+fossology-scheduler: script-in-etc-init.d-not-registered-via-update-rc.d etc/init.d/fossology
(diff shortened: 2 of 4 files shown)


ag4ums:
somehow its not working for me... but the code looks fine to me

PR #1372, opened 2019-06-10
refactor(fossology): Refactor modularity

GMishx:
## Description

When installing FOSSology on a cluster, all packages are not required to be installed on all machines.

1. There was a dependency of `fossology-db` and `fossology-decider` on `fossology-web` which were unnecessary if someone is running database on a different host.
1. There was a missing dependency of `fossology-deciderjob` on `fossology-monkbulk` and `fossology-decider` which is required.
1. Common twig templates defined in `src/www/ui/template/components` and `src/www/ui/template/include` were required by some other agents as well which creates unnecessary dependency of `fossology-web` on them.
    - `lib/php/UI/Component/Menu.php` depends on `components/menu.html.twig`
    - `lib/php/UI/template/include/upload.html.twig` and `reportImport/ui/template/ReportImportPlugin.html.twig` depends on `components/select-folder.html.twig`.
    - Files like `lib/php/Plugin/FO_Plugin.php` and `DefaultPlugin.php`, `copyright/ui/template/copyrighthist.html.twig`, `monk/ui/admin-monk-revision.php`, etc requires `include/base.html.twig`

### Changes

1. Fix dependencies defined above in point 1 & 2.
1. Move `src/www/ui/template/components` and `src/www/ui/template/include` to `src/lib/php/UI/template` so they can be made available by `fossology-common`.
1. Update the path of twig templates.
1. Fix template error in `install/fo-postinstall.in`.
1. Update `src/lib/php/Makefile` to ignore test files under `src/lib/php`.

## How to test

1. Make a normal install to check if movement of templates does not effect any working.
1. Do a cluster install with minimal packages to check if everything works.


ag4ums:
@GMishx please rebase the PR.

GMishx:
> 
> 
> @GMishx please rebase the PR.

Done, please check

Issue #1322, opened 2019-04-03
Agents crash when Version is > 23 characters


NicolasToussaint:
### Description
When building Fossology, the Version is [determined automatically](https://github.com/fossology/fossology/blob/2c57a6629585abe7696489d4ef9ea1d41b33b1f3/Makefile.conf#L135) using Git information, or can be forced the version using `make VERSION=<my-version>`

When this version is too long (more than 23 characters), every agent crashes when Fossology starts.

I tested precisely everything is fine with 23 characters, and it breaks at 24 characters.

#### How to reproduce

Build Fossology with a version information longer than 23 characters (`make VERSION="123456789012345678901234"`)

### Versions

* Last commit id on master: 
  - commit 87c4435fdef8134fd1701f2276403a9cfe1eaf54
  - Date:   Tue Apr 2 11:23:22 2019 +0200
* Operating System (lsb_release -a):
```
LSB Version:    core-9.20160110ubuntu0.2-amd64:core-9.20160110ubuntu0.2-noarch:printing-9.20160110ubuntu0.2-amd64:printing-9.20160110ubuntu0.2-noarch:security-9.20160110ubuntu0.2-amd64:security-9.20160110ubuntu0.2-noarch
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.5 LTS
Release:        16.04
Codename:       xenial
```

### Logs

Any logs (if any) generated in

#### FOSSology logs

Logs generated under /var/log/fossology/fossology.log
```
2019-04-03 18:36:15 scheduler [1] :: NOTE: *****************************************************************
2019-04-03 18:36:15 scheduler [1] :: NOTE: ***                FOSSology scheduler started                ***
2019-04-03 18:36:15 scheduler [1] :: NOTE: ***        pid:      1                                        ***
2019-04-03 18:36:15 scheduler [1] :: NOTE: ***        verbose:  3                                        ***
2019-04-03 18:36:15 scheduler [1] :: NOTE: ***        config:   /usr/local/etc/fossology                 ***
2019-04-03 18:36:15 scheduler [1] :: NOTE: *****************************************************************
2019-04-03 18:36:15 scheduler [1] :: ERROR agent.c.322: agent localhost.copyright has been invalidated, removing from agents
2019-04-03 18:36:15 scheduler [1] :: JOB[-3].copyright[107.localhost]: agent didn't send version information: "FATAL libfossscheduler.c.138: unable to insert into agent table: ERROR:  value too long for 
[...]

PR #1388, opened 2019-06-27
Force the VERSION variable to be < 23 chars

NicolasToussaint:
## Description

The issue https://github.com/fossology/fossology/issues/1322 shows that the scheduler crashed if fossology was built with a VERSION variable longer than 23 characters.

This PR simply forces the variable down to 23 characters directly in the Makefile.
There might be a way, more desirable, to make it not crash at all, but at least this solution will prevent the unwanted behaviour for the moment.

### Changes

Simply forcibly truncated the `VERSION`vairable in the file `Makefile.conf`:

`override VERSION:= $(shell echo ${VERSION} | cut -b1-23)`

## How to test

Build Fossology with: `make VERSION=V2345678901234567890123456789 install` and check that:
- the scheduler does not crash at startup
- the top banner shows the truncated version: `Version: [V2345678901234567890123]`

### DIFF
--- a/Makefile.conf
+++ b/Makefile.conf
@@ -136,6 +136,9 @@ VERSION=`git describe --tags > /dev/null 2>&1 && git describe --tags | head -1 |
 BRANCH=`git rev-parse --abbrev-ref HEAD > /dev/null 2>&1 && git rev-parse --abbrev-ref HEAD | head -1 || echo "unknown"`
 COMMIT_HASH=`git show > /dev/null 2>&1 && git show | head -1 | awk '{print substr($$2,1,6)}' || echo "unknown"`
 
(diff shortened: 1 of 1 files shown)


GMishx:
Can you please change the commit message to follow the contribution guidelines?

Maybe like "chore(version): Force the VERSION variable to be < 23 chars"

NicolasToussaint:
Thanks Gaurav ;

> Can you please change the commit message to follow the contribution guidelines?
> 
> Maybe like "chore(version): Force the VERSION variable to be < 23 chars"

I attempted a commit message change, I hope it works OK

GMishx:
It looks fine now. Thanks for updating it.

PR #1392, opened 2019-07-01
fix(deps): Fix dependencies for Debian Buster

GMishx:
## Description

Fix the dependencies for Debian Buster (10).

## How to test

1. Check if all the listed dependencies can be installed without conflicts on all supported distros (at least on stretch and buster).
1. Check if FOSSology can be compiled successfully.


ag4ums:
composer is breaking while installing from source
image

GMishx:
> composer is breaking while installing from source

This appears to be an issue with composer itself not being compatible with PHP 7.3. Try upgrading composer using `composer self-update` or manually installing it.

ag4ums:
well then why not updating the composer in the utils/install_composer.sh

GMishx:
> well then why not updating the composer in the utils/install_composer.sh

I have bumped the composer version to 1.9.0 in the latest commit.

I tested it with PHP 5.6, 7.1 and 7.3 and it worked without any issue. Can you please test now?

Issue #1301, opened 2019-03-09
The FOSSology `/jobs` rest entpoint should contain more useful information


maxhbr:
The `/jobs` endpoint does not contain useful information like
- kind of job
- is it already finished (i.e. status)
- how fast was it
- time of end
- dependencies
- ...

### Details

Currently (https://github.com/fossology/fossology/commit/69629cc1f45f2cbde7e1318dba8cf96a3ac3de8a) the request to `.../jobs` returns something like
```
$ curl -k -s -S -u fossy:fossy -X GET http://localhost:8081/repo/api/v1/jobs
[
[...]
  {                                                                                                                                                                                                                                                           
    "id": 17,                                                                                                                                                                                                                                                 
    "name": "zlib-1.2.11.tar.gz",                                                                                                                                                                                                                             
    "queueDate": "2019-03-09 17:37:46.174642+00",                                                                                                                                                                                                             
    "uploadId": 9,                                                                                                                                                                                                                                            
    "userId": 3,                                                                                                                                                                                                                                              
    "groupId": 3                                                                                                                                                                                                                                              
  },                                                                                                                                                                                                                                                          
  {                                                                                                                                                                                                                                                           
    "id": 18,                                                                                                                                                                                                                                                 
    "name": "zlib-1.2.11.tar.gz",                                                                                                                                                                                                                             
    "queueDate": "2019-03-09 17:43:16.34307+00",                                                                                                                                                                                                              
    "uploadId": 9,                                                                                                                                                                                                                                            
    "userId": 3,                                                                                                                                                                                                                                              
    "groupId": 3                                                                                                                                                                                                                                              
  },                                                                                                                                                                                                                                                          
  {                                                                                                                                                                                                                                                           
    "id": 19,                                                                                                                                                                                                                                                 
    "name": "zlib-1.2.11.tar.gz",                                                                                                                                                                                                                             
    "queueDate": "2019-03-09 17:57:53.938163+00",                                                                                                                                                                                                             
    "uploadId": 9,                                                                                                                                                                                                                                            
    "userId": 3,                                                                                                                                                                                                                                              
    "groupId": 3                                                                                                                                                                                                                                              
  },                                                                                                                                                                                                                                                          
  {                                                                                                                                                                                                                                                           
    "id": 20,                                                                                                                                                                                                                                                 
    "name": "zlib-1.2.11.tar.gz",                                                                                                                                                                                                                             
[...]

PR #1415, opened 2019-08-07
feat(api): Return job's ETA and status

GMishx:
## Description

The jobs endpoint was not returning the job status which confuses users if the job was finished or not.

This PR also intoduces filtering jobs based on upload id, pagination on jobs and missing ojo analysis to REST API.

### Changes

1. Add ETA and job status to the job object returned by `/jobs` endpoint.
    1. The job ETA is in seconds and can be 0 for completed/failed jobs.
    1. The job status can have one of the following values:
        - Completed
        - Failed
        - Queued
        - Processing
1. New filter based on upload id in `/jobs` endpoint.
    - The filter is passed as query parameter `upload` with the request.
1. Added pagination to the `/jobs` endpoint.
    - The request will return total number of pages by `X-Total-Pages` header in response.
    - Accept page number from the request headers along with limit.
1. Added missing agent `ojo` to the analysis on POST request to `/jobs`.

## How to test

1. Check if the job status and ETA are returned for every job.
1. Check if the limit and page works as expected and returns `X-Total-Pages` header.
1. Check if the jobs can be filtered based on upload id.
1. Check if ojo can be scheduled using the rest API.

Related issue: #1301

### DIFF
--- a/src/lib/php/services.xml.in
+++ b/src/lib/php/services.xml.in
@@ -191,7 +191,8 @@ without any warranty.
             <argument type="service" id="dao.folder"/>
             <argument type="service" id="helper.dbHelper"/>
             <argument type="service" id="helper.authHelper"/>
-            <argument type="service" id="logger"/>
+            <argument type="service" id="dao.job"/>
+            <argument type="service" id="dao.show_jobs"/>
         </service>
     </services>
 </container>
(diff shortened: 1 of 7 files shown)


ag4ums:
code looks good, 
@GMishx rebase with current master, to get rid of this  unwanted output on-screen
image

GMishx:
> code looks good,
> @GMishx rebase with current master, to get rid of this unwanted output on-screen

Done. Please check

Issue #1397, opened 2019-07-09
Unpack failure when a filename contains single quotes


NicolasToussaint:
### Description

Unpack failure when a filename contains the character `'`.

#### How to reproduce

1. Upload a package containing a file with a quote in its name (eg. `foo'bar.baz`)
1. The Unpack stage will fail with following logs:
```
2019-07-09 12:49:12 ununpack [0] :: JOB[102].ununpack[1621.localhost]: "sh: 1: Syntax error: Unterminated quoted string"
2019-07-09 12:49:12 ununpack [0] :: JOB[102].ununpack[1621.localhost]: "FATAL utils.c.1585: Unable to calculate SHA256 of /srv/fossology/repository/6764497f58c4/ununpack/469540l/94c2650a0e7a9b7b7ad839860ccb1d9231232b4c.7475b9a3872350af0d2c2300446b0445.240.dir/upload_01.dir/upload_01/foo'bar.baz"
2019-07-09 12:49:12 ununpack [0] :: JOB[102].ununpack[1621.localhost]: agent failed with error code 56
2019-07-09 12:49:12 ununpack [0] :: JOB[102].ununpack[1621.localhost]: agent failed, code: 56
```

The problem come from this line, as the filename is included unprotected:
https://github.com/fossology/fossology/blob/a26c0cd8ec0dca5063248667d734f86cc5c23f08/src/ununpack/agent/utils.c#L1576

Replacing single quotes fixes with `"sha256sum \"%s\""` fixes the single-quote problem, but of course now it fails with filenames containing double quotes :)

I don't know if there is already an escape function available in the source code to reuse ; but I'll look for a quick workaround for now.

#### Screenshots

![image](https://user-images.githubusercontent.com/22956329/60890553-cad21500-a25b-11e9-8757-47d34975d2c6.png)

### Versions

* Last commit id on master: https://github.com/fossology/fossology/commit/53849883c7a95f699f8b2198a1849aa6926dee6f
* Operating System (lsb_release -a): Docker build

GMishx:
Thanks for pointing this out @NicolasToussaint. There is an issue running `sha256sum` as a child process if the path contains any kind of special characters.
This was a quick fix to enclose the path in `'` to accommodate special characters like @, space, etc. But it will fail for another `'`.

The best way will be to escape such characters before constructing the path.

NicolasToussaint:
[...]

PR #1414, opened 2019-08-06
fix(src/ununpack/agent/utils.c) attempt to escape correct characters in filenames

NicolasToussaint:
Fixes #1397 

## Description

Enclosing the filenames in single quotes, and escaping the existing single quotes do not work, somehow (in my Docker environment).
I had to enclose in double quotes and escape `"`, `\` and `$`.

The fix works fine, except for filenames containing '\' characters, see #1397 
I also make additional checks on filename length that were missing before.

### Changes

I copy the filename, character by character, escaping ", \ and $, to the new array, checking for array size.

In the changes, I am not sure which exit code should be used if the (escaped) filename length exceeds the array length.

## How to test

Uploaded a package with the following files ; all but those with `\` are unpacked:
image

PR #1425, opened 2019-08-08
fix(src/ununpack/agent/utils.c) Embed SHA256 implementation,

NicolasToussaint:
Source code under BSD License
Taken from github.com/ogay/sha2 - see also ouah.org/ogay/sha2/

## Description

This PR replaces https://github.com/fossology/fossology/pull/1414 which became a mess.
It fixes https://github.com/fossology/fossology/issues/1397

This PR is clean, and no change in CFLAGS option required anymore.

### Changes

I replaced the shell-based SHA256SUM computation by an embedded implementation

The source implementation is https://github.com/ogay/sha2 / http://ouah.org/ogay/sha2/, unser BSD Licence.

## How to test

I didn't add a unit test (yet), not sure how to do it yet.
I tested with this package: 
[Issue_1397.zip](https://github.com/fossology/fossology/files/3480813/Issue_1397.zip)


fogninid:
The code is much better with the embedded C implementation!

Any reason why you did not join all the checksums computations and add the SHA256 digest to the Cksum struct? https://github.com/fossology/fossology/blob/e1d60cc7fb08509c4afe71bb06bf3c338de4d4fb/src/ununpack/agent/checksum.h#L39

If you do it in checksum.c the unit tests will just be a trivial extension of:
https://github.com/fossology/fossology/blob/e1d60cc7fb08509c4afe71bb06bf3c338de4d4fb/src/ununpack/agent_tests/Unit/test_Checksum.c#L105

GMishx:
There is one catch here @fogninid, there can be a case where a pfile has MD5 and SHA1 calculated and stored in DB but the SHA256 is null. Now if someone uploads the same file again, the SHA256 will be calculated by ununpack agent but it will be missing from DB.
Now, this will make ununpack think that the file is new and tries to insert it in DB but pfile has a unique constraint on MD5, SHA1 and size where this new insert will result in a failure.

fogninid:
> There is one catch here @fogninid, there can be a case where a pfile has MD5 and SHA1 calculated and stored in DB but the SHA256 is null. Now if someone uploads the same file again, the SHA256 will be calculated by ununpack agent but it will be missing from DB.
> Now, this will make ununpack think that the file is new and tries to insert it in DB but pfile has a unique constraint on MD5, SHA1 and size where this new insert will result in a failure.

@GMishx The code, as it currently is, unconditionally computes the SHA256, I am suggesting moving the computation to the same place where the other checksums are computed.

I don't understand the implementation of the optionality of SHA256 :/

As it currently is every call to `AddToRepository` will be performed with the SHA256 in the Fuid

https://github.com/fossology/fossology/blob/e1d60cc7fb08509c4afe71bb06bf3c338de4d4fb/src/ununpack/agent/utils.c#L1660

And if the DB Insert inside it fails, the repofile imported here will be leaked:

https://github.com/fossology/fossology/blob/e1d60cc7fb08509c4afe71bb06bf3c338de4d4fb/src/ununpack/agent/utils.c#L1421

Instead of relying on DB constraints, the existence check should be extended to consider equals two pfiles if the SHA256 in the DB is NULL:

e.g.:
`WHERE pfile_sha1 = '%.40s' AND pfile_md5 = '%.32s' AND (pfile_sha256 = '%.64s' OR pfile_sha256 IS NULL) AND pfile_size = '%s';`

https://github.com/fossology/fossology/blob/e1d60cc7fb08509c4afe71bb06bf3c338de4d4fb/src/ununpack/agent/utils.c#L1139

(It could be made even better by setting the pfile_sha256 entry if it was null)

What am I missing here?

GMishx:
> @GMishx The code, as it currently is, unconditionally computes the SHA256, I am suggesting moving the computation to the same place where the other checksums are computed.

I also agree with you there.

> I don't understand the implementation of the optionality of SHA256 :/

SHA256 can be made required in future versions but for now, the recalculation of hash for old records is quite a lengthy process and might not be pleasing for everyone. Therefore it is kept as optional.
Also, there is a special case where you have an entry in pfile but the actual file is removed from the FS, where recalculation of SHA256 can not be done, which is currently marked as NULL.

> Instead of relying on DB constraints, the existence check should be extended to consider equals two pfiles if the SHA256 in the DB is NULL:
> 
> e.g.:
> `WHERE pfile_sha1 = '%.40s' AND pfile_md5 = '%.32s' AND (pfile_sha256 = '%.64s' OR pfile_sha256 IS NULL) AND pfile_size = '%s';`

That will be a better option.

> (It could be made even better by setting the pfile_sha256 entry if it was null)

It is being implemented here in the PR #1400:

https://github.com/fossology/fossology/blob/75fc1252f14b8d1a3375f84f737c7690a5c3fcb3/src/ununpack/agent/utils.c#L1190

fogninid:
> > @GMishx The code, as it currently is, unconditionally computes the SHA256, I am suggesting moving the computation to the same place where the other checksums are computed.
> 
> I also agree with you there.
> 

Ah, I misinterpreted your comment as saying that it could not be done there. 

> ...
> 
> SHA256 can be made required in future versions but for now, the recalculation of hash for old records is quite a lengthy process and might not be pleasing for everyone. Therefore it is kept as optional.

I was suggesting updating it when ununpack encounters a file with the same (md5,sha1,lenght): it could set the sha256 it just computed. But as you said this is not required for it to work.

> 
> It is being implemented here in the PR #1400:
> 

That's what I was missing, with the fix in #1400 it makes perfect sense! Ty

NicolasToussaint:
There is a problem with the PR as it is .... I don't fully understand.

When I produce **upper case** sha256sums:
- I get this error from the ununpack agent `ununpack [0] :: JOB[10].ununpack[1631.localhost]: "row number 0 is out of range 0..-1"`
- the upload fails
- I can see one file inserted in the DB, with **lowercase** checksum.
```
 pfile_pk |            pfile_md5             |                pfile_sha1                |                           pfile_sha256                           | pfile_size | pfile_mimetypefk 
----------+----------------------------------+------------------------------------------+------------------------------------------------------------------+------------+------------------
        2 | C03EFE875296A7DDAA4CC5FC1D7E3D75 | A4BA656342147FF65DB8CBCE6BEB811230A0C4A0 | 0506e676bc4bb132cbdc646bb6833dd05f13808f9eed172b689fc5657605a631 |     110051 |                 
```
When I produce **lower case** sha256sums:
- everything works fine
- lowercase sha256sums are inserted to the DB

I got confused after looking at PR https://github.com/fossology/fossology/pull/1400, where we *"Uppercase already existing hashes"* : that made me think that all checksums should be stored uppercase (other checksums are stored uppercase).

Any idea on the origin of the problem ?

GMishx:
@NicolasToussaint, I tested your branch on my local system and it unpacked zlib-1.2.8 very well without any error

image

I also tried to upload the same file again and it went through successfully.

Then I droped the `pfile_sha256` column and added it back again so it contains no value and upload same package again. And again ununpack **did not** fail.

I am not sure why it failed for you.

The only thing is that the `pfile_sha256` is not updated for the pfiles which I deleted. Maybe it can be done by the way @fogninid suggested.

NicolasToussaint:
Mmh :-/

Looking at the [conflict in the file utils.c](https://github.com/fossology/fossology/pull/1425/conflicts) above, I guess I didn't test with the code forcing the upper case ... and I suppose that you had that code already.

image


I will update my master and test again.

NicolasToussaint:
I fixed the conflict by removing the code where the SHA256 is forced to uppercase, since it is now directly computed this way (`...printf "%2.X"`).

Just tested the branch, everything is in working order, all checksums stored in uppercase in the DB.
Can we merge the PR ?

NicolasToussaint:
> Can we merge the PR ?

oh no, Travis errors :-(

GMishx:
> > Can we merge the PR ?
> 
> oh no, Travis errors :-(

Restarted the failed job.

src/ununpack/agent/utils.c
```diff
@@ -21,6 +21,12 @@
 #include "ununpack.h"
 #include "externs.h"
 #include "regex.h"
+#include "sha2.h"
+
+#define NEW_SHA256
+#ifdef NEW_SHA256
```
GMishx: Do we still need these macros?

NicolasToussaint: No we don't,
I'm sorry, that's a leftover, and should be removed

NicolasToussaint: I remove it in latest commit

NicolasToussaint:
Idea and code look good - I merge this patch to my branch then - is that what you meant ?

PR #1427, opened 2019-08-12
Fix first fossology start failure with vagrant file

andi8086:
Update systemd configuration before starting the init.d script,
so that systemd generates the missing fossology.service in time.

Signed-off-by: Andreas J. Reichel <andreas.reichel@tngtech.com>

## Description

Vagrant:
Reload systemd configuration before first startup of fossology

### Changes

Added a systemctl daemon-reload

## How to test

vagrant destroy
vagrant up

Fossology now starts correctly. Before it did not.


tobemedia:
this change fixed the problem for me.
+1 for merge

PR #1430, opened 2019-08-13
fix(ojo): Remove call to omitEndingLineFeed on json cpp < 0.7

GMishx:
## Description

`libjsoncpp` prior to 0.7.0 does not support `omitEndingLineFeed()` which removes extra `\n` at end.

But since version 1.4.0, `FastWriter` class itself is deprecated. So, till version 1.3.x, remove the `\n` at end explecitly and from 1.4.0 onwards, use `StreamWriterBuilder` class.

Also use boost range iterator since `recursive_directory_iterator` does not have begin and end on version 1.55 (Debian 8).

### Changes

1. Use `FastWriter` only if json cpp is < 1.4.0, use `StreamWriterBuilder` since 1.4.0.
1. Wrap `recursive_directory_iterator` with `iterator_range`.

## How to test

1. Build `ojo` on Debian 8
1. Check if the JSON output is correct.
1. Check if the directory scan is working.

### DIFF
--- a/src/ojo/agent/OjoUtils.cc
+++ b/src/ojo/agent/OjoUtils.cc
@@ -351,8 +351,16 @@ void appendToJson(const std::string fileName,
     bool &printComma)
 {
   Json::Value result;
-  Json::FastWriter jsonBuilder;
-  jsonBuilder.omitEndingLineFeed();
+#if JSONCPP_VERSION_HEXA < ((1 << 24) | (4 << 16))
+  // Use FastWriter for versions below 1.4.0
+  Json::FastWriter jsonWriter;
+#else
+  // Since version 1.4.0, FastWriter is deprecated and replaced with
+  // StreamWriterBuilder
+  Json::StreamWriterBuilder jsonWriter;
+  jsonWriter["commentStyle"] = "None";
+  jsonWriter["indentation"] = "";
+#endif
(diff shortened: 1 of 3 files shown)


src/ojo/agent/OjoUtils.cc
```diff
@@ -352,7 +352,11 @@ void appendToJson(const std::string fileName,
 {
   Json::Value result;
   Json::FastWriter jsonBuilder;
+#if JSONCPP_VERSION_MAJOR > 0 && JSONCPP_VERSION_MINOR > 6
```
andi8086: Don't we need an alternative function then for older versions? Otherwise we can get different results with older versions.

GMishx: Nice idea. Thanks.

GMishx: I have made the required changes to keep the output same on all versions. Can you please review it again?

mcjaeger:
this has been fixed and is in the new package for Debian 8?

GMishx:
> this has been fixed and is in the new package for Debian 8?

The Debain 8 package for 3.6.0-rc1 does not contain the latest changes from this branch. Should I update the package or we can wait for the branch to be merged and it will automatically be in new release.

PR #1456, opened 2019-08-26
fix(ajaxExplorer): Reduce view creation

GMishx:
## Description

Since PR #1314, the number of times `UploadTreeProxy` is materialized is increased (2 per folder in the view). Each call takes significant amount of time depending on the entries in DB.

This PR tires to reduce these calls to only 2 per UI load.

The average amount of time taken to load `?mod=ajax_explorer` for an upload with 10 files is reduced from ~600ms to ~350ms.

### Changes

1. Create upload view proxy in DB only once with the complete `ItemTreeBounds`.
2. Fix PHP warnings for `AgentLicenseEventProcessor.php` when browsing license for an empty file.

## How to test

1. Upload a huge package and check the loading time for License Browser table.
1. Install the fix and brows the same package.
    - The load time should reduce.

### DIFF
--- a/src/lib/php/BusinessRules/AgentLicenseEventProcessor.php
+++ b/src/lib/php/BusinessRules/AgentLicenseEventProcessor.php
@@ -181,7 +181,11 @@ class AgentLicenseEventProcessor
         $this->latestAgentMapCache[$uploadId][$agentName] = $agentMap;
       }
     }
-    return $this->latestAgentMapCache[$uploadId];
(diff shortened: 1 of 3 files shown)


mcjaeger:
I was testing it and compared it previous 3.6.0 rc2 code. For the fossology master I was getting about 1000 msec on average for 'ajax_explorer', and with this patch consistent about 600 msec. So speed increase is working.

PR #1471, opened 2019-09-10
feat(notice_import): Import notice file content into acknowledgement

### LINKED ISSUE #1465: Find notice files and add them as acknowldgement support
Adding content from notice files is much manual work and thus automation would be help fun.

Today, if a project has a notice file and the license is asking for adding this, then the contents is added manually to the file-license conclusion as acknowledgements.

### Description

in the dialog:

Screenshot 2019-09-04 at 14 41 09

There should be also the proposal to add contents from a file. The user can then select the files found in the repo which are named 'NOTICE' or 'NOTICE*'.

Then, the contents of the file will be added from acknowledgment. (please keep line breaks)

andi8086:
Add a second text input modal dialog specifically for acknowledgement
texts.
Add a button to open a file browser accepting text files.
Read selected text file and set the text in the textarea input field
according to the file content.

Closes #1465


mcjaeger:
(from discussion)
There is one case where the automation could make sense:
* we have about 25% of apache-licensed components
* so for the case where
  * one file `NOTICE` or `notice.*` is found
  * and only apache license is found as scan result
* the agent would
  * conclude apache license
  * add content of the notice file to acknowledgment text

A very simple agent, that could be additionally selected at the upload dialogues
* REST endpoint should also have this option
* Agent needs to be after the scanner (like the reuse agent)

mcjaeger:
Adding to that the endpoint in question is a POST request on the `jobs` resource "Schedule an Analysis of an existing upload" ...
 
API description is here:

https://raw.githubusercontent.com/fossology/fossology/master/src/www/ui/api/documentation/openapi.yaml

PR #1479, opened 2019-09-20
fix(groups): add validations and remove CONSTRAINTS

shaheemazmalmmd:
## Description

* Add validation in case of removing all users from group.
* Remove some constraints .
* add notes 

## How to test

* Create a group. add some users.
* Try Remove all users from that group from "manage group users".
* it should not allow you to remove all users.

### DIFF
--- a/src/www/ui/page/AdminGroupUsers.php
+++ b/src/www/ui/page/AdminGroupUsers.php
@@ -67,9 +67,16 @@ class AdminGroupUsers extends DefaultPlugin
     }
 
     $gum_pk = intval($request->get('gum_pk'));
+    $text = "";
     if ($gum_pk) {
       $perm = intval($request->get('perm'));
-      $this->updateGUMPermission($gum_pk, $perm);
+      $atleastOneUserShouldBePart = $dbManager->getSingleRow("SELECT count(*) cnt FROM group_user_member WHERE group_fk = (SELECT group_fk FROM group_user_member WHERE group_user_member_pk = $1)",
+      array($gum_pk), $stmt = __METHOD__ . ".atleastOneUserShouldBePart");
+      if ($atleastOneUserShouldBePart['cnt'] <= 1) {
+         $text = _("Error: atleast one user should be part of a group.");
+      } else {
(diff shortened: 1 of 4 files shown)


src/www/ui/template/admin_group_delete.html.twig
```diff
@@ -15,6 +15,10 @@
     {{ macro.select('grouppk',groupMap,'grouppk','','class="ui-render-select2"') }}
     <input type="submit" value="{{ 'Delete'|trans }}"/>
   </form>
+<br/>
+<br/>
+<b>{{ 'Note: After deleting group, you will loose uploads That are associated with deleted group.'|trans }}</b>
```
GMishx:
```suggestion
<b>{{ 'Note: After deleting group, you will loose access to uploads that are associated with deleted group.'|trans }}</b>
```

src/www/ui/template/admin_group_users.html.twig
```diff
@@ -38,7 +38,10 @@
 <br/>
 {{ 'All user permissions take place immediately when a value is changed.  There is no submit button.'|trans }}
 <br/>
-{{ 'Add new users on the last line.'|trans }}  
+{{ 'Add new users on the last line.'|trans }}
+<br/>
+<br/>
+<b>{{ 'Note: By removing users You may loose uploads, That are uploaded by them.'|trans }}</b>
```
GMishx:
```suggestion
<b>{{ 'Note: By removing users, you may loose access to uploads, that are uploaded by them.'|trans }}</b>
```

src/www/ui/page/AdminGroupUsers.php
```diff
@@ -137,6 +144,9 @@ function setNewPermUrl(newperm){
     }
 
     $vars['scripts'] = $scripts;
+    if(!empty($text)) {
+      $vars['scripts'] .= displayMessage($text);
```
GMishx: Won't it be be better to have something like this?
```suggestion
      $vars['message'] .= $text;
```

It will also give consistent view for error messages.

PR #1480, opened 2019-09-20
feat(pbconf): adapt to pb 0.15 and new fossology 3.3+

bcornec:
## Description

Support pb 0.15

### Changes

New pb 0.15 is now using YAML file formats so adding these files in order to build fossology again using this new version
Fixing rpm build (Mageia based, but generic fixes) to take in account new files needing inclusion in the package.

## How to test

pb -p fossology sbx2pkg


bcornec:
In preparation of my call with @mcjaeger

PR #1487, opened 2019-09-26
feat(upload_vcs.html.twig) make Git the default VCS rather than SVN

### LINKED ISSUE #1485: Irrelevant documentation
Throughout the whole code, subversion is referenced as cvs. Don't we use git as default?

NicolasToussaint:

## Description

fixes #1485 

Just reorder the VCS in the drop down list so that Git is selected by default.

## How to test

Choose 'Upload From Version Control System', verify that Git is selected by default.

### DIFF
--- a/src/www/ui/template/upload_vcs.html.twig
+++ b/src/www/ui/template/upload_vcs.html.twig
@@ -19,8 +19,8 @@
   </label>
   <br/>
   <select name="{{ vcstypeField }}">
-    <option value='SVN'>SVN</option>
     <option value='Git'>Git</option>
+    <option value='SVN'>SVN</option>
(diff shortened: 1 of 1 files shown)


src/www/ui/template/upload_vcs.html.twig
```diff
@@ -19,8 +19,8 @@
   </label>
   <br/>
   <select name="{{ vcstypeField }}">
-    <option value='SVN'>SVN</option>
     <option value='Git'>Git</option>
+    <option value='SVN'>SVN</option>
```
shaheemazmalmmd: i think `<option selected>` would be better instead of changing the order.

PR #1488, opened 2019-09-27
feat(spdx): add user found copyrights to SPDX reports

### LINKED ISSUE #1362: Adding Copyright entry from GUI (copyright-view)
### Description

On copyright-view page it is possible do deactivate or edit copyright entry found in file. 
It would be also useful to have possibility to add new copyright directly from the GUI - in case that there was no coyright entry in file but we know that specific file should have one.

I suppose it could look like this:
1. Find file without proper copyright entry
2. Go to Copyright/Email/Url/Author menu entry
3., Enter proper Copyright text to some text field (probably existing with 'Replace' label can be used
4. Click new Add button - new copyright entry should be added and persisted for this specific file

#### Screenshots

Please find GUI page I'm writing about:

image

piotreke12:
###  Description

PR adds user found Copyright entries to SPDX reports

Fixes #1362 

### Changes

src/spdx2/agent/spdx2.php - read copyright entries from copyright_decision table and add them to result array (merged with scanner results)

## How to test

1. Add new Copyright in GUI
2. Generate SPDX report
3. Check if previously added copyright can be found on report

### DIFF
--- a/src/spdx2/agent/spdx2.php
+++ b/src/spdx2/agent/spdx2.php
@@ -527,9 +527,13 @@ class SpdxTwoAgent extends Agent
     /* @var $copyrightDao CopyrightDao */
     $copyrightDao = $this->container->get('dao.copyright');
     $uploadtreeTable = $this->uploadDao->getUploadtreeTableName($uploadId);
-    $allEntries = $copyrightDao->getAllEntries('copyright', $uploadId, $uploadtreeTable, $type='statement');
-    foreach ($allEntries as $finding) {
+    $allScannerEntries = $copyrightDao->getScannerEntries('copyright', $uploadtreeTable, $uploadId, $type='statement', $extrawhere=null);
+    $allEditedEntries = $copyrightDao->getEditedEntries('copyright_decision', $uploadtreeTable, $uploadId, $decisionType=null);
+    foreach ($allScannerEntries as $finding) {
       $filesWithLicenses[$finding['uploadtree_pk']]['copyrights'][] = \convertToUTF8($finding['content'],false);
+    }
+    foreach ($allEditedEntries as $finding) {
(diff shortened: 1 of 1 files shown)


src/spdx2/agent/spdx2.php
```diff
@@ -527,10 +527,14 @@ protected function addCopyrightResults(&$filesWithLicenses, $uploadId)
     /* @var $copyrightDao CopyrightDao */
     $copyrightDao = $this->container->get('dao.copyright');
     $uploadtreeTable = $this->uploadDao->getUploadtreeTableName($uploadId);
-    $allEntries = $copyrightDao->getAllEntries('copyright', $uploadId, $uploadtreeTable, $type='statement');
-    foreach ($allEntries as $finding) {
+    $allScannerEntries = $copyrightDao->getScannerEntries('copyright', $uploadtreeTable, $uploadId, $type='statement', $extrawhere=null);
```
shaheemazmalmmd: You can use the existing function getAllEntriesReport from copyrightDao to get both scanner and edited entries.

piotreke12: Hello,

thanks for your comment.
I tried to use getAllEntriesReport, but this function merges scannerEntries and editedEntries arrays in a way it's hard to use them in the report (For example: we have 1 scanner and 5 user-added copyrights - result of this function is array having 5 rows with user-edited copyrights but for each row there is also 5 times duplicated scanner copyright)
I hope this explanation is clear enought - if not, I can show you an example :)

That's why I decided to use 2 functions separately (funcion getAllEntriesReport function is used also by Unified Report and i decided no to modify it)

shaheemazmalmmd: ok

PR #1491, opened 2019-10-01
fix(geekyscan): make full job report link more descriptive

### LINKED ISSUE #1346: At job detail page - replace 3 dots which gives access to full report
### Description

In order to see the details of a job in the job overview page, the job number needs to be clicked to open the "Geeky Scan Details" page. Scrolling down, the report is cut at some line. Below the report, there are 3 dots displayed, which - after clicking - gives access to the full report. I did not recognize the dots to represent a link. 

I would propose to replace the 3 dots with text "display full log" in order to make it clear for the user, that the information is available but hidden. See link displayed at the bottom left of the screenshot. It should be clear, that a link is represented by the dots...
Screenshot-LogLink

prondubuisi:
fixes issue #1346 

## Description

This pull request changes the three dots(...) in the view full log hyper link to show "display full log" which is more descriptive

### Changes

changed [this file](https://github.com/fossology/fossology/blob/9d0765c2eda0b72c4a5aaac640322d4f59ffd7f4/src/www/ui/async/AjaxShowJobs.php#L156-L166)

## How to test

Upload a project as compressed file, allow job to complete, click on job to view details, if log is too long you should see the "Download full log" link instead of "..."

Closes #1346

### DIFF
--- a/src/www/ui/async/AjaxShowJobs.php
+++ b/src/www/ui/async/AjaxShowJobs.php
@@ -159,7 +159,7 @@ class AjaxShowJobs extends \FO_Plugin
           }
           if (filesize($row[$field]) > self::MAX_LOG_OUTPUT) {
             $value = "<pre>" .file_get_contents($row[$field],false,null,-1,self::MAX_LOG_OUTPUT)."</pre>"
-                    .'<a href="'.Traceback_uri() . '?mod=download&log=' . $row['jq_pk'] . '">...</a>';
+                    .'<a href="'.Traceback_uri() . '?mod=download&log=' . $row['jq_pk'] . '">Download full log</a>';
           } else {
             $value = "<pre>" . file_get_contents($row[$field]). "</pre>";
           }


prondubuisi:
@GMishx can you please review. Thanks

src/www/ui/async/AjaxShowJobs.php
```diff
@@ -159,7 +159,7 @@ protected function getGeekyScanDetailsForJob($job_pk)
           }
           if (filesize($row[$field]) > self::MAX_LOG_OUTPUT) {
             $value = "<pre>" .file_get_contents($row[$field],false,null,-1,self::MAX_LOG_OUTPUT)."</pre>"
-                    .'<a href="'.Traceback_uri() . '?mod=download&log=' . $row['jq_pk'] . '">...</a>';
+                    .'<a href="'.Traceback_uri() . '?mod=download&log=' . $row['jq_pk'] . '">display full log</a>';
```
GMishx:
```suggestion
                    .'<a href="'.Traceback_uri() . '?mod=download&log=' . $row['jq_pk'] . '">Download full log</a>';
```

GMishx: Since the link provides a file to download, the display should be replaced with download in the message.

PR #1496, opened 2019-10-11
feat(rest): Provide API version as an endpoint

GMishx:
## Description

Provide information about REST API via an endpoint.

### Changes

Create a new endpoint `/version` to get following information about REST API:
    - Version
    - Authentication methods allowed

## How to test

1. Check if the current API version returned by `/version` endpoint matches with `src/www/ui/api/documentation/openapi.yaml`.
1. Check if other endpoints are not effected.

Closes #1474

PR #1498, opened 2019-10-14
feat(rest): Provide group ID during POST upload

### LINKED ISSUE #1468: REST API: Associate new upload to a GroupID
### Description

Today, the uploads triggered via the REST API are associated to the User & Group for which the token was created.

We would need to add the possibility associate the upload to an alternate GroupID.

Use REST command such as this to create a new Upload, triggering a new Git Clone

POST .../api/v1/uploads -H 'uploadGroupId="<my-alternate-group>"' (...)

GMishx:
## Description

The PR allows user to define the group ID to be used while creating a new upload via REST API.

Depends on #1496 for API version.

### Changes

1. Read the group ID via header ~`uploadGroupId`~ `groupId`.
1. Pass the group ID in the request object to upload pages.

## How to test

1. Do a normal upload without group id and check if the upload is done under default group (last chosen by user in UI).
1. Do an upload providing a valid group id and check if the upload is done under the provided group id.
1. Do an upload providing an invalid/inaccessible group id and check if the request fails.

Closes #1468

### DIFF
--- a/src/www/ui/api/Controllers/UploadController.php
+++ b/src/www/ui/api/Controllers/UploadController.php
@@ -26,6 +26,7 @@ namespace Fossology\UI\Api\Controllers;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Fossology\Lib\Auth\Auth;
+use Fossology\UI\Page\UploadPageBase;
 use Fossology\UI\Api\Models\Info;
 use Fossology\UI\Api\Models\InfoType;
 use Fossology\UI\Api\Helper\UploadHelper;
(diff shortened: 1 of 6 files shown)


NicolasToussaint:
Great, thanks @GMishx !

Testing your PR now.

NicolasToussaint:
Works fine!

And, well, now ... we will need to do the same to search a folder and trigger  scans ;-)

GMishx:
> Works fine!

Thank you for your prompt testing.

> And, well, now ... we will need to do the same to search a folder and trigger scans ;-)

Should I rename the header field `uploadGroupId` to something more generic like `groupId` so it can be made common among all other endpoints?

NicolasToussaint:
> Should I rename the header field `uploadGroupId` to something more generic like `groupId` so it can be made common among all other endpoints?

That sounds like a good idea to me.

Also, I realised that a user can trigger a scan on an upload that belongs to another user and group.  
I'm not sure if it's desirable, or a bug.  
Ex. 
- UserA & UserB have no group in common.
- UserA uploads File1
- UserB can trigger a scan on that upload

So:
- If it's a feature, no need to add a new `groupId` parameter for the `/jobs` command
- If it's a bug, I guess it needs being fixed + add the  `groupId` parameter

GMishx:
Oh, it is a bug then as you can not do it from UI and REST API should follow the same rules. Can you please create an issue for this?

PR #1499, opened 2019-10-14
fix(rest): getUploads - invoke getRows with proper parameters 

### LINKED ISSUE #1481: REST API: Incoherent /search result
### Description

There seems to be a problem with the REST API `/search` feature.

When uploading several times the same files, searching the file returns the correct number of entries, but all information is identical (match the first upload)

Tested via the REST API.

#### How to reproduce

- Upload a zip file
  - `{
  "code": 201,
  "message": 31,
  "type": "INFO"
}
`
- Add a file in the zip archive, and upload again
  - `{
  "code": 201,
  "message": 32,
  "type": "INFO"
}
`
- Search for the file:
  - `curl -k -s -S -X GET https://<server>/api/v1/search -H "Authorization:Bearer eyJ..." -H filename:reuse_3.zip | jq .`
- The results show twice the first upload:
```
[
  {
    "upload": {
      "folderid": 11,
      "foldername": "Reuse_D",
      "id": 31,
      "description": "Uploaded via REST",
      "uploadname": "reuse_3.zip",
      "uploaddate": "2019-09-20 09:19:29.074775+00",
      "filesize": 1144
    },
    "uploadTreeId": 269,
    "filename": "reuse_3.zip"
  },
  {
    "upload": {
      "folderid": 11,
      "foldername": "Reuse_D",
      "id": 31,
      "description": "Uploaded via REST",
      "uploadname": "reuse_3.zip",
      "uploaddate": "2019-09-20 09:19:29.074775+00",
      "filesize": 1144
    },
    "uploadTreeId": 275,
    "filename": "reuse_3.zip"
  }
]
```

### Versions

Recent version fro Master

piotreke12:
## Description

This PR fixes problem with incoherent search results using REST API
Fixes #1481 

### Changes

src/www/ui/api/Helper/DbHelper.php - call to dbManager.getRows method passes correct parameters (query parameters in array and statementName) to enable running multiple calls from one method with different params

## How to test

when: using api/v1/search to find file that has multiple occurences in different uploads

before: for files in different uploads, result JSON lists the same upload details (same as first element)

after fix: upload details are correct in result JSON

### DIFF
--- a/src/www/ui/api/Helper/DbHelper.php
+++ b/src/www/ui/api/Helper/DbHelper.php
@@ -86,8 +86,10 @@ FROM upload
 INNER JOIN folderlist ON folderlist.upload_pk = upload.upload_pk
 INNER JOIN folder ON folder.folder_pk = folderlist.parent
 INNER JOIN pfile ON pfile.pfile_pk = upload.pfile_fk
-WHERE upload.user_fk = " . pg_escape_string($userId) . "
+WHERE upload.user_fk = $1
 ORDER BY upload.upload_pk;";
+      $statementName = __METHOD__ . ".getAllUploads";
+      $params = [$userId];
     } else {
       $sql = "SELECT
 upload.upload_pk, upload.upload_desc, upload.upload_ts, upload.upload_filename,
@@ -96,12 +98,13 @@ FROM upload
 INNER JOIN folderlist ON folderlist.upload_pk = upload.upload_pk
 INNER JOIN folder ON folder.folder_pk = folderlist.parent
 INNER JOIN pfile ON pfile.pfile_pk = upload.pfile_fk
-WHERE upload.user_fk = " . pg_escape_string($userId) . "
-AND upload.upload_pk = " . pg_escape_string($uploadId) . "
+WHERE upload.user_fk = $1
+AND upload.upload_pk = $2
 ORDER BY upload.upload_pk;";
+      $statementName = __METHOD__ . ".getSpecificUpload";
+      $params = [$userId,$uploadId];
     }
-
-    $result = $this->dbManager->getRows($sql);
+    $result = $this->dbManager->getRows($sql, $params, $statementName);
     $uploads = [];
     foreach ($result as $row) {
       $upload = new Upload($row["folder_pk"], $row["folder_name"],


GMishx:
Thank you @piotreke12 for the fix.

Can you please change the commit message to something more compliant to [contributing guideline](https://github.com/fossology/fossology/blob/master/CONTRIBUTING.md)?

Maybe something like: `fix(rest): getUploads - invoke getRows with proper parameters`

piotreke12:
Hello @GMishx - sorry for this.

How, in your oppinion, is the best way to change this commit name?
As I can see, there is also one additional change to be done (wrong indentation)

piotreke12:
I can try to change commit name locally and then push changes with force to PullRequest source branch "Orange-OpenSource:feat/orange-opensource/1481/api_incoherent_search_results_fix", but I'm not sure if it is the best possible way to do it and if there won;t be any problems connected with such change

GMishx:
We also prefer changing the commit name using `--ammend` and force push to the branch as you've mentioned. It helps keeping the changelog clean and uniform in master.

piotreke12:
> We also prefer changing the commit name using `--ammend` and force push to the branch as you've mentioned. It helps keeping the changelog clean and uniform in master.

Thanks - commit updated.
I also changed indents to fix one of Travis's errors

src/www/ui/api/Helper/DbHelper.php
```diff
@@ -86,8 +86,10 @@ public function getUploads($userId, $uploadId = null)
 INNER JOIN folderlist ON folderlist.upload_pk = upload.upload_pk
 INNER JOIN folder ON folder.folder_pk = folderlist.parent
 INNER JOIN pfile ON pfile.pfile_pk = upload.pfile_fk
-WHERE upload.user_fk = " . pg_escape_string($userId) . "
+WHERE upload.user_fk = $1
 ORDER BY upload.upload_pk;";
+      $statementName = __METHOD__ . ".getAllUploads";
+      $params = [pg_escape_string($userId)];
```
GMishx: I think the use of `pg_escape_string` is not required here as once a statement is prepared, it's query plan cannot be changed, thus can be attacked by SQL injection.

What are the views of others on this?

GMishx: Instead of using `pg_escape_string`, it will better to use `intval` as `pg_escape_string` will do no good even if a string gets slipped during the call somehow.

piotreke12: Ok, @GMishx  - if it's final decision, I can update source branch with this.

If i understood well our previous discussion, source branch should still contain only 1 commit, so I should use `-amend` commit option

maxhbr: I think the use of `pg_escape_string` is even harmful here, since it results in double escaping. Just drop it since it is not necessary for prepared statements.

piotreke12: Hello @GMishx, @maxhbr - PR updated as You requested (deleted `pg_escape_string`)

GMishx: Approved the changes. Thank you.

PR #1508, opened 2019-10-22
chore(docker-compose): harmonize versions with sw360chores

maxhbr:
The SW360chores docker compose project uses version 3.5. Bumping the version
here has no negative effects (except for probably old distributions having not a
new enough docker-compose, needs Docker Engine 17.12.0+).

The advantage is, that this allows the deployments to work in sync and one can
start both at the same time e.g. via:
```
./sw360chores.pl --build -- -f path/to/fossology/docker-compose.yml up
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@
 #
 # Description: Recipe for setting up a multi container FOSSology
 #              Docker setup with separate Database instance
-version: '3'
+version: '3.5'
 services:
   scheduler:
     build:


mcjaeger:
I am not much into this, but if we narrow it down to 3.5 instead of 3.*, could we have maybe a lockout on specific distros because of the mapping between compose and docker itself? ... Because there is a mapping between file format version and docker engine (which is then imposed at 17.06?)

maxhbr:
3.5 means, that we allow a special feature set in the compose files, which requires that the corresponding docker engine also supports that (and thus has at least a specific version). Since docker recommends to install itself directly, instead of via the package manager, the version should not be a problem and all distros should have access to a recent docker.

The version 3.5 also introduced features, which we require for sw360chores, but were not present in 3.4. So if we harmonize the version it has to be at least 3.5.

PR #1525, opened 2019-11-13
chore(installdeps): remove php-yaml from os level and add it to composer

shaheemazmalmmd:
## Description

- Remove php-yaml from os level and add it to composer
- Update composer.lock file with latest changes
- Correct failing yaml file
- Remove php-yaml dependency from debian/control file

## How to test

* build Debian packages and test if the installation works file without errors.

### DIFF
--- a/src/www/ui/api/Controllers/VersionController.php
+++ b/src/www/ui/api/Controllers/VersionController.php
@@ -25,6 +25,8 @@ namespace Fossology\UI\Api\Controllers;
 
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\Yaml\Parser;
+use Symfony\Component\Yaml\Exception\ParseException;
 
 /**
  * @class VersionController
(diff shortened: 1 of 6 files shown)


ag4ums:
Please fix the indentation issue : 
image

shaheemazmalmmd:
> Please fix the indentation issue :
> image

Done. thanks for pointing.

src/www/ui/api/Controllers/VersionController.php
```diff
@@ -42,8 +44,12 @@ class VersionController extends RestController
    */
   public function getVersion($request, $response, $args)
   {
-    $yamlDocArray = yaml_parse_file(__DIR__ .
-      "/../documentation/openapi.yaml");
+    try {
+      $yaml = new Parser();
+      $yamlDocArray = $yaml->parse(file_get_contents(__DIR__ ."/../documentation/openapi.yaml"));
```
GMishx: Won't it be good to use the syntax (`Yaml::parseFile(__DIR__ ."/../documentation/openapi.yaml")`) mentioned in the documentation of symfony/yaml instead of calling the underlying functions?

It will keep the code more readable and easy to maintain.

shaheemazmalmmd: Yes. 
for the version v3.3.18 the proposed syntax `(Yaml::parseFile(__DIR__ ."/../documentation/openapi.yaml"))` is not supported.
so i have chosen this way.

GMishx: Ah, I must have checked the latest documentation then.

Why not using the syntax for v3.3 (`Yaml::parse(file_get_contents('config.yml'));`)?

shaheemazmalmmd: Did you already check if it is working ??

shaheemazmalmmd: I don't this there is much logic which is confusing the code. apart from initializing the class. i think the code is fine.

PR #1526, opened 2019-11-14
fix(highlight): highlight for reference text that exists in different…

shaheemazmalmmd:
## Description

Add page number to the URL which shows highlight.

## How to test

* Upload a package with multiple license findings for a single page.
* Go to admin -> Customize -> Change "Chars per page in text view" to 100 or 50 depending upon your license texts length( different license references should fall under different pages). 
* Click on reference number under source for particular agent finding.
* Check if it takes you to the page where license reference text exists.

### DIFF
--- a/src/lib/php/Dao/HighlightDao.php
+++ b/src/lib/php/Dao/HighlightDao.php
@@ -196,4 +196,24 @@ class HighlightDao
     $highlightEntries = array_merge(array_merge($highlightDiffs,$highlightKeywords),$highlightBulk);
     return $highlightEntries;
   }
+
+  /**
+   * @param licenseMatchId
+   * @return page number
+   */
+  public function getPageNumberOfHighlightEntry($licenseMatchId)
+  {
+    $row = $this->dbManager->getSingleRow(
+      "SELECT FLOOR(
+                (
+                  SELECT start FROM highlight WHERE fl_fk=$1 ORDER BY start ASC LIMIT 1
+                ) / (
+                  SELECT conf_value FROM sysconfig WHERE variablename LIKE 'BlockSizeText'
+                )::numeric
+              )
+       AS page;",
+      array($licenseMatchId)
+    );
+    return $row['page'];
+  }
 }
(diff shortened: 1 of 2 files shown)


src/lib/php/Dao/HighlightDao.php
```diff
@@ -196,4 +196,24 @@ public function getHighlightEntries(ItemTreeBounds $itemTreeBounds, $licenseId =
...
+                (
+                  SELECT start FROM highlight WHERE fl_fk=$1 ORDER BY start DESC LIMIT 1
+                ) / (
+                  SELECT conf_value FROM sysconfig WHERE variablename LIKE 'BlockSizeText'
+                )::numeric
+              )::bigint
+       AS page;",
+      array($licenseMatchId)
+    );
+    return $row['page']-1;
```
GMishx: If the `page` is `0`, the function will return `-1`.

So, if there are no pages for the file, or the highlight is on the first page itself, the link generated will return a blank page.

GMishx: Or, if the license is on page 1, I am getting page 0.

shaheemazmalmmd: Yes, the pagination starts from 0 for licenses.

GMishx: For some reason it started with 1 on my Ubuntu Xenial machine and with 0 on Debian Buster.

I will investigate and let you know.

src/lib/php/Dao/HighlightDao.php
```diff
@@ -196,4 +196,24 @@ public function getHighlightEntries(ItemTreeBounds $itemTreeBounds, $licenseId =
...
+  /**
+   * @param licenseMatchId
+   * @return page number
+   */
+  public function getPageNumberOfHighlightEntry($licenseMatchId)
+  {
+    $row = $this->dbManager->getSingleRow(
+      "SELECT (
+                (
+                  SELECT start FROM highlight WHERE fl_fk=$1 ORDER BY start DESC LIMIT 1
```
GMishx:
```suggestion
                  SELECT start FROM highlight WHERE fl_fk=$1 ORDER BY start ASC LIMIT 1
```

Can we do a ascending sort so that the user lands at the beginning of the match and manually move forward instead of finding the end of the match and moving backward to find the starting the match? It just feels more natural.

src/lib/php/Dao/HighlightDao.php
GMishx: I just recalled what I reported first.
To rephrase,

If there are no page on the file (file is loaded at a one go), the `page` returned is `-1` which is invalid.

So maybe a check `<0` will be good.

Issue #1492, opened 2019-10-02
Change DB encoding from SQL_ASCII to UTF8


alpianon:
### Description

I was in need to export some data (copyright notices) from Fossology DB to another application and I ran into the problem of character encoding. Fossology DB uses SQL_ASCII, which is pure evil when you have to export data somewhere else (or when you want to use tools like phppgadmin).

In the end I decided to convert the Fossology DB into UTF8, with the help of [this blog post](https://www.endpoint.com/blog/2017/07/21/postgres-migrating-sqlascii-to-utf-8). After that, I ran a lot of copyright and license scans on that machine, and I got no errors whatsoever due to character encoding problems. So it seems that no Fossology component actually depends on or requires SQL_ASCII (or that I have been lucky so far...).

**Use of SQL_ASCII encoding in Fossology DB is probably due to legacy reasons, but it should be time to switch to UTF8 encoding**. Dan Stangel said that in a [comment](https://github.com/fossology/fossology/blob/828d8d523edff3264e27e0eab00f3e71ed654247/src/testing/db/create_test_database.php#L286) within Fossology code about 7 years ago...

Probably the only issue would be for updates from previous installations, but in that case the old SQL_ASCII encoding may be kept.

What do you think?

GMishx:
Hello @alpianon,

I do changed the FOSSology DB encoding to UTF-8 (following [this blog post](http://blog.e-shell.org/134) on my development machine but now I am running into problems.

If you try to scan some binary files, I am trying with [NuGet.Client](https://github.com/NuGet/NuGet.Client) in particular some test dlls like https://github.com/NuGet/NuGet.Client/blob/dev/test/EndToEnd/Packages/InstallPackageRespectReferencesAccordingToDifferentFrameworks/files/RefPackage.1.0.0/lib/sl40/fear.dll, which contains text like `Copyright ©`. These texts are picked up by Copyright agent.

Now, since this file is binary and not text file, the data captured by the agent also contains some binary information. And if agent tries to insert it into UTF-8 DB, this binary string can not be stored as a `Text` in `copyright.content` column as it is not a valid UTF-8 text. Which results in data loss.

If you check the [character set support documentation](https://www.postgresql.org/docs/9.3/multibyte.html), it says:
> When the server character set is SQL_ASCII, the server interprets byte values 0-127 according to the ASCII standard, while byte values 128-255 are taken as uninterpreted characters.

[...]

PR #1528, opened 2019-11-16
fix(copyright): Remove non utf8 strings

GMishx:
## Description

Earlier versions of PostgreSQL used `SQL_ASCII` for default text encoding which stores everything as is and ignore everything that is not ASCII. New versions uses `UTF-8` as default encoding and rejects everything which is not UTF8 compatible.

This PR tries to solve that issue using [icu library](http://site.icu-project.org/home).

Read #1492 for more info.

Since copyright is the only agent which captures text from input file and store it in DB, this PR only removes non UTF-8 strings in copyright agents.

### Changes

1. Use `libicu` to clean matches in copyright agent.

## How to test

1. Install PostgreSQL with UTF-8 as default character encoding for FOSSology.
1. Install the branch and upload a package with many files which allows more chances of copyright being captured with binary strings in between.
1. Check the copyright agent logs from geeky scan view.
1. Check the copyright captured in copyright view.
1. Check if the copyrights appear correctly in reports.

## Notes

After working with ICU library, I realized that we can make the whole copyright agent unicode compatible since the beginning (the regex scan stage). This can be done in a separate PR.

I will open an issue for the same for further discussion before making the changes.

Probable fix for #1492


GMishx:
Package used for testing the branch: [NuGet.Client](https://github.com/NuGet/NuGet.Client)

ag4ums:
there seems to be a issue with copyrights as the number of copyrights are more after removing non UTF8 character,  need to find out what could be the issue.

GMishx:
Hello @ag4ums,

I've made some changes in the logic.
1. Earlier the data was read as a `signed char` (8 bit) which effectively have only 7 bit. Now the data will be processed as `unsigned char`.
1. Earlier there was only one check for UTF-8. Now if UTF-8 fails, there is additional check for UTF-16.

Can you please check with new patch?

PR #1531, opened 2019-11-22
chore(composer.json): updated symfony/dependency-injection version

ag4ums:
Signed-off-by: Anupam Ghosh <anupam.ghosh@siemens.com>

## Description

Should remove the below warning
image

### Changes

Updated Symfony/DependencyInjection version to 3.4.26.
and to resolve conflict updated the Symfoney/yaml  from 3.3.18 to  "symfony/yaml" : "v3.4.0"

## How to test


src/composer.json
```diff
@@ -22,8 +22,8 @@
         "slim/slim": "3.12.0",
         "symfony/config" : "v3.3.18",
         "symfony/http-foundation" : "v3.3.18",
```
GMishx: Please update the version for other symfony components as well.

ag4ums: good point ... 👍

PR #1532, opened 2019-11-25
fix(obligation): Associate all candidate licenses with same name

GMishx:
## Description

Associate all candidate licenses with same name to the obligations.

### Changes

1. Get the list of all licenses which shares the same name `ObligationMap`.
1. Associate all licenses with the given license while creating the obligation map in DB.

## How to test

1. Create a candidate license in `Group - 1`.
1. Create another candidate license in `Group - 2` with same shortname.
1. Create an obligation using `Group - 1` and associate the candidate license.
1. Edit the same obligation using `Group - 2`. The candidate license will create problems saving the obligation reliably.

Install the branch.

1. Repeat the steps above.
    1. Both candidate licenses should be associated with the obligation (no issue while updating the license in UI).
    1. Check if there are different entries for the candidate licenses in `obligation_candidate_map` table.
1. During clearing, select the candidate license and generate the report.
    - Check if the report contains the obligations.


shaheemazmalmmd:
Tested With duplicate candidate licenses. everything works fine. except that the duplicate Associated Licenses licenses are getting displayed in added licenses. Needs to create a issue.

PR #1541, opened 2019-12-02
fix(ojo): Handle dual-license and SPDX new naming

### LINKED ISSUE #1540: ojo output needs to be improved
### Description

1. Currently, if the file contains license info like
`SPDX-License-Identifier: License A <AND/OR> License B`
ojo only reports `License A` and `License B`. Since the file is dual licensed, ojo should report `Dual-license` as well like nomos.
1. In case the upload contains license `GPL-2.0-only` or `GPL-2.0-or-later`, since FOSSology default comes with `GPL-2.0` and `GPL-2.0+`, the logic to compare this from DB in ojo is not working correctly.

GMishx:
## Description

1. Report Dual-license if SPDX identifier contains or, and or with.
2. Normalize "-or-later" with "+" and "-only" with SPDX old and new naming conventions.

### Changes

1. Added new regex to find dual licenses.
1. Replace `std::basic_string::replace` with `std::basic_string::erase`.

## How to test

1. Upload a test file with SPDX License Identifiers.
1. Check if ojo is reporting "Dual-license" if identifier contains **or, and or with**.
1. Check if ojo is handling licenses with "-or-later" or "-only" at end (new SPDX naming scheme) with old naming scheme.
1. Check if ojo is reporting "Dual-license" for CLI  execution as well.

Closes #1540


shaheemazmalmmd:
While testing i have found that the OJO agent fails. please refer the attached screenshot.

screenShot1

ag4ums:
seems ojo is crashing
image

GMishx:
There is a new regex added in `src/ojo/agent/ojoregex.hpp`. For some reason, make does not acknowledge this change and rebuild everything.
If you do a `make clean-ojo build-ojo install-ojo`, it should work.

PR #1554, opened 2019-12-19
feat(unifiedReport): exclude scanner found copyrights of irrelevent f…

shaheemazmalmmd:
## Description

Exclude scanner found copyrights from report for irrelevant files.

### Changes

Get list of all irrelevant files from clearing decisions and remove copyrights by comparing uploadtreeId.

## How to test

* Upload a package
* Go to folder view and click on edit for any folder.
* Click on Mark files as irrelevant for that folder.
* Generate unified report and check copyrights the irrelevant files shouldn't appear.

### DIFF
--- a/src/lib/php/Dao/CopyrightDao.php
+++ b/src/lib/php/Dao/CopyrightDao.php
@@ -30,6 +30,8 @@ class CopyrightDao
   private $dbManager;
   /** @var UploadDao */
   private $uploadDao;
+  /** @var ClearingDao */
+  private $clearingDao;
   /** @var Logger */
   private $logger;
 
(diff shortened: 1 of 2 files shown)


src/lib/php/Dao/ClearingDao.php
```diff
@@ -872,7 +872,7 @@ public function removeMainLicense($uploadId, $groupId, $licenseId)
    */
   function getIrrelevantFilesFolder(ItemTreeBounds $itemTreeBounds, $groupId, $onlyCurrent=true)
   {
-    $statementName = __METHOD__;
+    $statementName = __METHOD__.rand();
```
GMishx: This is a bad practice to add a random string to a unique statement name.

The function `getRelevantDecisionsCte()` alters the value for statement name based on the parameters, I don't think there are additional changes required.

shaheemazmalmmd: Actually this is not required, Somehow pushed. thanks for pointing.

PR #1557, opened 2019-12-23
feat(migrate): Program to make file UTF-8 compatible

GMishx:
## Description

Old PostgreSQL with text collation `SQL-ASCII` will have some non-printable characters captured by copyright agent. While migrating to a new PostgreSQL DB with collation `UTF-8`, values in copyright table will cause a failure since they are not UTF-8 compatible.
This program will help to remove such characters from dump created by `pg_dump` and thus, it can be safely imported on new DB.

### Changes

1. Created a new CPP program to remove non UTF-8 characters from a file or a stream.

## How to test

1. On a database with `SQL-ASCII` collation, upload a package (tested with [Nuget.Client](https://github.com/NuGet/NuGet.Client) and run copyright agent.
1. Dump the database using `pg_dump` and import on a DB with `UTF-8` collation.
    1. Import will fail reporting invalid characters.
1. Use the program to remove such non UTF-8 characters and import the DB again.
    1. It should pass.

---

1. Check if the program can read from STDIN if no input file is provided.
1. Check if the program can output to STDOUT if no output file is provided.
1. Check if the program does not remove other valid UTF-8 characters from the file.
    - Example test file with valid characters: [non-ascii.txt](https://github.com/fossology/fossology/files/4122009/non-ascii.txt)

### DIFF
--- a/utils/FossologyUnicodeClean.cc
+++ b/utils/FossologyUnicodeClean.cc
@@ -1,0 +1,272 @@
+/*
+ * Copyright (C) 2019, Siemens AG
+ * Author: Gaurav Mishra <mishra.gaurav@siemens.com>
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * version 2 as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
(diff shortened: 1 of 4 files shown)


mcjaeger:
maybe try the following package:

https://github.com/unicode-org/icu/releases/tag/release-64-2

GMishx:
Will reuse the test file added in #1528 and add it to `.travis.yml` once #1528 is merged.

utils/FossologyUnicodeClean.cc
```diff
@@ -0,0 +1,272 @@
...
+ * @return UTF-8 valid string
+ */
+const icu::UnicodeString FossologyUnicodeClean::removeNonUtf(const string &input)
+{
+  int len = input.length();
+  const unsigned char *in = reinterpret_cast<const unsigned char *>(input.c_str());
+
+  icu::UnicodeString out;
+  for (int i = 0; i < len;) {
+    UChar32 c;
```
ag4ums: use meaningful variable name

ag4ums:
@GMishx , please rebase the branch

utils/FossologyUnicodeClean.cc
GMishx: Done

GMishx:
> 
> 
> @GMishx , please rebase the branch

Done, please check.

PR #1559, opened 2020-01-09
fix(view-license): Browse file without scanner ars

GMishx:
## Description

Packages without any scanners scheduled tends to fail while opening the
view-license since getLatestSuccessfulAgentIds() returns empty array.

### Changes

Check if getLatestSuccessfulAgentIds() returns empty array then fix the
SQL accordingly.

## How to test

1. Upload a package without scheduling any scanner.
1. Browse the package and open a single file.
    - The view will fail to load.
1. Install the branch and repeate the steps above.
    - The view should load successfully.
1. Add a user decision to any file, the clearing counters on top should
update accordingly.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -957,17 +957,17 @@ INSERT INTO clearing_decision (
     $scanJobProxy = new ScanJobProxy($GLOBALS['container']->get('dao.agent'), $uploadId);
     $scanJobProxy->createAgentStatus(array('nomos', 'monk', 'ninka', 'reportImport', 'ojo'));
     $latestAgentIds = $scanJobProxy->getLatestSuccessfulAgentIds();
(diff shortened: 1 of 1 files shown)


src/lib/php/Dao/ClearingDao.php
```diff
@@ -957,7 +957,11 @@ public function getTotalDecisionCount($uploadId, $groupId)
     $scanJobProxy = new ScanJobProxy($GLOBALS['container']->get('dao.agent'), $uploadId);
     $scanJobProxy->createAgentStatus(array('nomos', 'monk', 'ninka', 'reportImport', 'ojo'));
     $latestAgentIds = $scanJobProxy->getLatestSuccessfulAgentIds();
-    $agentIds = "ANY(VALUES(" . implode("),(", $latestAgentIds) . "))";
```
ag4ums: may be  ANY($1::int[]) is a better approach

GMishx: Done, please check.

PR #1560, opened 2020-01-10
feat(Decisions): add new decision type do not use

shaheemazmalmmd:
## Description

- Add new decision type **do not use**

### Changes

- Single file and file-tree contains new decision types "DO NOT USE".
- Change the color of clearing status for no_license_found of scanner finding to grey.
- List "DO NOT USE" cases in unified report.

## How to test

- End to end clearing with new decision type.


src/lib/php/Dao/ClearingDao.php
```diff
@@ -599,6 +599,15 @@ public function isDecisionTBD($uploadTreeId, $groupId)
     return ($latestDec['decision_type'] == DecisionTypes::TO_BE_DISCUSSED);
   }
 
+  public function isDecisionDNU($uploadTreeId, $groupId)
+  {
+    $sql = "SELECT decision_type FROM clearing_decision WHERE uploadtree_fk=$1 AND group_fk = $2 ORDER BY date_added DESC LIMIT 1";
```
GMishx: There is no index on `clearing_decision.date_added`. So it will be better to sort with `clearing_decision.clearing_decision_pk` as they both are related.

Also, will it be okay to break the lines to keep the 80 characters line limit?

src/lib/php/Report/LicenseDNUGetter.php
```diff
@@ -0,0 +1,89 @@
+<?php
+/*
+ Copyright (C) 2019, Siemens AG
```
GMishx: Please update the copyright year.

src/lib/php/Dao/ClearingDao.php
```diff
@@ -732,29 +741,39 @@ function getClearedLicenseIdAndMultiplicities(ItemTreeBounds $itemTreeBounds, $g
    * @param int $groupId
    * @param int $userId
    */
-  public function markDirectoryAsIrrelevant(ItemTreeBounds $itemTreeBounds,$groupId,$userId)
+  public function markDirectoryAsDecisionType(ItemTreeBounds $itemTreeBounds, $groupId, $userId, $decisionMark)
   {
-    $this->markDirectoryAsIrrelevantIfScannerDetected($itemTreeBounds, $groupId, $userId);
-    $this->markDirectoryAsIrrelevantIfUserEdited($itemTreeBounds, $groupId, $userId);
+    if ($decisionMark == "doNotUse") {
```
GMishx: Why not pass the internal type (`DecisionTypes::DO_NOT_USE`) instead of strings?

src/www/ui/async/AjaxExplorer.php
```diff
@@ -402,6 +402,17 @@ private function createFileDataRow($child, $uploadId, $selectedAgentId, $pfileLi
...
 
+    // override green/red flag with greenRed flag in case of single file with decision type "Do Not Use"
+    $isDecisionDNU = $this->clearingDao->isDecisionDNU($childUploadTreeId, $groupId);
+    $img = $isDecisionDNU ? 'redGreen' : $img;
+
+    // override green/red flag with grey flag in case of no_license_found scanner finding
+    if (!empty($licenseList) && empty($editedLicenseList)) {
+      $noLicenseFound = LicenseDao::NO_LICENSE_FOUND;
+      $splitLicense = str_split($licenseList, strlen($noLicenseFound));
+      $img = (strcmp("No_license_found", $splitLicense[0]) <= 0) ? 'grey' : $img;
```
GMishx: In my opinion, will be better to use `!== 0` to get the exact match `No_license_found`.

GMishx: Instead of hard coding `"No_license_found"`, the variable `$noLicenseFound` can be reused.

src/lib/php/Dao/ClearingDao.php
```diff
@@ -732,29 +741,39 @@ function getClearedLicenseIdAndMultiplicities(ItemTreeBounds $itemTreeBounds, $g
...
    * @param ItemTreeBounds $itemTreeBounds
    * @param int $groupId
    * @param int $userId
    */
-  public function deleteIrrelevantDecisionsFromDirectory(ItemTreeBounds $itemTreeBounds,$groupId,$userId)
+  public function deleteDecisionTypeFromDirectory(ItemTreeBounds $itemTreeBounds, $groupId, $userId, $decisionMark)
   {
-    $this->markDirectoryAsIrrelevantIfScannerDetected($itemTreeBounds, $groupId, $userId, true);
-    $this->markDirectoryAsIrrelevantIfUserEdited($itemTreeBounds, $groupId, $userId, true);
+    if ($decisionMark == "doNotUse") {
```
GMishx: As per `src/www/ui/change-license-processPost.php`, the string sent is `deleteDoNotUse`. Please check.

src/www/ui/async/AjaxExplorer.php
GMishx: `strpos` will be better here instead of using `str_split` as you may get the required string split in between to two different elements.

GMishx:
Did a small testing, have following remarks:

    - snap1
    - If the job finish successfully, the functionality works fine.
1. The red-green icon does not show up if folder contains "Do not use" files. It is still green. (Is it intended behavior?)
1. Everything else (Change concluded License & Unified report looks good).

shaheemazmalmmd:
> Did a small testing, have following remarks:
> 
> 1. Changing license using `[Edit]` on **License Browser**. The job failed due to scheduler error (ajax call response has `end_bits:2`), got following error in my browser and page did not reload (licenses were applied to the files):

    Please create a ticket as this behavior exists in master.

> 2. The red-green icon does not show up if folder contains "Do not use" files. It is still green. (Is it intended behavior?)

It is an intended behavior.

GMishx:
Created issue https://github.com/fossology/fossology/issues/1595

Thank you for feedback @shaheemazmalmmd

NicolasToussaint:
Hi @shaheemazmalmmd ; What is the exact intent of the new "decision type" ?

I don't really understand the tooltip, and didn't find info elsewhere:
image

I would suspect that the files should be ignored altogether (for not distributed files for ex.), is that correct ?
Are these files included in the reports ?

shaheemazmalmmd:
Hello @NicolasToussaint 

Actually the new Decision type is to mark file which should not be used in the source code. ex: dll files, some jar files etc.

"Do not use" decision type will be listed in the unified report in separate section. With comments.

Actually Tool-tips needs to be updated with specific definitions. as the current definitions are too generic.

Thanks

PR #1566, opened 2020-01-13
fix(bulk): Fix dropdown bulk on folder level

GMishx:
## Description

The dropdown in modal view for bulk option in aggregated license view was not rendering.

This PR uses `dropdownParent` property of `select2` to fix the issue.

### Changes

1. Change the `dropdownParent` of any `ui-render-select2` element which has a parent class `.modal`.
1. Change the call to `tooltip` class from `$(document)` to `$('img')`.

## How to test

1. Upload a package and goto aggregated license browser.
1. Select `[Bulk]`.
    1. Check if the dropdown renders properly.
1. Check if other occurances of `select2` are not effected.

### DIFF
--- a/src/lib/php/UI/template/include/foot.html.twig
+++ b/src/lib/php/UI/template/include/foot.html.twig
@@ -10,9 +10,16 @@
 <script type="text/javascript">
   function renderSelect2() {
     if(!$('.ui-render-select2').attr("size") > 0) {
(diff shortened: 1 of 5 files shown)


shaheemazmalmmd:
Tested : when i clicked on clean text in bulk view of tree view it throws below error. 
Please fix. 

screenShot4

PR #1568, opened 2020-01-14
feat(rest): Add group context (groupName param) for REST Api calls

### LINKED ISSUE #1529: Rest API: calling API in given groupId context
### Description

Im not sure if it is a bug or just a behaviour, so I report this as feature request

# Intro
Groups in Fossology work rather like context than standard group of users.
In the GUI, User has to change Group context (upper right corner of page) to do things for uploads he has access because of being in specific group, while "standard" group behaviour should always show and allow user to do things with all uploads he has access to within all his groups.

# Problem

Because of this Group behavior, there is a problem when calling REST API.
In general, REST API calls uses (stores session parameters) always with GroupID set as "default" in *users* table (group_fk column).
The thing is, this groupId stored in database changes each time when given user changes "selectMemberGroup" item on GUI

Few weeks ago, @GMishx solved #1468 issue, and added option to store new upload for specific groupId given as REST call parameter.
It works ok, but there is a problem when:
- user has 2 groups access: group A and B
- user on GUI selected group **B** (upper-right select box), so this value is stored in DB
- user calls REST API (with his own token) and creates upload sending groupID pointing to group **A** (he has access, so he can) and this upload ic created in group **A** context
- user tries to run jobs via REST API (his own token) and he **can not** (no access error) - because /jobs endpoint uses groupId stored in database, which points to group **B** selected in GUI 

In my oppinion, there should be option for all REST API calls to pass specific groupId for which you want to do stuff 

In fact, I did some research, and if this explanation is clear for You and you also think there is a bug to solve, I have a proposal :)

## Resolve proposal

In RestAuthMiddleware class, there is a method `$authHelper->updateUserSession` call that sets current Session parameters. This method uses `$row = $this->userDao->getUserAndDefaultGroupByUserName($user["name"]);` call that reads user default groupId from DB.

In my oppinion, we should change this method and when someone sends groupId in request, we should update session with groupId from request (if there is no groupId provided, we could still use this value from DB)

I think this change would also be valid for /upload endpoint, so changes done by @GMishx in #1498 would be no longer needed - but im not 100% sure.

**To be honest, i made some work on my side to resolve this issue as described above (as we are going to use this REST API for our CI/CD intergation and we need this to work correctly ASAP) so I think I can prepare PullRequest for your verification if you accept this way of doing it**

By the way, I think it would be nice to add also groupName parameter, as GroupId is quite hard to find for standard GUI user

piotreke12:
## Description

PR adds option to provide groupName for REST  API calls, to do stuff in given groupName context (not the one set in DB as user-default group)

Fixes #1529 , while previous PR #1498 will no longer be needed.

### Changes

REST API accepts additional parameter (groupName) which changes behaviour only for subset of possible paths (please see openapi.yaml file)

RestAuthMiddleware verifies given group existence and user access - if user has no access to given group or group does not exist, error is returned.
In case of positive verification, AuthHelper.updateUserSession method is called with proper paremetrs, to alter and set correct group for specific user session

## How to test

Try to call REST APi with/without additional groupName parameter to verify if things are done for proper group (passed or not by user), using also group that does not exist or group that user has no access to.

### DIFF
--- a/src/www/ui/api/Middlewares/RestAuthMiddleware.php
+++ b/src/www/ui/api/Middlewares/RestAuthMiddleware.php
@@ -70,8 +70,21 @@ class RestAuthMiddleware
         $tokenValid = new Info(403, "Do not have required scope.", InfoType::ERROR);
       }
       if ($tokenValid === true) {
-        $authHelper->updateUserSession($userId, $tokenScope);
-        $response = $next($request, $response);
+        $groupName = "";
+        $groupName = strval($request->getHeaderLine('groupName'));
+        if (!empty($groupName)) { // if request contains groupName
+          $userHasGroupAccess = $authHelper->userHasGroupAccess($userId, $groupName);
+          if ($userHasGroupAccess === true) {
+            $authHelper->updateUserSession($userId, $tokenScope, $groupName);
+            $response = $next($request, $response);
+          } else { // no group access or group does not exist
+            $response = $response->withJson($userHasGroupAccess->getArray(),
+            $userHasGroupAccess->getCode());
+          }
+        } else { // no groupName passed, use defult groupId saved in DB
+          $authHelper->updateUserSession($userId, $tokenScope);
+          $response = $next($request, $response);
+        }
       } else {
         $response = $response->withJson($tokenValid->getArray(),
           $tokenValid->getCode());
--- a/src/www/ui/api/Helper/AuthHelper.php
+++ b/src/www/ui/api/Helper/AuthHelper.php
@@ -186,13 +186,18 @@ class AuthHelper
    *
    * @param int    $userId User id from the JWT.
    * @param string $scope  Scope of the current token.
+   * @param string $groupName  Name of the group to update session with.
    * @sa updateSession()
    */
-  public function updateUserSession($userId, $scope)
+  public function updateUserSession($userId, $scope, $groupName = null)
   {
     $authPlugin = $GLOBALS["container"]->get("helper.restHelper")->getPlugin('auth');
     $user = $this->dbHelper->getUsers($userId)[0];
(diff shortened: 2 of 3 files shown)


src/www/ui/api/Middlewares/RestAuthMiddleware.php
```diff
@@ -70,8 +70,21 @@ public function __invoke($request, $response, $next)
         $tokenValid = new Info(403, "Do not have required scope.", InfoType::ERROR);
       }
       if ($tokenValid === true) {
-        $authHelper->updateUserSession($userId, $tokenScope);
-        $response = $next($request, $response);
+        $groupName = "";
+        $groupName = strval($request->getHeaderLine('groupName'));
+        if (!empty($groupName)) { // if request contains groupName
+          $userHasGroupAccess = $authHelper-> userHasGroupAccess($userId, $groupName);
```
GMishx: Just a little extra space.
```suggestion
          $userHasGroupAccess = $authHelper->userHasGroupAccess($userId, $groupName);
```

piotreke12:
Hello @GMishx - thank for Your comments.

- you are right, I somehow missed that /reports are also group dependent - I corrected it
- extra space deleted
- API descriptor version updated - unfortunately there are conflicts as in the meantime version changed

Is it possible that You will resolve this conflict or I have to do it somehow?

GMishx:
Thank you for the changes @piotreke12. I think it would be better if you resolve the conflicts. And the way we resolve them is by rebasing the branch to master and force push the commit again.

Would it be okay for you to do it?

piotreke12:
Hi @GMishx - thanks for explanation.

I've rebased my feature branch - I hope it is ok now :)

GMishx:
Created revert PR #1609. Will merge this PR after the merger of revert branch.

PR #1576, opened 2020-01-19
perf(ui): Reduce load time for license browser

GMishx:
## Description

The PR tries to improve the load time of tree view by optimizing some SQL queries, reducing calls to DB and introducing new indexes.

### Changes

1. Update CTE provided by `ClearingDao::getRelevantDecisionsCte()` to sort and unique only if required and reduce CTE count from 2 to 1 (will benefit PG < 12).
2. Remove `getClearingDecisionsCount()` and `getClearingDecisionsCount()` as they are no longer helpful.
3. Update `UploadTreeProxy::getQueryCondition()` to use simple SORT and DISTINCT instead of WindowAgg.
4. In `AjaxExplorer`, new function `updateFilesToBeCleared()` to cache itemtree bounds which are already calculated to reduce DB reads.
5. Updated index `clearing_decision_event_clearing_fk_idx` to include `clearing_event_fk` making join on cde as `INDEX ONLY SCAN`.
6. New index `clearing_event_pk_removed_idx` to optimize ajax view query.

## How to test
1. Upload a package.
1. Do some clearing with both Global and Local decisions.
1. Upload the same package again.
    1. Check if global decisions are reflected.
    1. Previous local decisions are not effected.
    1. New local decisions on the package does not effect previous package.
    1. New global decisions on new package are reflected on previous package.
1. Check that **to be discussed** decisions are not counted while finding the number of files cleared in single file view.
1. Overall significant improvement on load time.
1. Folders with single file (example `Folder/Folder/File`) should show respective license names and clearing status as they are expanded by License Browser.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -71,41 +71,40 @@ class ClearingDao
     $sql_upload = "";
     if ('uploadtree' === $uploadTreeTable || 'uploadtree_a' === $uploadTreeTable) {
       $params[] = $itemTreeBounds->getUploadId(); $p = "$". count($params);
-      $sql_upload = "ut.upload_fk=$p AND ";
+      $sql_upload = " AND ut.upload_fk=$p";
     }
     if (!empty($condition)) {
       $statementName .= ".(".$condition.")";
-      $condition .= " AND ";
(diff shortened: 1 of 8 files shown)


shaheemazmalmmd (approved):

PR #1597, opened 2020-02-03
fix(apache): Enable fossology on source install

### LINKED ISSUE #182: Apache configuration
---

Author Name: **Alex Norton** 
Original Redmine Issue: 3003, http://www.fossology.org/issues/3003
Original Date: 2012/06/20

---

The way that we instruct people to change the Apache configuration for FOSSology during a source install is very hard and prone to error. Our current method is found in install/INSTALL.

I propose changing it to adding a symlink to the fossology url configuration file in the apache2/conf.d directory. This has been a much more consistent and easier to perform method of Apache configuration that what we currently use.

GMishx:
## Description

Enable fossology site on source install.

### Changes

1. Check if the default location for Debian based distro exists, then
   install conf file Debian style. Otherwise check for RedHat directory
   and install RedHat style.

## How to test

1. Do a clean install of FOSSology from source on Debian and RedHat
   based Linux Distros.
1. Check if FOSSology (`http://localhost/repo`) is enabled.

Closes #182

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ VERSION
 .vagrant/
 *.orig
 .#*
+tags
 
 # compiled or derived files
 install/db/dbcreate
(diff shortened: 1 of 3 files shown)


ag4ums:
code looks good,
working fine with debian,
need to check with redhat system

sjha2048:
tested on Pop Os! 19.10, working fine.

ag4ums:
Checked with centos,

PR #1598, opened 2020-02-03
fix(rest): Remove hostname from JWT

### LINKED ISSUE #1512: Why is the host encoded in the JwtToken?
The corresponding code is:
https://github.com/fossology/fossology/blob/23a2687cf81f565ec76a225dab5090afe261c657/src/www/ui/api/Controllers/AuthController.php#L124-L126
and
https://github.com/fossology/fossology/blob/23a2687cf81f565ec76a225dab5090afe261c657/src/www/ui/api/Helper/AuthHelper.php#L211-L222

# My Problem
In my Setup I have fossology running via docker compose and some client running in another container. And I want to setup a connection between these two. If I go to FOSSology on `localhost` as usual, I can only generate a token which can not be used by the client. This token is only valid for clients, which access a fossology on localhost.

To generate a working token, I have to go to `172.17.0.2:8081/repo/...` and generate a token, since `172.17.0.2` is the IP used by my client to connect to FOSSology.

## But the problem gets more annoying: 
The IP space `172.17.0.*` might be used by another setup, and docker might choose to setup everything on `172.18.0.*` and the token is now invalid again.

# My Question:

- What is the reason for having the hostname in the token?

It might be good to add a help message to the application, which explains that.

GMishx:
## Description

Removed hostname from JWT token as requested in #1512

### Changes

1. Removed hostname from all token related functions
1. Removed hostname check while authenticating with token.

## How to test

1. Generate a new token from UI and REST API
    - Inspect it using jwt.io and verify if `iss` and `aud` claims are not in the token.
1. Use old token to access FOSSology (with `iss` and `aud`)
    - Should work without any issue
1. Use new token to access FOSSology (without `iss` and `aud`)
    - Should work without any issue

Closes #1512

### DIFF
--- a/src/www/ui/api/Controllers/AuthController.php
+++ b/src/www/ui/api/Controllers/AuthController.php
@@ -122,8 +122,7 @@ class AuthController extends RestController
           }
           if (isset($jti['jti']) && ! empty($jti['jti'])) {
             $theJwtToken = $this->restHelper->getAuthHelper()->generateJwtToken(
-              $request->getUri()->getHost(), $expire, $jti['created_on'],
-              $jti['jti'], $scope, $key);
+              $expire, $jti['created_on'], $jti['jti'], $scope, $key);
             $returnVal = $response->withJson([
               "Authorization" => "Bearer " . $theJwtToken
             ], 201);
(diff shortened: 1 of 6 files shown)


ag4ums:
code looks good

GMishx:
> code looks good

Can you please give this comment as an approval?

PR #1602, opened 2020-02-05
fix(scheduler): Check if max_run is -1 (unlimited)

GMishx:
## Description

With PR #1305, a check for max agent run was set.
With the change, if agent has max limit as -1 (unlimited), the scheduler will not run another instance of the agent which is not expected.

### Changes

Check if max_run = -1 in `isMaxLimitReached`.

## How to test

1. Schedule multiple instances of agent with `max = -1` in agent's conf file.


GMishx:
Needs a different approach. Will work on it later.

PR #1604, opened 2020-02-10
fix(reuser): Run decider after reuser

GMishx:
## Description

If an upload is done with reuse but the new upload has new findings, the reuser simply copied the previous decision ignoring the conflicts.

### Changes

1. Add scanner dependency to reuser.
2. Add reuser as dependency to decider if monkbulk decisions are being reused.

## How to test

1. On master branch. upload a package and do clearing (both manual and with monkbulk).
1. From the previous upload, copy a text portion of a file and make it as a new license so monk will add new license for next upload.
1. Install this branch to change the agent IDs to include newly created license.
1. Reupload the package but reusing the previously cleared package.
    1. Upload with "bulk phrases from reused packages"
    1. Upload without "bulk phrases from reused packages"
1. Check if files with new license findings (new license added in step 2) have WIP or no decisions.

### DIFF
--- a/src/decider/ui/DeciderAgentPlugin.php
+++ b/src/decider/ui/DeciderAgentPlugin.php
@@ -101,6 +101,7 @@ class DeciderAgentPlugin extends AgentPlugin
         case 'reuseBulk':
           $dependencies[] = 'agent_nomos';
           $dependencies[] = 'agent_monk';
+          $dependencies[] = 'agent_reuser';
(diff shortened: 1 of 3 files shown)


shaheemazmalmmd:
@GMishx Can you please rebase the branch with master ?

GMishx:
@shaheemazmalmmd , rebased. Please check

PR #1617, opened 2020-02-18
feat(unifiedReport): include DNU information in assesment summary if …

shaheemazmalmmd:
## Description

include Do not use information in assessment summary of report, If Do not use files found.

## How to test

- Upload a component.
- Mark some files do not use from clearing view.
- Check the link in assessment summary of unified report with a internal link.

### DIFF
--- a/src/unifiedreport/agent/unifiedreport.php
+++ b/src/unifiedreport/agent/unifiedreport.php
@@ -298,6 +298,7 @@ class UnifiedReport extends Agent
 
     $otherStatement = $this->otherGetter->getReportData($uploadId);
     $this->heartbeat(empty($otherStatement) ? 0 : count($otherStatement));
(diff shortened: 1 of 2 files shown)


src/unifiedreport/agent/unifiedreport.php
```diff
@@ -791,6 +792,12 @@ private function writeReport($contents, $uploadId, $groupId, $userId)
 
     /* Display Do not use license files */
     $heading = "Do not use Files";
+    // adding an internal bookmark
+    $rightColStyleBlue = array("size" => 11, "color" => "0000A0", 'underline' => 'single');
```
GMishx: Can you please change the variable name?

shaheemazmalmmd: Done. Thank you

PR #1612, opened 2020-02-16
fix(spasht): Made required UI changes

vivekaindia:
## Description

Made Useful UI changes and Fixed ui bugs

### Changes

1. Fixed spasht packaging
2. Fixed MicroMenu
3. Fixed UI Bugs
4. Fixed Advance Search View
5. Removed Liccense Tab

## How to test

`sudo make install`

PR #1622, opened 2020-02-19
Fix/spasht/advance search

vivekaindia:
## Description

Added Functionality to Advance Search.
Advance search is designed in such a way that all the fields are not necessary.
Even by entering just extra field will give you the results based on the requirements.

### Changes

1.  Added Advance search feature (method).
2.  Fixed UI bugs.
3.  Fixed UI response, when no result is found based on advance search.

## How to test

`sudo make install`


GMishx:
Cherry-picked in #1612
Closing here

PR #1651, opened 2020-03-06
feat(DB): Recode copyright tables to UTF-8

### LINKED ISSUE #1492: Change DB encoding from SQL_ASCII to UTF8

GMishx:
## Description

Current encoding FOSSology DB is `SQL_ASCII`. This PR adds a migration script to recode the contents of `copyright`, `keyword` and `ecc` tables to `UTF-8`.

**Note:** This PR also updates the encoding of database to `UTF-8`.

### Changes

1. Add a new migration script to recode `copyright`, `keyword` and `ecc` table contents.
1. Move the UTF-8 filter logic to CPP library.
1. Move the `fo_unicode_clean` to `src/copyright/agent`.
1. Do not run the script if total number of records are `> 500,000` (~15 minutes on 4 vCPU, 16GB vRAM)

## How to test

1. On an old FOSSology, upload files with invalid copyright texts (see #1528 and #1557).
1. Install the branch and run `fo-postinstall`.
    1. The encoding of the DB should be changed to `UTF-8`.
    1. All invalid characters should be removed from the tables.
1. Check if the support script works (`/usr/local/lib/fossology/dbmigrate_change_db_encoding.php`).

Closes #1492


install/db/dbmigrate_3.7-3.8.php
```diff
@@ -0,0 +1,322 @@
...
+  "copyright",
+  "ecc",
+  "keyword"
+);
+
+/**
+ * Calculate the total number of records in concerned tables
+ * @param DbManager $dbManager
+ */
+function calculateNumberOfRecordsInCopyright($dbManager)
```
shaheemazmalmmd: Can you change the naming as the count you get is not specific to only copyright table.

install/db/dbmigrate_3.7-3.8.php
```diff
@@ -0,0 +1,322 @@
...
+ * @return int Count of updated rows
+ */
+function updateRecodedValues($dbManager, $table, $rows)
+{
+  if (empty($rows)) {
+    return 0;
+  }
+
+  $values = array();
+  foreach ($rows as $id => $content) {
```
shaheemazmalmmd: can we use Array_map here ??

install/db/dbmigrate_3.7-3.8.php
GMishx: But the function is specific to copyright and sister tables.

Do you have any suggestions for it?

install/db/dbmigrate_3.7-3.8.php
GMishx: I don't think that will be necessary as this is just 2 lines of code and `array_map` requires a callback function to work on.

Also, to pass the keys, I will have to call `array_keys` to get the keys which will have it's own overhead.

install/db/dbmigrate_3.7-3.8.php
shaheemazmalmmd: :D . some thing like calculateNumberOfRecordstablestoencode ?

install/db/dbmigrate_3.7-3.8.php
shaheemazmalmmd: Actually inbuilt functions will be faster compared to loops. can improve speed a bit.

install/db/dbmigrate_3.7-3.8.php
GMishx: Done, please check.

install/db/dbmigrate_3.7-3.8.php
GMishx: Done, please check.

GMishx:
@shaheemazmalmmd done. Please check.

shaheemazmalmmd:
@GMishx :  

Can you please fix the below issue.

```
PHP Warning:  pg_execute(): Query failed: ERROR:  invalid byte sequence for encoding "UTF8": 0xa9 in /usr/share/fossology/lib/php/Db/Driver/Postgres.php on line 70
[2020-03-17 18:48:07] /usr/share/fossology/lib/php/libschema.php.CRITICAL: ERROR:  invalid byte sequence for encoding "UTF8": 0xa9 [] []
*** Something went wrong. Try again! ***
Something went wrong. Try running postinstall again!
PHP Fatal error:  Uncaught Exception: too much transaction rollbacks in /usr/share/fossology/lib/php/Db/DbManager.php:82
Stack trace:
#0 /usr/lib/fossology/dbmigrate_3.7-3.8.php(320): Fossology\Lib\Db\DbManager->rollback()
#1 /usr/lib/fossology/fossinit.php(399): Migrate_37_38(Object(Fossology\Lib\Db\ModernDbManager), '/usr/share/foss...')
#2 {main}
  thrown in /usr/share/fossology/lib/php/Db/DbManager.php on line 82

```

ag4ums:
Seems that PR require some additional steps as re-encoding is failing during installation

image

mcjaeger:
(from discussion) mayber for this, user needs really to acknowledge that something is wrong. Proposed idea is to interrup the user flow and let the user type a letter in order to continue. The better proposal would be to have a seperate script which switches the user from `fossy` to `postgresql` in order to change the setting to `UTF-8` encoding

GMishx:
I have updated the code to do a `su` to `postgres` user before making changes to `pg_database`.

install/db/dbmigrate_3.7-3.8.php
```diff
@@ -0,0 +1,335 @@
...
+ * Check if the DB encoding is SQL_ASCII, change it to UTF-8
+ * @param DbManager $dbManager
+ */
+function updateDbEncoding($dbManager)
+{
+  $dbName = $GLOBALS["SysConf"]["DBCONF"]["dbname"];
+  $sql = "SELECT pg_encoding_to_char(encoding) AS encoding " .
+    "FROM pg_database WHERE datname = '$dbName';";
+  $row = $dbManager->getSingleRow($sql);
+  if (strcasecmp($row["encoding"], 'sql_ascii') == 0) {
```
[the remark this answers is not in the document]
GMishx: Done. Please check.

shaheemazmalmmd:
@GMishx : can you please fix this notice .
```
PHP Notice:  Undefined variable: dbName in /usr/local/lib/fossology/dbmigrate_3.7-3.8.php on line 230
PHP Stack trace:
PHP   1. {main}() /usr/local/lib/fossology/fossinit.php:0
PHP   2. Migrate_37_38() /usr/local/lib/fossology/fossinit.php:399
PHP   3. recodeTables() /usr/local/lib/fossology/dbmigrate_3.7-3.8.php:336
PHP   4. updateDbEncoding() /usr/local/lib/fossology/dbmigrate_3.7-3.8.php:315
```

shaheemazmalmmd:
@GMishx Please create a wiki for conversion of non-UTF-8 to UTF-8 through DB dump ??

shaheemazmalmmd:
Code looks good. Tested and works as expected.

GMishx:
> @GMishx Please create a wiki for conversion of non-UTF-8 to UTF-8 through DB dump ??

@shaheemazmalmmd Here is the wiki: https://github.com/fossology/fossology/wiki/Migration-to-UTF-8-DB

PR #1658, opened 2020-03-11
fix(git) : add php.ini to gitignore

adityabisoi:
## Description

Add generated php.ini to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ VERSION
 *.orig
 .#*
 tags
+php.ini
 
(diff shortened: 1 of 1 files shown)


GMishx:
Thank you for your contribution @adityabisoi. Can you please update your commit message to be more inline with the [contributing guidelines](https://github.com/fossology/fossology/blob/master/CONTRIBUTING.md)?

adityabisoi:
I've updated the commit message. Thanks.

PR #1662, opened 2020-03-13
fix(AdviceLicense):show error message on failed merge

adityabisoi:
## Description

Fix white page showing when **merge** is clicked, when advice license and a main license have the same license text.

## Changes

An error message stating `the license text already exists` is shown if the above case occurs.

### DIFF
--- a/src/www/ui/page/AdminLicenseCandidate.php
+++ b/src/www/ui/page/AdminLicenseCandidate.php
@@ -134,7 +134,12 @@ class AdminLicenseCandidate extends DefaultPlugin
         $vars['message'] = 'Short name must be unique';
         break;
       case 'merge':
-        $ok = $this->mergeCandidate($rf,$suggest,$vars);
+        try {
+          $ok = $this->mergeCandidate($rf,$suggest,$vars);
+        } catch (\Throwable $th) {
+          $vars['message'] = 'The license text already exists.';
+          break;
+        }
         if ($ok) {
           $vars = array(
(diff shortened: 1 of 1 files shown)


adityabisoi:
@GMishx Please review and suggest any changes. Thanks.

GMishx:
Merge always results in failure. After some more digging, I found that the code to merge license was referring to wrong table.
Can you please apply the following patch to fix it:
```diff
diff --git a/src/www/ui/page/AdminLicenseCandidate.php b/src/www/ui/page/AdminLicenseCandidate.php
index 326730776..b245c26d9 100644
--- a/src/www/ui/page/AdminLicenseCandidate.php
+++ b/src/www/ui/page/AdminLicenseCandidate.php
@@ -245,7 +245,7 @@ class AdminLicenseCandidate extends DefaultPlugin
     /** @var DbManager */
     $dbManager = $this->getObject('db.manager');
     $tableColumnMap = array("license_file"=>"rf_fk",
-        "license_ref_bulk"=>"rf_fk",
+        "license_set_bulk"=>"rf_fk",
         "clearing_event"=>"rf_fk");
     foreach ($tableColumnMap as $table=>$column) {
       $dbManager->prepare($stmt=__METHOD__.".$table","UPDATE $table SET $column=$1 WHERE $column=$2");
```

GMishx:
Thank you for pushing the changes @adityabisoi , but can you please update the commit message of your last commit to be compliant with the contributing guidelines?
Since the changes are small, it would be better if you can squash the two commits as well.

PR #1666, opened 2020-03-16
feat(ui): Conversion of timestamps to browser timezone

Dany-T:
## Description

This pull request adds the capability to convert the timestamps shown in the FOSSology WebUI to the current timezone of the users browser. This makes it easier for the users to track when which action was executed without manually converting the time.

### Changes

**src/lib/php/UI/template/components/login-form.html.twig** - get the current users timezone during login
**src/lib/php/common-ui.php** - added function "Convert2BrowserTime" to be called wherever in the UI a timestamp is being displayed
**src/www/ui/core-auth.php** - if no timezone is provided by the browser it falls back to the server timezone

The remaining changed files are now calling the function "Convert2BrowserTime" for displaying the timestamps.

## How to test

After deployment of the application the user logging in to the application should see all timestamps in his current timezone instead of whichever timezone is currently configured on the server hosting FOSSology.


shaheemazmalmmd:
@d4nnywh0 :  Please rebase this branch and remove merge commits from it ?

Dany-T:
@shaheemazmalmmd : Sorry for that I will create a new PR with the cleaned up history.

PR #1669, opened 2020-03-20
style(images): A unused image is removed

Kaushl2208:
## Description

There was an unused image present in the folder.

Screenshot from 2020-03-20 15-16-04

## How to test

No test required.


Kaushl2208:
@sjha2048 Nice catch! Wait I will update the PR,

mcjaeger:
Hello, I am not sure, I thought it was easter egg from the days of fossology 1.0 I have seen the image too and did not remove it because I understood it as such. I think there is some text somewheresaying that license scanning is like putting a lipstick on an animal ...

PR #1673, opened 2020-03-26
feat(globalDecision): show warning if the candidate license is added

shaheemazmalmmd:
## Description

* check if the candidate license is added to the list 
* if added the show message

## How to test

* Upload a package X.
* Create a candidate license.
* Add the candidate license to license table in single file view through user decision.
* Check "Apply decision to all future occurrences of this file" option.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -333,6 +333,10 @@ class ClearingDao
    */
   public function createDecisionFromEvents($uploadTreeId, $userId, $groupId, $decType, $scope, $eventIds)
   {
+    if ( ($scope == DecisionScopes::REPO) &&
+         !empty($this->getCandidateCountForItemId($uploadTreeId))) {
+      throw new \Exception( _("Cannot add candidate license as global decision\n") );
+    }
     $this->dbManager->begin();
 
     $this->removeWipClearingDecision($uploadTreeId, $groupId);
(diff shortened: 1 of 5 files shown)


src/lib/php/Dao/ClearingDao.php
```diff
@@ -1022,4 +1022,24 @@ public function getTotalDecisionCount($uploadId, $groupId)
     $foundCounter = $this->dbManager->getSingleRow($sql, $params, $statement);
     return $foundCounter['cnt'];
   }
+
+  /**
+   * @param uploadTreeId
+   * @return count
+   */
+  public function getCandidateCountForItemId($uploadTreeId)
```
GMishx: Can you please add this check to `createDecisionFromEvents()` as well so we can make sure the check happens at backend as well?

https://github.com/fossology/fossology/blob/b5a389be8ef413fc488b56700ea23bdf3ff9c73a/src/lib/php/Dao/ClearingDao.php#L334

GMishx: Also for `markDirectoryAsDecisionType()`

https://github.com/fossology/fossology/blob/b5a389be8ef413fc488b56700ea23bdf3ff9c73a/src/lib/php/Dao/ClearingDao.php#L753

src/lib/php/Dao/ClearingDao.php
```diff
@@ -333,6 +333,10 @@ public function removeWipClearingDecision($uploadTreeId, $groupId)
    */
   public function createDecisionFromEvents($uploadTreeId, $userId, $groupId, $decType, $scope, $eventIds)
   {
+    if ( ($scope == DecisionScopes::REPO) &&
+         !empty($this->clearingDao->getCandidateCountForItemId($itemId))) {
```
GMishx, line 337:
```suggestion
         !empty($this->getCandidateCountForItemId($uploadTreeId))) {
```

shaheemazmalmmd: Oops. Missed it copy .

GMishx:
@shaheemazmalmmd , can you please modify the commit message so that we don't get the weird `...` at the end?

PR #1553, opened 2019-12-19
fix(libschema): Schema fix for PostgreSQL 12

GMishx:
## Description

While installing FOSSology on PostgreSQL 12, following error was thrown:
```
PHP Warning:  pg_prepare(): Query failed: ERROR:  column attrdef.adsrc does not exist
LINE 7:         attrdef.adsrc AS default,
                ^ in /usr/local/share/fossology/lib/php/Db/Driver/Postgres.php on line 60
[2019-12-19 16:00:22] /usr/local/share/fossology/lib/php/libschema.php.CRITICAL: ERROR:  column attrdef.adsrc does not exist LINE 7:         attrdef.adsrc AS default,                 ^ [] []
PHP Fatal error:  Uncaught Fossology\Lib\Exception: error executing: SELECT class.relname AS table,
        attr.attnum AS ordinal,
        attr.attname AS column_name,
        type.typname AS type,
        attr.atttypmod-4 AS modifier,
        attr.attnotnull AS notnull,
        attrdef.adsrc AS default,
        col_description(attr.attrelid, attr.attnum) AS description
      FROM pg_class AS class
      INNER JOIN pg_attribute AS attr ON attr.attrelid = class.oid AND attr.attnum > 0
      INNER JOIN pg_type AS type ON attr.atttypid = type.oid
      INNER JOIN information_schema.tables AS tab ON class.relname = tab.table_name
        AND tab.table_type = 'BASE TABLE'
        AND tab.table_schema = 'public'
      LEFT OUTER JOIN pg_attrdef AS attrdef ON adrelid = attrelid AND adnum = attnum
      ORDER BY class.relname,attr.attnum -- fo_libschema::addTables

ERROR:  column attrdef.adsrc does not exist
LINE 7:         attrdef.adsrc AS default,
                ^ in /usr/local/share/fossology/lib/php/Db/DbManager.php:144
Stack tr in /usr/local/share/fossology/lib/php/Db/DbManager.php on line 144
```

## Changes
1. Use information_schema.columns (available since PG 7.4) to fetch the required details.
~1. Update the datatype in `core-schema.dat`.~

PR #1674, opened 2020-03-30
fix(postgresql): libschema to adopt Postgresql 12 changes

heliocastro:
Postgresql API interface changed after release 12 breaking ad column
name. Extended the Driver base class to provide db version.

Signed-off-by: Helio Chissini de Castro <helio.chissini-de-castro@bmw.de>

## Description

TEst postgres version and use the proper string for ad column.

### Changes

* Add version to Driver.php abstract class
* Add version to Postgres.php class
* Modified libschema.php to use the proper entrry

## How to test

Do fossology base installation with Postgresql release 12 or later

### DIFF
--- a/src/lib/php/Db/Driver.php
+++ b/src/lib/php/Db/Driver.php
@@ -49,6 +49,11 @@ interface Driver
   public function isConnected();
 
   /**
+   * @return int
+   */
+  public function version();
+  
+  /**
    * @return string
    */
   public function getLastError();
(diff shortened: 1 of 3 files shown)


GMishx:
Hello @heliocastro ,

Thank you for your contribution. However, have you seen PR #1553? It is to achieve the same goal with a different approach.

heliocastro:
> Hello @heliocastro ,
> 
> Thank you for your contribution. However, have you seen PR #1553? It is to achieve the same goal with a different approach.

Very good, his approach is way better. I will withdraw my commit. 
I'm testing that pr on my side.
Thanks for mention it

PR #1677, opened 2020-04-02
feat(maintenance): Remove orphan log files

### LINKED ISSUE #292: maintagent - add feature to remove orphan log files
---

Author Name: **Bob Gobeille** 
Original Redmine Issue: 6903, http://www.fossology.org/issues/6903
Original Date: 2014/04/14
Original Assignee: Bob Gobeille

---

Delagent defect #6897 causes deleted uploads to leave their jobqueue log files.  Add option to remove orphan log files.

GMishx:
## Description

Remove job log files created by agents for the uploads which were deleted.

### Changes

1. Added new function `removeOrphanedLogFiles()` with flag `-L`.
1. List the `jobqueue` entries for uploads where Delete action has been performed and remove all associated log files.
1. The removed file locations gets replaced with `removed`.

## How to test

1. Upload some package with binary files (images, etc.) which will cause monk to create log.
1. Delete the upload (`delagent` will also create a log file).
1. Save the `jobqueue.job_log` entries for the upload for verification.
1. Run the maintenance agent with relevant options.
1. Check if the `jobqueue.job_log` entries have been updated to `removed`.
1. Check if the previously saved files are removed from FS.

Closes #292

### DIFF
--- a/src/maintagent/agent/process.c
+++ b/src/maintagent/agent/process.c
@@ -25,6 +25,7 @@
 
 /**********  Globals  *************/
 PGconn* pgConn = NULL;        ///< the connection to Database
+fo_dbManager* dbManager;      ///< fo_dbManager object
 
 /**
  * \brief simple wrapper which includes PQexec and fo_checkPQcommand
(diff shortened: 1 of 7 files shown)


src/maintagent/agent/process.c
```diff
@@ -449,3 +450,77 @@ FUNCTION void removeOrphanedRows()
...
+  int countTuples;
+  int removedCount = 0;
+  int jobQueueId;
+  char* logPath;
+  long startTime, endTime;
+  fo_dbManager_PreparedStatement* updateStatement;
+  struct stat statbuf;
+
+  char *SQL = "SELECT jq_pk, jq_log FROM job ja "
+    "INNER JOIN job jb ON ja.job_upload_fk = jb.job_upload_fk "
```
shaheemazmalmmd, line 473: Joining same table ?? i think the below query is enough.

```
SELECT jq_pk, jq_log FROM job ja 
    INNER JOIN jobqueue jq ON ja.job_pk = jq.jq_job_fk 
    WHERE ja.job_name = 'Delete' AND jq_log IS NOT NULL 
    AND jq_log != 'removed';
```

GMishx: This was done as the `Delete` job as a separate entry in `job_queue` and `job` table.

The same table was joined back on `upload_fk` to get all the jobs run on the uploads with `delete` jobs.

Issue #1678, opened 2020-04-03
When installing FOSSology using the single container method, containers do not detach


jeremiah:
### Description

When installing FOSSology via single container method, the resulting container does not detach from the console. 

#### Steps followed and expected result
Install FOSSology with docker according to the installation instructions.

#### Screenshots

N/A

dineshr93:
Even after `docker-compose up -d`?

jeremiah:
Well, `docker-compose up -d` is not the instruction listed; https://hub.docker.com/r/fossology/fossology/ nor on the GitHub wiki. I will test to see if that does work and if it does, I strongly feel that we should add the `-d` flag because detaching is likely the expected behavior. (Of course, if it's not the expected behavior please let me know.) 

For the other approach listed in the installation page, I think that the -d flat will not work. This is based on my simple testing. I think what is happening is that detaching from the container means that certain scripts are not run which leaves the container in an unfinished configuration state which is suboptimal. The container ends up dead. I do think that reviewing the apache2ctl -D FOREGROUND command at the bottom of the script might help, but that may require a separate bug.

jeremiah:
So I've submitted a PR #1679 that includes some changes to the installation instructions. One of the changes is to add the ```-d``` flag which works as expected on my machine and leads to what I believe is a more production ready install. I strongly feel that the Docker compose solution should be made the preferred, and perhaps the default, installation means, but I'll file another issue for that.

Issue #1630, opened 2020-02-25
Documentation of precise steps to configure external postgres DB container & related infos


dineshr93:
### Description

After `docker pull postgres`

Precise steps to 
1. configure the postgresql container
2. Increase the size of the postgresql db container
3. how to monitor the postgres container Resource. (check used & remaining size)

On Fossology container:
4. how to monitor the performance of the fossology docker container.
Resource usage etc.
5. Backup & security aspects of the fossology container

All this would make it robust & complete.

GMishx:
Thanks for the suggestions @dineshr93. Can you please fork FOSSology and create a wiki page for the same? We can import the same to our wiki.

GMishx:
If you are okay with it, I can assign the issue to you.

dineshr93:
@GMishx  I can document it. But I myself dont know the answers for the above . If you give me the pointers to the above points ill update the document

GMishx:
That's good news. For postgresql, if you see the `docker-compose.yml`, you can see that it already has a mount volume defined. You can check the documentation of the official postgres image and check [Where to Store Data](https://hub.docker.com/_/postgres).

For FOSSology container, there is another mount point for the container. You can try to play with it and see how it can grow and some tools which you can use to monitor the containers.

bharathkumar522:
Hi @dineshr93,
           if you prepared any rough document. please share

PR #1679, opened 2020-04-07
WIP: fix(Changes) to install section of README for clarity

jeremiah:
## Description
Changes to install section of README for clarity

### Changes

## How to test

- Changes to the installation section of the README.md file in an effort to make it easier to understand and clearer. 
- Added the -d switch to the description of the docker compose method and 
- Separated installation section into separate instructions because the previous instructions ran together.

I think at the moment, it is not as clear to install via containers and it might be. One of the reasons is that if one chooses to install a single container, you get a scary warning that data is not persistent. It looks like there's a lot to do in that single container setup to make data persist, so my recommendation would be to use the Docker compose method as the sole container method for installation. 

This address, but does not close completely, #1630 and #1678


maxhbr:
Additionally to the database, you might explain how the repository is handled in each case

README.md
```diff
@@ -25,26 +25,29 @@ See https://github.com/fossology/fossology/releases for source code download of
...
-use, since the the standalone image does not take care of data persistency.
+use, since the the single container does not handle data persistence.
+
+FOSSology comes with a [Dockerfile](https://github.com/fossology/fossology/blob/master/Dockerfile) which creates a single container which runs all three services; web-based UI, the FOSSology scheduler, and the PostgresSQL database. A pre-built Docker image is available from [Docker Hub](https://hub.docker.com/r/fossology/fossology/) and can be run using following commands:
 
-A pre-built Docker image is available from [Docker Hub](https://hub.docker.com/r/fossology/fossology/) and can be run using following command:
 ``` sh
 docker run -p 8081:80 fossology/fossology
 ```
+The docker image's web-base UI can then be used viewed http://IP_OF_DOCKER_HOST:8081/repo user fossy password fossy.
```
maxhbr, line 37: I think that should be `http://IP_OF_DOCKER_HOST:80/repo` or `http://localhost:8081/repo` due to the `-p 8081:80`

maxhbr:
```suggestion
The docker image's web-base UI can then be used viewed http://IP_OF_DOCKER_HOST:8081/repo user `fossy` password `fossy`.
```

README.md
```diff
@@ -25,26 +25,29 @@ See https://github.com/fossology/fossology/releases for source code download of
 
 For installation instructions see [Github Wiki](https://github.com/fossology/fossology/wiki)
 
-## Docker
-
-FOSSology comes with a Dockerfile allowing the containerized execution
-both as single instance or in combination with an external PostgreSQL database.
+### Install as a single container with all services combined
 **Note:** It is strongly recommended to use an external database for production
```
maxhbr, line 29: You also need to preserve the repository, without that you only have the metadata but not the files!

README.md
```diff
@@ -25,26 +25,29 @@ See https://github.com/fossology/fossology/releases for source code download of
...
-A pre-built Docker image is available from [Docker Hub](https://hub.docker.com/r/fossology/fossology/) and can be run using following command:
 ``` sh
 docker run -p 8081:80 fossology/fossology
 ```
+The docker image's web-base UI can then be used viewed http://IP_OF_DOCKER_HOST:8081/repo user fossy password fossy.
 
-The docker image can then be used using http://IP_OF_DOCKER_HOST:8081/repo user fossy passwd fossy.
-
-Execution with external database container can be done using Docker Compose, via the following command:
+### Install separate services with docker-compose
```
maxhbr, line 39: It might be worth noting, that here the same image is used, as in the single container variant.

README.md
```diff
@@ -25,26 +25,29 @@ See https://github.com/fossology/fossology/releases for source code download of
...
+Separate containers for the web-based UI, the FOSSology scheduler, and the PostgresSQL database can be set up with Docker Compose, via the following command:
 ``` sh
-docker-compose up
+1. git clone https://github.com/fossology/fossology.git
+2. cd fossology/
+3. docker-compose up -d
 ```
+This set up allows for an separate container for the database which in turn configures data persistency on the host. You will not get data peristency with the single container set up, you have to use the Docker compose set up.
+
+### Database configuration
```
maxhbr, line 48: Please add a section how to extract and preserve the repository

jeremiah:
> Additionally to the database, you might explain how the repository is handled in each case
Good feedback Maximillian, thanks. 

What do you mean exactly? Do you mean an explanation for how the repository of source code is handled on disk? Do you mean how the various containers interact with the repo?

README.md
jeremiah: I'm not sure what you mean here. Could you explain?

README.md
jeremiah: Yeah, that might be a good idea. But are the images identical? It appears that they are both built by the same Dockerfile but I thought there was some separate configuration, I'll have to dig deeper to understand more.

maxhbr: The Image is just build from the dockerfile, no configuration there. One could also start the docker-compose setup with the released image on dockerhub.

README.md
maxhbr: The data of the FOSSology application is stored in two places:
- the postgres DB, that contains the relational data and
- the repository at `/srv/fossology/repository/` containing the files, generated reports and other stuff

So for production you need to externalize the DB and the repostory.

maxhbr:
> What do you mean exactly? Do you mean an explanation for how the repository of source code is handled on disk? Do you mean how the various containers interact with the repo?

In the single container case, the repository is either in the ephemeral docker container, or it can be mounted to the host or replaced by a persistent volume.

In the compose setup there are two volumes defined, one for the database content and one for the repository. So to use the docker-compose setup in production one also has to preserve  and backup these volumes. This might be worth to be mentioned.

README.md
jeremiah: But the point is that the single instance Docker install doesn't provide an *external* db connection. Thus we get a warning and this is why we should likely deprecate it. I feel the installation instructions using this method should be removed as they are not production ready.

README.md
jeremiah: But the instructions do not describe or support that separate process. Don't you think the instructions should be as clear as possible an support limited, well-defined, supported processes? This reduces the out-of-process installs and improves UX greatly.

jeremiah:
> In the single container case, the repository is either in the ephemeral docker container, or it can be mounted to the host or replaced by a persistent volume.
> 
> In the compose setup there are two volumes defined, one for the database content and one for the repository. So to use the docker-compose setup in production one also has to preserve and backup these volumes. This might be worth to be mentioned.

Yes - I agree the access, backup, and mounting of the database ought to be addressed, good point. But should that be in the "Install" section? Or should there be a link from the install section where users can go to review that documentation?

README.md
maxhbr: There are several ways of getting it production ready:
- running a DB on the host, mounting the repository to the host and starting it with `--net=host`
- overwriting the DB connection via environment variables and putting the repository into a volume

maxhbr: The single docker instance provides the possibility to configure an external db connection via env variables

jeremiah:
I'll take a look.

README.md
jeremiah: Visiting this a year later and it is unclear what you're after @maxhbr - if you're interested in the description of how one configures and database that is on a separate host (i.e. not _in_ the container nor on the host), then perhaps that is a separate set of documentation?

PR #1691, opened 2020-04-20
feat(download): Limit source code download for admin users only

### LINKED ISSUE #1304: Prevent a user from downloading the source code
### Description

In the browser view, an entry in the drop-down list proposes to download the material scanned in a given upload.
The need is to remove the possibility for non-admin users to download this data - although users will still be able to see the files contents by browsing them one by one (or by creating a specific web craweler), the idea is to remove this obvious way to get hold of the whole upload.

A system-wide configuration entry (in Admin -> Customize page) would:
- remove the Download entry from the drop down list
- result in an error if trying to access the download URL directly

We can think of smarter configuration options (for each user, for each upload, etc.), but I suppose a general system-wide option should be enough.
 
#### Screenshots

image

piotreke12:
## Description

Fix proposal for #1304 

This PR is now in initial version, final solution should be confirmed

Now, all read/write users can download source code for uploads they have access to - it is not always desired situation.
Changing line below could not be the best solution as it will also prevent non-admin users from downloading anything else that uses this plugin - for example reports (and we do not want that)
https://github.com/fossology/fossology/blob/b5a389be8ef413fc488b56700ea23bdf3ff9c73a/src/www/ui/ui-download.php#L42

In this PR we simply do not show **Download** buttons when user is not an admin.
I'm aware that it isn't ready to be ths final solution, but I can see 2 options:

- add option in Customize to choose for which access right download button should be shown
- separate somehow (I don't know how yet) required permission level for:
    - downloading source code
    - everything else

### Changes (final version)

src/lib/php/common-sysconfig.php - additionl configuration option (Admin>Customize page) - possibility to choose which user level should be able to download source code

src/lib/php/common-license-file.php, src/www/ui/ui-download.php - showing Download buttons depending on user-level set on Admin>Customize page

### How to test

Depending on selected option:
image

User with specified access rights (or higher) should see or not see Download button for source code

Closes #1304

### DIFF
--- a/src/lib/php/common-sysconfig.php
+++ b/src/lib/php/common-sysconfig.php
@@ -16,6 +16,8 @@
  51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 ***********************************************************/
 
+use Fossology\Lib\Auth\Auth;
+
 /**
  * \file
  * \brief System configuration functions.
@@ -391,6 +393,14 @@ function Populate_sysconfig()
   $mimeTypeDesc = _("add  comma (,) separated mimetype to exclude files from scanning");
   $valueArray[$variable] = array("'$variable'", "null", "'$mimeTypeToSkip'",
     strval(CONFIG_TYPE_TEXT), "'Skip'", "1", "'$mimeTypeDesc'", "null", "null");
+
+  $perm_admin=Auth::PERM_ADMIN;
+  $perm_write=Auth::PERM_WRITE;
+  $variable = "SourceCodeDownloadRights";
+  $SourceDownloadRightsPrompt = _('Acces rights required to download source code');
+  $SourceDownloadRightsDesc = _('Choose which acces level will be required for user to be able to download source code.');
+  $valueArray[$variable] = array("'$variable'", "'$perm_write'", "'$SourceDownloadRightsPrompt'",
+  strval(CONFIG_TYPE_DROP), "'DOWNLOAD'", "1", "'$SourceDownloadRightsDesc'", "null", "'Administrator{{$perm_admin}}|Read_Write{{$perm_write}}'");
 
   /* Doing all the rows as a single insert will fail if any row is a dupe.
    So insert each one individually so that new variables get added.
--- a/src/lib/php/common-license-file.php
+++ b/src/lib/php/common-license-file.php
@@ -285,6 +285,7 @@ function Level1WithLicense($agent_pk, $rf_shortname, $uploadtree_pk, $PkgsOnly=f
 function FileListLinks($upload_fk, $uploadtree_pk, $napk, $pfile_pk, $Recurse=True, &$UniqueTagArray = array(), $uploadtree_tablename = "uploadtree", $wantTags=true)
 {
   $LinkStr = "";
+  global $SysConf;
 
   if ($pfile_pk) {
     $text = _("View");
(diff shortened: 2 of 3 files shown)


piotreke12:
I forgot to mention - please feel free to comment this solution and my proposals - at this moment I'm not sure which version is best :)

If You have any comments/remarks/proposals - I'm waiting for your feedback :)

GMishx:
The approach looks good @piotreke12 .
To make this option configurable, have you considered the **Admin > Customize** page? You can modify the `Populate_sysconfig()` in `src/lib/php/common-sysconfig.php` and use it value to get which level of users should be allowed to download the file.

piotreke12:
> The approach looks good @piotreke12 .
> To make this option configurable, have you considered the **Admin > Customize** page? You can modify the `Populate_sysconfig()` in `src/lib/php/common-sysconfig.php` and use it value to get which level of users should be allowed to download the file.

Yes, I will use Admin -> Customize page. 
Thanks for this hint concerning `Populate_sysconfig()` :)

src/lib/php/common-sysconfig.php
```diff
@@ -386,6 +386,12 @@ function Populate_sysconfig()
   $valueArray[$variable] = array("'$variable'", "30", "'$patTokenValidityPrompt'",
     strval(CONFIG_TYPE_INT), "'PAT'", "1", "'$patTokenValidityDesc'", "null", "null");
 
+  $variable = "SourceCodeDownloadRights";
+  $SourceDownloadRightsPrompt = _('Acces rights required to download source code');
+  $SourceDownloadRightsDesc = _('Choose which acces level will be required for user to be able to download source code.');
+  $valueArray[$variable] = array("'$variable'", "'3'", "'$SourceDownloadRightsPrompt'",
+    strval(CONFIG_TYPE_DROP), "'DOWNLOAD'", "1", "'$SourceDownloadRightsDesc'", "null", "'Administrator{10}|Read_Write{3}'");
```
GMishx, line 393: Instead of writing `10` and `3`, will it possible to use the values from `Fossology\Lib\Auth\Auth` class?

https://fossology.github.io/d9/dbd/Auth_8php_source.html#l00060

piotreke12: Yes, definitely it'd be better - I thought about it but to be honest I eventually forgot to find a class with those values and change it.

Thanks for this tip - I will correct this

GMishx: I guess when you are done with change, you can make the PR ready for review. The approach looks good enough to me.

piotreke12:
Changes applied as requested - PR ready to review

GMishx:
While testing the PR, I found that from the UI there is no was user was getting a download link/dropdown. The user can still download the reports.

But if I have the URL to download the file, even though I;m not allowed to do so, I can download it.
- URL to download a file (dependent on `SourceCodeDownloadRights`): `http://localhost/repo/?mod=download&upload=32&item=23650`
- URL to download a report (should always work): `http://localhost/repo/?mod=download&report=524`

NicolasToussaint:
> But if I have the URL to download the file, even though I;m not allowed to do so, I can download it.

Agreed, and we need to make this extra clear in this PR too.
I wrote in #1304 , this solution "does not prevent a non-admin user from downloading the source code by guessing and entering the download URL directly"

NicolasToussaint:
So, @GMishx and others, do you think we should merge the PR as is, knowing it is not ideal, as a first step?
Or should we only only add a feature that completely prevents downloads ?

GMishx:
@NicolasToussaint , I think the solution is pretty simple.

Won't adding
https://github.com/fossology/fossology/blob/fee95f6e6e5afdc4ea1b313da7e311110663ce9c/src/www/ui/ui-download.php#L53
to following function
https://github.com/fossology/fossology/blob/40eef9267c21dbb52d02174c4cd72db73fe0ffee/src/www/ui/ui-download.php#L203
and throwing an Exception solve the issue?

Also, please rebase the branch with master as we have some merge conflicts.

piotreke12:
Hi - branch rebased.

I unfortunately don't get your proposal, @GMishx 

If I understand well this plugin, there is no way to distunguish which download try is done for source code and which is for report - or maybe I miss something?

GMishx:
> I unfortunately don't get your proposal, @GMishx
> 
> If I understand well this plugin, there is no way to distunguish which download try is done for source code and which is for report - or maybe I miss something?

The plugin uses two different parameters `item` and `report` to distinguish between source download and report download, respectively.

Maybe instead of modifying the `downloadFile()`, you can implement the logic in 
https://github.com/fossology/fossology/blob/fee95f6e6e5afdc4ea1b313da7e311110663ce9c/src/www/ui/ui-download.php#L131

piotreke12:
Yes - it looks like you are perfectly right, @GMishx 
I somehow missed out this detail :)

piotreke12:
Hi @GMishx,

changes done - additional verification step in getPathAndName() function.
Still, I left changes concerning "Download" buttons as previously - I thing it is nice no to have them when you are not allowed to download source code.
It works ok on my side, please verify on yours :)

Issue #1631, opened 2020-02-25
Several dependencies are missing in fo-installdeps 


sjha2048:
### Description

While installing fossology from the source It came into my sight that there are several missing dependencies which can be added into fo-installdeps so that the installation process will be more smooth.
these dependencies can be added:
libpq-dev
libjsoncpp-dev
libboost-all-dev
libssl-dev
libjson-c-dev

### Versions

* Operating System (lsb_release -a): Pop!_OS 19.10

GMishx:
Hello @sjha2048, the dependencies are also distributed per agent basis under `src/<agent>/mod_deps` files. You had a look there?

sjha2048:
sure I'll have a look

GMishx:
Did you get a chance to check the dependencies @sjha2048 ?

sjha2048:
sorry for the late reply, I was only able to find two of these packages in mod_deps files
* libjsoncpp-dev
* libjson-c-dev
the rest are missing

GMishx:
Hello @sjha2048

Here are the locations of each dependency:

* libpq-dev : Will be installed by postgres package but should be listed in dependency list.
* libjsoncpp-dev : https://github.com/fossology/fossology/blob/b5a389be8ef413fc488b56700ea23bdf3ff9c73a/src/copyright/mod_deps#L80
* libboost-all-dev : (note, only required packages are installed instead of `all`) https://github.com/fossology/fossology/blob/b5a389be8ef413fc488b56700ea23bdf3ff9c73a/src/copyright/mod_deps#L80
* libssl-dev : (will be removed by #1545) https://github.com/fossology/fossology/blob/b5a389be8ef413fc488b56700ea23bdf3ff9c73a/src/delagent/mod_deps#L79
* libjson-c-dev : https://github.com/fossology/fossology/blob/b5a389be8ef413fc488b56700ea23bdf3ff9c73a/src/nomos/mod_deps#L80

sjha2048:
shall I add libpq-dev then?

GMishx:
Sure, please go ahead

PR #1708, opened 2020-04-30
fix(fo-installdeps): added a missing dependency

sjha2048:
## Description

as per discussion on #1631 one dependency was missing, So I have added it in fo-installdeps and Debian control file

Fixes: #1631


GMishx:
Thank you for the fix @sjha2048. Could you please add `libpq-devel` to Red Hat based systems in `fo-installdeps` as well?

sjha2048:
Sure thing.

utils/fo-installdeps
```diff
@@ -148,7 +148,7 @@ if [[ $BUILDTIME ]]; then
             postgresql-devel \
             gcc make file libxml2 \
             perl-Text-Template subversion \
-            boost-devel php-mbstring libicu-devel
+            boost-devel php-mbstring libicu-devel libpq-devel
```
GMishx, line 151: Can you please add it for Fedora as well?

PR #1709, opened 2020-05-06
feat(jobs): Show all running jobs

GMishx:
## Description

Add a new page to show all running/pending jobs on server.
The page also shows current status of scheduler (Running/Stopped).

This will help users to understand the reason if their jobs are stuck.

### Changes

1. Add new function in `ShowJobsDao.php` to get running and pending jobs per agent.
1. Change the default port on `common-scheduler.php` to `24693` (as per the `fossology.conf`).
1. New async page `AjaxAllJobStatus.php` to get job status and scheduler status.
1. New page `AllJobStatus.php` and template `all_job_status.html.twig` to show the data from `AjaxAllJobStatus.php`.

**Note:**
1. The page is added under Admin >> Dashboards >> All Jobs
1. The page will work **without login** (as no sensitive information is displayed).

## How to test

1. Check if the new page appears under Admin >> Dashboards >> All Jobs
1. Check if the page reloads in the time intervals defined by **ShowJobs Auto Refresh Time** in **Customize** page.
1. Disable/Enable the scheduler to check the scheduler status updates on page.

### DIFF
--- a/src/lib/php/Dao/ShowJobsDao.php
+++ b/src/lib/php/Dao/ShowJobsDao.php
@@ -372,4 +372,21 @@ class ShowJobsDao
       return array();
     }
   }
+
+  /**
+   * Get the status of all pending or running jobs.
+   * @return array Containing job name, number of jobs pending and running
+   */
+  public function getJobsForAll()
+  {
+    $sql = "SELECT jq_type AS job, jq_job_fk, job_upload_fk AS upload_fk, " .
+      "CASE WHEN (jq_endtext IS NULL AND jq_end_bits = 0) THEN 'pending' " .
+      "WHEN (jq_endtext = ANY('{Started,Restarted,Paused}')) THEN 'running' " .
+      "ELSE '' END AS status " .
+      "FROM jobqueue INNER JOIN job " .
+      "ON jq_job_fk = job_pk " .
+      "WHERE jq_endtime IS NULL;";
+    $statement = __METHOD__ . ".getAllUnFinishedJobs";
+    return $this->dbManager->getRows($sql, [], $statement);
+  }
 }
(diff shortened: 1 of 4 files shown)


src/lib/php/common-scheduler.php
```diff
@@ -45,7 +45,7 @@ function fo_scheduler_connect($IPaddr='', $Port='', &$ErrorMsg="")
     $IPaddr = '127.0.0.1';
   }
   if (empty($Port)) {
-    $Port = 5555;
+    $Port = 24693;
   }
   if (($sock = socket_create(AF_INET, SOCK_STREAM, SOL_TCP)) === false) {
     $ErrorMsg = "socket_create() failed: reason: " . socket_strerror(socket_last_error()) . "<br>\n";
```
shaheemazmalmmd, line 51: Is this a required change for this PR ??

GMishx: Not required but will be helpful.
I just stumbled across the code and thought it would be worth a fix to help someone working on it for the first time.

shaheemazmalmmd: Can we make it separately on other PR ? if it is fixing something ?

GMishx: I'll take this out as it does not fix anything for a running system, but it will save some time if they are running scheduler on a different port so I'll create a separate PR for it.

shaheemazmalmmd:
Some suggestions: 

* Can we also add the total ETA for this jobs to finish ??

GMishx:
> Some suggestions:
> 
>     * Can we also add the total ETA for this jobs to finish ??

Thank you for the suggestion @shaheemazmalmmd. Implemented, please check.

PR #1711, opened 2020-05-07
fix(spdxReport): add missing artifact to file path in spdx reporting

### LINKED ISSUE #1703: SPDX tag-value export ignores top-level zips in path
### Description

When an uploaded archive contains sub-archives in the top-level directory within the main archive, in at least some cases the name of the sub-archive file is being dropped from the FileName field during SPDX tag-value export.

Additionally, if multiple archives contain similar paths, then this can result in invalid SPDX documents such as by having duplicate file paths within one package.

#### How to reproduce

Upload file `test2.zip` (attached: [test2.zip](https://github.com/fossology/fossology/files/4533573/test2.zip)) which contains the following rather complicated archive structure:
```
test2.zip
├─ test1.zip
│  └─ test1/
│     ├─ f1.txt
│     ├─ f2.zip
│     │  ├─ f21.txt
│     │  └─ f2/
│     │     ├─ f3.zip
│     │     │  └─ f3/
│     │     │     └─ f31.txt
│     │     └─ f4.tar.gz
│     │        └─ f4.tar
│     │           └─ f4/
│     │              └─ f41.txt
│     └─ f5.tar.gz
│        └─ f5.tar
│           └─ f5/
│              └─ f51.txt
└─ test1a.zip
   └─ test1/
      ├─ f1.txt
      ├─ f2.zip
      │  ├─ f21.txt
      │  └─ f2/
      │     ├─ f3.zip
      │     │  └─ f3/
      │     │     └─ f31.txt
      │     └─ f4.tar.gz
      │        └─ f4.tar
      │           └─ f4/
      │              └─ f41.txt
      ├─ f5.tar.gz
      │  └─ f5.tar
      │     └─ f5/
      │        └─ f51.txt
      └─ f6.zip
         └─ f6/
            └─ f7/
               └─ f7.txt
```

Export SPDX tag-value file for the above archive. (In my case, I ran nomos / monk / ojo agents first and cleared the results, before exporting as SPDX).

Review the SPDX file that is exported (also attached here; renamed to .txt because GitHub won't allow .spdx extensions for uploads): [test2.txt](https://github.com/fossology/fossology/files/4533572/test2.txt)

For subdirectories, you'll see that as expected, .zip and .tar.gz files will include the .zip / .tar.gz filename within the file path -- e.g. `test1/f2.zip/f2/f4.tar.gz/f4.tar/f4/f41.txt`

However, for the two top-level .zip files -- `test1.zip` and `test1a.zip` -- the zip files' names are not included in the path for the exported SPDX file.

Moreover, because each of those top-level .zip files contained exactly one directory -- which was called `test1/` in both of them -- the exported SPDX file has duplicate filenames. E.g., there are multiple entries for `test1/f2.zip/f2/f21.txt` which is invalid as Files must have unique paths. Given the way that Fossology handles archive files, these should probably be:
* `test1.zip/test1/f2.zip/f2/f21.txt` and
* `test1a.zip/test1/f2.zip/f2/f21.txt`

which would retain the top-level zip file name in the exported file path, and therefore ensure that file names remain unique.

### Versions

This was seen on a Fossology 3.6.0 server.

* Last commit id on master: fc1b3c
```
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 9.11 (stretch)
Release:        9.11
Codename:       stretch
```

### Logs

I haven't attached logs as I don't think any are relevant to this issue, but let me know if any are needed and I'll dig them out (or re-generate them).

shaheemazmalmmd:
closes #1703 

## Description

Path in SPDX files is getting chopped because of which there are few files missing in the report in rare cases.

### Changes

add parent and enable file path with artifacts.

## How to test

Please see #1703

### DIFF
--- a/src/spdx2/agent/spdx2.php
+++ b/src/spdx2/agent/spdx2.php
@@ -443,7 +443,7 @@ class SpdxTwoAgent extends Agent
       if (($filesProceeded&2047)==0) {
         $this->heartbeat(0);
       }
-      $fullPath = $treeDao->getFullPath($fileId,$treeTableName,0,true);
+      $fullPath = $treeDao->getFullPath($fileId,$treeTableName,0);
       if (!empty($licenses['concluded']) && count($licenses['concluded'])>0) {
         $this->toLicensesWithFilesAdder($licensesWithFiles,$licenses['concluded'],$licenses['copyrights'],$fileId,$fullPath);
       } else {
@@ -689,7 +689,7 @@ class SpdxTwoAgent extends Agent
         $lastValue = $filesProceeded;
       }
       $hashes = $treeDao->getItemHashes($fileId);
-      $fileName = $treeDao->getFullPath($fileId, $treeTableName, 0, true);
+      $fileName = $treeDao->getFullPath($fileId, $treeTableName, 0);
       if (!is_array($licenses['concluded'])) {
         $licenses['concluded'] = array();
       }


src/spdx2/agent/spdx2.php
```diff
@@ -443,7 +443,8 @@ protected function toLicensesWithFiles(&$filesWithLicenses, $treeTableName)
       if (($filesProceeded&2047)==0) {
         $this->heartbeat(0);
       }
-      $fullPath = $treeDao->getFullPath($fileId,$treeTableName,0,true);
+      $parentId = $treeDao->getMinimalCoveringItem($uploadId, $treeTableName);
```
GMishx, line 446: I think the value of `$parentId` is going to be constant for each iteration of the loop. Wouldn't it be better if it is initialized outside the loop only once?

src/spdx2/agent/spdx2.php
```diff
@@ -676,7 +677,8 @@ protected function generateFileNodesByFiles($filesWithLicenses, $treeTableName,
         $lastValue = $filesProceeded;
       }
       $hashes = $treeDao->getItemHashes($fileId);
-      $fileName = $treeDao->getFullPath($fileId, $treeTableName, 0, true);
+      $parentId = $treeDao->getMinimalCoveringItem($uploadId, $treeTableName);
```
GMishx, line 680: Same change to initialize `$parentId` only once...

GMishx:
Also, please update the commit message to remove `;`

GMishx:
I just uploaded the test file attached in #1703 and did some clearing and generated the SPDX tag:value report. The issue still persists. Please have a look at the report: [SPDX2TV_test2.zip_1591875331.spdx](https://github.com/fossology/fossology/files/4764452/SPDX2TV_test2.zip_1591875331.txt)

PR #1722, opened 2020-05-14
feat(decisions): auto deactivate copyrights

shaheemazmalmmd:
## Description

Automatically deactivate the copyright statement if the user applies the license decision to irrelevant.  

## How to test

- Upload a package.
- Apply irrelevant decisions. from file level as well as folder level and check the copyrights.
- As this changes may affect the clearing decision part, please test every clearing part.


ag4ums:
from discussion, Need to have more efficient irrelevant handling,... (further discussion require) 
eg: 
>bulk scanning also update the folders set as irrelevant for licenses, while the licenses for the files are identified, but the copyrights are irrelevant/deactivated, may become consistency issue . 
>another case could be how to handle manually enabled/disabled copyrights, when making folder as irrelevant/not irrelevant.

Issue #1494, opened 2019-10-03
Directory with "Cleared" and "To be discussed" files should be tagged  "To be discussed"


NicolasToussaint:
### Description

In the browser, a directory containing files with a *clearing Status* to either "Cleared" (green) or "To Be Discussed" (Yellow) should be show as Yellow rather than red (open).

#### Steps followed and expected result

Take a directory containing two files:
- file_a has a clearing status: "cleared" -> show with a green tag
- file_b has a clearing status: "To be Discussed" -> show with a yellow tag

Then the directory itself is displayed with a Red tag, while it should be Yellow.

#### Screenshots

In the directory "Dir_2", there are two files: one "green" and one "yellow":
![image](https://user-images.githubusercontent.com/22956329/66125341-fd837380-e5e6-11e9-996a-b5a53e440553.png)

The directory itself, has a Red tag:
![image](https://user-images.githubusercontent.com/22956329/66125407-26a40400-e5e7-11e9-8045-bfe9ab135cc2.png)

NicolasToussaint:
Unless it's been fixed in recent version, I think there may be a nice effort / UI improvement ratio here.

shaheemazmalmmd:
We(me and @EttingerK) have tested it. so the issue is still there.

Satjeet2005:
hey @shaheemazmalmmd @NicolasToussaint, I have tried to solve it. I have uploaded two files in the directory same as in the screenshot, and made one file "cleared" and other one "to be discussed", but I am unable to get the directory view as in screenshot. The video below shows how i am trying to do that. I have watched the YouTube showing end to end workflow but couldn't get that.


https://github.com/user-attachments/assets/8399c13c-0b80-4a53-b0a2-452dc2c81f68

NicolasToussaint:
@Satjeet2005, It looks to me like you are testing with 2 different _uploads_ containg one file, rather than one upload that contains a _folder_ that contains two files.

I could be mistaken since I have not used recent versions of Fossology

Satjeet2005:
Ok, got it. Thanks @NicolasToussaint, for helping out.

nikhilsingh7777:
Hey,  does this still need to fix ?  should i look into this ?

PR #1732, opened 2020-05-29
fix(clearingCount): do not add count as cleared in case of to be disc…

shaheemazmalmmd:
## Description

To be discussed files appearing as Grey in license browser. And count of cleared files varies in case of to be discussed.

### Changes

* fix count exclude to be discussed in single file view.

## How to test

* Single file view should have same count as tree-view in case of "To be discussed" decisions.

* For "no license found" files. if we have a decision as "To be discussed" in tree-view it should be with yellow dot.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -974,7 +974,7 @@ INSERT INTO clearing_decision (
       $statementName, $params);
 
     $statementName = __METHOD__ . $statementName;
-    $sql = "$cte SELECT COUNT(*) AS cnt FROM decision;";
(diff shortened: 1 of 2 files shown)


NicolasToussaint:
Nice, so that fixes my old issue #1494

GMishx:
> Nice, so that fixes my old issue #1494

Not exactly. The fix does not change the clearing status at folder level. It's a fix at file level status only.

NicolasToussaint:
> > Nice, so that fixes my old issue #1494
> 
> Not exactly. The fix does not change the clearing status at folder level. It's a fix at file level status only.

oh, too bad ;-)

Thanks @GMishx   for clarifying

PR #1733, opened 2020-05-31
 fix(spasht-ui): Removed extension from the spasht search

Akash-Sareen:
## Description

Fix Spasht search value. Removed file extension name.

### Changes

Spasht search  recommendation was having file extension. Removed file extension from the spasht search box.

## How to test

`sudo make install-spasht`

### DIFF
--- a/src/spasht/ui/template/agent_spasht_search.html.twig
+++ b/src/spasht/ui/template/agent_spasht_search.html.twig
@@ -8,7 +8,7 @@
 
 <form id="searchForm" method="post" action="">
   <span class="searchBoxText">Name: </span>
-    <input type="text" name="patternName" value={{ uploadName }} size="30"/>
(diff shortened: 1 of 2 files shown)


GMishx:
Hello @Akash-Sareen, thank you for your contribution.

I have few comments on your PR:
1. Please update the PR title to match your commit title.
2. Please squash your commits to one as the second commit is a very minute change.
3. How do you expect to handle cases like `zlib-1.2.3.tar.gz` and `zlib-1.2.3.zip`? As per your current changes, I would get `zlib-1` which is of slightly more work for user to add `.2.3` to the search string.

I would suggest to use the regex `ANY_VERSION` and `ARCHIVE_EXT` from
https://salsa.debian.org/debian/devscripts/-/blob/master/lib/Devscripts/Uscan/WatchFile.pm#L106-110

These regex are used in uscan script of Debian to match the version and extension of file.

PR #1740, opened 2020-06-10
fix(bulk): add class to show text highlighted for matched page

shaheemazmalmmd:
## Description

* Add missing class to highlight text for bulk. in case of matched file.

## How to test

* Upload a package and run bulk for multiple files with different text phrases.
* Check if in each file in bulk history if the matched text is highlighted.

### DIFF
--- a/src/www/ui/template/bulk-history.html.twig
+++ b/src/www/ui/template/bulk-history.html.twig
@@ -28,13 +28,15 @@
             {% endif %}
           </table>
       </td>
-      <td>
       {% if bulkHistoryHighlightUri and row.id > 0 and row.matched %}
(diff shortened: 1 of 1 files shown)


GMishx:
In my opinion, highlighting the whole row looks much nicer than highlighting just the link.

What do you say @shaheemazmalmmd ?

image

image

**P.S.:** The class `hi-bulk` needs to include `!important` to override the alternating row color scheme.

shaheemazmalmmd:
@GMishx : Had discussed with @ag4ums and added the class .

i think it would be better just to highlight the link. because it matches the text in the file.

@ag4ums any thoughts ?

PR #1748, opened 2020-06-18
feat(rest): Get file info from hash

### LINKED ISSUE #1255: Rest end point serving upload id for Hash value
### Description

Fossology should be able to return search results based on the hash value of the uploaded package.

- User should be able to provide a hash value as search string
- Fossology should return list of uploads/upload with the matching hash
- User should be able to review the scanned results and regenerate the reports, SPDX rdf file and CLI files.

### LINKED ISSUE #1611: Rest API: Get license conclusions for individual files by their hash value.
### Description

Add an API endpoint to get file level license conclusions by entering their hash values. This would help with integrating Fossology into different workflows.

#### Approaches

As the number of files that one may want to request license conclusions for may be in tens of thousands, the most efficient way would probably be for the API to accept a list of hash values as an input, while providing a similar list with license conclusions of the files added as an output.  

Another approach, maybe more in line with the current API design, would be to accept API calls with individual hash values as input. If the networking costs for individual API calls is significant, this may get somewhat expensive in many use cases.  

Many other approaches could also work here. The desired outcome is to input hash values of files and receive license conclusions for the files, be it as a list or individual values.

GMishx:
## Description

Get following file info from hash:
- Scanner findings
- License conclusions
- Upload ID if hash represents a package
- Main license if hash represents a package

The query can be done with any combination of MD5, SHA1 and SHA256 checksum.

### Note

There are some additional changes in how hash and license findings and conclusions are structured. It results in two new models `Hash` and `Findings`.

The checksum returned by `GET /uploads` endpoint has been replaced with `hash` model and the license findings returned by `GET /uploads/{id}/licenses` has been replaced with `findings` model.

These two changes are breaking changes for existing REST API clients, thus the API version has been increased to 1.1.0.

#### Request

The endpoint expects data as a POST body (to allow large data in single call) in following format:
```json
[
   {
      "sha1": "9D541987F64E5E4EE4FF45B422788F63A3F4F46F"
   },
   {
      "sha256": "291FA73C6508C0DFF86456F8FAFECC91939F411C39A1631121CF9A586686601E"
   },
   {
      "md5": "e2fc714c4727ee9395f324cd2e7f331f",
      "sha1": "81fe8bfe87576c3ecb22426f8e57847382917acf"
   }
]
```

The checksum is case insensitive and can be of an individual file or of a whole package archive.

#### Response

Typical response looks like:
```json
[
    {
        "hash": {
            "sha1": "9D541987F64E5E4EE4FF45B422788F63A3F4F46F",
            "md5": "10C0582A2217D5150F2FC8CFCA5D36CB",
            "sha256": "56009FCCA16DB2837C5DB8111E1D1CC359855B4C090B428497710B88622C21A4",
            "size": 1593386
        },
        "findings": {
            "scanner": [],
            "conclusion": [
                "GPL-2.0"
            ]
        },
        "uploads": [
            122, 123, 124
        ]
    },
    {
        "hash": {
            "sha1": "68CC6F96B47A3DE753DCE354319DE416182AB9C8",
            "md5": "69B3C4D5B5A6AD9A2BA021193CB66015",
            "sha256": "291FA73C6508C0DFF86456F8FAFECC91939F411C39A1631121CF9A586686601E",
            "size": 4056
        },
        "findings": {
            "scanner": [
                "LGPL-2.1",
                "GPL-2.0"
            ],
            "conclusion": [
                "GPL-2.0"
            ]
        },
        "uploads": null
    },
    {
        "hash": {
            "sha1": null,
            "md5": "gibberishchecksum",
            "sha256": null,
            "size": null
        },
        "message": "Not found"
    },
    {
        "hash": {
            "sha1": null,
            "md5": null,
            "sha256": null,
            "size": null
        },
        "message": "Invalid keys"
    }
]
```

1. If the requested checksum was a package archive
    1. A list of accessible uploads will be returned in `uploads` key.
    1. The `conclusion` of `findings` will be the list of global licenses for the upload.
1. If the requested checksum was an individual file
    1. The `uploads` key will be `null`.
    1. The `scanner` and `conclusion` will be filled accordingly.
1. If the requested checksum was not found
    1. The appropriated key in the `hash` will be filled with provided values for identification.
    1. No `findings` and `uploads` will be returned.
    1. The field `message` will contain value `Not found`.
1. If the requested object was malformed
    1. The request will be malformed if it does not contain any of `md5`, `sha1` or `sha256` key or it contains any other key.
    1. The values in `hash` will be `null` as they could not be parsed.
    1. No `findings` and `uploads` will be returned.
    1. The field `message` will contain value `Invalid keys`.

### Changes

1. New endpoint `POST /filesearch`.
1. New models `Hash` and `Findings`.
1. Change model for `Upload` model and `UploadLicenses` model.

## How to test

1. Test the `/filesearch` endpoint.
1. Test the `/uploads` endpoint.
1. Test the `/uploads/{id}/licenses` endpoint.

Closes #1611
Closes #1255


mmurto:
This looks good! A couple of observations from my testing:

1. Indexing pfile on the hash columns speeds up the queries quite a lot. Without indexing, a query of 50 sha256-values takes about 8 seconds, with `CREATE INDEX pfile_sha256 ON pfile (pfile_sha256);` the same query takes about 350 ms, including network latency.
2. If the same file exists in more than one upload with different license conclusions, the API gives no indication of this and only returns one of the license conclusions. I'm not sure how much of and edge case this is, or if different conclusions for same files actually come up in real use.

deveaud-m:
Just tested everything using the Python wrapper, looks good to me.

PR is read: https://github.com/fossology/fossology-python/pull/34

(I'm not so familiar with the review features of Github, but this PR is ok from my side)

GMishx:
Thanks for testing and providing your observations @mmurto 

> 1. Indexing pfile on the hash columns speeds up the queries quite a lot. Without indexing, a query of 50 sha256-values takes about 8 seconds, with `CREATE INDEX pfile_sha256 ON pfile (pfile_sha256);` the same query takes about 350 ms, including network latency.

Thanks for pointing that out. I've updated constraint `pfile_md5_sha1_size_ukey` to include `pfile_sha256` as well.

> 2. If the same file exists in more than one upload with different license conclusions, the API gives no indication of this and only returns one of the license conclusions. I'm not sure how much of and edge case this is, or if different conclusions for same files actually come up in real use.

Actually, if the file exists in more than one upload, all the conclusions should be clubbed together and returned (if that not working for you, it needs to be fixed). It will be quite complicated task to segregate the results and remove redundant data.
Also, in my opinion, if two files shares shame checksum, should contain same text and should have same license conclusion. However, I can imagine a scenario where this file exists in a package where the license is determined by another file like `LICENSE.md`. I am also not very sure how often that happens in real life.

mmurto:
> Actually, if the file exists in more than one upload, all the conclusions should be clubbed together and returned (if that not working for you, it needs to be fixed). It will be quite complicated task to segregate the results and remove redundant data.
> Also, in my opinion, if two files shares shame checksum, should contain same text and should have same license conclusion. However, I can imagine a scenario where this file exists in a package where the license is determined by another file like `LICENSE.md`. I am also not very sure how often that happens in real life.

I uploaded a test project twice and concluded one of the files as MIT in one and as ISC in the other. The API only returns MIT as a conclusion. I've attached some pictures. 

I don't really know what the best way to separate the data would be. A logical way would be to have a array of arrays for the conclusion-field, but it seems a little clumsy as huge majority of the files would only have one item, like [["MIT"]]. 

is-promise1
is-promise2
file_info
request

GMishx:
Thanks for pointing it out @mmurto , I've fixed the issue now. You should get `["MIT","ISC"]` now for the `conclusion`.

mmurto:
I noticed a small problem while trying to process the output. The API gives no indication of license decisions that remove scanner findings. For example, if a file has a scanner hit of MIT and no clearing decision has been made, `conclusion` is empty. If a user determines that the scanner hit was a false hit and creates a clearing decision to remove the hit, `conclusion` is also empty. In SPDX terms, the first empty should probably be a `NOASSERTION` and the second empty a `NONE`. 

One solution to this would be to return some indication of the clearing status of the file; if `conclusion` is empty and clearing status is closed, the file is `NONE`, if `conclusion` is empty and clearing status is open, the file is `NOASSERTION`. Another solution would be to return some indication of `NONE` or `NOASSERTION` in the `conclusion` field. What do you think would be best?

GMishx:
Hello @mmurto, I've pushed changes. Now, if there are no conclusions done by user, the `conclusion` will be empty array. But if the user has removed all the licenses or marked file as irrelevant, the `conclusion` will be `["NOASSERTION"]`.

Hope the results are much more useful.

mmurto:
@GMishx great, that should work! Just one thing, if the user has removed the licenses or marked the file as irrelevant, the it should be NONE rather than NOASSERTION. NOASSERTION would work for example for a situation where the SPDX file creator has made no attempt to determine the license. [Here's the relevant part of the spec.](https://spdx.github.io/spdx-spec/4-file-information/#45-concluded-license)

GMishx:
Thanks for the clarification @mmurto , I understood it the other way round. I pushed the changes, please test/review them.

ag4ums (approved):
code looks good

Issue #1535, opened 2019-11-28
Rest API: Create a group


NicolasToussaint:
### Description

We need the possibility to create a group through the REST API.

We could imagine a new command `/group`, similar to `/user`, with the extra possibility to create a new group with a `PUT` verb.

#### Steps followed and expected result

1. User creates a token
2. User creates a new group

Command would provide in return the `groupID` of the new group.

NicolasToussaint:
PR merged, closing issue

PR #1764, opened 2020-07-14
feat(rest): Filter uploads by folder id

### LINKED ISSUE #1720: Rest API: extend uploads GET with groupName and folderId
Currently an API user is only able to see his "own" uploads with GET.
Some other methods support the parameter `groupName` but whether GET nor PUT.

It would be very helpful to be able to see and manage uploads from a given group.

Additionally, uploads listing shall accept a `folder` parameter to restrict listing of uploads in a dedicated folder.

GMishx:
## Description

This PR allows user to limit uploads results to given folder id, provide an option to list uploads recursively, paginate the results if more than 100 uploads exists (can be controlled with `limit` parameter in header).

The default folder id is the user's root folder id and recursive is set to true.

In the earlier implementation, the uploads listing was limited to the user accessing the endpoint meaning that you can list only the uploads done by you. However, the UI lists the uploads based on group id. The REST API now reflects the same behavior.

The listing of `groupName` parameter is updated in the `openapi.yaml` to reduce the redundancy and add the endpoints which were missing it.

### Changes

1. Add new parameters `folderId`, `recursive` and `page` to `/uploads` endpoint.
1. List uploads based on groupId, not userId in `/uploads`.
1. `/uploads` response is paginated and can be controlled with `page` and `limit` parameters in headers.
1. Refactor `openapi.yaml` for `groupName`.

## How to test

1. Check if you can list uploads with `/uploads`.
1. Check if you can limit the responses with `folderId` and `recursive` parameters.
    - Also, bad values for the parameters should not break the code.
1. Check if `/folders` endpoint and `/search` endpoints are working.
1. Check if general UI of FOSSology is working.

Closes #1720

### DIFF
--- a/src/lib/php/Dao/FolderDao.php
+++ b/src/lib/php/Dao/FolderDao.php
@@ -296,7 +296,7 @@ WHERE fc.parent_fk = $1 AND fc.foldercontents_mode = " . self::MODE_UPLOAD . " A
     return !empty($cycle);
   }
 
-  protected function getContent($folderContentId)
+  public function getContent($folderContentId)
   {
     $content = $this->dbManager->getSingleRow('SELECT * FROM foldercontents WHERE foldercontents_pk=$1',
       array($folderContentId),
@@ -444,13 +444,14 @@ WHERE fc.parent_fk = $1 AND fc.foldercontents_mode = " . self::MODE_UPLOAD . " A
   /**
    * Get the folder contents id for a given child id
    * @param integer $childId Id of the child
+   * @param integer $mode    Mode of child
    * @return NULL|integer Folder content id if success, NULL otherwise
    */
-  public function getFolderContentsId($childId)
+  public function getFolderContentsId($childId, $mode)
   {
     $folderContentsRow = $this->dbManager->getSingleRow(
       'SELECT foldercontents_pk FROM foldercontents '.
-      'WHERE child_id = $1', [$childId]);
+      'WHERE child_id = $1 AND foldercontents_mode = $2', [$childId, $mode]);
     if (!$folderContentsRow) {
       return null;
     }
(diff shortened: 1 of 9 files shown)


GMishx:
> Some overall comments on using the group feature via the API:
> 
>     * `groupName` option is difficult to use and test in an automated context because group creation/management is not possible via the API

We have an issue for that and will be implementing it soon:  #1535

>     * uploads a file with the `groupName` header doesn't work if the group doesn't exists, the API call shall return a 404 with a meaningful message, e.g. "The group you requested doesn't exists"

API currently gives following response if the group does not exists on every endpoint. I am not sure why it is not working for you.
```json
{
    "code": 403,
    "message": "Provided group:123123 does not exist",
    "type": "ERROR"
}
```

> Some specific feedback to this PR:
> 
>     * using `page` as **query parameter** is not consistent with the paging options available for `/jobs` where **headers** are required

Thanks for pointing that out. Fixed it.

>     * an additional `page_size` option would also make sense to be aligned with other paginated API calls

Added as `limit` in **headers** to match `/jobs` endpoint.

> I'm preparing a PR for [fossology-python](https://github.com/fossology/fossology-python/pull/32) in parallel to catch up with the new features.

That's super-fast! You should also have a look on #1748 as it is a breaking change.

deveaud-m:
> We have an issue for that and will be implementing it soon: #1535

Great! I'll keep on testing non-existing groups for now. 

> API currently gives following response if the group does not exists on every endpoint. I am not sure why it is not working for you.

I overlooked this because it's not visible in the OpenAPI documentation.
I reworked all of my endpoint wrappers to handle 403.

> Thanks for pointing that out. Fixed it.

Works :+1: 

> That's super-fast! You should also have a look on #1748 as it is a breaking change.

Ok, will have a look at this too.

ag4ums (approved):
code looks good

Issue #1760, opened 2020-07-09
Dropping support for Debian 8 and PHP5


GMishx:
### Description

Since Debian 8 (Jessie) has reached it's EOL, we can remove it from the supported distro list.

Debian 8 was the only distro supported by FOSSology which was using PHP5 (PHP 7.4 is the latest release). Meaning that we can even drop support for PHP5 and put minimum requirement of PHP to 7.0 (because Ubuntu Xenial (16.04) still supports it).

Dropping PHP5 will allow upgrading of many composer dependencies and introducing static analysis for PHP using [PHPStan](https://github.com/phpstan/phpstan) (minimum requirement PHP 7.0).

In the same process, support for Ubuntu Focal (20.04) can be introduced which was released last month.

#### Steps

- [ ] Drop support for Debian Jessie from all dependency scripts.
- [ ] Include Ubuntu Focal in all dependency scripts.
- [ ] Update PHP requirement to PHP 7.0 in composer.json
- [ ] Update required composer packages.
- [ ] Introduce PHP Stan for static code analysis.

sjha2048:
Hi @GMishx , can I pick up this one?

GMishx:
Hello @sjha2048, it would be great.
I will assign the issue to you then.

PR #1765, opened 2020-07-14
update(org): drop debian 8 support and php 5

sjha2048:
## Description

- [x] Drop support for Debian Jessie from all dependency scripts.
- [ ] Include Ubuntu Focal in all dependency scripts.
- [x] Update PHP requirement to PHP 7.0 in composer.json
- [x] Update required composer packages.
- [x] Introduce PHP Stan for static code analysis.

Fixes: #1760

### DIFF
--- a/src/composer.json
+++ b/src/composer.json
@@ -10,35 +10,35 @@
         "without any warranty."
     ],
     "require" : {
-        "easyrdf/easyrdf" : "0.9.0",
+        "easyrdf/easyrdf" : "v0.9.1",
         "ext-gettext": "*",
         "ext-mbstring": "*",
         "ext-json" : "*",
         "ext-xml" : "*",
-        "flexsounds/slim-symfony-di-container": "1.0.3",
-        "firebase/php-jwt": "v5.0.0",
-        "monolog/monolog" : "1.24.0",
+        "flexsounds/slim-symfony-di-container": "v1.0.3",
+        "firebase/php-jwt": "v5.2.0",
+        "monolog/monolog" : "v1.25.4",
         "phpoffice/phpword" : "v0.17.0",
-        "slim/slim": "3.12.0",
-        "symfony/config" : "v3.4.26",
-        "symfony/http-foundation" : "v3.4.26",
-        "symfony/dependency-injection" : "v3.4.26",
-        "symfony/yaml" : "v3.4.26",
-        "twig/twig" : "v1.38.4",
+        "slim/slim": "v3.12.3",
+        "symfony/config" : "v3.4.42 || v3.3.6",
+        "symfony/http-foundation" : "v3.4.42 || v3.3.6",
+        "symfony/dependency-injection" : "v3.4.42 || v3.3.6",
+        "symfony/yaml" : "v3.4.42 || v3.3.6",
+        "twig/twig" : "v2.12.5",
         "twig/extensions" : "v1.5.4",
-        "guzzlehttp/guzzle": "6.3.3"
+        "guzzlehttp/guzzle": "v6.5.5"
     },
     "require-dev" : {
-        "doctrine/instantiator" : "1.0.5",
+        "doctrine/instantiator" : "v1.0.5",
         "ext-posix" : "*",
         "ext-sqlite3" : "*",
-        "hamcrest/hamcrest-php" : "v2.0.0",
-        "mockery/mockery" : "1.2.2",
-        "myclabs/deep-copy" : "1.7.0",
-        "php-coveralls/php-coveralls" : "v2.1.0",
+        "hamcrest/hamcrest-php" : "v2.0.1",
+        "mockery/mockery" : "v1.3.2",
+        "myclabs/deep-copy" : "v1.7.0",
+        "php-coveralls/php-coveralls" : "v2.2.0",
         "phpunit/phpunit" : "5.7.27 || 6.5.14",
-        "sebastian/phpcpd" : "3.0.1",
-        "squizlabs/php_codesniffer" : "3.4.1",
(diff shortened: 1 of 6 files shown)


sjha2048:
luckily all the dependencies were supported by PHP 7.0 in composer.json 
phpstan requires PHP ^7.1 for the latest version hence PHP stan 0.90 is used as of now for PHP 7.0

src/composer.json
```diff
@@ -53,7 +54,7 @@
     },
     "config" : {
         "platform" : {
-            "php" : "5.6"
+            "php" : "7.0.8"
```
GMishx, line 57: Is it possible to make more general like `7.0`? Because not every distro on PHP 7.0 will have 7.0.8.

utils/fo-installdeps
```diff
@@ -174,27 +170,24 @@ if [[ $RUNTIME ]]; then
             subversion git \
             dpkg-dev
          case "$CODENAME" in
-            jessie)
-               apt-get $YesOpt install postgresql-9.4 php5-pgsql libapache2-mod-php5 php5 php5-pgsql \
-                 php5-cli php5-curl heirloom-mailx libboost-program-options1.55.0 libboost-regex1.55.0 libicu52;;
             xenial)
-               apt-get $YesOpt install postgresql-9.5 php7.0 php7.0-pgsql libapache2-mod-php7.0 php7.0-pgsql \
+               apt-get $YesOpt install postgresql-9.5 php7.0 php7.0-pgsql libapache2-mod-php7.0 php7.0-pgsql php7.0-sqlite3 \
```
GMishx, line 174: Please remove sqlite from runtime dependency. It is required only for testing.

GMishx:
> luckily all the dependencies were supported by PHP 7.0 in composer.json just the addition of php-sqlite3 was required in runtime dependencies

The idea was to update the versions of dependencies. Currently many dependencies are on lower version as new version requires PHP 7.
And is there any particular reason to add php-sqlite3 dependency?

> phpstan requires PHP ^7.1 for the latest version hence PHP stan 0.90 is used as of now for PHP 7.0

Should work just good.

sjha2048:
> And is there any particular reason to add php-sqlite3 dependency?

without this composer was throwing the following error
```
Problem 1
    - The requested PHP extension ext-sqlite3 * is missing from your system. Install or enable PHP's sqlite3 extension.
```
shall I remove it?

> The idea was to update the versions of dependencies. Currently many dependencies are on lower version as new version requires PHP 7.

I have updated all the dependencies for PHP 7.0, but one strange thing is happening, dependencies like` symfony/config` needs to be downgraded for in order to work with PHP 7.0 but work absolutely fine for 7.0.8, can you have a look at this please?

src/composer.json
sjha2048: updated

GMishx:
> > And is there any particular reason to add php-sqlite3 dependency?
> 
> without this composer was throwing the following error
> 
> ```
> Problem 1
>     - The requested PHP extension ext-sqlite3 * is missing from your system. Install or enable PHP's sqlite3 extension.
> ```
> 
> shall I remove it?

If you run composer with `--no-dev` flag, you should not get this message. IMO, it would be better to have php-sqlite3 as a dev dependency and not runtime dependency. If you see in the install scripts, dev dependencies are ignored: https://github.com/fossology/fossology/blob/6502333cc1fe8c98b227d090f134593a4cefee7a/src/Makefile#L88

> > The idea was to update the versions of dependencies. Currently many dependencies are on lower version as new version requires PHP 7.
> 
> I have updated all the dependencies for PHP 7.0, but one strange thing is happening, dependencies like` symfony/config` needs to be downgraded for in order to work with PHP 7.0 but work absolutely fine for 7.0.8, can you have a look at this please?

Strange, but I just check that the version of PHP7.0 supported by both Ubuntu and Debian is PHP 7.0.33 and it appears to be the same case with CentOS and Fedora. It should be okay to tag PHP 7.0.8 in composer.json in that case.

GMishx:
Few more suggestions in terms of dependencies,

1. As per PR #1469, for `easyrdf/easyrdf`, the version `0.9.1` is not stable. Please revert back to `0.9.0`.
1. For `phpunit/phpunit`, you can remove version `5.7.27` and keep only `6.5.14`
1. Please tag `phpstan/phpstan` to `0.9.3`.

sjha2048:
I am not sure that `composer.lock` was supposed to be in the pull request, is that the case?

GMishx:
Yes @sjha2048, `composer.lock` should be in repo and since you are updating `composer.json`, lock file should be updated as well.

sjha2048:
can you review the changes now?

GMishx:
Sure. I will.
But your tests are failing. Can you please check? And please remove PHP 5 from Travis tests as this PR drops it's support.

sjha2048:
sure thing!

sjha2048:
```
1) Fossology\SpdxTwo\Test\SchedulerTest::testSpdxForNormalUploadtreeTable
Hamcrest\AssertionError: Expected: a value greater than <0>
     but: <0> was equal to <0>
```
any pointers for this? is this because of the PHP version in `composer.json`

GMishx:
Hello @sjha2048 , I was wondering if you are still working on this PR. Do you have any update or need any help?

sjha2048:
Hi @GMishx , it slipped off mind thank you for reminding me I'll start working on it again and revert back to you asap

sjha2048:
Planning to bifurcate this pull request into three parts for easy debugging

1. Dropping Jessie support 
2. Adding Focal Fossa Support 
3. Updating PHP version

hence closing this pull request.

PR #1766, opened 2020-07-16
fix(report): Don't group results with custom text

GMishx:
### Description

In case a same license has been identified in different files with different license text, the report shows only one entry with only one license text.

### Changes

1. Add the missing `"license"` parameter to `getCleared()`.

## How to test

1. In a package, identify 3 files with same license (say Public-domain).
1. In all files, change the **License Text** with 3 different text.
1. Generate the Unified report and check for the license.
    - It should show 3 entries for all the 3 license texts with different files.


shaheemazmalmmd:
@GMishx : Can you please add the same sorting for fetching acknowledgements ?

GMishx:
> @GMishx : Can you please add the same sorting for fetching acknowledgements ?

Hello @shaheemazmalmmd, I've pushed the changes. Please check

shaheemazmalmmd:
Tested working as expected.

PR #1768, opened 2020-07-20
feat(build): Get build version number from git

GMishx:
## Description

Use the information provided by `git describe --tags` and generate the build version number.

The `git describe --tags` also displays the number of commits since last tags which is used here to get the build number. So let's say, last tag was 3.8.1 and there are 88 commits since then, the version now will be 3.8.1.88.
(See the **Example** in [man git-describe](https://linux.die.net/man/1/git-describe) for info on format)

Also, created a new script `utils/fo-debuild` to generate Debian packages and source tar with option to skip signing of packages with `--no-sign` flag.

### Changes

1. Get the no. of commits from `git describe --tags` using `sed` and generate build number using it.
1. Fix the `fo-mktar` to consider `.gitignore` and use new version.
1. New script `fo-debuild` to generate Debian packages.

## How to test

1. Check if `make VERSIONFILE` generates `VERSION` without errors and the version in `VERSION` file is correct.
1. Check if the script works even on tags (no commits since tag, no modification in version).
1. Check if `utils/fo-debuild` works with and without `--no-sign` option.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -60,14 +60,10 @@ jobs:
         echo PACKAGE_NAME=$(echo "FOSSology-${VERSION}-$(lsb_release -si | tr [:upper:] [:lower:])-$(lsb_release -sc | tr [:upper:] [:lower:]).tar.gz") >> $GITHUB_ENV
 
     - name: Build Debs
-      run: dpkg-buildpackage -I
+      run: ./utils/fo-debuild --no-sign
 
-    - name: Package
-      run: |
-        find .. -type f -name "*-dbgsym*" -exec rm -rf {} \;
-        mkdir packages
-        mv ../*.deb packages/
-        tar -czvf ${PACKAGE_NAME} ./packages
+    - name: Rename package
+      run: mv fossology_*.tar.gz ${PACKAGE_NAME}
 
     - name: Upload Release Asset
       id: upload-release-asset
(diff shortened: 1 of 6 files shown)


GMishx:
(from discussion): Need to add a new field called "build" and add the number there instead of changing the version number. Also, change the "Version" to "Release".

Doing so, we can still control the version number from git tagging/release in case of major bug fix (similar to release 3.8.1) and get the human readable way to show increment in builds.

ag4ums:
update : @GMishx need to set the flag for pipefail

GMishx:
> update : @GMishx need to set the flag for pipefail

Thanks for the update @ag4ums , pushed the changes. Please check.

mcjaeger:
> (from discussion): Need to add a new field called "build" and add the number there instead of changing the version number. Also, change the "Version" to "Release".
> 
> Doing so, we can still control the version number from git tagging/release in case of major bug fix (similar to release 3.8.1) and get the human readable way to show increment in builds.

"major bug fix" means incrementing the middle of last version number of the three? (because the iddle is called minor version number already and it could be understood that a "major bugfix" could result in a increment in the minor version


I think we are already working with the intermediate master states and thus there should be a version bumber. I am not sure if a separate build variable would solve the problem. Because also the issue that we have discussed was about the packages generated. I found the following quite useful:

https://en.wikipedia.org/wiki/Software_versioning

one scheme is there: `major.minor[.build[.revision]] (example: 1.2.12.102)`

GMishx:
I have changed the implementation to add the number of commits at the end as `major.minor.build.revision`. This should not cause issues while tagging fix patches like `3.9.1` as it will endup with `3.9.1.0`

ag4ums (approved):
Changes loos good, working as expected.

PR #1780, opened 2020-08-18
fix(licenseref): handle errors license errors

GMishx:
## Description

While installing the fossology, there might be some conflicts while importing licenses. This PR tries to catch them and print them in a neat format instead of throwing an exception.

### Changes

1. Store all failed modifications and print them in a log format.
2. While merging candidate license, check if the same license exists:
    a. Same shortname and md5, then delete the candidate and update it's references with main license.
    b. If the same license exists but with different shortname or md5, skip it.

## How to test

1. Add a new candidate license with shortname and license text which exists in main license list.
1. Add a new candidate license with a new shortname (exists in `licenseRef.json` but not in `license_ref`) but same license text which exists in main license list.
1. Add a new candidate license with same shortname which exists in main license list but different license text.
1. Add few decisions with these new candidate licenses.
1. Delete one license from main list and update another license with the deleted license's text (to create duplicate `rf_md5`).
1. Running the `fo-postinstall` should:
    a. Do not throw exception, instead print a text output.
    b. The license from pt. 1 should be merged but the decisions should be preserved (`rf_fk` updated).
    c. The licenses from pt. 2 and 3 should not be updated in any manner.
    d. The license from pt. 5 should cause a print of error log.

### DIFF
--- a/install/fossinit.php
+++ b/install/fossinit.php
@@ -222,6 +222,8 @@ else
   print "Database schema update completed successfully.\n";
 }
 
+$FAILED_LICENSE_IMPORT = array();
+
 /* initialize the license_ref table */
 if ($UpdateLiceneseRef)
 {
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):
Code looks Good.

ag4ums (approved):
changes looks good and working as expected

Issue #1786, opened 2020-08-23
Create and show Error,Fatal and warning log counter on Fossdash-Dashboard


darshank15:
### Description

- Include code to maintain a counter for a number of **ERROR**, **FATAL**, and **WARNING** of fossology.
- Need to figure out the central logging mechanism.
- Fetch these data from our fossdash publisher script and push it into influxDB.
- Creates these counter Stats charts on the grafana dashboard.

darshank15:
++ @NicolasToussaint @GMishx,

NicolasToussaint:
@darshank15, two separate remarks, not anwsering your auqtion, I know :+1: 

1. Not sure how you were seeing things, but I think it would make more sense to separate the error counts from the scheduler and PHP (and maybe database?)
They could then be aggregated in Grafana to get a consolidated view.


1.  Somehow the graph show a full green bar when the value is zero... couldn't find why :-/
![image](https://user-images.githubusercontent.com/22956329/91068487-a63e2580-e634-11ea-9365-046e0ea1f6b3.png)

NicolasToussaint:
Looking at the source code, I'm not sure anymore if tracking the error count (this way at least) is worth the development effort, deep changes in the scheduler and particularly nomos source code, and the impact on performance (writing to an extra file at every log)...

Maybe error detection, and self-healing is a topic that should be tackled more generally within Fossology, and show data on the existing Dashboard -> Overview page before being integrated to Fossdash.

PR #1788, opened 2020-08-26
Central logging mechanism in Fossology and easy to count different errors

darshank15:
## Description
This was a POC for fossology proposed as part of GSOC-2020.
As of now, we don't have a single and central logging mechanism.
As per my knowledge, there is three different logging file there as per uses.
* `src/scheduler/agent/logging.h`
* `src/lib/c/libfossscheduler.h`
* `src/lib/c/libfossdbmanager.c`
* `src/nomos/agent/standalone.h`

Also at some other place, we have defined our custom login MACROS.
this central logging mechanism makes it easy to maintain different log counts
If we want to have a count of number of error, fatal, warning log of fossology,.

Related Issue :  https://github.com/fossology/fossology/issues/1786


darshank15:
This doesn't requires to be merge. Work is not completed. We need extend to do centralised logging in fossology first. This is future task.

GMishx:
Closing the PR as it does not solve #1786

The issue can be solved in another PR.

Please reopen if required.

PR #1793, opened 2020-08-29
fix(licenseRef): Fix type in array_map

GMishx:
## Description

Fix typo in `array_map` call.

### Changes

Fix position of `)` in `fossinit.php`


shaheemazmalmmd (approved):

PR #1794, opened 2020-09-01
feat(utils): Filter inputs for unicode ctrl chars

GMishx:
## Description

At every possible place where user input is taken as text and the input is written to a file (reports), remove non-printable unicode control characters.

### Changes

1. New function `replaceUnicodeControlChar` is added to `StringOperation` class.

## How to test

1. In an upload, add some license text and acknowledgements with invalid control characters.
1. Also, add some valid unicode characters like &copy; and text from languages with non-latin script.
1. Check if the invalid characters are removed before inserting the data in DB.
1. Perform same operation on License text update, new license creation, acknowledgement with bulk finding, etc.

Example of test strings:
```
4.	“IDENT”.  some text and it’s
copyrights ©
Agreement.

invals
```
```
There is something below this line.

Do copy with care.
```
```
नमस्ते
你好
こんにちは
```

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -33,6 +33,7 @@ use Fossology\Lib\Proxy\UploadTreeProxy;
 use Fossology\Lib\Proxy\ScanJobProxy;
 use Monolog\Logger;
 use Fossology\Lib\Data\AgentRef;
+use Fossology\Lib\Util\StringOperation;
 
 class ClearingDao
 {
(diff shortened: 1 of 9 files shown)


shaheemazmalmmd (changes requested):
While adding the licenses and obligations for testing i have found this notices, Can you please fix them ?

GMishx:
> While adding the licenses and obligations for testing i have found this notices, Can you please fix them ?

Done @shaheemazmalmmd

PR #1796, opened 2020-09-02
feat(nomos): Apache detection

GMishx:
## Description

If the source contain words **Apache =SOME= license** and not **license =SOME= Apache**, nomos fails to detect it.

### Changes

Additional check for such cases added in `parse.c` with test file `Apache-2.0-nourl.txt` (just removal of url and license header made nomos fail in detecting actual license text).


ag4ums (approved):
code looks good, working as expected

PR #1797, opened 2020-09-03
fix(ClearingDao): Get uploadtree table name

GMishx:
## Description

`getItemTreeBounds()` returned default `uploadtree` table causing wrong join with copyright table.

### Changes

1. Get the correct uploadtree table name and recreate the `$itemTreeBounds` to fix it.
1. As a fail safe, check for `uploadtree` in ClearingDao as well.


shaheemazmalmmd (approved):
Code looks good. needs test.

Issue #1280, opened 2019-02-06
Improve Search option to limit the results to one project


piotreke12:
### Description

Trying to find files in Search window, it would be nice to be able to limit search scope to only one upload. Now, in case of many uploads visible for one user, search results could be huge and uneasy to read. 

#### Steps followed and expected result

- User should be able to choose upload in which he wants to search for files
- User should also be able to search in all visible uploads if he wants to
- Search results should be limited to upload choosed by user

PR #1804, opened 2020-09-22
feat(search) - possibility to search in selected upload only

piotreke12:
This feature adds possibility to search only in specified upload while currently it is possible to do it in all available uploads

Closes: #1280 

## Description

Search option for all available uplads only is not very useful when there is more than 3-5 uploads user has access to.
This feature adds possibility to choose in which upload user wants to search in.
List od available uploads for user is dynamically loaded from DB.

### Changes

- Added option-field on template `src/www/ui/search-helper.php`  filled with all uploads user has access to (plus option to choose to search in All uploads)
- Added method for `src/lib/php/Dao/UploadDao.php` to retrieve all uploads 
- modified `src/www/ui/search-helper.php#GetResults` to filter search results only for selected upload

## How to test

Search results should be shown only for Upload selected from dropdown menu.
For --All uploads -- results from all uploads should be shown (as previously, before this feature)

### DIFF
--- a/src/lib/php/Dao/UploadDao.php
+++ b/src/lib/php/Dao/UploadDao.php
@@ -101,6 +101,20 @@ class UploadDao
     return $row ? Upload::createFromTable($row) : null;
   }
 
+  public function getActiveUploadsArray()
+  {
+    $stmt = __METHOD__;
+    $queryResult = $this->dbManager->getRows("SELECT * FROM upload where pfile_fk IS NOT NULL",
+        array(), $stmt);
+
+    $results = array();
+    foreach ($queryResult as $row) {
+      $results[] = Upload::createFromTable($row);
+    }
+
+    return $results;
+  }
+
   /**
    * @param $itemId
    * @param $uploadTreeTableName
--- a/src/www/ui/search-helper.php
+++ b/src/www/ui/search-helper.php
@@ -36,7 +36,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
  *         contains uploadtree_pk, parent, upload_fk, pfile_fk, ufile_mode, and
  *         ufile_name
  */
-function GetResults($Item, $Filename, $tag, $Page, $SizeMin, $SizeMax, $searchtype,
+function GetResults($Item, $Filename, $Upload, $tag, $Page, $SizeMin, $SizeMax, $searchtype,
                     $License, $Copyright, $uploadDao, $groupID, $PG_CONN)
 {
   $MaxPerPage  = 100;  /* maximum number of result items per page */
@@ -163,6 +163,14 @@ function GetResults($Item, $Filename, $tag, $Page, $SizeMin, $SizeMax, $searchty
       $SQL .= " AND";
     }
     $SQL .= " ufile_name ilike '". pg_escape_string($Filename) . "'";
+    $NeedAnd=1;
+  }
+
+  if ($Upload != 0) {
+    if ($NeedAnd) {
+      $SQL .= " AND";
+    }
+    $SQL .= " upload_fk = ". pg_escape_string($Upload) . "";
     $NeedAnd=1;
   }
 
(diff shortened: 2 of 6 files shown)


src/lib/php/Dao/UploadDao.php
```diff
@@ -101,6 +101,15 @@ public function getUpload($uploadId)
     return $row ? Upload::createFromTable($row) : null;
   }
 
+  public function getActiveUploadsArray()
+  {
+    $stmt = __METHOD__;
+    $result = $this->dbManager->getRows("SELECT * FROM upload where pfile_fk IS NOT NULL",
+        array(), $stmt);
+
```
GMishx, line 109: As a suggestion, how do you think about returning an array of [Upload](https://github.com/fossology/fossology/blob/b7a846ca391b091407b8173ec0f74a1dbab5bc46/src/lib/php/Data/Upload/Upload.php#L38) objects?

src/www/ui/search.php
```diff
@@ -48,6 +48,22 @@ function PostInitialize()
     return $this->State;
   }
 
+  function loadUploads()
+  {
+    $allUploadsPre = $this->uploadDao->getActiveUploadsArray();
+    $filteredUploadsList = array();
+
+    foreach($allUploadsPre as $row)
```
GMishx, line 56: How about using `array_filter()` instead of a `foreach` loop?

src/www/ui/search.php
```diff
@@ -141,9 +161,26 @@ function Output()
...
+    if ($Upload != 0) 
+    {
+      $CriteriaCount++;
+      $GETvars .= "&upload=" . urlencode($Upload);
+      $this->vars["Upload"] = $Upload;
+      foreach($uploadsArray as $row)
+      {
+        if($row['upload_pk'] == $Upload)
+        {
+          $SelectedUploadName = $row['upload_filename'] . " from " .  substr($row['upload_ts'],0,-10);
```
GMishx, line 175: How about using the `Convert2BrowserTime()` from #1667?

GMishx (changes requested):
Few suggestions.

GMishx:
There are few linting failures

image

piotreke12:
> There are few linting failures
> 
>
Yes, sorry - forgot to push updated version - done

src/lib/php/Dao/UploadDao.php
piotreke12: Thanks for this tip - corected as you suggested

src/www/ui/search.php
piotreke12: Changed

src/www/ui/search.php
piotreke12: Changed

src/www/ui/search-helper.php
```diff
@@ -36,7 +36,7 @@
  *         contains uploadtree_pk, parent, upload_fk, pfile_fk, ufile_mode, and
  *         ufile_name
  */
-function GetResults($Item, $Filename, $tag, $Page, $SizeMin, $SizeMax, $searchtype,
+function GetResults($Item, $Filename, $Upload, $tag, $Page, $SizeMin, $SizeMax, $searchtype,
```
GMishx, line 39: I just realized that this function is used by REST API as well and this will cause failure for it.

Can you please add additional header for the `uploadId` in `SearchController`?

piotreke12: Good point, I completely forgot about RAST API.
Of course I'll add it.

The thing is that when there is no uploadId supplied, SearchController should send `0` as `$Upload` parameter.
Unless you have another method, I will do it this way.

piotreke12: Done

GMishx: Awesome!

GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working fine.

PR #1807, opened 2020-09-30
fix(gcc-10): Fix errors and warnings

GMishx:
## Description

Fix warning and errors reported by GCC-10 which will be shipped by next Debian release (https://packages.debian.org/bullseye/gcc)

### Changes

Nomos and monk: Make shared variables are extern
Others: Fix string warnings

## How to test

Build fossology on various GCC versions.


shaheemazmalmmd (approved):

PR #1810, opened 2020-10-05
feat(password): Create password policy

GMishx:
## Description

Allow admins to enforce password policies from Customize page. Following options can be chosen to generate policy:
1. Enable/disable policy check.
2. Fixing length of passwords (min and max).
3. Requiring minimum one of lowercase character.
4. Requiring minimum one of uppercase character.
5. Requiring minimum one digit.
6. Requiring minimum one of special characters from a predefined set.

This policy is used to generate RegEx and the password will be checked while creating a new user or editing a user.

The checks are added to the backend as well as frontend.

Also, if password policy is enabled, do not allow empty passwords for new users.

### Changes

1. `common-auth.php`: Two new functions `generate_password_policy()` and `generate_password_policy_string()` to generate policy regex and string respectively.
1. `admin-config.php`: New config type `CONFIG_TYPE_BOOL` to generate checkbox.
1. `password-policy-check.js.twig`: JS functions to handle checks in UI.
1. `user-add.php` and `user-edit.php`: Changes required for frontend checks.

### Screenshots

1. Configuration
image
1. Check at frontend
image

## How to test
1. Make changes to the policy from Customize page and check if the policy string and regex changes.
1. Disable policy and check if everything works.
1. While trying to submit non conformant password, UI should report error.
1. User edit page should work if password was not changed.

### DIFF
--- a/src/lib/php/Dao/UploadPermissionDao.php
+++ b/src/lib/php/Dao/UploadPermissionDao.php
@@ -60,7 +60,10 @@ class UploadPermissionDao
 
     $perm = $this->dbManager->getSingleRow('SELECT perm FROM perm_upload WHERE upload_fk=$1 AND group_fk=$2',
         array($uploadId, $groupId), __METHOD__);
-    return $perm['perm']>=Auth::PERM_WRITE;
+    if (! empty($perm) && array_key_exists('perm', $perm)) {
+      return $perm['perm']>=Auth::PERM_WRITE;
+    }
+    return false;
   }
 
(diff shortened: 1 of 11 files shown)


shaheemazmalmmd:
Please do not match the passwords if the first password column doesn't match the password policy.
screenShot29

GMishx:
> Please do not match the passwords if the first password column doesn't match the password policy.

Done, please check.

shaheemazmalmmd (approved):

PR #1815, opened 2020-10-13
feat(decisions): add new decision type non-functional

### LINKED ISSUE #1114: Support for Distributing Non-Distributing Files
Right now, persons use the irrelevant feature for marking files that do not go into the distribution. Intention of this feature is to save work, when analyzing the files. However, for some cases, the files do not go into the distribution, but are relevant for licensing. 

Therefore a new clearing decision type "Non-Functional", "Not-Distributing" would be required.

shaheemazmalmmd:
### Changes

add new decision "non functional" to decision types

## How to test

* Upload a component.
* Apply Non functional decision from both single file view and tree view.
* Check if the results are attached properly to the files.
* Check unified report for non-functional files and comments.

Closes #1114

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -795,7 +795,7 @@ INSERT INTO clearing_decision (
    * @param int $groupId
    * @param int $userId
    */
-  public function markDirectoryAsDecisionType(ItemTreeBounds $itemTreeBounds, $groupId, $userId, $isRemoval, $decisionMark)
+  public function markDirectoryAsDecisionType(ItemTreeBounds $itemTreeBounds, $groupId, $userId, $decisionMark)
   {
     $decisionMark = $this->getDecisionType($decisionMark);
     $this->markDirectoryAsDecisionTypeRec($itemTreeBounds, $groupId, $userId, false, $decisionMark);
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -1,6 +1,6 @@
 <?php
 /*
-Copyright (C) 2014-2018,2020,2022, Siemens AG
+Copyright (C) 2014-2018,2020-2022, Siemens AG
 Author: Johannes Najjar
 
 This program is free software; you can redistribute it and/or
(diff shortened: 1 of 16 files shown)


NicolasToussaint:
Hi @shaheemazmalmmd ,

Can you share on the usage behind the new "non-functional"' decision type ?

shaheemazmalmmd:
Hello @NicolasToussaint ,

We often have the situation that FOSSology (correctly!) detects licenses
in files, but these are are not part of the source code that is really used
by the component.

Examples : 
build files
documentation files
unit test files
examples

Today we can only decide to either list them in the clearing report (type = identified)
or do not list them by marking them a irrelevant. But irrelevant does not really describe
the situation. We would need something that allows use to add these licenses/files in the
clearing report.

NicolasToussaint:
Thanks

I couldn't find place where all decision types are described, and if they end up in the reports or not  (and, by the way, are all report types share the same behaviour ?)

[This page](https://www.fossology.org/get-started/basic-workflow/) explains the first four types.

"Do not use" is missing, but purpose is explained [in this issue](https://github.com/fossology/fossology/pull/1560#issuecomment-581778559)  (and the tooltip was not updated, in 3.8.0)
> Actually the new Decision type is to mark file which should not be used in the source code. ex: dll files, some jar files etc.
> "Do not use" decision type will be listed in the unified report in separate section. With comments.

shaheemazmalmmd:
Actually not all reports imply same behavior but unified report. i think the wiki needs to be updated. and for tooltip lets do it in this release.

GMishx:
@shaheemazmalmmd , can you please update the tooltip as requested (recently done in #2142).

#1114 is the related issue and describes the behavior of 'Non-functional' decision.

GMishx:
Should rebase after merge of #2145

shaheemazmalmmd:
Found some issue while testing this: 

* Upload a package **XNAME1**
* Go to tree view and apply decision as "DO NOT USE" using edit.
* From the same tree view remove "Non Functional decision"
* "DO NOT USE" is getting removed.

GMishx:
> Found some issue while testing this:
> 
>     * Upload a package **XNAME1**
> 
>     * Go to tree view and apply decision as "DO NOT USE" using edit.
> 
>     * From the same tree view remove "Non Functional decision"
> 
>     * "DO NOT USE" is getting removed.

Should be resolved now @shaheemazmalmmd . Please check once.

shaheemazmalmmd:
> > Found some issue while testing this:
> > ```
> > * Upload a package **XNAME1**
> > 
> > * Go to tree view and apply decision as "DO NOT USE" using edit.
> > 
> > * From the same tree view remove "Non Functional decision"
> > 
> > * "DO NOT USE" is getting removed.
> > ```
> 
> Should be resolved now @shaheemazmalmmd . Please check once.

Tested. Works as expected. Perfect !

ag4ums:
code looks good...  working as expected... need non-functional entry for clixml report as well

NicolasToussaint:
Good improvement, thanks :-)

I'm wondering though if and where  the decision types are documented.
This intro page in the [Wiki (in Creating a Clearing Decision section)](https://www.fossology.org/get-started/basic-workflow/) describes the first few ones only.

could this page, or another, be updated to reflect the new _decision types_ ?

PR #1822, opened 2020-10-22
feat(copyright): Search and replace with regex

GMishx:
## Description

Add feature to search using regex, create placeholders and replace in Copyright and sister UIs. Same feature can be used to search for specific terms and replace.
A test table has been added as well where user can see sample results before comiting to it.

For example, you can search for `Copyright(*) rights reserved(*)` which can match anything like:
```
Copyright 2012-2016 Example Inc, All rights reserved
Copyright 2018 Example Inc, All rights reserved some small text snipit
Copyright 2014-2019 Example Inc, All rights reserved and some random text picked up by the agent
```

And replacement can be done with `Copyright$1 rights reserved` (notice the `$1` placeholder) which will transform all text to:
```
Copyright 2012-2016 Example Inc, All rights reserved
Copyright 2018 Example Inc, All rights reserved
Copyright 2014-2019 Example Inc, All rights reserved
```

Also, another simple search like `Example Inc` and replacement to `Example Incorporation` will transform text to:
```
Copyright 2012-2016 Example Incorporation, All rights reserved
Copyright 2018 Example Incorporation, All rights reserved some small text snipit
Copyright 2014-2019 Example Incorporation, All rights reserved and some random text picked up by the agent
```

### Changes

1. Add new buttons to `HistogramBase.php`
1. Add JS calls to `*-hist.php`
1. Add helper functions to `copyrighthist_scripts.html.twig`
1. Add corresponding calls to `histTable.js.twig`

## How to test

1. On an upload, search for a small set of copyrights which needs to be fixed.
1. Create a regex in Advance search box and get the corresponding sample replace text using `Create replacement text` button.
1. Check results with `Test replacement` button.
1. Submit the result and check if everything works as expected.
1. Any other regex symbols should not have any effect in Advance search box.

### DIFF
--- a/src/copyright/ui/HistogramBase.php
+++ b/src/copyright/ui/HistogramBase.php
@@ -79,19 +79,35 @@ abstract class HistogramBase extends FO_Plugin {
       $typeDescriptor = $description;
     }
     $output = "<h4>Activated $typeDescriptor statements:</h4>
-               <div><table border=1 width='100%' id='copyright".$type."'></table></div>
-               <br/><br/>
-               <div>
-                 <span>
-                   Replace: <input type='text' id='replaceText".$type."' style='width:80%'>
-                 </span>
-               <br/><br/>
-               <a style='cursor: pointer; margin-left:10px;' id='replaceSelected".$type."' class='buttonLink'>Mark selected rows for replace</a>
-               <a style='cursor: pointer; margin-left:10px;' id='deleteSelected".$type."' class='buttonLink'>Mark selected rows for deletion</a>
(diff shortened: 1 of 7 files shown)


GMishx:
Right now, if there is a `<` or `>` in a copyright which is being replaced, their corresponding HTML entities are saved instead (`&lt;`, `&rt;`)

Also, there is an idea to allow empty text in the replace text box so a particular text can be deleted.

shaheemazmalmmd (approved):
Code looks good. Tested Works as expected

PR #1831, opened 2020-11-09
fix(ununpack): Correct the mimetype for deb files

GMishx:
## Description

The mimetype for .deb Debian packages was wrong, added the new mimetype.

### Changes

Added the new mimetype for `.deb` files (`application/vnd.debian.binary-package`).

## How to test

Upload a `.deb` package, it should be unpacked.


ag4ums (approved):
Code looks good. 
working as expected

PR #1838, opened 2020-11-26
chor(release): Release changelog for 3.9.0

GMishx:
Change log updates for release 3.9.0

PR #1839, opened 2020-11-27
fix(spasht): Fix advance search

GMishx:
## Description

The data type used to store coordinates was changed from array to an object which was causing advance search to fail.

Check https://github.com/fossology/fossology/pull/1838#issuecomment-734678329 for logs.

### Changes

Change the data access in `checkAdvanceSearch()`.

## How to test

1. Search for a package in Spasht tab.
1. Do an advance search for a package in Spasht tab.
1. Schedule Spasht agent.


mcjaeger (approved):
code looks OK, works with advance search

PR #1851, opened 2020-12-20
refactor(login): updated password hashing algorithm

### LINKED ISSUE #841: Update password hashing algorithm
SHA-1 was officially deprecated by NIST in 2011
<b>src/www/ui/core-auth.php</b>
``` php
if (!empty($row['user_seed']) && !empty($row['user_pass']))
{
      $passwordHash = sha1($row['user_seed'] . $password);
      if (strcmp($passwordHash, $row['user_pass']) != 0)
      {
        return false;
      }
}
```

Aman-Codes:
## Description

Currently fossology uses SHA-1 which is now officially deprecated. 
Also as per the [official website](https://www.php.net/manual/en/function.password-hash.php), the salt option of PASSWORD_BCRYPT has been deprecated as of PHP 7.0.0

### Changes

Updated password hashing algorithm from SHA1 to PASSWORD_DEFAULT (Use the bcrypt algorithm)

Closes #841

### DIFF
--- a/src/cli/fo_usergroup.php
+++ b/src/cli/fo_usergroup.php
@@ -72,8 +72,8 @@ if ($user !== false) {
 
 if ($uName && !$user) {
   $pass = array_key_exists('upasswd', $opts) ? $opts['upasswd'] : '';
-  $seed = rand() . rand();
-  $hash = sha1($seed . $pass);
+  $options = array('cost' => 10);
+  $hash = password_hash($pass, PASSWORD_DEFAULT, $options);
   $desc = 'created via cli';
   $perm = array_key_exists('accesslvl', $opts) ? intval($opts['accesslvl']) : 0;
   if (array_key_exists('folder', $opts)) {
@@ -89,7 +89,7 @@ if ($uName && !$user) {
   }
   $agentList = userAgents();
   $email = $emailNotify = '';
-  add_user($uName, $desc, $seed, $hash, $perm, $email, $emailNotify, $agentList, $folderid);
+  add_user($uName, $desc, $hash, $perm, $email, $emailNotify, $agentList, $folderid);
   $user = $userDao->getUserByName($uName);
   print "added user $uName\n";
 }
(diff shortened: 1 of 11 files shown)


Aman-Codes:
Can you guide me which function to call in below 2 cases to save new hash in database
[here](https://github.com/Aman-Codes/fossology/blob/login/src/www/ui/core-auth.php#L316) and [here](https://github.com/Aman-Codes/fossology/blob/login/src/lib/php/common-auth.php#L167)

GMishx:
> Can you guide me which function to call in below 2 cases to save new hash in database
> [here](https://github.com/Aman-Codes/fossology/blob/login/src/www/ui/core-auth.php#L316) and [here](https://github.com/Aman-Codes/fossology/blob/login/src/lib/php/common-auth.php#L167)

Hey, unfortunately there is not function right now to update passwords. You can create a new function in `src/lib/php/common-users.php` and use that (note, the function will be included automatically in all PHP files).

install/db/dbmigrate_2.1-2.2.php
```diff
@@ -217,4 +217,4 @@ function Migrate_21_22($Verbose)
   pg_free_result($result);  
 
   return 0;  // success
-} // Migrate_21_22
+} // Migrate_21_22
```
GMishx, line 220: Please revert unnecessary change.

GMishx (changes requested):

install/db/dbmigrate_2.1-2.2.php
Aman-Codes: Did the change

GMishx (approved):

GMishx:
FOSSology allows users without passwords and to accomplish that, it checks for empty `seed`.
https://github.com/fossology/fossology/blob/479322e49e87f94f41d846452d6167ec85f63ea9/src/www/ui/core-auth.php#L310

So @Aman-Codes, please put some default values like `Seed` for the `user_seed` whenever the password is either updated or a new user is created. With the current implementation, if I edit an account and change it's password, I can login with that account without a password as seed is empty for it.

Aman-Codes:
@GMishx  Updated the default seed value

GMishx (approved):
Tested, working as expected.

PR #1735, opened 2020-06-08
test(rest): Unit test cases for REST API

GMishx:
## Description

Adding unit test cases for REST API.

### Changes

1. Some changes are done in `Contollers` to use Helper class to get userID and groupID.
1. Created new folder `src/www/ui_tests/api/` containing all the test cases.
1. Added test suite for REST API to `src/phpunit.xml`.

## How to test

1. Run `make test-www` or `src/vendor/bin/phpunit -csrc/phpunit.xml --testsuite="Fossology PhpUnit Test Suite"` or check the Travis log.

Partial work on #1590

PR #1854, opened 2020-12-22
feat(rest): get copyright info for file hash

mmurto:
## Description

Modify REST API's `/filesearch` endpoint to export copyright information in addition to license data. 

### Changes

Added list of copyright statements to the findings object returned by the API. I'm not super familiar with the copyright aspects of Fossology, so would any feedback on the draft would be very welcome. I'm only querying the `copyright` table, should `copyright_decision` also be queried? 

## How to test

Query `/filesearch` endpoint with a hash of a source file uploaded. Should return the copyright statements of the file.

### DIFF
--- a/src/lib/php/Dao/PfileDao.php
+++ b/src/lib/php/Dao/PfileDao.php
@@ -197,5 +197,31 @@ WHERE rnk = 1 AND removed = false AND decision_type = " .
     }
     return false;
   }
+
+  /**
+   * Get the list of copyrights for given pfile
+   * @param integer $pfileId Pfile to search
+   * @return array Array of copyrights found and not disabled
+   */
+  public function getCopyright($pfileId)
+  {
+    $statement = __METHOD__ . ".getCopyright";
+    $sql = "SELECT content " .
+      "FROM copyright " .
+      "WHERE (pfile_fk = $1) AND (is_enabled = TRUE) " .
+      "UNION " .
+      "SELECT textfinding " .
+      "FROM copyright_decision " .
+      "WHERE (pfile_fk = $1) AND (is_enabled = TRUE);";
+    $params = [$pfileId];
+    $rows = $this->dbManager->getRows($sql, $params, $statement);
+    if (!empty($rows) && array_key_exists('content', $rows[0])) {
+      $copyright = array_column($rows, 'content');
+      natcasesort($copyright);
+      return array_values($copyright);
+    } else {
+      return [];
+    }
+  }
 }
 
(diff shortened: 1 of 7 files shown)


GMishx:
The PR looks super helpful. Couple of feedbacks:
1. The `copyright` table holds entries unique to a trio of `pfile_fk`, `hash` and `agent_fk`. So with the current query, you can get duplicate copyrights for the given `pfile_fk` if it has been scanned by multiple agents in different uploads. It will be better to add `DISTINCT` to the query or handle it in PHP.
2. The table` copyright_decision` holds manually added copyrights (they are unique to just `hash` and `pfile_fk`). It will be helpful to get those results as well. Would adding another key in the `Findings` model to hold manually added copyrights sounds useful or merging both results is okay?

mmurto:
Thanks for the feedback @GMishx! I changed the query to account for both. Regarding 2, I think merging of the results makes more sense, as the the `copyright` table also holds some decisions in the form of `is_enabled`. If it was purely scanner results and all manual decisions were in `copyright_decisions`, splitting them to two keys in the `Findings` model would probably make more sense.

GMishx:
The idea sounds reasonable. Can you please update the patch version of the API in `openapi.yaml` to denote a change?

mmurto:
Sure! What's the versioning scheme for the API? Are breaking changes denoted somehow, and do new keys count as breaking? They're probably often fine, but might break if deserializing the responses with strict rules.

GMishx:
Currently any addition keys is not counted as breaking changes so an update from `1.1.1` to `1.1.2` will be fine.
Any removal will be considered as breaking and should update the minor version, e.g. from `1.1.1` to `1.2.0`.

GMishx:
Is the PR is ready to be reviewed?

mmurto:
Sure, haven't written any tests though!

GMishx:
Oh, I shall keep the WIP label then.

mmurto:
The PR is now ready for review.

GMishx (approved):

mmurto:
Are there some specifics I should write tests for while https://github.com/fossology/fossology/pull/1735 is not yet merged?

GMishx (approved):
Tested, working as expected.

GMishx:
> Are there some specifics I should write tests for while #1735 is not yet merged?

Hello @mmurto , I will be better to write the test cases after #1735 is merged. It will create merge conflicts if done in other branches.
That said, I should rebase the branch as there are many changes in REST API since the PR was open.

PR #1857, opened 2020-12-23
fix(ack): add missing uploadid in getresults function

shaheemazmalmmd:
## Description

Include uploadId in GetResults function to search for notice file.

## How to test

* Upload a package with notice file.
* Select notice file as acknowledgement while clearing.


GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

PR #1861, opened 2020-12-31
fix(db): add indexes to pfile on sha1 and sha256

### LINKED ISSUE #1859: No indexes in pfile table on sha1 or sha256
### Description

Table `pfile` is not indexed on `pfile_sha1` or `pfile_sha256`, making API queries with hash values extremely slow. The indexes were supposed to be added with https://github.com/fossology/fossology/pull/1748, but the index created by the UNIQUE constraints seems to only be effective on the first column.

Manual indexes on `pfile_sha1` and `pfile_sha256` fix the issue.

#### How to reproduce

Run `SELECT * FROM pfile WHERE pfile_md5 = 'ABC';` and `SELECT * FROM pfile WHERE pfile_sha1 = 'ABC';` and compare the analysis.

Select on `pfile_md5`, finished in 0.03 ms:

```
Index Scan using pfile_md5_sha1_sha256_size_ukey on public.pfile as pfile (cost=0.54..8.56 rows=1 width=155) (actual=0.03..0.03 rows=0 loops=1)
Index Cond: (pfile.pfile_md5 = 'ABC'::bpchar)
```

Select on `pfile_sha1`, finished in 18.144 ms:

```
Seq Scan on public.pfile as pfile (cost=0..4029.25 rows=1 width=155) (actual=18.144..18.144 rows=0 loops=1)
Filter: (pfile.pfile_sha1 = 'ABC'::bpchar)
Rows Removed by Filter: 109540
```

mmurto:
## Description

Create indexes for columns `pfile_sha1` and `pfile_sha256` on table `pfile`.

## How to test

Run and analyze `SELECT * FROM pfile WHERE pfile_sha1 = 'ABC'` and `SELECT * FROM pfile WHERE pfile_sha256 = 'ABC'`. The queries should utilize the indexes.

Closes #1859

### DIFF
--- a/src/www/ui/core-schema.dat
+++ b/src/www/ui/core-schema.dat
@@ -2281,6 +2281,8 @@
   $Schema["INDEX"]["personal_access_tokens"]["user_fk_active_idx"] = "CREATE INDEX user_fk_active_idx ON personal_access_tokens USING btree (user_fk, active);";
 
   $Schema["INDEX"]["pfile"]["pfile_mimetypefk_idx"] = "CREATE INDEX pfile_mimetypefk_idx ON pfile USING btree (pfile_mimetypefk);";
+  $Schema["INDEX"]["pfile"]["pfile_sha1_idx"] = "CREATE INDEX pfile_sha1_idx ON pfile USING btree (pfile_sha1);";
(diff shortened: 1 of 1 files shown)


GMishx (approved):
Changes looks good.

PR #1865, opened 2021-01-07
fix(login): Do not set group_fk if empty

GMishx:
## Description

If the default group of a user was not updated in the form request, do not set the `group_fk` key which causes issue with DB update.

### Changes

Check if `default_group_fk` was sent before updating the `$UserRec['group_fk']` in `src/www/ui/user-edit.php`.

## How to test

1. Login as a non admin user.
2. Goto Admin > Users > Edit User Account and update a field. Submit the form and the update should be recorded.
3. Login as an admin user.
4. Goto Admin > Users > Edit User Account and update the default group of another user.


ag4ums (approved):
Changes look good.
Working as expected

PR #1869, opened 2021-01-13
fix(globaldecision): do not add decisions if the events have no change

shaheemazmalmmd:
## Description

Do not add decisions i case of global decision if events wont change.

### Changes

* Make use of isDecisionWip function to check if decisions have been changed.

## How to test

* Upload a component X and make global clearing to few files.
* Create another user and Upload component X by new users login.
* Scroll through files which have global clearing decisions.
* Check the history so that new decisions should not be added.
* Do end to end clearing testing, So that other clearing areas shouldn't break.


ag4ums (approved):
code looks good

PR #1887, opened 2021-02-05
fix(ojo): New license as candidate

GMishx:
## Description

Create new OJO licenses as candidate licenses rather than main license as ojo looks for regex and can insert malicious licenses in DB.
Also, the auto decision of ojo findings was set to global which was causing some troubles, this PR also sets it back to local decisions only.

### Changes

1. Insert new licenses by ojo in candidate table.
2. **Do not make ojo auto decisions as global.**
3. Fix UI and Unified report to show candidate licenses.
4. Fix ojo to not rescan files.

## How to test

1. Create a sample file with `SPDX-License-Identifier` of a new license.
2. Check if the new license was detected my OJO but ends-up in candidate list.
3. Check if the reported license is shown in all UI views.
4. Check if the reported license is shown in reports.
5. Check if reuser can understand this new candidate license.
6. Check if running decider on upload does not mark decisions as global.

### DIFF
--- a/src/decider/agent/DeciderAgent.php
+++ b/src/decider/agent/DeciderAgent.php
@@ -281,7 +281,14 @@ class DeciderAgent extends Agent
     }
 
     if ($licenseMatchExists) {
-      $this->clearingDecisionProcessor->makeDecisionFromLastEvents($itemTreeBounds, $this->userId, $this->groupId, DecisionTypes::IDENTIFIED, $global=true);
+      try {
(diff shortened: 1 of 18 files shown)


ag4ums:
Segmentation fault while try to scan with Ojo
image

ag4ums:
may be the segmentation fault was due to the some installation issue.

ag4ums (approved):
code looks good...working as expected

Issue #1614, opened 2020-02-17
Detect the MTA binary based on distro


GMishx:
### Description

The MTA package chosen for mail delivery was `heirloom-mailx` which was then renamed to `s-nail`.

The MTA binary in the `fossology.conf` right now is hardcoded to `/usr/bin/mailx`. However, for following distros (Debian based), it changes:

| Distro | Package name | Path |
| :--- | :--- | :--- |
| Debian Jessie (8) | `heirloom-mailx` | `/usr/bin/heirloom-mailx` |
| Debian Stretch (9) | `heirloom-mailx` and `s-nail` | `/usr/bin/heirloom-mailx` and `/usr/bin/s-nail` |
| Debian Buster (10) | `s-nail` | `/usr/bin/s-nail` |
| Debian Sid (unstable) | `s-nail` | `/usr/bin/s-nail` |
| Ubuntu Xenial (16.04) | `heirloom-mailx` and `s-nail` | `/usr/bin/heirloom-mailx` and `/usr/bin/s-nail` |
| Ubuntu Bionic (18.04) | `s-nail` | `/usr/bin/s-nail` |
| Ubuntu Focal (20.04) | `s-nail` | `/usr/bin/s-nail` |

The idea is to use `which` to find if `heirloom-mailx` is installed or `s-nail` and update the `client` https://github.com/fossology/fossology/blob/8f692a66355e5c3acbd022b251ff39b7edf366a8/install/defconf/fossology.conf.in#L92

#### How to reproduce

On Debian Jessie, Stretch and Ubuntu Xenial, the mail functionality works but breaks on other distros.

### Versions

* Last commit id on master: 8f692a66355e5c3acbd022b251ff39b7edf366a8
* Operating System (lsb_release -a):
```
Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster
```

#### FOSSology logs

```
2020-02-14 14:10:39 scheduler [31591] :: NOTE: *****************************************************************
2020-02-14 14:10:39 scheduler [31591] :: NOTE: ***                FOSSology scheduler started                ***
2020-02-14 14:10:39 scheduler [31591] :: NOTE: ***        pid:      31592                                    ***
2020-02-14 14:10:39 scheduler [31591] :: NOTE: ***        verbose:  1                                        ***
2020-02-14 14:10:39 scheduler [31591] :: NOTE: ***        config:   /usr/local/etc/fossology                 ***
[...]

PR #1889, opened 2021-02-09
fix(email): Update email command for s-nail

Aman-Codes:
## Description

For sending email in Ubuntu 20.04, s-nail is used as a dependency by fossology (#1614 ). 
In [S-nail v14.9.6](http://manpages.ubuntu.com/manpages/bionic/man1/s-nail.1.html) `smtp-auth-user` and `smtp-auth-password` flag is replaced by `mta`. Also `-S v15-compat` flag should be added for forward compatibility with s-nail v15

### Changes

Updated the email command for [s-nail](http://manpages.ubuntu.com/manpages/bionic/man1/s-nail.1.html)

## How to test

- Pull this branch
- Start PostgreSQL and apache 2
- Build and install from src in ubuntu 20.04
- Upload a file 
- Check if email notification for completion of fossology scan is received

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -1043,6 +1043,39 @@ void database_job_priority(scheduler_t* scheduler, job_t* job, int priority)
 }
 
 /**
+ * @brief Find s-nail version to check if mta is supported 
+ *
+ * \return 1 if mta is supported, 0 if not
+ */
+int check_mta_support()
+{
+  char cmd[] = "dpkg -s s-nail | grep -i 'Version' | awk '{print$2}' | cut -c -4";
+  char version_str[128];
+  char buf[128];
+  float version_float;
+  FILE *fp;
+
+  fp = popen(cmd, "r");
+  if(!fp)
+  {
+    WARNING("Unable to run the command '%s'.\n", cmd);
+    return 0;
+  }
+  while(fgets(buf, sizeof(buf), fp) != NULL)
+  {
+    strcpy(version_str,buf);
+  }
+  pclose(fp);
+  sscanf(version_str, "%f", &version_float);
+
+  if(version_float - 14.8 > 0.0001)
+  {
+    return 1;
+  }
+  return 0;
+}
+
+/**
  * \brief Build command to run to send email
  * \param scheduler  Current scheduler object
  * \param user_email Email id to send mail to
@@ -1076,8 +1109,6 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
   SafePQclear(db_result_smtp);
   if(g_hash_table_contains(smtpvariables, "SMTPHostName") && g_hash_table_contains(smtpvariables, "SMTPPort"))
   {
-    g_string_append_printf(client_cmd, " -S smtp=\"%s:%s\"", (char *)g_hash_table_lookup(smtpvariables, "SMTPHostName"),
-        (char *)g_hash_table_lookup(smtpvariables, "SMTPPort"));
     if(g_hash_table_contains(smtpvariables, "SMTPStartTls"))
     {
       temp_smtpvariable = (char *)g_hash_table_lookup(smtpvariables, "SMTPStartTls");
@@ -1097,23 +1128,15 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
       {
         g_string_append_printf(client_cmd, " -S smtp-auth=plain");
       }
+      else if(g_strcmp0(temp_smtpvariable, "N") == 0)
+      {
+        g_string_append_printf(client_cmd, " -S smtp-auth=none");
+      }
     }
-
-    if(g_hash_table_contains(smtpvariables, "SMTPAuthUser"))
-    {
-      g_string_append_printf(client_cmd, " -S smtp-auth-user=\"%s\"" ,
-          (char *)g_hash_table_lookup(smtpvariables, "SMTPAuthUser"));
-    }
-
     if(g_hash_table_contains(smtpvariables, "SMTPFrom"))
     {
       g_string_append_printf(client_cmd, " -S from=\"%s\"",
           (char *)g_hash_table_lookup(smtpvariables, "SMTPFrom"));
-    }
-    if(g_hash_table_contains(smtpvariables, "SMTPAuthPasswd"))
-    {
-      g_string_append_printf(client_cmd, " -S smtp-auth-password=\"%s\"",
-          (char *)g_hash_table_lookup(smtpvariables, "SMTPAuthPasswd"));
     }
     if(g_hash_table_contains(smtpvariables, "SMTPSslVerify"))
     {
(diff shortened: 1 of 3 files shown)


src/scheduler/agent/database.c
```diff
@@ -1132,6 +1118,30 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
         g_string_append(client_cmd, "warn");
       }
     }
+    if(g_hash_table_contains(smtpvariables, "SMTPAuthUser") && g_hash_table_contains(smtpvariables, "SMTPAuthPasswd")
+          && g_hash_table_contains(smtpvariables, "SMTPHostName") && g_hash_table_contains(smtpvariables, "SMTPPort"))
```
GMishx, line 1122: This will result in some issue. For example, with current implementation, if I do not provide `SMTPAuthPasswd` (as my server do not required one), I will never receive an email. With the current implementation, that is not the case.

I would suggest to modify current code and build up a server string with the existing checks (if password is not provided, don't add it to the string). And then use the string in the command as value for `-S mta=\"%s\"`.

src/scheduler/agent/database.c
```diff
@@ -1132,6 +1118,30 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
...
+    {
+      if(system("s-nail --help") == 0)
```
GMishx, line 1124: It won't be required to check if `s-nail` is installed or not. As with all the OS supported by FOSSology at the moment, it is available. So you can directly modify this value to point to `s-nail` https://github.com/fossology/fossology/blob/45e186a68fc3c3fbf42dad0a8e92dce3e39922f7/install/defconf/fossology.conf.in#L92

GMishx: Also, you can remove `heirloom-mailx` from here: https://github.com/fossology/fossology/blob/45e186a68fc3c3fbf42dad0a8e92dce3e39922f7/debian/control#L82

GMishx (changes requested):

src/scheduler/agent/database.c
```diff
@@ -1132,6 +1118,30 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
...
+      {
+        g_string_append(client_cmd, " -S v15-compat");
+        g_string_append(client_cmd, " -S mta=\"smtps://");
+        g_string_append(client_cmd, (char *)g_hash_table_lookup(smtpvariables, "SMTPAuthUser"));
+        g_string_append(client_cmd, ":");
+        g_string_append(client_cmd, (char *)g_hash_table_lookup(smtpvariables, "SMTPAuthPasswd"));
```
GMishx, line 1130: Sorry, I missed it. It will be nice to use `smtps` only if port is not `25` or `SMTPStartTls` is provided. Otherwise, `smtp://` should be good enough as per the s-nail's man page.

Also, will it be possible to do a url encoding on the `username:password`? https://manpages.debian.org/stretch/s-nail/s-nail.1.en.html#On_URL_syntax_and_credential_lookup

Aman-Codes: Did the asked changes.
For URL encoding, I took reference from [here](http://geekhideout.com/urlcode.shtml)

GMishx: @Aman-Codes , it will be fine to have the code from geekhideout. Please provide the link to the website. It is already mentioned to be in Public Domain. (Please do mention the mentioned statement from the webpage: **To the extent possible under law, Fred Bulback has waived all copyright and related or neighboring rights to URL Encoding/Decoding in C. This work is published from: United States.**)

image

Aman-Codes: @GMishx Added the link to their [website](http://geekhideout.com/urlcode.shtml) and the mentioned statement from their webpage

GMishx (approved):
Changes looks good. Needs testing.

src/scheduler/agent/database.c
```diff
@@ -1132,7 +1169,38 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
...
+      g_string_append_printf(client_cmd, " -S mta=\"smtps://");
+    }
+    else
+    {
+      g_string_append_printf(client_cmd, " -S mta=\"smtp://");
+    }
+    if(g_hash_table_contains(smtpvariables, "SMTPAuthUser"))
+    {
+      temp_smtpvariable = (char *)g_hash_table_lookup(smtpvariables, "SMTPAuthUser");
+      temp_smtpvariable2 = url_encode(temp_smtpvariable);
```
GMishx, line 1187: I was just looking at glib and found this function [g_string_append_uri_escaped()](https://developer.gnome.org/glib/stable/glib-Strings.html#g-string-append-uri-escaped)

It seems to work just fine with `reserved_chars_allowed` marked as `NULL` and `allow_utf8` set as `TRUE`. This is eliminate requirement of additional `url_encode()` function and can save one temporary variable too.

src/scheduler/agent/database.c
```diff
@@ -1098,27 +1147,15 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
         g_string_append_printf(client_cmd, " -S smtp-auth=plain");
```
GMishx, line 1147: Seems like new `s-nail` does not like empty password. So if `SMTPAuth` is set to `N` (None), ` -S smtp-auth=none` needs to be added.

GMishx (commented):
Please change the client from `mailx` to `s-nail` in https://github.com/fossology/fossology/blob/45e186a68fc3c3fbf42dad0a8e92dce3e39922f7/install/defconf/fossology.conf.in#L92

GMishx:
I did a small testing on different OS:

| OS | Finding |
|:--|:--|
| Ubuntu Xenial | `s-nail` requires `sendmail`, but don't work even with it |
| Ubuntu Bionic | Works fine |
| Ubuntu Focal | Works fine |
| Debian Stretch | `s-nail` requires `sendmail`, but don't work even with it |
| Debian Buster | Works fine |

/cc @ag4ums and @NicolasToussaint , can you please help us what we can do with the Xenial and Stretch?

And I am sorry that I impulsed on `tls-verify`, the only OS that supports it at the moment is Debian Bullseye which is still not released. Other OS completely ignored `tls-verify`. So it needs to be reverted to `ssl-verify` (it only cause warning in Debian Bullseye).

ag4ums:
> I did a small testing on different OS:
> 
> OS	Finding
> Ubuntu Xenial	`s-nail` requires `sendmail`, but don't work even with it
> Ubuntu Bionic	Works fine
> Ubuntu Focal	Works fine
> Debian Stretch	`s-nail` requires `sendmail`, but don't work even with it
> Debian Buster	Works fine
> /cc @ag4ums and @NicolasToussaint , can you please help us what we can do with the Xenial and Stretch?
> 
> And I am sorry that I impulsed on `tls-verify`, the only OS that supports it at the moment is Debian Bullseye which is still not released. Other OS completely ignored `tls-verify`. So it needs to be reverted to `ssl-verify` (it only cause warning in Debian Bullseye).

my suggestion: for Xenial and Stretch add sendmail as dependency and the configuration information as part of the wiki.

GMishx:
A new finding, the `s-nail` version in Ubuntu Xenial and Debian Stretch do not have `mta` internal variable. Instead it was called as `smtp`. But with the new versions in other OS, the `smtp` is marked as obsolete and generates following warning
```
-nail: Warning: variable superseded or obsoleted: smtp
s-nail: Obsoletion warning: please do not use *smtp*, instead assign a smtp:// URL to *mta*!
```

So if just for the two OS, we can use `smtp` flag instead of `mta`, the issue will be solved.

/cc @ag4ums

GMishx:
Hello @Aman-Codes , instead of checking `lsb_release`, you can check the `s-nail` version to start with `14.8.` to use `smtp` flag or `mta` otherwise as `lsb_release` is not a run time dependency for FOSSology.
This will also safe guard you against source installation of `s-nail`.

GMishx:
Also, the use of `smtp-auth-password` and `smtp-auth-user` flags is not required on older versions. Just changing the name of `mta` or `smtp` flag is enough.

src/scheduler/agent/database.c
```diff
@@ -1042,6 +1042,39 @@ void database_job_priority(scheduler_t* scheduler, job_t* job, int priority)
...
+    strcpy(version_str,buf);
+  }
+  pclose(fp);
+  sscanf(version_str, "%f", &version_float);
+
+  if(version_float - 14.8 > 0.0001)
+  {
+    return 0;
+  }
+  return 1;
```
GMishx, line 1075: Looks like the return values are swapped. Does not work on Xenial and Stretch until I swap the values.

GMishx (changes requested):
Hopefully last changes.

src/scheduler/agent/database.c
Aman-Codes: My bad 😅. Updated the code now

GMishx: Anyway, this is an interesting approach `version_float - 14.8 > 0.0001`

GMishx (approved):
Tested, working as expected on Xenial, Bionic, Focal, Stretch, Buster, Bullseye.

PR #1895, opened 2021-02-22
feat(export): Consolidating results per file or directories

JustinWonjaePark:
## Description

Consolidate same results from multiple agents into one for each file.
Consolidate scanning results per directories.

### Changes

- Consolidating options are added on Export Lists

## How to test

- Select multiple agents at Export Lists.
- Select one of consolidate options
- Generate list or download csv file

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ Consider the usual best practice for writing issues, among them:
 - Screenshots are a great help
 - Providing example files (in case for example scanning crashes)
 - Please determine the version, better the commit id
-- Details on operating system you are using
+- Details on the operating system you are using
 
 ## Code Guidelines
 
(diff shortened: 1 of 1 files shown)


src/www/ui/ui-export-list.php
```diff
@@ -522,6 +534,85 @@ private function printLines($lines, $copyright=false)
...
+   */
+  private function consolidateResult($lines)
+  {
+    $newLines = [];
+    foreach ($lines as $row) {
+      $consolidatedLicenses = array();
+      if ($row['conclusions'] !== null) {
+        foreach ($row['agentFindings'] as $key => $value) {
+          unset($row['agentFindings'][$key]);
+        }
```
GMishx, line 550: This can be optimized as `$row['agentFindings'] = array();` if you want to delete everything inside `agentFindings`.

JustinWonjaePark: Thanks @GMishx I just changed the code as you suggested.

GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

GMishx:
Just a quick question, I notice all files **tensorflow/tools/ci_build/** directory are marked as **Apache-2.0** and thus all directories show as a single entry.
This can further be consolidated to show only one **tensorflow/tools/ci_build/: Apache-2.0** entry instead of listing 10 inner folders. Is this left like this intentionaly?
snap

PR #1897, opened 2021-02-25
fix(uploadDao): Fetch status based on group id

GMishx:
## Description

`getStatus()` in `UploadDao` currently fetches a random entry. This causes confusion. Updated the function to match `getListOfUploadsOfFolder()` which fetches status based on upload id and group id.

### Changes

Update `UploadDao::getStatus()` to match the group id as well.
**Note**: Default status returned will be 1 (open).

## How to test

1. Upload a package from a user which has access to minimum 2 groups.
2. With group 1, change the upload status, to say rejected.
3. With group 2, change the upload status, to say closed.
4. The UI shows correct status to respective groups.
5. From the REST API, fetch the upload summary once with `groupName: group1` and once with `groupName: group2`. The `clearingStatus` should change based on group selected.
6. Fetch the upload summary from a third group with no status changes, the `clearingStatus` should say `Open`.

### DIFF
--- a/src/lib/php/Dao/UploadDao.php
+++ b/src/lib/php/Dao/UploadDao.php
@@ -226,14 +226,20 @@ class UploadDao
   }
 
   /**
-   * @brief unused function
+   * @brief Get the upload status.
+   * @param int $uploadId Upload to get status for
+   * @param int $groupId  Effective group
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):

PR #1901, opened 2021-03-01
fix(UploadTreeProxy): Get if candidate license

GMishx:
## Description

Fix the query to show result only if license is not a "No_license_found" or is a candidate license.

### Changes

1. Change query to `WHERE (lr.rf_shortname NOT IN ('No_license_found', 'Void') OR lr.rf_shortname IS NULL) WHERE (lr.rf_shortname NOT IN ('No_license_found', 'Void') OR (lr.rf_pk IS NULL AND lc.rf_pk IS NOT NULL))`


shaheemazmalmmd (approved):

PR #1905, opened 2021-03-03
fix(globalDecision): fix global decision prevent adding history in ca…

shaheemazmalmmd:
## Description

Do not add decisions in case of global decision if events wont change.

### Changes

* Make use of isDecisionWip function to check if decisions have been changed.

## How to test

* Upload a component X and make global clearing to few files.
* Create another user and Upload component X by new users login.
* Scroll through files which have global clearing decisions.
* Check the history so that new decisions should not be added.
* Do end to end clearing testing, So that other clearing areas shouldn't break.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -643,6 +643,15 @@ INSERT INTO clearing_decision (
     return ($latestDec['decision_type'] == DecisionTypes::DO_NOT_USE);
   }
 
+  public function getClearingType($uploadTreeId, $groupId, $type)
+  {
+    $sql = "SELECT decision_type, scope FROM clearing_decision
+              WHERE uploadtree_fk=$1 AND group_fk = $2
+            ORDER BY clearing_decision_pk DESC LIMIT 1";
+    $latestDec = $this->dbManager->getSingleRow($sql,
+                 array($uploadTreeId, $groupId), $sqlLog = __METHOD__);
+    return $latestDec;
+  }
 
   public function isDecisionIrrelevant($uploadTreeId, $groupId)
   {
--- a/src/www/ui/ui-clearing-view.php
+++ b/src/www/ui/ui-clearing-view.php
@@ -217,7 +217,11 @@ class ClearingView extends FO_Plugin
     $lastItem = GetParm("lastItem", PARM_INTEGER);
 
     if (!empty($lastItem)) {
-      $this->updateLastItem($userId, $groupId, $lastItem, $uploadTreeId);
+      $currentUploadtreeId = $lastItem;
+      if ($lastItem == $uploadTreeId) {
+        $currentUploadtreeId = $uploadTreeId;
+      }
+      $this->updateLastItem($userId, $groupId, $lastItem, $currentUploadtreeId);
     }
 
     $uploadTreeTableName = $this->uploadDao->getUploadtreeTableName($uploadId);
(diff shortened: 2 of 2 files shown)


src/lib/php/Dao/ClearingDao.php
```diff
@@ -643,6 +643,18 @@ public function isDecisionDNU($uploadTreeId, $groupId)
...
+  {
+    $sql = "SELECT decision_type, scope FROM clearing_decision
+              WHERE uploadtree_fk=$1 AND group_fk = $2
+            ORDER BY clearing_decision_pk DESC LIMIT 1";
+    $latestDec = $this->dbManager->getSingleRow($sql,
+                 array($uploadTreeId, $groupId), $sqlLog = __METHOD__);
+    if ($latestDec === false) {
+      return false;
+    }
+    return $latestDec;
```
GMishx, line 656: This seems like an unnecessary conditional check to me.

src/www/ui/ui-clearing-view.php
```diff
@@ -217,7 +217,12 @@ function Output()
     $lastItem = GetParm("lastItem", PARM_INTEGER);
 
     if (!empty($lastItem)) {
-      $this->updateLastItem($userId, $groupId, $lastItem, $uploadTreeId);
+      if ($lastItem == $uploadTreeId) {
+        $currentUploadtreeId = $uploadTreeId;
+      } else {
+        $currentUploadtreeId = $lastItem;
+      }
+      $this->updateLastItem($userId, $groupId, $lastItem, $currentUploadtreeId);
```
GMishx, line 225: Can we simplify it saying `$currentUploadtreeId = $lastItem` and eliminate `$currentUploadtreeId` by passing `$lastItem` to `updateLastItem()`?

GMishx (commented):
Two suggestions, simplifying the code.

src/www/ui/ui-clearing-view.php
```diff
@@ -350,8 +355,18 @@ protected function updateLastItem($userId, $groupId, $lastItem, $currentUploadtr
     $global = GetParm("globalDecision", PARM_STRING) === "on" ? 1 : 0;
     $uploadTreeTableName = $this->uploadDao->getUploadtreeTableName($lastItem);
     $itemBounds = $this->uploadDao->getItemTreeBounds($lastItem, $uploadTreeTableName);
-    $isDecisionWip = $this->clearingDao->isDecisionWip($currentUploadtreeId, $groupId);
-    if ($isDecisionWip || !$global) {
+    if ($global) {
+      $isDecisionWip = $this->clearingDao->isDecisionWip($currentUploadtreeId, $groupId);
+      $hasChangedClearingType = $this->clearingDao->getClearingType($currentUploadtreeId, $groupId, $type);
+      if ($isDecisionWip
+           || !$hasChangedClearingType['scope']
```
GMishx, line 362: I just have a question here. Noting that `clearingDao->getClearingType()` can return `false`, won't that make this `|| !$hasChangedClearingType['scope']` check throw a warning (checking array index on bool)??

GMishx (approved):
Tested, working as expected.

GMishx:
There is a small issue I found. If I upload a package and on any file, I do the following operations, no result is saved:
Open any file -> Select a decision type -> Select global checkbox -> hit submit or go left/right.

If I do it in two steps, it works:
Open any file -> Select a decision type -> hit submit -> Select global checkbox -> hit submit or go left/right

Can you please look into it @shaheemazmalmmd?

GMishx (approved):
Changes looks good. Tested, working fine.

PR #1906, opened 2021-03-03
fix(rest): missing Group component in API documentation

### LINKED ISSUE #1900: Missing component Group in API documentation
### Description

While testing a new PR, I discovered following error in Swagger Editor.

image

Looks like the component information for `#/components/schemas/Group` was missed in #1872.

#### How to reproduce

Load the `src/www/ui/api/documentation/openapi.yaml` in Swagger editor.

### Versions

* Last commit id on master: 45e186a68fc3c3fbf42dad0a8e92dce3e39922f7

piotreke12:
Signed-off-by: Piotr Pszczola <piotr.pszczola@orange.com>

## Description

Fixes #1900 

### Changes

Added Group component to src/www/ui/api/documentation/openapi.yaml
Few naming fixes in src/www/ui/api/Models/Group.php

## How to test

Swagger integration verification should not show any errors

### DIFF
--- a/src/www/ui/api/documentation/openapi.yaml
+++ b/src/www/ui/api/documentation/openapi.yaml
@@ -1560,6 +1560,20 @@ components:
           nullable: true
           example:
             3
+    Group:
+      description: Group object
+      type: object
+      properties:
+        id:
+          description: Id key
+          type: integer
+          example:
+            126
+        name:
+          description: Name of a group
+          type: string
+          example:
+            "Main group"    
   responses:
     defaultResponse:
       description: Some error occured. Check the "message"
--- a/src/www/ui/api/Models/Group.php
+++ b/src/www/ui/api/Models/Group.php
@@ -18,7 +18,7 @@
  ***************************************************************/
 /**
  * @file
- * @brief Folder model
+ * @brief Group model
  */
 
 namespace Fossology\UI\Api\Models;
@@ -37,7 +37,7 @@ class Group
   private $name;
 
   /**
-   * Folder constructor.
+   * Group constructor.
    *
    * @param int $id
    * @param string $name


src/www/ui/api/documentation/openapi.yaml
```diff
@@ -15,7 +15,7 @@ openapi: 3.0.2
 info:
   title: FOSSology API
   description: Automate your fossology instance using REST API
-  version: 1.2.1
+  version: 1.2.2
```
GMishx, line 18: I don't think it is necessary to update the API version. It is just a documentation patch. What do you think?

piotreke12: This is a thing I also thought about. Finally I decided to change it as I don't know Your versioning policy very well. 
In fact, last part in semver policy says that last section in version number is for patch purposes, thats why I decided to change it.
All in all, I can revert this change if you wish so.

GMishx: Yeah it is true that following semver policy, it should change to 1.2.2 but for the REST API, we tried to keep it simple.
1. If a change **does not** break backward compatibility (in terms of API call) but adds something new, we change the patch version (1.2.1 => 1.2.2).
2. If a change **does break** backward compatibility, we like to make it explicit by changing the minor version (1.2.1 => 1.3.0).
3. We have not yet faced or decided anything to change the major version of API :neutral_face:

GMishx: So yeah, these changes does not do anything on the API so it should be okay to leave the version as is.

piotreke12: Ok, Your explanation is clear.
Update reverting API version change is on it's way

GMishx (approved):
Changes looks good. Working fine with Swagger editor.

PR #1908, opened 2021-03-08
fix(username): update session variable on username change

### LINKED ISSUE #1112: Problem while changing current user's username
## Current issue
When I try to change my username by going to Admin->Users->Edit user account and change the value of username, the session is not update. This creates problem while accessing FOSSology pages as the user information is taken from the current session's cookie which does not reflect updated values. It won't be fixed until I log out and login again.

## How to reproduce
1.  Go to Admin->Users->Edit user account.
2.  Change the value of username for the current user (with active session).
3.  Hit Update Account.

    image

4.  Goto Upload->(any page which allows selection of agents)
image

## Possible solutions
1.  Update the session cookie while updating username of current user.
2.  Don't allow to change username.

Aman-Codes:
## Description
* Update session variable on changing the username 
* Only allow the admin to change the username
* If the user provides an empty username then a blank white screen appears due to an internal server error

Closes #1112

### Changes

Updated `$_SESSION['User']` if `$UserRec['user_name']` does not match `$_SESSION['User']`

## How to test

1. Go to Admin->Users->Edit user account.
2. Change the value of the username for the current user (with active session).
3. Hit Update Account.
4. Goto Upload->(any page which allows selection of agents)

### DIFF
--- a/src/www/ui/template/user_edit.html.twig
+++ b/src/www/ui/template/user_edit.html.twig
@@ -21,7 +21,7 @@
   <table style="border:1px solid black; border-collapse: collapse;" width="100%">
     <tr class="classic">
       <th width="25%">{{ "Username."|trans }}</th>
-      <td><input type="text" value="{{ userName|e }}" name="user_name" size="20"></td>
+      <td><input type="text" value="{{ userName|e }}" name="user_name" size="20" {% if not isSessionAdmin %}disabled="disabled"{% endif %}></td>
     </tr>
     <tr class="classic">
       <th width="25%">{{ "Description (name, contact, or other information)."|trans }} {{ "This may be blank."|trans }}</th>
--- a/src/www/ui/user-edit.php
+++ b/src/www/ui/user-edit.php
@@ -109,7 +109,13 @@ class UserEditPage extends DefaultPlugin
 
         /* Reread the user record as update verification */
         $UserRec = $this->CreateUserRec($request, $UserRec['user_pk']);
+        if ($user_pk == $user_pk_to_modify) {
+          $_SESSION['User'] = $UserRec['user_name'];
+        }
       } else {
+        if (empty($UserRec['user_name']) || $_SESSION['User'] != $UserRec['user_name']) {
+          $UserRec = $this->CreateUserRec($request, $UserRec['user_pk']);
+        }
         $vars['message'] = $rv;
       }
     } else {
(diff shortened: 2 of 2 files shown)


GMishx:
Thank you for your contribution!
Can you please add a new twig variable so that the following field becomes disable for users other than admin?
https://github.com/fossology/fossology/blob/f6ba00362685916e099a5337e3b195e0b9925cef/src/www/ui/template/user_edit.html.twig#L24

Aman-Codes:
> Can you please add a new twig variable so that the following field becomes disable for users other than admin?

Do you mean to change it to   
```twig
{%  if isSessionAdmin %}
    <tr class="classic">
      <th width="25%">{{ "Username."|trans }}</th>
      <td><input type="text" value="{{ userName|e }}" name="user_name" size="20"></td>
    </tr>
{% endif %}
```

or something else?

GMishx:
Something like this:
```twig
<input type="text" value="{{ userName|e }}" name="user_name" size="20" {% if !isSessionAdmin %} disabled="disabled"{% endif %} />
```

And keeping the PHP check in case someone edited their DOM.

GMishx:
Woops, unable to load Edit User Account page
```
PHP Fatal error:  Uncaught Twig\\Error\\SyntaxError: Unexpected character "!". in /usr/local/share/fossology/www/ui/template/user_edit.html.twig:24
Stack trace:
#0 /usr/local/share/fossology/vendor/twig/twig/src/Lexer.php(281): Twig\\Lexer->lexExpression()
#1 /usr/local/share/fossology/vendor/twig/twig/src/Lexer.php(182): Twig\\Lexer->lexBlock()
#2 /usr/local/share/fossology/vendor/twig/twig/src/Environment.php(542): Twig\\Lexer->tokenize()
#3 /usr/local/share/fossology/vendor/twig/twig/src/Environment.php(595): Twig\\Environment->tokenize()
#4 /usr/local/share/fossology/vendor/twig/twig/src/Environment.php(408): Twig\\Environment->compileSource()
#5 /usr/local/share/fossology/vendor/twig/twig/src/Environment.php(381): Twig\\Environment->loadClass()
#6 /usr/local/share/fossology/lib/php/Plugin/DefaultPlugin.php(298): Twig\\Environment->loadTemplate()
#7 /usr/local/share/fossology/www/ui/user-edit.php(143): Fossology\\Lib\\Plugin\\DefaultPlugin->render()
#8 /usr/local/share/fossology/lib/php/Plugin/DefaultPlugin.php(239): UserEditPage->han in /usr/local/share/fossology/www/ui/template/user_edit.html.twig on line 24, referer: http://localhost/repo/
```

Aman-Codes:
Updated the code. It should work fine now

GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

PR #1909, opened 2021-03-09
fix(copyright): add uploadId in case of fetching deactivated copyrights

shaheemazmalmmd:
## Description

Add upload-id incase of fetching deactivated copyright statements.

### Changes

Change names of foreign key for ECC, keyword and author tables for respective event

## How to test

* Upload package 1 do some deactivation.
* Upload same package again and see if deactivated copyrights appear.
* Check Keyword, ECC and email/url/author pages.

### DIFF
--- a/install/db/dbmigrate_copyright-event.php
+++ b/install/db/dbmigrate_copyright-event.php
@@ -56,12 +56,13 @@ function insertCopyrightEventTables($dbManager)
     $sql = "SELECT count(*) AS cnt FROM ";
     $statement = __METHOD__ . ".getCountsFor";
     $length = 0;
-    $row = $dbManager->getSingleRow($sql . $table ."WHERE is_enabled=false;", array(), $statement . $table);
+    $row = $dbManager->getSingleRow($sql . $table ." WHERE is_enabled=false;", array(), $statement . $table);
     $length = intval($row['cnt']);
-    
-    echo "*** Inserting $length records from $table to $tableEvent table ***\n";
-    
+    if (!empty($length)) {
+      echo "*** Inserting $length records from $table to $tableEvent table ***\n";
+    }
     $tablePk = $table."_pk";
+    $tableFk = $table."_fk";
     $sql = "SELECT DISTINCT ON ($tablePk) $tablePk, uploadtree_pk, upload_fk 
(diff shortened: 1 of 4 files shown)


GMishx:
There are several issues with the migration script. I did some investigation into it.
1. The `ENCODE_TABLES` should have been `TABLE_NAMES` https://github.com/fossology/fossology/blob/a29e8027555822414cc14410db5bde8339ce19b4/install/db/dbmigrate_copyright-event.php#L103
2. There should be a white space before the `WHERE` in https://github.com/fossology/fossology/blob/a29e8027555822414cc14410db5bde8339ce19b4/install/db/dbmigrate_copyright-event.php#L59
3. If the `is_enabled=true` is done, then I think the `OFFSET` is not required in the loop's select query as the total number of rows are reduced by `MAX_SIZE_OF_ROW` already. This sends it into infinite loop.
4. The query inside the loop has hard-coded `copyright_pk` which will not work for ECC and Keyword tables.

GMishx:
Any activation/deactivation/edit on tables other than Copyright seems to be not working.

And while testing the branch, I did following:
1. Uploaded Atarashi package.
2. Deactivate every copyright other than Aman Jain's, mine and Siemens.
3. Edited my copyright and change it to my official email id.
4. Uploaded atarashi again.

Here are the findings:
1. The edited entries are not getting clubbed.
    image
2. In the second upload, nothing was deactivated as expected but the search was still working on original copyright text and table was showing edited entry.
    - For example, original copyright was there with my personal email (gmishx@gmail.com) which I changed to Siemens email. Now if I search for gmishx, this is what I am getting.
    image

Just to be sure, I did the tests on `Version: [3.10.0.63], Branch: [contrib/fix/copyrightdeactivated], Commit: [#6f9852] 2021/03/09 18:47 IST built @ 2021/03/10 11:39 IST`

GMishx:
It will be a good idea to add indexes to `*_event` tables (referencing from copyright table).

GMishx (changes requested):
Please check the comments above.

shaheemazmalmmd:
Closing this PR as i have pulled the changes to here https://github.com/fossology/fossology/pull/1914

Issue #1913, opened 2021-03-12
Migrate from Travis to Github Actions


NishealJ:
### Description
Fossology currently uses Travis, but moving into GitHub action will be more convenient and versatile for building+running tests.
Read more at, https://github.com/fossology/fossology/wiki/Google-Summer-of-Code-Proposals-2021#migrate-from-travis-to-github-actions

The list of jobs that has to be migrated from Travis to Github Actions are:

- [x] Syntax Check (#1919)
- [x] Static Code Analysis (#1919)
- [x] Copy/Paste Detector (#1919)
- [x] PHP Codesniffer (#1919)
- [ ] Docker Tests
- [x] Docker builds (#2076)
- [ ] C/C++ agent tests
- [ ] PHPUnit tests
- [x] GitHub Page Release (#1917)
- [ ] Implementing Caching in workflows
- [ ] Implement source install (see #207)

#### Steps followed and expected result
https://github.com/fossology/fossology/actions

GMishx:
Everything was moved eventually. Thank you for starting it! Closing the issue as done.

PR #1917, opened 2021-03-12
feat(CI): Migrate API docs generation and deployment to GitHub Actions

avinal:
## Description

Migrate API docs generation and deployment to GitHub Actions from Travis CI.

### Changes

- Added new workflow script for GitHub Actions: pages-deploy.yml
- Removed pages deployment from Travis CI: .travis.yml
- Modified pages-deployment script for GHA: pages-deploy.sh

## How to test

Check the current deployment at http://avinal.space/checkdeployment/

Fixes part of #1913

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -1,0 +1,39 @@
+name: GitHub Pages Deploy
+
+on:
+  push:
+    branches: 
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -y git doxygen graphviz
+
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+
+      - name: Execute docs generation
+        run: |
+          chmod +x utils/deploy-pages.sh
+          bash utils/deploy-pages.sh
+        env:
+          GH_REPO_REF: ${{ secrets.GH_REPO_REF }}
+
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          personal_token: ${{ secrets.GH_TOKEN }}
+          external_repository: ${{ secrets.GH_REPO_REF }}
+          publish_branch: master
+          publish_dir: ./code_docs
+          force_orphan: true
+          user_name: 'github-actions[bot]'
+          user_email: 'github-actions[bot]@users.noreply.github.com'
+          full_commit_message: ${{ github.event.head_commit.message }}
--- a/.travis.yml
+++ b/.travis.yml
@@ -179,26 +179,5 @@ jobs:
             - postgresql-client-10
             - *default_packages
       before_script: *postgres-before-script
-    - stage: GitHub Page Release
-      name: GitHub Page deploy
-      addons:
-        apt:
-          packages:
-            - doxygen
-            - graphviz
-      script:
-        - doxygen -v
-        - bash utils/deploy-pages.sh
-      deploy:
-        provider: pages
-        cleanup: false
-        skip_cleanup: true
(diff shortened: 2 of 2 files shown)


avinal:
I am a bit confused about if this should run on each `push` or `release`. The release is a better choice. Please let me know I will change it then.

GMishx:
How about using https://github.com/marketplace/actions/github-pages-action to handle the commit pushes? It keeps the code clean.
It also allows to use the default `GITHUB_TOKEN` with `github_token` option.

It also allows to set `external_repository`, `force_orphan`, `commit_message`, `user_name` and `user_email`.

> I am a bit confused about if this should run on each `push` or `release`.

It should be done whenever there is a merge (`push`) to master not on releases as the documents get updated with the code.

avinal:
Yeah, I considered using this (https://github.com/marketplace/actions/github-pages-action) initially, but I thought maybe this project needs some customizations and fewer dependencies, so I created a custom one. (I should have asked) No Worries. I will refactor the workflow to use it. 

But still, I don't think you can use `GITHUB_TOKEN` to deploy to another repo. It is only repo scoped. You can see it that action's readme too: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-deploy-to-external-repository-external_repository
Official Doc: https://docs.github.com/en/actions/reference/authentication-in-a-workflow

`push` is already the default option :)

GMishx:
It is okay to use dependencies as long as they are compatible with GPL-2.0 license. It takes some load off from maintaining the code.

And for `GITHUB_TOKEN`, I missed that point. Thanks for pointing it. Please change it to `personal_token` and will setup one in the secrets.

avinal:
New Changes:
- Page deployment is being handled by [peaceiris/github-pages-action](https://github.com/peaceiris/actions-gh-pages)
- `external_repository: ${{ secrets.GH_REPO_REF }}` - assuming `GH_REPO_REF` is set to **fossology/fossology.github.io** 
- Authentication set to `personal_token` - Please add a secret named `GH_TOKEN`

I think it is all complete now 👍

GMishx (approved):
Changes looks good.

GMishx:
Just saw your deployment and found the version number missing:
image

It is fetched from git and requires the history to find the first tag. Maybe the checkout depth needs to be increased?
https://github.com/fossology/fossology/blob/eb0319dced8f0b4dd4fd7e5db5d654e0cd64bd4e/utils/deploy-pages.sh#L31-L33

avinal:
I think it will work here because my deployment is based on my fork only and it has no tags, so it can't get any. For a test, we can do one thing add an `echo $CURR_TAG` after line 33 and see the logs it detects or not. If yes then we are good to go.

GMishx:
> I think it will work here because my deployment is based on my fork only and it has no tags, so it can't get any.

Can you please try to create some tags on your fork? As per the current state (https://github.com/fossology/fossology/releases/tag/3.10.0-rc1), there are at least 66 commits since last tag. So a tag 70 commits back from your test commit should be fine.

> For a test, we can do one thing add an `echo $CURR_TAG` after line 33 and see the logs it detects or not. If yes then we are good to go.

What you suggest is to remove the **"Run on master"** condition and print `CURR_TAG` for this PR?

avinal:
I saw some solutions on StackOverflow, one of the solutions is definitely increasing the fetch depth, so I will hardcode it for a test and see if it works. 

> What you suggest is to remove the **"Run on master"** condition and print `CURR_TAG` for this PR?

Kinda yes, because in my test run this is the error `fatal: No tags can describe '3e89fc845add6a7970c9aa2296fab163a7eb84e2'` I will get back after the test shortly.

avinal:
image
Increasing fetch-depth worked, adding that in a moment.

GMishx (approved):
Tested at https://avinal.space/checkdeployment/

GMishx:
Yay! Deployment successful: https://github.com/fossology/fossology/actions/runs/679189817
https://fossology.github.io/

PR #1918, opened 2021-03-12
fix(report): Do not merge ack text

GMishx:
## Description

Do not merge entries if two licenses have same acknowledgements.

### Changes

1. Overridden `ClearedGetterCommon::getCleared()` in `LicenseClearedGetter`.
2. Added new function `LicenseClearedGetter::groupStatementsSpecial()` to merge entries if only acknowledgements or comments are requested. Use the `ClearedGetterCommon::groupStatements()` otherwise.
    - The merge is done taking **license_name.acknowledgement** or **license_name.comment** as a unique key.

## How to test

1. In a package identify one file with **License1** and add **Acknowledgement1** as acknowledgement.
2. In other file, identify **License2** and add **Acknowledgement1** as acknowledgement.
3. Add an identification **License1** and add **Comment1** as comment.
4. Add another identification **License2** and add **Comment1** as comment.
5. Add some normal identifications with various combinations of acknowledgements and comments.
6. Generate Unified Report and Readme OSS.
    - Even if there are duplicate acknowledgements, there should be separate entries for each license.
    - There should be separate entries for comments in Notice section as well.
    - Other parts of the report should not be effected.

### DIFF
--- a/src/lib/php/Report/LicenseClearedGetter.php
+++ b/src/lib/php/Report/LicenseClearedGetter.php
@@ -27,6 +27,7 @@ use Fossology\Lib\Data\DecisionTypes;
 use Fossology\Lib\Proxy\ScanJobProxy;
 use Fossology\Lib\Data\License;
 use Fossology\Lib\Data\AgentRef;
+use Fossology\Lib\Agent\Agent;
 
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):
code looks good and works as expected

PR #1919, opened 2021-03-13
feat(CI): Migrate Static Checks and Analysis to GitHub Actions from Travis CI

avinal:
## Description

This PR migrates the following Checks and Analysis to GitHub Actions from Travis CI
- Syntax Checks
- Static Code Analysis
- Copy/Paste Detector
- PHP Codesniffer

### Changes

- Added new workflow configuration for static checks and analysis: static-checks.yml
- Removed static checks and analysis from Travis CI configuration: .travis.yml

## How to test

NA - can only be tested after merging, however, a test run is available here: https://github.com/avinal/fossology/actions/runs/648349454

Fixes parts of issue #1913

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -1,3 +1,6 @@
+# Copyright 2021 Avinal Kumar <avinal.xlvii@gmail.com>
+# SPDX-License-Identifier: GPL-2.0 AND LGPL-2.1
+
 name: GitHub Pages Deploy
 
 on:
--- a/.github/workflows/static-checks.yml
+++ b/.github/workflows/static-checks.yml
@@ -1,0 +1,79 @@
+# Copyright 2021 Avinal Kumar <avinal.xlvii@gmail.com>
+# SPDX-License-Identifier: GPL-2.0 AND LGPL-2.
+
+name: Static checks and analysis
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  syntax-check:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y php7.4-cli
+          export PATH="/usr/lib/ccache/:$PATH"
+          export COMPOSER_HOME="$HOME/.composer/"
+          sudo update-alternatives --set php /usr/bin/php7.4
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - name: Syntax Check
+        run: |
+          bash src/testing/syntax/syntaxtest.sh
+          composer validate --no-check-all --working-dir=src --strict
+
+  code-analysis:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cppcheck
+          export PATH="/usr/lib/ccache/:$PATH"
+
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - name: Static Code Analysis
+        run: cppcheck -q -isrc/nomos/agent_tests/testdata/NomosTestfiles/ -isrc/testing/dataFiles/ src/
+
+  codesniffer:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y php-cli
+          export PATH="/usr/lib/ccache/:$PATH"
+          export COMPOSER_HOME="$HOME/.composer/"
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - name: PHP Codesniffer
+        run: |
+          composer install --prefer-dist --working-dir=src
+          src/vendor/bin/phpcs --standard=src/fossy-ruleset.xml -p
+  
+  copy-paste:
+    runs-on: ubuntu-20.04
+    continue-on-error: true
+    steps:
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y php-cli
+          export PATH="/usr/lib/ccache/:$PATH"
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - name: Copy/Paste detector
+        run: |
+          composer install --prefer-dist --working-dir=src
+          sudo phpdismod -v ALL -s ALL xdebug
+          src/vendor/bin/phpcpd src/cli/ src/copyright/ src/decider*/ src/lib/ src/monk/ src/nomos/ src/readmeoss/ src/spdx2/ src/www/ || echo -e "\033[0;31mDuplication detected: Task Failed"
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,22 +56,6 @@ services: postgresql
 jobs:
   fast_finish: true
   include:
-    - name: Syntax Check
-      script:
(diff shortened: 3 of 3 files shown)


avinal:
There are some additional explanations:
- Syntax Checks: This project is using `php7.x` and is not compatible with the latest version. The workflow currently uses `ubuntu-16.04` as the virtual environment and will need to be updated if the project is upgraded to the latest PHP.
- Copy/Paste Detector: As usual this check is failing, and is marked as `continue-on-error: true`. So it doesn't affect the overall status of checks.
- Static Code Analysis and PHP Codesniffer now runs on the latest Ubuntu(focal-fossa)
- Instead of a single job the workflow has been divided into 4 separate jobs, this allows checks to run in parallel.

.github/workflows/static-checks.yml
```diff
@@ -0,0 +1,76 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  syntax-check:
+    runs-on: ubuntu-16.04
```
GMishx, line 11: It would be nice to run this test on `ubuntu-latest` as well. Are there any difficulties with latest PHP?

avinal: The project is based on PHP 7.x and by default `ubuntu-latest` picks PHP 8.x. There have been some major changes(syntax) in PHP 8.0. We can try a workaround by forcing the Action to install PHP 7.x but it doesn't solve any purpose. The project will still be outdated in terms of PHP. 

Changes in PHP 8.0: https://www.php.net/manual/en/migration80.incompatible.php

You can see details of the test run here : https://github.com/avinal/fossology/runs/2097843501?check_suite_focus=true#step:4:206

I would suggest, we should migrate to PHP 8.x, can't say if this is desirable or not.

GMishx: It will be big of a task to make FOSSology PHP 8 compatible. It will require big effort on migration and testing with 7.0 and 8.x. Lets keep it for a separate PR.

As per the documentation, Ubuntu 20.04 image comes with PHP 7.4 and 8.0: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#php

Can you please check if you can use the `update-alternatives` to update php to point to php-7.4?

Also, now that I think about it, changing the tag from `ubuntu-latest` to `ubuntu-20.04` will be better making sure the pipeline does not break spontaneously in future. What are your thoughts on it?

avinal: May I open an issue for PHP upgrade??

I tried using `update-alternatives`, it is working on my local machine but not on GitHub Actions for some reasons. I will try once again with some modifications and let you know.

Since LTS versions are released after every two years, `ubuntu-latest` doesn't changes that often, but yeah for assurance `ubuntu-20.04` is surely a better choice. I will update that.

avinal: Great, It worked, I might have missed something while testing :)

avinal:
Should I open an issue for PHP upgrade??

GMishx:
> Should I open an issue for PHP upgrade??

Sure, that will be good.

avinal:
I have added all the suggestions. this PR is complete now from my side :)

GMishx:
Hi @avinal , the changes look good but there will be a change required for the `phpcpd` test. Even though there is `continue-on-error` tag, the job ends with a :x: which causes confusion.
Can you please try adding something like `|| echo "Test failed"` at end of the command so that the Copy-paste test remains green and we get a good :heavy_check_mark: for all test cases?

avinal:
Yeah sure, doing in few time 👍

GMishx:
Oh, btw, did you see this: https://github.com/fossology/fossology/pull/1919/checks?check_run_id=2118747995#step:4:202 ?
Can it be fixed?

avinal:
> 
> 
> Oh, btw, did you see this: https://github.com/fossology/fossology/pull/1919/checks?check_run_id=2118747995#step:4:202 ?
> Can it be fixed?

Yes, I tried fixing that. But the problem I am getting is: `/home/runner/work/fossology/fossology/src/vendor/sebastian/phpcpd/src/CLI/Application.php:109`, there is no `vendor` folder in `fossology/src`, that means it must be produced by some other steps, This is also causing errors in PHPUnit and C++ agent test. I need a way to explore that folder, I will see into that again. 
It would be a great help if you could tell a shortcut to install every dependency of this project, I have installed using the script, but still every now and then it needs something else 😥

avinal:
I think there is some problem with `composer` itself. The readme says this https://github.com/sebastianbergmann/phpcpd#installation. I will try using the archive version.

GMishx:
The `vendor` folder is generated by `composer`. So you can do `composer install --working-dir=src` to get it.

I investigated a bit more and found this is coming from here: https://github.com/sebastianbergmann/phpcpd/blob/3.0.1/src/CLI/Application.php#L109

And as per XDebug3, the function is removed: https://xdebug.org/docs/upgrade_guide#Changes-to-Functions

So disabling the xdebug plugin from PHP will be the best option as we can not update the dependency (need to support PHP7.0).

GMishx:
Hi @avinal , how about using `sudo phpdismod -v ALL -s ALL xdebug`?

avinal:
image
I tried the `phar`, it ran successfully but still failed 🙄. 

> 
> Hi @avinal , how about using `sudo phpdismod -v ALL -s ALL xdebug?

Okay, let us try.

GMishx:
> I tried the `phar`, it ran successfully but still failed roll_eyes.

The job is expected to fail, that's why we allow it to fail in Travis. It is here just as a recommendation, not as a rule like other tests.
That's the reason I asked you to add the `|| echo...` so the return code is always 0.

avinal:
Ok so I will add `echo` so now should I stick to `phar` or try this method

GMishx:
It will be preferred to use the composer (without downloading the `phar`) to keep the tests consistent for every developer.

GMishx (approved):
Changes looks good.
The test outputs looks good.

avinal:
image
Finally all green.

GMishx:
@avinal , sorry I forgot about it in your last PR. Would you like to add your copyright and license information at the top of the YAML files you are contributing? Example https://github.com/fossology/fossology/blob/20d41796515191bb44611944199dafd7e61d48f9/.travis.yml#L1-L2

avinal:
Hehe, I don't have much preference about this, But it feels nice to have a signature in a big open source project 😋. So I would like to add it.

GMishx:
We encourage you to do so as you've put in your efforts to get this this running. ✨

avinal:
I have added the copyrright header to both the previous and this file.

.github/workflows/deploy-pages.yml
```diff
@@ -1,3 +1,6 @@
+# Copyright Avinal Kumar 2021
```
GMishx, line 1: Just a suggestion, it will be good to add your email to, e.g. `Copyright 2021 Avinal Kumar <185067@nith.ac.in>`

.github/workflows/deploy-pages.yml
```diff
@@ -1,3 +1,6 @@
...
+# SPDX-License-Identifier: GPL-2.0 LGPL-2.1
```
GMishx, line 2: It will be nice to write it as
```suggestion
# SPDX-License-Identifier: GPL-2.0 AND LGPL-2.1
```

GMishx (commented):
Some suggestion if you like it.

GMishx (approved):
Changes looks good.

GMishx (approved):
Test results are here: https://github.com/fossology/fossology/actions/runs/682467635

PR #1926, opened 2021-03-22
fix(ui): Show error message for invalid license id

GMishx:
## Description

While trying to read wrong license from `?mod=admin_license&rf_pk=<wrong_rf_pk>`, PHP throws error and displays an empty page.
This PR fixes that by showing an error message when the license id is invalid.

### Changes

1. Add a new Twig variable `error` for license admin.
    - If it is empty, show the page normally.
    - If it is not empty, show the error message only.
2. Create the required `error` variable in case license ID is invalid.

## How to test

1. Open `?mod=admin_license&rf_pk=<wrong_rf_pk>`, should show an error message.
2. Edit a valid license, should work as expected.
3. Create a new license, should work as expected.


shaheemazmalmmd (approved):
Code looks good. needs testing

Issue #1932, opened 2021-03-26
Postgres is unavailable  - SCRAM authentication requires libpq >= 10


rkh-hash:
### Description

I'm using the docker image `fossology/fossology:3.9.0`.

Authentication failed when `scram-sha-256` is configured for password encryption on the server side (PostgreSQL server v12.2).

The reason: `SCRAM authentication requires libpq version 10 or above`.

 As specified into PostgreSQL documentation,

> Note that older clients might lack support for the SCRAM authentication mechanism, and hence not work with passwords encrypted with SCRAM-SHA-256. See Section 20.5 for more details.

(cf https://www.postgresql.org/docs/12/runtime-config-connection.html)

```shell
$ docker run --rm -it --entrypoint="" fossology/fossology:3.9.0 /bin/sh -c 'dpkg -l | grep libpq'
ii  libpq5:amd64                         9.6.19-0+deb9u1                    amd64        PostgreSQL C client library
```

#### How to reproduce

```yaml
version: '3.5'
services:
  scheduler:
    image: fossology/fossology:3.9.0
    restart: unless-stopped
    environment:
      - FOSSOLOGY_DB_HOST=db
      - FOSSOLOGY_DB_NAME=fossology
      - FOSSOLOGY_DB_USER=fossy
      - FOSSOLOGY_DB_PASSWORD=fossy
    command: scheduler
    depends_on:
      - db
    volumes:
      - repository:/srv/fossology/repository/
    healthcheck:
      test: ["CMD-SHELL", "true"]
      interval: 5s
      timeout: 5s
      retries: 2
  web:
    image: fossology/fossology:3.9.0
    restart: unless-stopped
    environment:
      - FOSSOLOGY_DB_HOST=db
      - FOSSOLOGY_DB_NAME=fossology
      - FOSSOLOGY_DB_USER=fossy
      - FOSSOLOGY_DB_PASSWORD=fossy
    command: web
    ports:
      - 8082:80
    depends_on:
      - db
      - scheduler
    volumes:
      - repository:/srv/fossology/repository/
    healthcheck:
      test: ["CMD-SHELL", "curl -sSf localhost/repo/api/v1/version || exit 1"]
      interval: 5s
      timeout: 5s
      retries: 5

  db:
    image: postgres:12.2
    restart: unless-stopped
    environment:
      - POSTGRES_DB=fossology
      - POSTGRES_USER=fossy
      - POSTGRES_PASSWORD=fossy
[...]

PR #1936, opened 2021-03-29
fix(Dockerfile): upgrade debian distribution

rkh-hash:
## Description

Debian distribution need to be upgraded in order to install a more recent version of `libpq`.

See issue: https://github.com/fossology/fossology/issues/1932 

### Changes

* Upgrade debian distrbution (strech -> buster) used during build stage.
* Upgrade `php7.0-cli` package to `php7.4-cli`

## How to test

Interact with a PostgreSQL instance with SCRAM authentication.

Closes #10


rkh-hash:
@GMishx , that's my proposal :)

There is no job to check if the image successfully built ?

rkh-hash:
@GMishx , I've bumped the version of `php-cli` package.

Does we need to add a new php unit test for this new version ?

GMishx:
Hi @rkh-hash , thanks for the contribution. Few points:
1. You want to write bumping php version to `php-7.3` in the description?
2. It will be good to update the Postgres to version 11 here: https://github.com/fossology/fossology/blob/aaec8d420e0d58f4e7663a9694e1fae44b42def6/docker-compose.yml#L69
3. It will be a nice idea to add PHP 7.3 to Travis tests.

rkh-hash:
Hello @GMishx , 

Oops, I've fixed the PR's description.

Additionally, I've made changes on the `docker-compose.yml` file and add a new php unit test with php 7.3 version.

GMishx:
Thanks. Will review the code.

It will be great help if @NicolasToussaint can test the docker container and composer file.

NicolasToussaint:
> Thanks. Will review the code.
> 
> It will be great help if @NicolasToussaint can test the docker container and composer file.

Will give it a try today

NicolasToussaint:
Hi ;
Been testing with a replicate of our production server, and it turns out that the database container refuses to start.

I produces these logs in a loop:
``` 
PostgreSQL Database directory appears to contain a database; Skipping initialization

2021-04-01 13:46:27.231 UTC [1] FATAL:  database files are incompatible with server
2021-04-01 13:46:27.231 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 9.6, which is not compatible with this version 11.9 (Debian 11.9-1.pgdg90+1).
```

My test:
- Run a recent build from Master (10 days ago or so), with *postgres:9.6*
- built image from branch: https://github.com/rkh-hash/fossology/tree/rkhhash/fix/1932
- redeployed new image with *postgres:11*

GMishx:
Thank you @NicolasToussaint . I think there is no getting around this issue other than system admin sitting and upgrading the DB.

What if we create a Wiki and attach it to the release notes when merging this branch? What you guys think about this?

rkh-hash:
Hi @GMishx ,

thank you @NicolasToussaint  for your review.

I'm not an expert about this fossology product.
But be advised that PostgreSQL version 9.6 will not been supported soon (at the end of this year).

Moreover, for supporting SCRAM authentication, the `libpq` package must be >= 10. 

I'm agree with you, in this case , the DB must be upgraded. Does it may exist a workaround ?

NicolasToussaint:
It looks indeed like some manual steps are required to upgrade postgresql ...
But because upgrading with Docker is so easy, I think an automated (or at least a helping) script will be necessary - otherwise we will have scores of questions about the upgrade failing, and how to fix it.

In any case, since Postgresql 9.6 is still supported in Debian Buster, I would propose to validate separately the upgrade to Debian Buster and Postgresql.

I tested Debian Buster + Postgresql 9.6, seems to work fine.

rkh-hash:
@NicolasToussaint ,

Initially, my goal was to upgrade the `libpq` package in order to support SCRAM authentication. To achieve that, I've just upgraded the Debian distribution to Buster and that's all.

So, as you said, If Debian Buster + **libpq >= 10** + PostgreSQL 9.6 is OK, that's enough at this time.

May I need to revert my changes on the `docker-compose.yaml` file ? @GMishx

GMishx:
Hi @rkh-hash , sorry I did not anticipated it to be that complex of a change. It will be good to revert the changes in `docker-compose.yaml` then.

Can you please open another issue to update the Postgres version to 11 in the compose file so we do not miss it?

rkh-hash:
Hi @GMishx , ok understood.

I've applied requested changes.

GMishx:
Thank you for creating #1951 @rkh-hash !

Can you please squash the commits to help us keep the commit history clean?

rkh-hash:
@GMishx done,

GMishx (approved):
Changes looks good.

Tested by @NicolasToussaint

rkh-hash:
@GMishx , a test with gcc-8 has failed, is that normal ?

https://travis-ci.org/github/fossology/fossology/jobs/766204229

GMishx:
Yeah, test case failures do occur often but they generally go away by a restart. But for some reason from a week, Travis is not allowing me to restart them.
image

NicolasToussaint:
I re-built the Docker image on the branch, and tested without noticing any problem.

NicolasToussaint (approved):

rkh-hash:
Hi @GMishx,

Do you think that this change would be included into the next release ?

GMishx:
I don't know. We can have it in this release as well. Just waiting for 3.10.0-rc2 release.

rkh-hash:
@GMishx ok, great.

I've enabled notifications :) .

rkh-hash:
@GMishx Hi, I guess that this feature would not been included into the future release 3.10.0.

Be advised that I can't consume my PostgreSQL instance because it require SCRAM authentication.
I think that I'm not alone in this case. 

Thank you.

PR #1948, opened 2021-04-05
feat(export) : Consolidating results per file or directories

JustinWonjaePark:
## Description

1) Consolidate findings of scanners into one.
2) Consolidate findings per file into per directory

### Changes

UI at Export Lists is slightly changed. There are three options. 1) raw, 2) per file, 3) per directory.
For 2) and 3), list of licenses at each row on the license list are merged into one.
For 3), list of files that has same result is merged into one row.

The result will be provided both on the browser and downloaded CSV file.

## How to test

Go to an upload > Export Lists
Select one of option at '5. Consolidate Results?'
Generate list. (You can print it on the browser or download the list.)

### DIFF
--- a/src/www/ui/template/ui-export-list.html.twig
+++ b/src/www/ui/template/ui-export-list.html.twig
@@ -1,4 +1,5 @@
 {# Copyright 2016-2017,2020 Siemens AG
+   Copyright (c) 2021 LG Electronics Inc.
 
    Copying and distribution of this file, with or without modification,
    are permitted in any medium without royalty provided the copyright notice and this notice are preserved.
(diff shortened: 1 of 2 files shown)


src/www/ui/ui-export-list.php
```diff
@@ -308,6 +309,13 @@ function Output()
     $exclude = GetParm("exclude", PARM_STRING);
     $formVars["exclude"] = $exclude;
 
+    $rawResult = (GetParm("consolidate", PARM_STRING) == "rawResult");
+    $formVars["rawResult"] = $consolidateLicenses;
```
GMishx, line 313:
```suggestion
    $formVars["rawResult"] = $rawResult;
```

src/www/ui/template/ui-export-list.html.twig
```diff
@@ -74,6 +75,31 @@
...
           </label>
         </li>
+        <li>
+          <span>
+            {{ "Consolidate Results?"|trans }}
+          </span>
+          <ul>
+            <li>
+              <lable>
+                <input type="radio" name="consolidate" value="rawResult" checked="checked" {% if rawResult %}checked{% endif %}>
```
GMishx, line 85: You might not wanna hard code `checked="checked"` in case `perFile` or `perDirectory` is true.

Try setting `$formVars["rawResult"] = true;` if `$consolidateLicenses` and `$consolidatePerDirectory` are `false` so that Raw result per file is checked on first load (or if nothing is selected by user).

GMishx (commented):
Few suggestions.

src/www/ui/template/ui-export-list.html.twig
JustinWonjaePark: Updated as suggested. Thanks @GMishx .

GMishx (approved):
Changes looks good.

GMishx:
Hi @JustinWonjaePark , can you please squash the commits into single commit?

GMishx (approved):
Tested, working as expected.

PR #1962, opened 2021-04-12
fix(decision): Create clearing events for irrelevant folder decisions

GMishx:
## Description

When someone marked a whole folder as irrelevant, the `clearing_event`s for the files went missing. This causes data inconsistency while getting the license for those `clearing_decision`s.

### Changes

1. Use `ClearingDecisionProcessor::makeDecisionFromLastEvents()` to generate `clearing_decision` and `clearing_event`.
2. The functions `markDirectoryAsDecisionTypeIfScannerDetected()` and `markDirectoryAsDecisionTypeIfUserEdited()` were no longer required to be separate and thus replaced with `markDirectoryAsDecisionTypeRec()`.
3. Handle errors in UI script if AJAX call fails.

## How to test

1. On old branch, upload a package and mark a folder as irrelevant using either `[Edit]` under Actions column or use the checkbox.
2. Check the clearing history for the files inside the folder, the license name will be missing.
3. Generate the Unified report and check the Irrelevant files table, the licenses will be missing from there as well.

1. Install the branch and repeat the steps above.
2. The clearing history should be correct and the unified report as well.
3. Repeat the test with `Mark decisions as do not use` button.
4. Try removing the decisions and check the clearing history/report.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -19,6 +19,8 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 namespace Fossology\Lib\Dao;
 
+use Fossology\Lib\BusinessRules\ClearingDecisionProcessor;
+use Fossology\Lib\Data\AgentRef;
 use Fossology\Lib\Data\Clearing\ClearingEvent;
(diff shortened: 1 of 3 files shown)


ag4ums (approved):
code looks good working as expected

Issue #1303, opened 2019-03-12
Create a job in maintenance agent to clean inactive tokens from DB.


GMishx:
### Description

Create a job in maintenance agent to clean inactive tokens from DB. As with current implementation, the expired tokens are kept in the DB and not deleted for audit purpose. This may result in quick build-up of huge amount of data in DB and may not be required for every instance.

One way to solve this is to create a new job in maintenance agent to clean such inactive tokens from DB which are expired **X** days ago.

#### How to reproduce

1. Install changes from #1290
1. Create some tokens.
1. Revoke tokens from UI.
1. Data remains in DB.

Aman-Codes:
Hello @GMishx, I would like to work on this issue. 
Just wanted to clarify that do we need delete all personal_access_tokens WHERE active = 'f' OR expire_on < today
Or we want an option to specify the value of X (delete tokens expired X days ago)

GMishx:
With the audit perspective, it is nice to keep the tokens for a predefined set of duration. So the "expire X days ago" sound a nice solution.

Aman-Codes:
@GMishx  What would be the right place to save that value?
1. In sysconfig table in database so that it's value can be edited from the Admin>Customize page.
But that may also require to create a dbmigrate script to add the new row in earlier versions of fossology.

2. In UI provide an input box to enter the value each time the maintagent job is scheduled.
In CLI provide the value as additional command line argument.

3. In fossology.conf file.

GMishx:
IMO, it should be put into the sysconfig table for easy access (you do not need migration script for this table).
Then the agent should accept it as a CLI parameter which will be fed by the UI reading from sysconfig table. So, if someone wants to run the agent from CLI, they can pass their own value without touching sysconfig.

What do you say?

Aman-Codes:
Yes that sounds good. Will add it in sysconfig table.

PR #1965, opened 2021-04-16
feat(maintagent): add job to remove expired tokens from database

Aman-Codes:
## Description

Created a new job in maintenance agent to clean inactive tokens from database which are expired X days ago.
Closes: #1303 

### Changes

- Added option to edit Max expired token retention period in Admin > Customize page
- Created option to trigger the new job from CLI by passing an additional parameter
- Option to trigger the job from GUI
- Fixed the broken link of Maintenance agent documentation

## How to test

- Run the agent from CLI ` sudo src/maintagent/agent/maintagent -t 30 `
- Run the agent from GUI

### DIFF
--- a/src/maintagent/agent/usage.c
+++ b/src/maintagent/agent/usage.c
@@ -41,6 +41,7 @@ FUNCTION void usage(char *name)
   printf("  -p   :: Verify file permissions (report only).\n");
   printf("  -P   :: Verify and fix file permissions.\n");
   printf("  -R   :: Remove uploads with no pfiles.\n");
+  printf("  -t # :: Remove personal access tokens expired # days ago.\n");
   printf("  -T   :: Remove orphaned temp tables.\n");
   printf("  -U   :: Process expired uploads (slow).\n");
   printf("  -Z   :: Remove orphaned files from the repository (slow).\n");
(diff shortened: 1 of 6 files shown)


Aman-Codes:
This new feature is working fine from CLI but the job never finishes from GUI and even no logs are generated for it.

GMishx:
Instead of adding jqargs in `src/maintagent/ui/maintagent.php`, try adding the time value (`PATMaxPostExpiryRetention`) when you receive `t` in this loop:
```php
$options = "-";
foreach ($_REQUEST as $key => $value) {
  if ($key == $value) {
    $options .= $value;
  }
}
```

So the `$options` become `-t30`. Can you please try that?

Aman-Codes:
Updated the code. Now it is working from UI as well

src/maintagent/agent/usage.c
```diff
@@ -30,8 +30,8 @@
 FUNCTION void usage(char *name)
 {
   printf("Usage: %s [options]\n", name);
-  printf("  -a   :: Run all non slow maintenance operations.\n");
-  printf("  -A   :: Run all maintenance operations.\n");
+  printf("  -a # :: Run all non slow maintenance operations where # is the retention period of expired tokens.\n");
+  printf("  -A # :: Run all maintenance operations where # is the retention period of expired tokens.\n");
```
GMishx, line 34: In my opinion, we can keep the removal of token as a separate process only and not mix with "all jobs". It will also keep the options `-a` and `-A` backwards compatible.

What do you think @ag4ums ?

mcjaeger:
(from discussion) ... pls remove the task from being run by default on fossology instances

GMishx (changes requested):

Aman-Codes:
@GMishx Updated it! Could you please review it again?

GMishx (approved):
Changes looks good.
Tested, working as expected.

PR #1975, opened 2021-05-02
perf(migration): remove offset to make the query faster

shaheemazmalmmd:
## Description

add a transaction and change limit and offset to equations.

## How to test

* In a large instance test migration of copyright_event


GMishx (approved):
Changes looks good.

Tested, working fine.

PR #1981, opened 2021-05-06
remove wrong 'extern "C"'

DerDakon:
glib headers may use C++ for some constructs, and they provide 'extern "C"' around the needed sections of there code already. Otherwise compilation may fail like this:

```
  In file included from /usr/include/glib-2.0/glib/gatomic.h:31,
                   from /usr/include/glib-2.0/glib/gthread.h:32,
                   from /usr/include/glib-2.0/glib/gasyncqueue.h:32,
                   from /usr/include/glib-2.0/glib.h:32,
                   from ../../../src/lib/c/libfossdbmanager.h:24,
                   from libfossdbmanagerclass.hpp:23,
                   from libfossAgentDatabaseHandler.cc:18:
  /usr/include/c++/10/type_traits:56:3: error: template with C linkage
     56 |   template<typename _Tp, _Tp __v>
        |   ^~~~~~~~
```

### DIFF
--- a/src/lib/c/libfossdbmanager.h
+++ b/src/lib/c/libfossdbmanager.h
@@ -23,6 +23,10 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include <stdarg.h>
 #include <glib.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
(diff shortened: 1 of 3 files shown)


GMishx (approved):
Changes looks good. Code complies successfully on GCC-10.

PR #1982, opened 2021-05-07
feat(monkbulk): Configurable irrelevant file scan

GMishx:
## Description

Allow user to choose if they want to ignore irrelevant files while doing a monkbulk scan or not.

### Changes

The existing scanner ignores irrelevant files by default but works on pfile. While inserting the data back, this pfile needs to be translated into `uploadtree_pk`. During this translation, if a pfile occurs multiple times and is marked as irrelevant in some places but not in other, will include all `uploadtree_pk` which is undesirable.

This PR fixes that by checking `clearing_decision` while inserting new results as well.

Also, depending on the user's selection of **Ignore Irrelevant files** checkbox, the queries will consider or ignore the `clearing_decision`.

image

## How to test

1. Upload a package where same file is copied in multiple folders.
2. Do a normal scan on the upload and mark some of the folders with duplicate file as irrelevant.
3. In other files, run the monkbulk scan ignoring irrelevant files.
    - The files marked as irrelevant should not be effected by this scan.
4. Run the monkbulk again, this time not ignoring the irrelevant files.
    - The irrelevant files should get this new result as well.
5. Try to reuse this upload with **"... bulk phrases from reused packages"** on decider and see if the ignoring of irrelevant files work there too.

### DIFF
--- a/src/decider/agent/BulkReuser.php
+++ b/src/decider/agent/BulkReuser.php
@@ -68,8 +68,8 @@ class BulkReuser
      */
     $deciderPlugin = plugin_find("agent_deciderjob");
     $dependecies = array();
-    $sql = "INSERT INTO license_ref_bulk (user_fk,group_fk,rf_text,upload_fk,uploadtree_fk) "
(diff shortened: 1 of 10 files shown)


shaheemazmalmmd (approved):

PR #1988, opened 2021-05-13
chore(os): Drop xenial support for eol

GMishx:
## Description

Drop support for Ubuntu Xenial as it [reached its EOL](https://help.ubuntu.com/community/EOL#Ubuntu_16.04_Xenial_Xerus). This means upgrading Travis jobs and Vagrant box too.

Also, there is an interesting observation that git does not print tag information in the format introduced in #1768 if the tag is on current commit. This causes Debian builds to fail.
The solution is to make format more flexible and add additional 0 revision.

# Note
This PR also updates the Vagrant box from Ubuntu Xenial to Ubuntu Focal!

### Changes

1. Drop support for Ubuntu Xenial.
2. Update Travis jobs from Xenial to Bionic
    - Not update to Focal due to limited PHP support.
    - See supported versions for Bionic: https://docs.travis-ci.com/user/reference/bionic/#php-support
    - See supported versions for Focal: https://docs.travis-ci.com/user/reference/focal/#php-support
3. Add test cases for GCC-10 and Postgres-13.
4. Remove `phpdbg` from PHP7.4 due to strict resource checks.
5. Handle case where building package for a tag (see description).
6. Revert changes from #1888 to remove signing from other build artifacts.
7. Other testcase fixes.

## How to test

Check the Travis logs.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -17,7 +17,6 @@ jobs:
           - 'debian:buster'
           - 'debian:stretch'
           - 'ubuntu:bionic'
-          - 'ubuntu:xenial'
           - 'ubuntu:focal'
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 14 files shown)


shaheemazmalmmd (approved):

PR #1994, opened 2021-05-19
fix(ui): Add scroll to NOTICE file modal

GMishx:
## Description

The Modal dialog to select a notice file for acknowledgement does not had max-height. Thus, if an upload contains many NOTICE files, it becomes very hard to scroll. Also, the inner elements of the dialog were draggable, which is not required.

### Changes

1. Add a scroll bar to notice file selection modal for acknowledgement, done by setting setting the max-height of the dialog to 70% of the viewport.
2. Remove the unnecessary draggable for the inner elements of modal.
3. Add additional UTF-8 checks to license browser.

## How to test

1. Upload a file with many NOTICE files.
2. Goto license browser and add an acknowledgement.
3. Click on "Select from Notice file".

- The modal dialog should not exceed 70% height of the page.
- The inner elements of the dialog should move with the modal, not separately.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -464,6 +464,7 @@ INSERT INTO clearing_decision (
       $row['acknowledgement'] = "";
     }
 
+    $changeTo = StringOperation::replaceUnicodeControlChar($changeTo, false);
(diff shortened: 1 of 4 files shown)


shaheemazmalmmd (approved):
Code looks good. Tested works as expected

PR #1996, opened 2021-05-21
fix(ui): License text editor

GMishx:
## Description

Since the license text is read from the "title" of the element, which is already [HTML encoded in PHP](https://github.com/fossology/fossology/blob/84a1897e32cc6556713f6efabd3c9ea3bd0dd4c7/src/www/ui/ajax-clearing-view.php#L282), the textarea should be decoded first.

### Changes

1. Use jQuery trickery to decode HTML with `htmlDecode()`.
2. Use the decoded value to fill textarea.

## How to test

1. Edit a license text and add some HTML characters like `&`, `"`, `'`, `<` or `>` in the license text.
2. Submit the decision and edit the same license text again.
    - In master branch, the special characters will be shows in HTML &amp; style.
    - In this branch, they should be shown in human readable format.
3. Check the above points for bulk license text as well.

### DIFF
--- a/src/www/ui/scripts/change-license-common.js
+++ b/src/www/ui/scripts/change-license-common.js
@@ -240,6 +240,14 @@ function removeMainLicense(uploadId,licenseId) {
   }
 }
 
+function htmlDecode(value) {
+    if (value) {
+        return $('<div/>').html(value).text();
+    } else {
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):

PR #1958, opened 2021-04-08
feat(export): Download results in spreadsheet (xlsx)

JustinWonjaePark:
## Description

Export list in Spreadsheet

### Changes

UI at Export Lists is slightly changed. There are three options. 
1) don't download license list (default)
2) download license list in CSV
3) download license list in Spreadsheet

## How to test

Go to an upload > Export Lists
Select one of option at '5. Download License List after creation?'
Generate list.

PR #2005, opened 2021-06-03
fix(debian): Add php-gd package as dependency

GMishx:
## Description

Add `php-gd` dependency to Debian `fossology-web` package required by `PhpOffice\PhpSpreadsheet`.

## How to test

- Build and install debian packages.
- Check if you can export spreadsheet of licenses.

See #1958 for more.


shaheemazmalmmd (approved):
code looks good.

PR #2008, opened 2021-06-07
feat(gitpod): Inital contribution

GMishx:
## Description

Initial commit for [gitpod.io](https://gitpod.io) integration.

#### What is working

- [x] The `make` installation
- [x] `fo-postinstall`
- [x] DB
- [x] Uploads
- [x] Other editor things

#### What is missing

- Way to do simple `make install`.
- The base image is tagged to `latest`, does not have specific tag for `focal`.
- Didn't check the behavior of `make test`.

#### Workaround used

- Since gitpod stores contents under `/workspace/` only, the FOSSology needs to be installed under it.
  + To implement this, following paths needed to be changes: `PREFIX`, `INITDIR`, `REPODIR`, `LOCALSTATEDIR`, `APACHE2_SITE_DIR`, `SYSCONFDIR`, `PROJECTUSER` and `PROJECTGROUP`.
- Changing the group of user in gitpod is difficult so FOSSology and DB are installed/setup under `gitpod:gitpod`.
- New aliases:
  + `fossrun` : Runs scheduler in shell, not as daemon
  + `fossinstallclean` : Runs `make clean` and then `make install` with all the appropriate path changes. Followed with `fo-postinstall`
  + `fossinstallnoclean` : Runs `make install` with all the appropriate path changes, no `clean`. Followed with `fo-postinstall`
  + `pg_start` : Starts postgres 12
  + `pg_stop` : Stops postgres

## Workflow
1. The `.gitpod.Dockerfile` is built and all the dependencies are installed. In the same image, the PHP settings are fixed too and Postgres 12 is setup. Also, create the helpful aliases.
2. Once the image is built, a workspace can be created.
3. Gitpod allows prebuilt which setups some commands on top of the Docker image and save boot time.
    - This includes installing some dependencies which can't be installed during image creation.
    - Compiling and installing FOSSology.
    - Setting up DB.
    - Setting up Apache.
4. When a user runs this workspace, another set of commands can be run:
    - Fixing some folders outside `/workspace`.
    - Restarting apache.
    - Running Scheduler in a separate terminal tab.

To take full advantage of prebuilds, the integration with gitpod needs to be setup in FOSSology repository. Doing so, will also allow to create a new workspace for reviewing PRs.

## How to test

Check https://gitpod.io/#https://github.com/GMishx/fossology

Review current PR: https://gitpod.io/#https://github.com/fossology/fossology/pull/2008

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,83 @@
+# FOSSology gitpod.io Dockerfile
+# Copyright (C) 2021 Siemens AG
+# Author: Gaurav Mishra <mishra.gaurav@siemens.com>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# version 2 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# Description: Docker container image recipe for gitpod.io workspace
+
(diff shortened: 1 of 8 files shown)


ag4ums:
code looks good... creating workspace in gitpod

PR #2009, opened 2021-06-08
fix(ui): Break long lines in copyright table

GMishx:
## Description

Break very long lines with the help of CSS properties.

### Changes

Created new class `wordbreaktable` which can be applied on tables and apply `word-break` & `overflow-wrap` properties to all `td`s.

## How to test
Scan an upload with very long copyrights which can not be broken.
I tested it with [Angular-9.1.1](https://github.com/angular/angular/releases/tag/9.1.1). Copyright captures a string with a very long base64 string which breaks the UI.


shaheemazmalmmd (approved):

PR #2010, opened 2021-06-09
fix(api): Add missing reuser options

GMishx:
## Description

Add the missing "Reuse report conf" and "Reuse deactivated copyright" options to the REST API.

### Changes

Added `reuse_report` and `reuse_copyright` to reuser object in REST API.

## How to test

Create a new job with REST API using new options.


shaheemazmalmmd (approved):

PR #2011, opened 2021-06-12
fix(copyright): Make check strict

GMishx:
## Description

While editing/fetching the copyrights, the association with uploadtree_pk was not very strong. It caused issue while editing copyrights from a single file view instead of aggregated table view.

This causes problem when same upload has duplicate files. Since `copyright` table holds information of `pfile_fk` and `agent_fk` only, there will be only one entry for a copyright in those files. But if someone made some changes, the `copyright_event` is involved which takes `uploadtree_fk` into account. This caused ambiguity on which copyright to show.

### Changes

1. Add checks for `uploadtree_fk` while fetching copyrights to get only relevant copyrights.
2. Update the check while inserting edits in `copyright_event` to make sure all `uploadtree_fk` are updated.

## How to test

1. Upload a package with duplicate files which contains copyright texts.
2. Update some copyrights from single file view.
3. See if the updates are reflected in the aggregated table view.
4. Editing copyrights from aggregated view should be consistent as well.

### DIFF
--- a/src/copyright/ui/ajax-copyright-hist.php
+++ b/src/copyright/ui/ajax-copyright-hist.php
@@ -278,21 +278,20 @@ class CopyrightHistogramProcessPost extends FO_Plugin
     $filterParms = $params;
     $searchFilter = $this->addSearchFilter($filterParms);
 
-    $activatedClause = "";
+    $activatedClause = "ce.is_enabled = 'false'";
     if ($activated) {
-      $activatedClause = "NOT";
+      $activatedClause = "ce.is_enabled IS NULL OR ce.is_enabled = 'true'";
     }
     $unorderedQuery = "FROM $tableName AS cp " .
         "INNER JOIN $uploadTreeTableName AS UT ON cp.pfile_fk = UT.pfile_fk " .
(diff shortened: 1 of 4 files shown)


shaheemazmalmmd:
Copyright loading issue with new changes. 

To Reproduce :

upload FCKeditor go to sample folder and default.html page(See highlight). See the entry in database(using uploadtree_pk) and not in UI.

screenShot17

GMishx:
> Copyright loading issue with new changes.
> 
> To Reproduce :
> 
> upload FCKeditor go to sample folder and default.html page(See highlight). See the entry in database(using uploadtree_pk) and not in UI.
> 
> screenShot17

Hello @shaheemazmalmmd , unfortunately I can not recreate the issue on my machine. I just uploaded another package [30-seconds-of-code](https://github.com/30-seconds/30-seconds-of-code) (can not find the source for FCKeditor) and this is what it says:
image

Also seeing from your snapshot, the branch and commit seems to be wrong. Can you please check that?

shaheemazmalmmd (approved):
Code looks good. Needs test

PR #2016, opened 2021-06-16
fix(auth): fix call to add_user() when login from external auth

NicolasToussaint:
## Description
When introducing the 'default visibility' feature, the call to `add_user` had not been updated, and therefore the user creating was failing.

### Changes

I added a configuration entry in `fossology.conf` to configure the default visibility for new user creations, and used it for new user creations.

## How to test

Configure external configuration, then log in with a user that is not already in the database.
The user should be assigned the "default visibility" that is configured in the `fossology.conf`  file

### DIFF
--- a/src/www/ui/core-auth.php
+++ b/src/www/ui/core-auth.php
@@ -289,7 +289,8 @@ class core_auth extends FO_Plugin
         $Email = $this->authExternal['emailAuthExternal'];
         /* Set default list of agents when a new user is created */
         $agentList = $GLOBALS['SysConf']['EXT_AUTH']['CONF_EXT_AUTH_NEW_USER_AGENT_LIST'];
-        add_user($User, $Desc, $Hash, $Perm, $Email, $Email_notify, $agentList, $Folder);
+        add_user($User, $Desc, $Hash, $Perm, $Email, $Email_notify,
+          $GLOBALS['SysConf']['EXT_AUTH']['CONF_EXT_AUTH_NEW_USER_UPLOAD_VISIBILITY'], $agentList, $Folder);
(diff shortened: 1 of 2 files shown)


src/www/ui/core-auth.php
```diff
@@ -289,7 +289,8 @@ function checkUsernameAndPassword($userName, $password)
         $Email = $this->authExternal['emailAuthExternal'];
         /* Set default list of agents when a new user is created */
         $agentList = $GLOBALS['SysConf']['EXT_AUTH']['CONF_EXT_AUTH_NEW_USER_AGENT_LIST'];
-        add_user($User, $Desc, $Hash, $Perm, $Email, $Email_notify, $agentList, $Folder);
+        add_user($User, $Desc, $Hash, $Perm, $Email, $Email_notify,
+          $GLOBALS['SysConf']['EXT_AUTH']['CONF_EXT_AUTH_NEW_USER_UPLOAD_VISIBILITY'], $agentList, $Folder);
```
GMishx, line 293: I think it will be better if we can fetch this value from the DB rather than the `.conf` file which is easier to update.

The value can be fetched with `$GLOBALS['SysConf']['SYSCONFIG']['UploadVisibility']`.

image

NicolasToussaint: oh, of course, thanks. Didn't realize that was configurable in the web interface.

I guess the whole ext auth settings should also be available from the GUI, that would be a nice "good first issue"

GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

PR #2022, opened 2021-06-18
fix(ui): Fix the modal height for small screens

GMishx:
## Description

Fix the modal box height for small screens.

### Changes

Change the `max-height` for non-popup modal to fix the content, rather than view-port.

## How to test

Open single file view and check the modals created for "User decision", "Bulk Recognition" and "Clearing History". The brown background should cover the whole content.


shaheemazmalmmd (approved):

PR #2036, opened 2021-06-30
feat(spasht): Show effective score

### LINKED ISSUE #1964: Include ClearlyDefined score in Spasht
That we now can access ClearlyDefined.io through FOSSology is pretty neat, but the data there is of mixed quality.

Luckily each package on ClearlyDefined.io includes a score. Filtering by that score would be really useful.

For example, if a package has a score of at least 87%, I would say that data is complete enough that a scan with FOSSology will not give me any new findings. If that could be integrated somehow into FOSSology, that’d be great.

GMishx:
## Description

Display scores from ClearlyDefined in Spahst configuration.
The effective is only considered here. Can be extended to tool scores.

### Changes

1. Update the Coordinate object to hold score information.
2. Update the DefinitionSummary ro hold score information.
3. Filter search results based on scores (desc) by default.
4. Update show_definitions and show_details tables with score.

## How to test

1. Upload a package and goto "Spasht" configuration page.
2. Search for package and check the score in the results.
  - They should be available on the result and details table.

Closes #1964

#### Screenshots
image
image

### DIFF
--- a/src/lib/php/Data/Spasht/Coordinate.php
+++ b/src/lib/php/Data/Spasht/Coordinate.php
@@ -55,6 +55,12 @@ class Coordinate
   private $namespace;
 
   /**
+   * @var integer $score
+   * Package score
+   */
+  private $score;
+
+  /**
    * Set the object based on array returned by API
    * @param array $obj Array containing the data
    * @throws \InvalidArgumentException If the input obj does not contain
@@ -132,6 +138,14 @@ class Coordinate
   }
 
   /**
+   * @return integer
+   */
+  public function getScore()
+  {
+    return $this->score;
+  }
+
+  /**
    * Helper function to generate Coordinate from string
    * @param string $coordinate Coordinates returned from definintions API
    * @throws \InvalidArgumentException If the coordinate string is not in valid
@@ -152,5 +166,15 @@ class Coordinate
       'revision' => $parts[4]
     ];
     return new Coordinate($obj);
+  }
+
+  /**
+   * Set the score for this coordinate of package.
+   *
+   * @param integer $score Score
+   */
+  public function setScore($score)
+  {
+    $this->score = intval($score);
   }
 }
--- a/src/lib/php/Data/Spasht/DefinitionSummary.php
+++ b/src/lib/php/Data/Spasht/DefinitionSummary.php
@@ -57,6 +57,12 @@ class DefinitionSummary
   private $discoveredLicenses;
 
   /**
+   * @var integer $score
+   * Package score
+   */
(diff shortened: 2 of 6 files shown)


silverhook:
Hi, @GMishx , I apologise for the late reply.

I’m on 3.10.0 (via Docker) and don’t see that functionality yet (in fact, currently Splasht fails accessing the ClearlyDefined API at all, error code 0).

But in general this info looks good. What I wonder though is which score that is. I suspect it’s the Effective score of the [ClearlyLicensed](https://github.com/clearlydefined/license-score/blob/master/ClearlyLicensedMetrics.md) part (and not ClearlyDefined part), but that might make sense to explicitly say in the UI.

As a side-note, is Splasht now able to parse PackageURL? That would make things quite a bit easier.

GMishx:
Thanks for the comments @silverhook 

> I’m on 3.10.0 (via Docker) and don’t see that functionality yet (in fact, currently Splasht fails accessing the ClearlyDefined API at all, error code 0).

I can provide a Docker image for this branch to try out. And can you please create issue if Spasht is currently not working?

> But in general this info looks good. What I wonder though is which score that is. I suspect it’s the Effective score of the [ClearlyLicensed](https://github.com/clearlydefined/license-score/blob/master/ClearlyLicensedMetrics.md) part (and not ClearlyDefined part), but that might make sense to explicitly say in the UI.

snap

For example, out of all these scores provided by ClearlyDefined, the "Overall Effective score" of 87 is displayed here. I am not very sure it's part of ClearlyLicensed or ClearlyDefined. If you can help, I will add it explicitly to the UI :-)

> As a side-note, is Splasht now able to parse PackageURL? That would make things quite a bit easier.

Right now no, only the syntax understood by ClearlyDefined.io is used. But yeah, it looks achievable as a separate feature request as PURL are also standard and can be translated to ClearlyDefined format.
For reference, here is the specification: https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst

silverhook:
> For example, out of all these scores provided by ClearlyDefined, the "Overall Effective score" of 87 is displayed here. I am not very sure it's part of ClearlyLicensed or ClearlyDefined. If you can help, I will add it explicitly to the UI :-)

Ah, that seems to be the average between the ClearlyDefined (i.e. is the package defined well enough to clearly identify it exactly) and the ClearlyLicensed (i.e. license and copyright info) scores.

Regarding the other questions, I’ll open up new tickets, if needed.

shaheemazmalmmd (approved):
code looks good tested works as expected.

PR #2029, opened 2021-06-22
feat(reso): new agent for REUSE.Software standard

NicolasToussaint:
Fixes #1833 

## Description

As discussed in issue #1833, implements a part of the RESUSE.Software specification, by detecting the `.license` suffixed files, and applying the license detected by the *Ojo* agent to the associated file.

### Changes
A new agent, named **ReSo**.

More details already available on Wiki page: [ReSo-(REUSE.Software)](https://github.com/fossology/fossology/wiki/ReSo-%28REUSE.Software%29)

Known limitations, bugs:
- Commas appear in the license table for files taht have been scanned multiple times. No incidence on workflow detected.
image
- the name of the agent displayed could be changed from '*reso* to *REUSE.Software*, to be discussed
- Copyrights not handled
- Next, [still-in-discussion](https://github.com/spdx/spdx-spec/issues/502) REUSE.yml files could be supported

## How to test

1. Start new upload - you may want to use example package available here: https://github.com/fossology/fossology/wiki/ReSo-%28REUSE.Software%29#testing
2. Select `REUSE.Software Analysis -> will also trigger a scan with Ojo agent` option
3. Select `... scanners matches if Ojo or REUSE.Software findings are no contradiction with other findings` option
4. image
5. Start Upload
6. Check that licenses are applied accordingly, while license conflicts are not concluded
7. image

PR #2040, opened 2021-07-02
feat(reso): new agent for REUSE.Software standard

### LINKED ISSUE #1833: Detect 'foobar.license' files, as part of RESUE.Software standard
### Description

Following discussion in #1592, this issue proposes a first step.

The idea would be to support the `file.license` way of declaring a license and copyright for binary files (or any file in which compliance text cannot be inserted, like `json`or `yaml` files)

Extract of the specification: https://reuse.software/spec/#comment-headers
> If a file is not a plain text file or does not permit the inclusion of comments, the comment header that declares the file’s Copyright and Licensing Information SHOULD be in an adjacent file of the same name with the additional extension .license (example: cat.jpg.license if the original file is cat.jpg).

I am not sure if the mechanism should be a new agent, or be implemented inside the existing Ojo agent.
I tend to like the idea of reusing Ojo since:
- The `file.license` files will contain SPDX tags
- The files for which compliance is described with a `file.license` file can be safely identified automatically, and this is the spirit of the Ojo agent.

I am not sure, either if the Copyright should be handled together, or separately by the Copyright agent.
If it should be handled separately, then I suggest that this issue concentrates on the License, and we will see the Copyright in an other issue.

Upload an archive containing:
```
config.json
config.json.license
volcano.jpg
voclano.jpg.license
```

The files `*.license` both contain someting like:
```
# SPDX-FileCopyrightText: 2016, 2018-2019 Jane Doe <jane@example.com>
# SPDX-FileCopyrightText: 2019 Example Company
#
# SPDX-License-Identifier: GPL-3.0-or-later
```

Then Fossology would  identify all 4 files as `GPL-3.0-or-later` automatically (and possibly the Copyrights too).

NicolasToussaint:
Replaces PR: https://github.com/fossology/fossology/pull/2029

Fixes #1833

### DIFF
--- a/src/www/ui/api/Models/Analysis.php
+++ b/src/www/ui/api/Models/Analysis.php
@@ -1,6 +1,7 @@
 <?php
 /***************************************************************
 Copyright (C) 2017 Siemens AG
+Copyright (C) 2021 Orange by Piotr Pszczola <piotr.pszczola@orange.com>
 
 This program is free software; you can redistribute it and/or
 modify it under the terms of the GNU General Public License
@@ -73,6 +74,11 @@ class Analysis
    * @var boolean $pkgagent
    * Whether to schedule package agent or not
    */
+  private $reso;
+  /**
+   * @var boolean $pkgagent
+   * Whether to schedule package agent or not
+   */
   private $pkgagent;
 
   /**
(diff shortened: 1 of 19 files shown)


src/www/ui/api/Models/Analysis.php
```diff
@@ -69,6 +70,11 @@ class Analysis
    * Whether to schedule ojo agent or not
    */
   private $ojo;
+  /**
+   * @var boolean $pkgagent
```
GMishx, line 74: Oops, the doc was updated for the variable.

GMishx (approved):
Changes looks good. Just one documentation needs update.

GMishx (approved):
Tested, working as expected.

src/www/ui/api/Models/Analysis.php
NicolasToussaint: well spotted ;-)

PR #2042, opened 2021-07-06
feat(expose-headers): added the expose headers option

### LINKED ISSUE #2041: Getting all the headers in response from the Rest APIs
The Access-Control-Expose-Headers response header indicates which headers can be exposed as part of the response by listing their names.
By default, only the 6 CORS-safelisted response headers are exposed:
`Cache-Control`, `Content-Language`, `Content-Type`, `Expires`, `Last-Modified`, `Pragma`
If we want clients to be able to access other headers, you have to list them using the Access-Control-Expose-Headers header.

We need to set `Access-Control-Expose-Headers` in our backend to expose our custom response headers.
Taken from [Access-Control-Expose-Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers)

Therefore, we have to code your backend to set Access-Control-Expose-Headers such that it returns:
Access-Control-Expose-Headers: `x-total-pages`

Shruti3004:
## Description

Added the `Access-Control-Expose-Headers`.
closes #2041 

### Changes

- Created the global variable for Allowed headers in `common-sysconfig.php`.
- Added the `Access-Control-Expose-Headers` property in `AuthController.php`.

### DIFF
--- a/src/www/ui/api/Controllers/AuthController.php
+++ b/src/www/ui/api/Controllers/AuthController.php
@@ -57,12 +57,7 @@ class AuthController extends RestController
   }
   public function optionsVerification($request, $response, $args)
   {
-    global $SysConf;
-    return $response->withStatus(204)
-      ->withHeader('Access-Control-Allow-Origin', $SysConf['SYSCONFIG']['CorsOrigins'])
-      ->withHeader('Access-Control-Allow-Headers', 'X-Requested-With, Content-Type, Accept, Origin, Authorization')
-      ->withHeader('Access-Control-Allow-Methods', 'GET, POST, PUT, DELETE, PATCH, OPTIONS')
-      ->withHeader('Access-Control-Allow-Credentials', 'true');
+    return $response->withStatus(204);
   }
   /**
    * Get the JWT authentication headers for the user
--- a/src/www/ui/api/Middlewares/RestAuthMiddleware.php
+++ b/src/www/ui/api/Middlewares/RestAuthMiddleware.php
@@ -48,6 +48,7 @@ class RestAuthMiddleware
    */
   public function __invoke($request, $response, $next)
   {
+    global $SysConf;
     $requestUri = $request->getUri();
     if (stristr($request->getMethod(), "options") !== false) {
       $response = $next($request, $response);
@@ -94,6 +95,11 @@ class RestAuthMiddleware
           $tokenValid->getCode());
       }
     }
-    return $response;
+    return $response
+      ->withHeader('Access-Control-Allow-Origin', $SysConf['SYSCONFIG']['CorsOrigins'])
+      ->withHeader('Access-Control-Expose-Headers', 'Look-at, X-Total-Pages, Retry-After')
+      ->withHeader('Access-Control-Allow-Headers', 'X-Requested-With, Content-Type, Accept, Origin, Authorization, action, active, copyright, Content-Type, description, filename, filesizemax, filesizemin, folderDescription, folderId, folderName, groupName, ignoreScm, license, limit, name, page, parent, parentFolder, public, reportFormat, searchType, tag, upload, uploadDescription, uploadId, uploadType')
+      ->withHeader('Access-Control-Allow-Methods', 'GET, POST, PUT, DELETE, PATCH, OPTIONS')
+      ->withHeader('Access-Control-Allow-Credentials', 'true');
   }
 }


src/lib/php/common-sysconfig.php
```diff
@@ -218,6 +218,13 @@ function Populate_sysconfig()
   $valueArray[$variable] = array("'$variable'", "'localhost:3000'", "'$prompt'",
     strval(CONFIG_TYPE_TEXT), "'PAT'", "3", "'$desc'", "null", "null");
 
+  /*  Expose Headers */
+  $variable = "Headers";
+  $prompt = _('Allowed headers for REST API');
+  $desc = _('"*" for anywhere');
+  $valueArray[$variable] = array("'$variable'", "'x-total-pages'", "'$prompt'",
```
GMishx, line 225: I don't think it needs to be a configuration. In order for API to function properly, the admin any way needs to allow all the headers. It will be better to have it in code.

src/www/ui/api/Controllers/AuthController.php
```diff
@@ -60,6 +60,7 @@ public function optionsVerification($request, $response, $args)
     global $SysConf;
     return $response->withStatus(204)
       ->withHeader('Access-Control-Allow-Origin', $SysConf['SYSCONFIG']['CorsOrigins'])
+      ->withHeader('Access-Control-Expose-Headers', $SysConf['SYSCONFIG']['Headers'])
```
GMishx, line 63: I was having discussion with @shaheemazmalmmd. He suggested, and also a recommended practice to send these headers for all requests. So how about moving all the CORS headers and put them at https://github.com/fossology/fossology/blob/1d7f5f9fc8c47781fcd327a46727b1a84789aff6/src/www/ui/api/Middlewares/RestAuthMiddleware.php#L97 instead?

That way, we can leave just the 204 response in this function and all the CORS headers will be available for all the responses from the API. What do you think?

GMishx (commented):
Some suggestions to harmonize CORS.

src/lib/php/common-sysconfig.php
Shruti3004: Okay so will pass the wildcard operator

src/www/ui/api/Controllers/AuthController.php
Shruti3004: Yes, surely we can do that.

Shruti3004: @GMishx updated, please check does this wok

GMishx: Awesome!

src/www/ui/api/Middlewares/RestAuthMiddleware.php
```diff
@@ -94,6 +95,11 @@ public function __invoke($request, $response, $next)
           $tokenValid->getCode());
       }
     }
-    return $response;
+    return $response
+      ->withHeader('Access-Control-Allow-Origin', $SysConf['SYSCONFIG']['CorsOrigins'])
+      ->withHeader('Access-Control-Expose-Headers', '*')
+      ->withHeader('Access-Control-Allow-Headers', 'X-Requested-With, Content-Type, Accept, Origin, Authorization')
```
GMishx, line 101: Currently, following headers are accessed by FOSSology. Please add them to the list as well.
`action`, `active`, `copyright`, `Content-Type`, `description`, `filename`, `filesizemax`, `filesizemin`, `folderDescription`, `folderId`, `folderName`, `groupName`, `ignoreScm`, `license`, `limit`, `name`, `page`, `parent`, `parentFolder`, `public`, `reportFormat`, `searchType`, `tag`, `upload`, `uploadDescription`, `uploadId` and `uploadType`.

GMishx: Now I understand why public APIs don't generally read parameters from headers :sweat_smile:

GMishx (changes requested):
A few corrections.

src/www/ui/api/Middlewares/RestAuthMiddleware.php
Shruti3004: But @GMishx I think we don't need to add all these headers `Access-Control-Expose-Headers` this is for the response headers so in response I guess we need to add `Look-at`, `X-Total-Pages` and `Retry-After`

Shruti3004: @GMishx updated as you suggested and thanks for providing the headers list :)

GMishx (approved):
Changes looks good.

Shruti3004:
> Changes looks good.

@GMishx Now, probably this pr will work you as well [https://github.com/fossology/FOSSologyUI/pull/81](https://github.com/fossology/FOSSologyUI/pull/81).

GMishx:
> > Changes looks good.
> 
> @GMishx Now, probably this pr will work you as well [fossology/FOSSologyUI#81](https://github.com/fossology/FOSSologyUI/pull/81).

Ah, magic! :tada:

shaheemazmalmmd (approved):
Changes looks good. Tested works as expected.

PR #2044, opened 2021-07-07
fix(ojo): Set candidate license creator

GMishx:
## Description

With the new audit information added for `license_candidate` table, add `rf_user_fk_created` and `rf_user_fk_modified` from ojo while creating new candidate license.

### Changes

Add the user id for `rf_user_fk_created` and `rf_user_fk_modified` while creating new candidate license.

## How to test

1. Create dummy package with `SPDX-License-Identifier` where the license name does not exists in main license table.
2. Ojo should create new candidate license for the file.
3. Edit the license and check the "User Created" and "User Modified" fields.


shaheemazmalmmd (approved):

PR #2048, opened 2021-07-08
feat(rest): Filter licenses by kind

### LINKED ISSUE #2045: REST API: Added the isCandidate parameter in getLicenses Api
### Description

Currently the api `/license` gives the all licenses whether it is candidate or not, and then for the candidate License page we are sorting the response by checking `isCandidate= true`, so in this way we are not able to properly get all the licenses as specified in the entries of candidate license page.

### Screenshots
Screenshot from 2021-07-08 09-03-17

Refrence: [https://github.com/fossology/FOSSologyUI/pull/83](https://github.com/fossology/FOSSologyUI/pull/83)

GMishx:
## Description

Filter licenses based on query from `/license` endpoint. Possible values:
- `?kind=all`
  + Get all licenses
  + Default option
- `?kind=candidate`
  + Get only candidate licenses
- `?kind=main`
  + Get only main licenses

### Changes

1. Added new query parameter `kind` to `/license` endpoint.

## How to test

1. List of licenses with and without `?kind=all` should be same from endpoint.
2. List candidate licenses with `?kind=candidate`.
3. List main licenses with `?kind=main`.

Closes #2045

### DIFF
--- a/src/www/ui/api/Controllers/LicenseController.php
+++ b/src/www/ui/api/Controllers/LicenseController.php
@@ -145,6 +145,7 @@ class LicenseController extends RestController
   public function getAllLicenses($request, $response, $args)
   {
     $retVal = null;
+    $query = $request->getQueryParams();
     $limit = $request->getHeaderLine(self::LIMIT_PARAM);
     if (! empty($limit)) {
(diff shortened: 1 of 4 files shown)


shaheemazmalmmd (approved):

PR #2052, opened 2021-07-13
Fix/openapi linting

deveaud-m:
Enforce OAS3.0 conformity for the OpenAPI spec via Github action.

Update the specification to pass the linting, mainly adding `operationId` to all endpoints and enhancing `description` when required.


GMishx (approved):
Changes looks good.

PR #2064, opened 2021-07-21
feat(rest): New info and health endpoints

GMishx:
## Description

New endpoints added:
1. `/info`
2. `/health`

Endpoint deprecated:
1. `/version`

Endpoint removed:
1. `/auth`

### Changes

1. `/info` endpoint
  - Return information about API including name, description, version, security, contact and license.
  - The existing endpoint `/version` is marked as deprecated and provides same information as `/info`.
2. `/health` endpoint
  - Proides information about DB and scheduler status.
    ```json
    {"status":"OK","db":{"status":"OK"},"scheduler":{"status":"OK"}}
    ```
  - In case DB is down, the status goes to "ERROR" mode.
  - If only scheduler is not working, the status is "WARN" and scheduler object has status "ERROR".
3. `/auth` endpoint
  - Endpoint is removed as it was deprecated in version 1.0.1

## How to test

1. Check the response of `/info`
2. Check the response of `/health`
  - With scheduler down
  - With DB down
3. Check the response of `/version`


deveaud-m:
This will introduce breaking changes in the Python wrapper but I guess this change makes a lot of sense.

Getting the state of the Fossology scheduler via the API is a very useful information.

LGTM :+1:

shaheemazmalmmd (approved):
code looks good. needs test.

deveaud-m (approved):

PR #2068, opened 2021-07-28
fix(ui/uploadVcsPage):Fix upload from VCS for parameterize agent

itssingh:
## Description

Upload from version control system was not working for parameterize
agent like reuser or decider. This commit will fix the same.

Signed-off-by: Sarita Singh <saritasingh.0425@gmail.com>

### Changes

image

## How to test

-  Upload from Version control system
-  Schedule a parameterize agent like decider and check Jobs
-  decider will be scheduled along with it's dependencies

### DIFF
--- a/src/www/ui/page/UploadVcsPage.php
+++ b/src/www/ui/page/UploadVcsPage.php
@@ -161,20 +161,8 @@ class UploadVcsPage extends UploadPageBase
       return array(false, _($ErrorMsg), $description);
     }
 
-    AgentCheckBoxDo($jobpk, $uploadId);
-
-    $msg = "";
-    /** check if the scheudler is running */
-    $status = GetRunnableJobList();
-    if (empty($status)) {
-      $msg .= _("Is the scheduler running? ");
-    }
-    $Url = Traceback_uri() . "?mod=showjobs&upload=$uploadId";
-    $text = _("The upload");
-    $text1 = _("has been queued. It is");
-    $msg .= "$text $Name $text1 ";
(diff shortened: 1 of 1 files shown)


src/www/ui/page/UploadVcsPage.php
```diff
@@ -161,20 +161,8 @@ protected function handleUpload(Request $request)
...
-    if (empty($status)) {
-      $msg .= _("Is the scheduler running? ");
-    }
-    $Url = Traceback_uri() . "?mod=showjobs&upload=$uploadId";
-    $text = _("The upload");
-    $text1 = _("has been queued. It is");
-    $msg .= "$text $Name $text1 ";
-    $keep =  "<a href='$Url'>upload #" . $uploadId . "</a>.\n";
-    return array(true, $msg.$keep, $description, $uploadId);
+    $message = $this->postUploadAddJobs($request, $Name, $uploadId);
```
GMishx, line 164: Please add `$jobpk` as the 4th parameter to prevent additional job being created.

GMishx (approved):
Changes looks good.
Tested, working as expected.

src/www/ui/page/UploadVcsPage.php
itssingh: Thanks @GMishx, Actually this solved the issue of scheduling agents with different name(once as `wget` and other with the `file name`). This issue was due to duplicate job created in `postUploadAddJobs` function as '$jobpk' wasn't passed.

PR #2076, opened 2021-08-09
ci(actions): Build Docker images in Actions

GMishx:
## Description

Since [free auto builds are no longer available on Docker Hub](https://www.docker.com/blog/changes-to-docker-hub-autobuilds/), use GitHub Actions to build the image and push to Docker Hub.

### Changes

1. Added new build workflow for builds on push to master and release which will build the "fossology/fossology:latest" and "fossology/fossology:scanner" images.
2. Modify release workflow to add build steps for "fossology/fossology:<release_version>" images.

## How to test

Check the build logs from: https://github.com/GMishx/fossology/runs/3277301909?check_suite_focus=true#logs

#### Note
Since builds are using QEMU and Buildx, there is a posibility to compile fossology for other architectures as well. I did [a trial with linux/arm64](https://github.com/GMishx/fossology/runs/3263935510?check_suite_focus=true#logs) but it was extremely slow (2h compared to 15m).
If anyone is in favour of it, following architectures are supported by QEMU on Actions and can produce builds from them:
- linux/amd64
- linux/arm64
- linux/riscv64
- linux/ppc64le
- linux/s390x
- linux/386
- linux/mips64le
- linux/mips64
- linux/arm/v7
- linux/arm/v6

Closes a task from #1913

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,44 @@
+# Copyright 2021 Siemens AG, Gaurav Mishra <mishra.gaurav@siemens.com>
+# SPDX-License-Identifier: GPL-2.0 AND LGPL-2.1
+name: Docker images - master
+
+on:
+  push:
+    branches:
+      - master
+  release:
+    types:
+      - published
+
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (approved):
Code looks good. works as expected.

PR #2077, opened 2021-08-09
fix(unifiedreport): Fix upload link for API

GMishx:
## Description

By default, unified report saves traceback URL in `jq_cmd_args`. For UI jobs, it will be `http://localhost/repo/` (or as server's location).
If the report is generated from REST API, it gets changed to `http://localhost/repo/api/v1/report`

This URL is then used to populate the "Fossology Upload" field in unified report.

Add filters to remove everything after `api/` from the traceback URL while generating jobs and report.

### Changes

1. Add filter to remove `api/.*` from URL while generating job.
2. Add the same filter while generating report.

## How to test

1. Before installing branch, create a unified report from UI and check the "Fossology Upload" field.
2. Create a unified report from REST API and check the "Fossology Upload" field.
3. Install the branch and generate report from REST API and check the "Fossology Upload" field.

### DIFF
--- a/src/unifiedreport/agent/unifiedreport.php
+++ b/src/unifiedreport/agent/unifiedreport.php
@@ -681,7 +681,9 @@ class UnifiedReport extends Agent
     $timestamp = $jobInfo['ts'];
     $packageUri = "";
     if (!empty($jobInfo['jq_cmd_args'])) {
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (approved):
Code looks good, tested works as expected.

PR #2080, opened 2021-08-11
fix(rest): fixed the pagination in apis

### LINKED ISSUE #2079: Rest: getting the response on page set to "0"
### Description

On api call we are getting response even on when the page is set to 0.

#### How to reproduce

Make resquest on `/uploads` with header as page 0.

#### Screenshots
Screenshot from 2021-08-11 07-35-02

Shruti3004:
## Description

Added the check for if page is 0 in the headers. Closes #2079

### Changes

Added the condition in `UploadController.php` and `LicenseController.php`

## How to test

Fetch data from `/uploads` and `/licenses`.

## Screenshots
Screenshot from 2021-08-11 07-33-41


GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

PR #2089, opened 2021-08-25
fix(php) Fix SQL query checking is a similar job is already scheduled

NicolasToussaint:
## Description

On my new Fossology instance, probably because networking or filesystem is slower than it used to be, the frontend keeps on rescheduling new jobs, for example when requesting report generation.

One report request would trigger from a few to 8 or 10 jobs - and until all are processed by the scheduler, the frontend would be totally non-responsive. Then all of a sudden, the web page would "unlock", and all jobs appear green in the job list.

It turns out that the queries used to detect that a similar job had a problem, as they include `job_pk=jq_job_fk`, meaning the query would never return any result.

### Changes
Changed both SQL queries to include `job_pk=$job_pk` instead

## How to test

Browse in Fossology, and generate reports. Reports are generated + job page appear immediately, when it would have taken up to 15 minutes before.

### DIFF
--- a/src/lib/php/common-job.php
+++ b/src/lib/php/common-job.php
@@ -391,11 +391,11 @@ function IsAlreadyScheduled($job_pk, $AgentName, $upload_pk)
    * it is unneccessary to reschedule ununpack and adj2nest, one time is enough
    */
   if ($AgentName == "ununpack" || $AgentName == "adj2nest") {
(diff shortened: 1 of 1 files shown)


NicolasToussaint:
Beware though ... although it has been working perfectly for a while, I am now experiencing crashed when displaying the job list, I might have either got something wrong here.

PR #2091, opened 2021-08-26
feat(maintagent): Implement deleteOrphanGold fn

GMishx:
## Description

Implement `deleteOrphanGold()` by recursively finding files under gold folder and comparing them against `pfile` table.

**Update**: Implement `removeOrphanedFiles()` reusing `deleteOrphanGold()` logic.

*Note:* Sending upper case SHA1 and MD5 to postgres uses `bpchar` datatype compared to using PG's `UPPER()` with `text` datatype, resulting in excution time of 0.027 ms vs 25125.819 ms respectively on PG 13.3

### Changes

1. Implement function `deleteOrphanGold()`
2. Implement function `removeOrphanedFiles()`
3. New functions `strtoupper()`, `recurseDir()`, `checkPFileExists()`, `deleteRepoFile()` in utils.c

## How to test

1. Copy some additional files in gold repo
2. Run the maintenance agent with the "Remove orphaned gold files." flag
3. Check if the additional files got removed.
4. Remove from files from `uploadtree` and run the "Remove orphaned files from the repository" option from maintenance agent (`-Z` flag).
  - If the file is not reused somewhere else, it should be deleted from file system.
  - If the file is used in other upload, it should not be deleted from file system.

### DIFF
--- a/src/maintagent/agent/maintagent.h
+++ b/src/maintagent/agent/maintagent.h
@@ -50,9 +50,9 @@ extern fo_dbManager* dbManager;
 
 /* File utils.c */
 void exitNow(int exitVal);
-void recurseDir(char* path, int level);
-void checkGoldExists(char* sha1, char* md5, long fsize);
-void deleteGoldFile(char* sha1, char* md5, long fsize);
+void recurseDir(const char* type, char* path, int level);
+void checkPFileExists(char* sha1, char* md5, long fsize, const char* type);
+void deleteRepoFile(char* sha1, char* md5, long fsize, const char* type);
 
 /* File usage.c */
 void usage(char *name);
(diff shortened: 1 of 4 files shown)


shaheemazmalmmd:
**Segmentation Fault**

To Reproduce
* Copy some file to `/srv/fossology/repository/localhost/gold/`
* Go to maintenance page and execute "Remove orphaned gold files."

screenShot25

GMishx:
> **Segmentation Fault**
> 

@shaheemazmalmmd The code is designed for the file in gold be in `sha1.md5.size` format. Was it the same?

Or do we need to take care of other files as well? This will be much difficult task as currently it's just like list files, test against DB and remove.

shaheemazmalmmd:
> > **Segmentation Fault**
> 
> @shaheemazmalmmd The code is designed for the file in gold be in `sha1.md5.size` format. Was it the same?
> 
> Or do we need to take care of other files as well? This will be much difficult task as currently it's just like list files, test against DB and remove.

I have simply copied a file to test, I have not maintained any format.  I think no need to take care of other files as well. Will correct the format and try again.

PR #2102, opened 2021-09-27
fix(upload): Fix upload description input

GMishx:
## Description

Fix the template issue for upload page.

### Changes

Fix syntax error in upload page form.

## How to test

1. Upload a file with description, check in browse page, the description will be missing.
2. Install the fix and repeat steps above, the description should be there.


ag4ums (approved):
changes look good

PR #2118, opened 2021-11-16
fix(login): Allow non-admin user to update

GMishx:
## Description

Currently, the username field for non-admin users are disabled. It cause any update request to fail.

### Changes

Added a check if field is empty and user is non-admin, update it with with user id from session.

## How to test

1. Login as admin user, goto Admin > Users > Edit User Account and update some field.
2. Login as non-admin user, goto same page and edit some fields (like password).


shaheemazmalmmd (approved):

PR #2132, opened 2021-12-23
feat(nomos): See file regex to include view

GMishx:
## Description

While looking out for files with `See-file.LICENSE` regex, look for "view" as an alternative for "see".

For example, "please view the LICENSE file" and "please see the LICENSE file" to be matched.

### Changes

Change the regex for SEE_FILE with `(see|view)` instead of `see`.

## How to test

Scan a file which contains wordings like "please view the license file for usage".

Tested with https://github.com/twigphp/Twig/tree/v1.38.4


shaheemazmalmmd (approved):

PR #2107, opened 2021-10-06
chore(composer): Update composer dependencies

GMishx:
## Description

Since least version of PHP on supported OS is 7.2.24, updating all composer dependencies to highest possible version.
This brings various new features and some vulnerability fixes as well.

A list of changes:
| Dependency | Old version | New version | Remarks |
| --- | --- | --- | --- |
| container-interop/container-interop | 1.2.0 | - | Abandoned, replaced with `psr/container`. |
| doctrine/instantiator               | 1.0.5   | 1.4.0   ||
| easyrdf/easyrdf                     | 0.9.0   | 1.1.1   ||
| firebase/php-jwt                    | v5.2.0  | v5.5.1  ||
| guzzlehttp/guzzle                   | 6.5.5   | 7.4.1   ||
| mockery/mockery                     | 1.3.3   | 1.3.5   ||
| monolog/monolog                     | 1.25.5  | 2.3.5   ||
| myclabs/deep-copy                   | 1.7.0   | 1.10.2  ||
| php-coveralls/php-coveralls         | v2.4.2  | v2.5.2  ||
| phpoffice/phpspreadsheet            | 1.8.2   | 1.19.0  ||
| phpoffice/phpword                   | 0.17.0  | 0.18.2  ||
| phpstan/phpstan                     | 0.9.3   | 1.2.0   ||
| phpunit/phpunit                     | 6.5.14  | 8.5.20  ||
| sebastian/phpcpd                    | 3.0.1   | 4.1.0   ||
| slim/slim                           | 3.12.3  | 4.8.1   ||
| squizlabs/php_codesniffer           | 3.5.8   | 3.6.1   ||
| symfony/config                      | v3.4.46 | v5.3.4  ||
| symfony/dependency-injection        | v3.4.46 | v5.3.7  ||
| symfony/http-foundation             | v3.4.46 | v5.3.7  ||
| symfony/yaml                        | v3.4.46 | v5.3.6  ||
| twig/extensions                     | v1.5.4  | v1.5.4  | Abandoned, replaced with `symfony/translate` for `trans` filter |
| twig/twig                           | v2.12.5 | v3.3.4 ||
| symfony/translation | - | v5.3.9 | Required for trans filter |
| symfony/mime | - | v5.3.8 | Required to guess mimetype of files to download |
| slim/psr7 | - | 1.4 | DI container for slim, rebuilds symfony/di |

Other changes:
* RDF patch is no longer required as dependency is updated.
* Fix PHPUnit test case signatures.
[...]

PR #2134, opened 2021-12-29
fix(vuln) upgrade phpoffice/phpspreadsheet" from 1.8.2 to 1.16.0

NicolasToussaint:
Proposing 

GitHub Dependabot shows a vulnerability warning about the [phpoffice/phpspreadsheet/](https://github.com/PHPOffice/PhpSpreadsheet) used in `src/composer.lock`

https://github.com/Orange-OpenSource/fossology/security/dependabot/src/composer.lock/phpoffice%2Fphpspreadsheet/open
image

I don't think that Fossology is impacted by the vulnerability, but it would still make sense to upgrade the dependency.
We could also upgrade directly to newest version 1.20.

I tested licences export and import feature, but I'm not sure where else the library is used

### DIFF
--- a/src/composer.lock
+++ b/src/composer.lock
@@ -828,11 +828,11 @@
         },
         {
             "name": "phpoffice/phpspreadsheet",
-            "version": "1.8.2",
+            "version": "1.16.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/PHPOffice/PhpSpreadsheet.git",
-                "reference": "0c1346a1956347590b7db09533966307d20cb7cc"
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
Hello @NicolasToussaint i have observed that this changes are already solved here https://github.com/fossology/fossology/pull/2107/files#diff-708d3319cde742c23125d07bdbacc4742dea301cb5cecc89164864b73daf3443R31 
Can we close this PR if you are ok with that ?

NicolasToussaint:
Oh, perfect then @shaheemazmalmmd, thanks !

PR #2137, opened 2021-12-31
feat(gdpr) New global option to make all user details read-only

NicolasToussaint:
## Description

A new option in the Customize page make all user details read-only (Username, email, Description)
Useful when details are retrieved from external auth server and should never be altered by hand.

### Changes

New option in Customize page 
image

## How to test

Create User, details are editable
image

Go to Admin -> Customize, down the page, click on `Make account details read-only`
image

Check value in database:
```
fossology=# select variablename,conf_value from sysconfig where variablename='UserDescReadOnly';
   variablename   | conf_value 
------------------+------------
 UserDescReadOnly | true
(1 row)
```

Accounts become read-only
image

Operation is reversible.

### DIFF
--- a/src/www/ui/template/user_edit.html.twig
+++ b/src/www/ui/template/user_edit.html.twig
@@ -21,15 +21,15 @@
   <table style="border:1px solid black; border-collapse: collapse;" width="100%">
     <tr class="classic">
       <th width="25%">{{ "Username."|trans }}</th>
-      <td><input type="text" value="{{ userName|e }}" name="user_name" size="20" {% if not isSessionAdmin %}disabled="disabled"{% endif %}></td>
+      <td><input type="text" value="{{ userName|e }}" name="user_name" size="20" {%- if not isSessionAdmin or userDescReadOnly == "true" -%}readonly="true"{%- endif -%}></td>
     </tr>
     <tr class="classic">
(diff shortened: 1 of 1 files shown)


NicolasToussaint:
Arg, my mistake : we actually identified this bug and realized that the fields should be tagged as _readonly_ (with `readonly="true"`) rather than disabled

image

That make the fields RO while preserving data when posting

NicolasToussaint:
Proposed this fix: https://github.com/fossology/fossology/pull/2137/commits/0c901c4b9a90a1880165eca31c23fdbfee87f4e8

Meaning that maybe the other fix https://github.com/fossology/fossology/pull/2118/commits/36a1573aee1e0633213986b152ef3d6b31fa7a21 can be reverted ?

If OK, I can propose yet anotyher PR because I think I messed the Git Log :-/

GMishx:
> Proposed this fix: 0c901c4b9a90a1880165eca31c23fdbfee87f4e8

Looks good. I missed the `readonly` attribute :-)

> Meaning that maybe the other fix 36a1573aee1e0633213986b152ef3d6b31fa7a21 can be reverted ?

Agree. Would you like to that in this PR or a separate one?

NicolasToussaint:
OK good ;
I will prepare a separate fresh PR

PR #2139, opened 2022-01-07
feat(rest): Provide server version on REST api

GMishx:
## Description

﻿Export version information about the server on REST API. Information includes:
- Version
- Branch Name
- Commit Hash
- Commit Date
- Build Date

The date stored in VERSION file is also changed to include timezone in `+hh:mm` format to avoid confusion with abbreveated timezones. PHP for some reason confuses `IST` and reports `Europe/Istanbul` (TRT), not `Asia/Kolkata`.

### Changes

1. Added new object `fossology` to `/info` endpoint with all the information.
2. Update `WriteVERSIONFile` function to use `%:z` timezone format.

## How to test

1. Install the branch and check if the UI shows new timezone format.
2. Call `/info` endpoint and check if the response contains all the required fields.


shaheemazmalmmd (approved):

PR #2141, opened 2022-01-17
fix(ui): Fix bulk modal with selectable bg

GMishx:
## Description

Fix bulk modal to go full screen with a selectable background.

### Changes

1. Revert changes to modal.
2. Add new event listener `shown.bs.modal` to update CSS properties of modal.
3. Update `draggable` function to not update width for `#bulkModal`.

## How to test

1. Open a file for clearing.
2. Open the bulk modal.
    1. The background text from file should still be selectable.
    2. Adding new license should work as expected with select 2 search bar.
    3. Other child modals like "License text" should work as expected.
3. On folder view, open bulk modal from `[Bulk]` link.
    - Everything should work as mentioned in point 2.


ag4ums (approved):
Code looks good, working as expected.

PR #2151, opened 2022-02-04
fix(report): Set content type header

### LINKED ISSUE #1767: Downloading unifiedreport from API results in unusable file, downloading from browser results in OK file
### Description

I am triggering a report via API. When downloading that report via API when it is done, the file I receive is not the same as the file I get if I download the report from the Web UI. The file from the API is not a valid DOCX file, and appears empty when opened despite being several kb. I have attached the obtained files for the exact same report here in case some insight can be gained from their structure. I am simply writing the bytes in the `data` field of the API response to a file.

[API.docx](https://github.com/fossology/fossology/files/4932110/API.docx)
[UI.docx](https://github.com/fossology/fossology/files/4932111/UI.docx)

#### How to reproduce

Trigger a report generation, download the report via Web UI, download the report via API.

### Versions
* Fossology: Version: [3.8.1], Branch: [heads/3.8.1], Commit: [#9103fa]
* Operating System: Ubuntu 16.04.3 LTS

GMishx:
## Description

Set the `Content-Type` headers while downloading reports.

Currently the header is missing causing issues with REST API clients.

### Changes

Check if report is docx or spdx and set `Content-Type` header to `application/vnd.openxmlformats-officedocument.wordprocessingml.document` or `application/xml` respectively. For other reports, set header to `text/plain`.

## How to test

Generate reports from UI and REST API and check for `Content-Type` header.

Closes #1767

### DIFF
--- a/src/www/ui/api/documentation/openapi.yaml
+++ b/src/www/ui/api/documentation/openapi.yaml
@@ -1,4 +1,4 @@
-# Copyright (C) Siemens AG 2017-2021
+# Copyright (C) Siemens AG 2017-2022
 # Copyright (C) 2021 Orange by Piotr Pszczola <piotr.pszczola@orange.com>
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (approved):

PR #2065, opened 2021-07-23
perf(sysconfig): Setup sysconfig at fossinit

GMishx:
## Description

1. Move update of sysconfig table from common-sysconfig to fossinit
2. Make sure sysconfig table is created by libschema

Changes are done to keep the table creation logic consistent and reduce unnecessary calls to Populate_sysconfig at every UI load.

### Changes

1. Split `ConfigInit()` into `get_pg_conn()` and `populate_from_sysconfig()`.
2. Remove `Populate_sysconfig()` from `ConfigInit()` and move it to `fossinit.php`
3. Remove the sysconfig table creation function.
4. Adjust `fossinit.php` to generate sysconfig table from core-schema.dat not `common-sysconfig`
5. Fix test cases for `common-sysconfig`.

## How to test

1. Check if sysconfig table is created at installation of fresh instance.
2. Check if changing configuration in `common-sysconfig` gets reflected in DB after running `fo-postinstall`.
3. Check if script still works on existing DBs.

PR #2154, opened 2022-02-11
fix(test): fix test failures caused by sysconfig

GMishx:
## Description

Fix test cases broken because of #2065

## How to test

Check GitHub Actions.


shaheemazmalmmd (approved):
Code looks good.

PR #2159, opened 2022-02-16
chore(composer): update composer form 1.9.0 to 2.2.6 version

shaheemazmalmmd:
### Changes

Update version and hash to version2.26

## How to test

* Install fossology with utils/fo-installdeps
* Check composer version using "**composer about**"
* Check if everything works on UI like report, clearing etc..

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2159"></a>

### DIFF
--- a/utils/install_composer.sh
+++ b/utils/install_composer.sh
@@ -7,9 +7,9 @@
 ## Configuration:
 # commit hash of https://github.com/composer/getcomposer.org
 
-current_github_hash="ba13e3fc70f1c66250d1ea7ea4911d593aa1dba5"
+current_github_hash="2417ac77de78cec5fd3e5eb55879c54b8c533812"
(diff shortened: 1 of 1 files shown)


GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

GMishx:
@shaheemazmalmmd , can we instead add `composer` as a dependency in `apt`/`yum` list? It will solve the issue #692 and make packaging easier.

shaheemazmalmmd:
> @shaheemazmalmmd , can we instead add `composer` as a dependency in `apt`/`yum` list? It will solve the issue #692 and make packaging easier.

Good amount of changes and testing needs to be done. can we take this PR and add the changes to some other PR.

Issue #1371, opened 2019-06-07
Admin can not delete other uploads


GMishx:
### Description

If an upload is done with the visibility only for active/all accessible group and the admin is not a member of the group, the upload can not be deleted.

#### How to reproduce

1. Create a group "Group-A" and add a user to it "User-A" but do not add one of an admin user "Admin-1".
1. With this new user "User-A", create an upload (which will not be accessible by the admin).
1. Login as the admin "Admin-1" and try to delete the upload.

User-A is member of group User-A (created by default) and Group-A.
Admin-1 is member of group Admin-1 and **not** Group-A or User-A.

#### Expected result

As the admin can modify any upload regardless of the group access, it should be able to delete any upload as well.

### Versions

* Last commit id on master: 5aa2ad521d951b1f13c53c292f8f24010e7ac518

#### Job logs

```
2019-06-07 14:49:14 delagent [0] :: JOB[249].delagent[8515.localhost]: "FATAL util.c.1081: Due to permission problems, the delagent was not able to list or delete the requested objects or they did not exist."
2019-06-07 14:49:14 delagent [0] :: JOB[249].delagent[8515.localhost]: "ERROR util.c.1097: Exiting with status 1"
2019-06-07 14:49:14 delagent [0] :: JOB[249].delagent[8515.localhost]: agent failed with error code 1
2019-06-07 14:49:14 delagent [0] :: JOB[249].delagent[8515.localhost]: agent failed, code: 1
```

Amit-Morade:
Can I work on this issue?

GMishx:
Yes sure, I will assign the issue to you :-)

Rahulrajak1710:
@GMishx I would like to work on this issue:)

rohitpandey49:
Raised PR #2168

PR #2168, opened 2022-02-24
fix(deleteupload): Admin can delete any upload

rohitpandey49:
Signed-off-by: rohitpandey49 <rohit.pandey4900@gmail.com>

## Description

Admin can delete any upload regardless of the group access.

## How to test

1. Create a group "Group-A" and add a user to it "User-A" but do not add one of an admin user "Admin-1".
2.  With this new user "User-A", create an upload (which will not be accessible by the admin).
3.  Login as the admin "Admin-1" and try to delete the upload.

Closing #1371


GMishx (changes requested):
Changes looks good. Need formatting changes.

GMishx (approved):
Tested, working as expected.

Issue #2114, opened 2021-11-01
Delete a license via REST API


dineshr93:
### Description

Ability to Delete a license via a REST API

#### Proposal and expected result

input: **license_short_name** at the end of url
```
url = "http://fossology.com:port/repo/api/v1/license/mit"
payload = ""
headers = {
    "accept": "application/json",
    "groupName": "fossy",
    "Authorization": bearer_token
}

response = requests.request("DELETE", url, data=payload, headers=headers)
print(response.text)
```
Output: delete the license

SanjaySinghRajpoot:
I would like to work on this issue. Could you please which files are to be modified?
 Thank You

Aditi-Roy-coder:
I want to work on this issue.

thegeekymuggle:
Hi, I would love to work on the issue.

Rahulrajak1710:
Hi,
I am rahul,a cse undergraduate.I am new to open source contributions but I am well aware of Java,C,Html,Css,JavaScript,React.js.I would love contribute in fossology 
 organization  could you please tell me how to get started?
Hoping to hear from you soon.
Regards
Rahul

shaheemazmalmmd:
@dineshr93 i think it is not safe to delete a license Because, 

- for actual licenses

lets say we have license "X". and some users have already added license "X" in to clearing results
if we delete license "X" then we will end up with errors in that file where license "X" got added. 

So it is not safe delete a license.

- for candidate licenses

we already have a delete functionality which can be reused in REST API.

This functionality will not allow you to directly delete the licenses but it shows you the list of files where this license "X" has been used. so one needs to add a alternate license and remove license "X" then only you would be able to delete it.

dineshr93:
@shaheemazmalmmd yes I agree, there was a use case where EPL-1.0 can be upgraded to EPL-2.0 (due to inbuilt clause) and it's very hard to do it manually.  So I felt the need of this functionality.

PR #2172, opened 2022-03-02
feat(API): Add an API endpoint to delete a license

its-sushant:
## Description

Add a new API endpoint to delete license

see https://github.com/fossology/fossology/issues/2114 for details.

### Changes

delete license/shortname endpoint added.

## How to test

Send a delete request to /license/shortname and check the response .

Closes https://github.com/fossology/fossology/issues/2114

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2172"></a>

### DIFF
--- a/src/lib/php/Dao/LicenseDao.php
+++ b/src/lib/php/Dao/LicenseDao.php
@@ -19,6 +19,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 namespace Fossology\Lib\Dao;
 
+use Fossology\Lib\Auth\Auth;
 use Fossology\Lib\BusinessRules\LicenseMap;
 use Fossology\Lib\Data\AgentRef;
 use Fossology\Lib\Data\License;
@@ -712,5 +713,35 @@ ORDER BY lft asc
       $this->dbManager->freeResult($result);
       return $ObligationRef;
     }
+  }
+
+  /**
+   * @param string $licenseShortname
+   * @param int|null $groupId
+   * @param int|null $userId
+   * @return null
+   */
+  public function deleteLicenseByShortName($licenseShortname, $groupId)
+  {
+    $extraCondition = "";
+    $condition = 'rf_shortname=$1';
+    $param = array($licenseShortname);
+    if (Auth::isAdmin()) {
+      $row = $this->dbManager->getSingleRow(
+        "DELETE FROM ONLY license_ref WHERE $condition RETURNING rf_shortname",
+        $param, __METHOD__ . ".$condition.only");
+    } else {
+      $row = false;
+    }
+    if (false === $row) {
+      $param[] = $groupId;
+      $row = $this->dbManager->getSingleRow(
+        "DELETE FROM license_candidate WHERE $condition AND group_fk=$".count($param)." RETURNING rf_shortname",
+        $param, __METHOD__ . ".$condition.group");
+    }
+    if (false === $row) {
+      return false;
+    }
+    return true;
   }
 }
--- a/src/www/ui/api/Controllers/LicenseController.php
+++ b/src/www/ui/api/Controllers/LicenseController.php
@@ -135,6 +135,43 @@ class LicenseController extends RestController
   }
 
   /**
+   * Delete the license information based on the provided parameters
+   *
+   * @param Request $request
+   * @param ResponseHelper $response
+   * @param array $args
+   * @return ResponseHelper
+   */
+  public function deleteLicense($request, $response, $args)
+  {
+    $shortName = trim($args["shortname"]);
+
+    if (empty($shortName)) {
+      $error = new Info(
+        400,
+        "Short name missing from request.",
+        InfoType::ERROR
+      );
+      return $response->withJson($error->getArray(), $error->getCode());
+    }
+
+    $result = $this->licenseDao->deleteLicenseByShortName($shortName,
+      $this->restHelper->getGroupId());
+
+    if ($result === false) {
+      $error = new Info(
+        404,
+        "No license found with short name '{$shortName}'.",
+        InfoType::ERROR
+      );
+      return $response->withJson($error->getArray(), $error->getCode());
+    }
+    $returnVal = new Info(200, "License deleted successfully",
+      InfoType::INFO);
+    return $response->withJson($returnVal->getArray(), $returnVal->getCode());
+  }
+
+  /**
    * Get list of all licenses, paginated upon request params
    *
    * @param Request $request
--- a/src/www/ui/api/documentation/openapi.yaml
+++ b/src/www/ui/api/documentation/openapi.yaml
@@ -15,7 +15,7 @@ openapi: 3.0.2
 info:
   title: FOSSology API
   description: Automate your fossology instance using REST API
-  version: 1.4.3
+  version: 1.4.4
   contact:
     email: fossology@fossology.org
   license:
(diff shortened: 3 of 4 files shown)


src/lib/php/Dao/LicenseDao.php
```diff
@@ -713,4 +713,34 @@ public function getLicenseObligations($licenseLists, $candidate = false)
...
+   * @return null
+   */
+  public function deleteLicenseByShortName($licenseShortname, $groupId=null)
+  {
+    $extraCondition = "";
+    $condition = 'rf_shortname=$1';
+    $param = array($licenseShortname);
+    $this->dbManager->deleteSingleRow(
+      "DELETE FROM license_ref WHERE $condition",
+      $param, __METHOD__ . ".$condition.only");
```
GMishx, line 729: This deletion should be conditional? Since `license_candidate` inherits `license_ref`, doing a delete here will make following condition redundant.

src/lib/php/Dao/LicenseDao.php
```diff
@@ -713,4 +713,34 @@ public function getLicenseObligations($licenseLists, $candidate = false)
...
+    if (isset($groupId)) {
+      $userId = (isset($_SESSION) && array_key_exists('UserId', $_SESSION)) ? $_SESSION['UserId'] : 0;
+      if (!empty($userId)) {
+        $param[] = $userId;
+        $extraCondition = "AND group_fk IN (SELECT group_fk FROM group_user_member WHERE user_fk=$".count($param).")";
+      }
+      if (is_int($groupId) && empty($userId)) {
+        $param[] = $groupId;
+        $extraCondition = "AND group_fk=$".count($param);
+      }
```
GMishx, line 739: Sorry but I am not sure what's happening here.
1. You are reading user id from session? It might fail for REST API calls, consider reading it as a parameter.
2. If user id exists, delete the license for all groups it might exist for me. Otherwise delete it just for one group.
  - Can you please explain the rational behind it?

src/lib/php/Dao/LicenseDao.php
```diff
@@ -713,4 +713,34 @@ public function getLicenseObligations($licenseLists, $candidate = false)
...
+      $this->dbManager->deleteSingleRow(
+        "DELETE FROM license_candidate WHERE $condition $extraCondition",
+        $param, __METHOD__ . ".$condition.group");
+    }
+    return null;
```
GMishx, line 744: Instead of returning null, returning `true` if license was deleted and `false` otherwise will make more sense. Please have a look at `RETURNING` statement for Postgres.

src/lib/php/Db/DbManager.php
```diff
@@ -402,4 +402,24 @@ public function existsColumn($tableName, $columnName)
...
     return $this->existsTable($tableName) && $this->dbDriver->existsColumn($tableName, $columnName);
   }
+
+  /**
+   * @param string $sqlStatement
+   * @param array $params
+   * @param string $statementName (optional)
+   * @return null
+   */
+  public function deleteSingleRow($sqlStatement, $params = array(), $statementName = "")
```
GMishx, line 412: Function is similar to `getSingleRow()` just without a `return` statement https://github.com/fossology/fossology/blob/fcf1de2f4beb714bd67131cbeeba2604f3ee1fb4/src/lib/php/Db/DbManager.php#L153

Not sure calling it `deleteSingleRow()` and working similar to `getSingleRow()` gives any benefit.

src/www/ui/api/Controllers/LicenseController.php
```diff
@@ -134,6 +134,45 @@ public function getLicense($request, $response, $args)
...
+   * Delete the license information based on the provided parameters
+   *
+   * @param Request $request
+   * @param Response $response
+   * @param array $args
+   * @return Response
+   */
+  public function deleteLicense($request, $response, $args)
+  {
+    $shortName = $args["shortname"];
```
GMishx, line 147: Please consider trimming the input to make it more robust.

src/www/ui/api/Controllers/LicenseController.php
```diff
@@ -134,6 +134,45 @@ public function getLicense($request, $response, $args)
     return $response->withJson($returnVal->getArray(), 200);
   }
 
+  /**
+   * Delete the license information based on the provided parameters
+   *
+   * @param Request $request
+   * @param Response $response
+   * @param array $args
+   * @return Response
```
GMishx, line 143:
```suggestion
   * @param ResponseHelper $response
   * @param array $args
   * @return ResponseHelper
```

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -1399,6 +1399,28 @@ paths:
                 $ref: '#/components/schemas/Info'
         default:
             $ref: '#/components/responses/defaultResponse'
+    delete:
```
GMishx, line 1402: Please consider updating the patch version of the API (line 18 of the file).

GMishx (changes requested):

src/lib/php/Dao/LicenseDao.php
its-sushant: @GMishx , thanks for your feedback. I am trying to make changes as per your suggestions.

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -1398,6 +1398,28 @@ paths:
...
+      summary: delete a specific license
+      description: >
+        delete information about a specific license
+      responses:
+        '200':
+          description: License deleted successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/License"
```
GMishx, line 1414: I believe when a license is successfully deleted, you are just returning an object of type `'#/components/schemas/Info'` and not the complete license info (as in GET).

its-sushant: Thanks @GMishx  for the review. I will just change it.

src/lib/php/Dao/LicenseDao.php
GMishx: Hey @its-sushant , I am still not clear about these operations. If you have the group id for the user, why not just delete the license for that particular group instead of deleting it for all groups?

In the OpenAPI doc, mention it explicitly that the license will only be deleted for active group.

src/lib/php/Dao/LicenseDao.php
```diff
@@ -713,4 +713,38 @@ public function getLicenseObligations($licenseLists, $candidate = false)
...
+      }
+      $row = $this->dbManager->getSingleRow(
+        "DELETE FROM license_candidate WHERE $condition $extraCondition RETURNING rf_shortname",
+        $param, __METHOD__ . ".$condition.group");
+    }
+    if (false === $row) {
+      return false;
+    }
+    return true;
+  }
```
GMishx, line 749: There is no return statement if logic did not enter `if (false === $row && isset($groupId))` condition. It will make function return `bool|null`.

src/www/ui/api/Controllers/LicenseController.php
GMishx: Please consider trimming.

src/www/ui/api/Controllers/LicenseController.php
```diff
@@ -134,6 +134,43 @@ public function getLicense($request, $response, $args)
...
+      $error = new Info(
+        400,
+        "Short name missing from request.",
+        InfoType::ERROR
+      );
+      return $response->withJson($error->getArray(), $error->getCode());
+    }
+
+    $result = $this->licenseDao->deleteLicenseByShortName($shortName,
+      $this->restHelper->getGroupId(), $this->restHelper->getUserId);
```
GMishx, line 159:
```suggestion
      $this->restHelper->getGroupId(), $this->restHelper->getUserId());
```

src/www/ui/api/Controllers/LicenseController.php
```diff
@@ -134,6 +134,43 @@ public function getLicense($request, $response, $args)
...
+
+    if ($result === false) {
```
GMishx, line 161: As suggested about, `$result` might also be `null` and it'll be considered as deleted (even though it is not).

GMishx (changes requested):

GMishx (approved):
Changes looks good.

GMishx:
@shaheemazmalmmd can we proceed with this PR?

shaheemazmalmmd:
Lets work on backend and UI for current functionality. then we can take in the REST API.

PR #2174, opened 2022-03-08
fix(api): Return full uploadId

TassiloPitrasch:
Removed unnecessary subindex to return full uploadId instead of first digit only

## Description

When uploading a package from URL via the API, the uploadId in the response is not complete, but only the first digit. This is caused by an unnecessary "subindex", which was removed. Maybe this issue was created with a change in the database structure.

### Changes

Removed subindex for array $uploadResponse.

## How to test

This can easily be tested by checking the response for an API call as described in https://www.fossology.org/get-started/basic-rest-api-calls/ under "Upload a package".

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2174"></a>

### DIFF
--- a/src/www/ui/api/Controllers/UploadController.php
+++ b/src/www/ui/api/Controllers/UploadController.php
@@ -391,7 +391,7 @@ class UploadController extends RestController
         $info = new Info(500, $message . "\n" . $statusDescription,
           InfoType::ERROR);
       } else {
-        $uploadId = $uploadResponse[3][0];
+        $uploadId = $uploadResponse[3];
         $info = new Info(201, intval($uploadId), InfoType::INFO);
       }
       return $response->withJson($info->getArray(), $info->getCode());
--- a/src/www/ui/api/Helper/UploadHelper/HelperToUploadFilePage.php
+++ b/src/www/ui/api/Helper/UploadHelper/HelperToUploadFilePage.php
@@ -44,6 +44,10 @@ class HelperToUploadFilePage extends UploadFilePage
    */
   public function handleRequest(Request $request)
   {
-    return $this->handleUpload($request);
+    $response = $this->handleUpload($request);
+    if ($response[0]) {
+      $response[3] = $response[3][0];
+    }
+    return $response;
   }
 }
--- a/src/www/ui_tests/api/Controllers/UploadControllerTest.php
+++ b/src/www/ui_tests/api/Controllers/UploadControllerTest.php
@@ -580,7 +580,7 @@ class UploadControllerTest extends \PHPUnit\Framework\TestCase
     $uploadHelper->shouldReceive('createNewUpload')
       ->withArgs([null, $folderId, $uploadDescription, 'protected', 'true',
         'vcs', false])
-      ->andReturn([true, '', '', [20]]);
+      ->andReturn([true, '', '', 20]);
 
     $this->folderDao->shouldReceive('getAllFolderIds')->andReturn([2,3,4]);
     $this->folderDao->shouldReceive('isFolderAccessible')


patrickwjh (approved):
I can approve the fix. By the second access "[0]" to the uploadResponse only the first character of the string is accessed, which causes the error that the complete UploadId is not returned. 

The fix was tested and the complete UploadId is returned again!

src/www/ui/api/Controllers/UploadController.php
```diff
@@ -391,7 +391,7 @@ public function postUpload($request, $response, $args)
         $info = new Info(500, $message . "\n" . $statusDescription,
           InfoType::ERROR);
       } else {
-        $uploadId = $uploadResponse[3][0];
+        $uploadId = $uploadResponse[3];
```
GMishx, line 394: This change will cause issues for file uploads as the third argument returned is an array and clients expect integer
https://github.com/fossology/fossology/blob/1f490d46d79ae9ea85ff0de279083b271147bbc5/src/www/ui/page/UploadFilePage.php#L201-L202

To fix that, following location must be updated
https://github.com/fossology/fossology/blob/1f490d46d79ae9ea85ff0de279083b271147bbc5/src/www/ui/api/Helper/UploadHelper/HelperToUploadFilePage.php#L47

to something like this
```php
$response = $this->handleUpload($request);
$response[3] = $response[3][0];
return $response;
```

GMishx (changes requested):

TassiloPitrasch:
@GMishx Thanks for the feedback! I updated the respective method.

GMishx (approved):
Changes looks good.

GMishx:
@TassiloPitrasch that's for fast response. Can you please rebase your branch and squash the commits so we do not get 2 merge commits?

GMishx (approved):
Tested, working as expected.﻿

GMishx:
> @GMishx I checked the failing test. As far as i understood, the test asserts an array with one single entry "20" for the uploadId. We changed that now to a single String. Maybe we need to adapt the test here...

I guess so too. The upload id `[20]` should now be single value due to change in upload helper
https://github.com/fossology/fossology/blob/fcf1de2f4beb714bd67131cbeeba2604f3ee1fb4/src/www/ui_tests/api/Controllers/UploadControllerTest.php#L583

TassiloPitrasch:
@GMishx But additionally, the returned ID was "1", not "20", where might this discrepancy come from?

GMishx:
I think PHP is translating `intval([20])` as `1`. There is a check for response code `201` and it returned only here
https://github.com/fossology/fossology/blob/7b3dab6b4ae7109242ff4d81fd4179ea66423f45/src/www/ui/api/Controllers/UploadController.php#L390-L397

TassiloPitrasch:
So, we will change intval als well and just use 20 as String for assertion?

GMishx:
No no, just change the return value of mock function to be simple integer and not array. The API should return integer and not string, therefore it is explicitly converted using `intval()`.

TassiloPitrasch:
@GMishx I guess, we're finished here!

GMishx:
Hey @TassiloPitrasch , is it possible to rebase the changes with latest `master` branch?
If I look at commits, I see a merge commit and this big commit https://github.com/fossology/fossology/pull/2174/commits/514e8cf13143101f1acb77eddb4e9b2eafce0063 which is not exactly limited to your changes.

If you can fix that, I can merge this PR.

TassiloPitrasch:
@GMishx I squashed the commit and reopened the Pull Request. If your fine with that, please merge :)

GMishx:
Perfect. Thank you @TassiloPitrasch

PR #2194, opened 2022-03-29
fix(mta): use swaks as an alternate to s-nail

GMishx:
## Description

Use [swaks](https://www.jetmore.org/john/code/swaks/) as an alternate to s-nail for MTA.

s-nail from past had many issues, there were some compatibility between versions causing troubles. The method used to design the authentication URL is not very good and causes issue with different setup. swaks provide different options to provide authentication information and server location which helps in separating the logic.

swaks looks like a good candidate to replace s-nail.

### Changes

1. Update the command generation string from s-nail format to swaks.
2. Change the options of `SMTPSslVerify` from `'Ignore{I}|Strict{S}|Warn{W}'` to simple `'Yes{1}|No{2}'`.

## How to test

1. Setup email configuration for your server.
3. Run a job and see if you receive email.


ChinmayShringi:
Hi @GMishx, since this PR is from your personal fork and I couldn't push changes directly to it, I've taken your changes, rebased them on the current master, and created a new PR #3319 that supersedes this one.
Thank you for the initial implementation!

Kaushl2208:
Reason to close is pretty similar to #3319 
Thank you for your contributions.

Issue #2031, opened 2021-06-24
REST API: Pagination feature in the Search request


Shruti3004:
### Description

Pagination feature for the `/search` request, on the UI we are seeing the files at different pages in the current implementation, but on making the request it is giving all the files at once and we have to apply filter on our UI but that will anyhow create the load on server.

#### Screenshots

![Screenshot from 2021-06-24 21-34-04](https://user-images.githubusercontent.com/56133783/123296061-efe8ec80-d533-11eb-9b59-d5a4c7858a4d.png)

Shruti3004:
@GMishx while making the search page i get to know actually it is giving 100 responses only but thre is no pagination thing mentioned in the swagger, and how we will get the further files.

GMishx:
> @GMishx while making the search page i get to know actually it is giving 100 responses only but there is no pagination thing mentioned in the swagger, and how we will get the further files.

Apparently, there is already a limit and pagination in the library. Just need to expose it from API.

https://github.com/fossology/fossology/blob/6fcd4005a4b5783e67377231fdca11e5d5dc7c4a/src/www/ui/search-helper.php#L42

coder-whale:
I can take up this feature request if the enhancement is still required. Please revert if that is indeed the case.

PR #2203, opened 2022-04-04
feat(API): Add pagination to search request

coder-whale:
## Description

Closes issue #2031

### Changes

- Added select input for page size in search page
- Added `$PageSize` parameter to `GetResults` function in `src/www/ui/search-helper.php` and updated all associated function calls

## How to test

Open search page on fossology and select different page sizes to test.


src/www/ui/api/Controllers/SearchController.php
```diff
@@ -89,7 +89,7 @@ public function performSearch($request, $response, $args)
     }
 
     $item = GetParm("item", PARM_INTEGER);
-    $results = GetResults($item, $filename, $uploadId, $tag, 0,
+    $results = GetResults($item, $filename, $uploadId, $tag, 0, 100,
```
GMishx, line 92: Wasn't the point of this PR is to read page size from the request?

coder-whale: I have hard coded size as 100 since page no. is hard coded to be 0 as well. I can change it so that both are fetched from the request.

coder-whale: Hard coding is only for the api, have added page size selection in the twig page for search.

src/www/ui/api/Controllers/SearchController.php
```diff
@@ -88,8 +90,31 @@ public function performSearch($request, $response, $args)
       return $response->withJson($returnVal->getArray(), $returnVal->getCode());
     }
 
+    /*
+     * check if page && pageSize are numeric, if existing
+     */
+    if ((! ($page==='') && (! is_numeric($page) || $page < 1)) ||
```
GMishx, line 96: Instead of using `===''`, it is better to use `empty()` as the variable can be `''` or null or 0 or does not exist. The `===` will explicitly check for a string `''` but `empty()` on the other hand will return `true` for all the things mentioned.

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -642,6 +642,20 @@ paths:
...
+          schema:
+            type: integer
+            minimum: 1
+        - name: pageSize
+          description: Size of the returned page
+          required: false
+          in: header
+          schema:
+            type: integer
+            minimum: 1
```
GMishx, line 658: Please add maximum 100 as well as per the code.

GMishx: Can add default values for `page` and `pageSize`.

GMishx (changes requested):
Some suggestions.

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -642,6 +642,20 @@ paths:
...
           schema:
             type: string
+        - name: page
+          description: Page number (starts from 1)
+          required: false
+          in: header
+          schema:
+            type: integer
+            minimum: 1
+        - name: pageSize
```
GMishx, line 652: Consider call it `limit` as done in other endpoints

https://github.com/fossology/fossology/blob/a3c93461b11aefc5a6b5d0ceb4a4a34ed8621cee/src/www/ui/api/documentation/openapi.yaml#L772-L777

GMishx: Also, will be nice (not required) if you can expose total number of pages

https://github.com/fossology/fossology/blob/a3c93461b11aefc5a6b5d0ceb4a4a34ed8621cee/src/www/ui/api/documentation/openapi.yaml#L794-L797

src/www/ui/api/Controllers/SearchController.php
coder-whale: I have written the code this way since `empty()` cannot differentiate between 0 and does not exist (i.e. parameter not provided by user).  If I use `!empty()`, it will not throw an error for 0. Parameter not existing is handled in a separate `empty()` check where default value is assigned.

src/www/ui/api/Controllers/SearchController.php
```diff
@@ -110,6 +135,6 @@ public function performSearch($request, $response, $args)
       $currentResult = new SearchResult($currentUpload, $uploadTreePk, $filename);
       $searchResults[] = $currentResult->getArray();
     }
-    return $response->withJson($searchResults, 200);
+    return $response->withHeader("X-Total-Pages", $count)->withJson($searchResults, 200);
```
GMishx, line 138: Total pages should be "count / limit" => thus pages. Currently I am getting total count of the result from query.

For reference: https://github.com/fossology/fossology/blob/25fa297d681dd70b744891b276a1b23a1e934120/src/www/ui/api/Controllers/LicenseController.php#L170

GMishx (approved):
PR is ready for merge.

PR #2207, opened 2022-04-07
fix(rest): fix CORS issue

GMishx:
## Description

The header 'Access-Control-Allow-Credentials' is used to expose credentials from JavaScript from other domains as well ([reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials)) which is not required for the application. By removing the header from CORS middleware, the allowed origin can be marked as '*'.

### Changes

1. Remove the `Access-Control-Allow-Credentials` header from REST middleware.
2. Set default value for `Access-Control-Allow-Origin` as `*`.

## How to test

1. Install the new branch and install the [FOSSologyUI project](https://github.com/fossology/FOSSologyUI).
2. Setup the correct endpoint in `.env` (`localhost/repo/api/v1`).
3. Test if using UI causes any CORS issues.

Closes https://github.com/fossology/FOSSologyUI/issues/202

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2207"></a>

### DIFF
--- a/src/lib/php/common-sysconfig.php
+++ b/src/lib/php/common-sysconfig.php
@@ -174,7 +174,7 @@ function Populate_sysconfig()
   $variable = "CorsOrigins";
   $prompt = _('Allowed origins for REST API');
   $desc = _('[scheme]://[hostname]:[port], "*" for anywhere');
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (approved):
Code looks good.

PR #2212, opened 2022-04-18
fix(copyright): do not update empty copyrights

GMishx:
## Description

Do not update the hash of copyrights where content is empty in `3.7-3.8` migration.
Empty/NULL content and hash are used to denote no copyrights found situations.
The migration script updates the hash which causes multiple collisions due to same hash.
(Same applies for ecc/keyword/author).

Add new migration script `dbmigrate_4.0-4.1.php` to revert content and hash to `NULL` if the hash was updated to `md5('')`.

### Changes

1. Update 3.7-3.8 script to exclude copyright where content is NULL or empty.
2. Add new script 4.0-4.1 to update content and hash of copyright where hash was `md5('') => d41d8cd98f00b204e9800998ecf8427e`.
3. Updated `ajax-copyright-hist.php` to use `empty()` instead of `!` to detect empty strings.

## How to test

1. On an existing server with copyright findings, run the installation of master branch.
    - Copyrights where content is empty will get hash `d41d8cd98f00b204e9800998ecf8427e`.
2. Install the branch
    - Copyrights where content is empty/null should have null hash as well.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2212"></a>


shaheemazmalmmd (approved):
Code looks good. Needs test

PR #2216, opened 2022-04-29
fix(jwt): explicitly declare jwk algorithm

GMishx:
## Description

Explicitly declare algorithm used by JWK key set to prevent token forgery.

Solves [CVE-2021-46743](https://nvd.nist.gov/vuln/detail/CVE-2021-46743).

### Changes

New parameter `OidcJwkAlgInject` to inject algorithm in jwk key set.

## Note
The vulnerability describes possibility of key forging by changing JWT header: https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/

The fix is to explicitly declare the algorithm used in signature. Some JWKs provider like 0auth already provide it, others do not. For providers who do not advertise algorithm for the keys, inject them by a value in sysconfig array.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2216"></a>

### DIFF
--- a/src/composer.json
+++ b/src/composer.json
@@ -15,7 +15,7 @@
         "ext-mbstring": "*",
         "ext-json" : "*",
         "ext-xml" : "*",
-        "firebase/php-jwt": "v5.5.1",
+        "firebase/php-jwt": "v6.1.2",
         "monolog/monolog" : "2.3.5",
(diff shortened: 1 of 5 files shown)


shaheemazmalmmd (approved):
Code looks good.

ag4ums (approved):
working as expected

PR #2217, opened 2022-05-10
chore(component-id): use package-url instead purl

GMishx:
## Description

Use component id "package-url" instead of purl as recommended value. It is added at index 5 and also is now the default value in database.

For SPDX reports, it is changed back to purl to be conformant with spec.

### Changes

1. Add new ComponentType::PACKAGEURL
2. Treat ComponentType::PACKAGEURL and ComponentType::PURL same in spasht and SPDX
3. Change default value of 'ri_component_type' from 0 to 5 in 'report_info' table.

## How to test

1. On existing packages with component ID set to purl, there should be no effect.
2. `purl` should now appear in "Legacy" section of Report Config page.
3. `package-url` should be "Recommended" in Report Config.
4. Changing `purl` to `package-url` in config should have no effect on SPDX reports. Spasht should not throw errors. Only CLIXML should have `package-url` in `<ComponentId><Type>`.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2217"></a>

### DIFF
--- a/src/clixml/agent/clixml.php
+++ b/src/clixml/agent/clixml.php
@@ -278,7 +278,6 @@ class CliXml extends Agent
       'organisation' => '',
       'componentHash' => strtolower($componentHash['sha1']),
       'contents' => $contents,
-      'packageIds' => $packageIds,
       'commentAdditionalNotes' => $notes,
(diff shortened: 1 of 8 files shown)


shaheemazmalmmd (approved):

PR #2218, opened 2022-05-10
fix(reuser): reuse all report columns

GMishx:
## Description

Reuse all columns for `report_info` table when reuser agent is used.

Also, overwrite all entries preventing duplicates.

### Changes

1. Fetch all columns of `report_info` table from PostgreSQL to be reused
2. Drop `report_info` if it already exists for current upload to replace the values and prevent multiple entries if someone reuse multiple times.

## How to test

1. Edit Conf section of an upload.
2. Upload new version of the package and run reuser.
  - Select "Reuse report configuration settings" while scheduling.
  - Check Report Conf section of new upload.
  - All values should be copied over.
3. Edit the new upload's Report Conf and run the reuser again.
  - Report Conf should have the same values as of reused upload.
4. Logs of Reuser agent should be clean.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2218"></a>

### DIFF
--- a/src/lib/php/Dao/UploadDao.php
+++ b/src/lib/php/Dao/UploadDao.php
@@ -444,7 +444,8 @@ class UploadDao
   /**
    * @param int $uploadId
    * @param int $groupId
-   * @return int
+   * @return array Assoc array of reused_upload_fk, reused_group_fk and
+   *               reuse_mode
    */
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):

PR #2219, opened 2022-05-12
fix(ui): restore license text for bulk modal

GMishx:
## Description

While adding/removing new license(s) to bulk modal, license text, acknowledgement and comments already entered were removed.

While refreshing the table with new row, restore the values.

### Changes

1. Update `updateTable()` in `ui-clearing-view_bulk.js` to read existing values while recreating the table.
2. Also, use template string instead of adding strings for the table to make it more presentable.

## How to test

1. Open a bulk window, add/remove a license.
2. Change the License text, Acknowledgement and Comment to it.
3. Add/remove another license.
  - The entered License text, Acknowledgement and Comment should persist.
4. Add/remove another license without making changes to fields.
  - "Click to add" should be displayed as usual.
5. Remove add/remove license row.
6. Check the console for warnings/errors.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2219"></a>

### DIFF
--- a/src/www/ui/scripts/ui-clearing-view_bulk.js
+++ b/src/www/ui/scripts/ui-clearing-view_bulk.js
@@ -22,18 +22,56 @@ var bulkFormTableContent = (function(){
     var s = "";
     var uploadTreeId = $('#uploadTreeId').val();
     for (i = 0; i < content.length; ++i) {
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):

PR #2262, opened 2022-07-15
feat(api): Added `users/{id}` PUT route to modify user details

### LINKED ISSUE #2257: Create a new PUT/PATCH route to modify user details via REST API
## Description
Currently, the only way to edit details of an existing user is via fossology's old UI. Since, the same feature needs to be migrated in the new UI, there is a requirement of a PUT/PATCH route at `/users/{id}` that will modify user details in the database.

## Steps to be followed
- Create a new PUT/PATCH route in the `index.php` file inside `www/ui/api` directory.
- Add a new method `editUser` in the `UserController`class.
- Try to re-use the methods that are already used in the static UI.

CC: @GMishx @Shruti3004 Please assign me this issue.

krishna9304:
Signed-off-by : Krishna Mahato <krishhtrishh9304@gmail.com>

## Description

Now, there is a PUT route which is available at `/api/v1/users/{id}` that will edit the user details of a given user id. 

## How to test

- Use any API platform like postman.
- Pull the changes from this PR.
- Provide the request body as following ----
Screenshot 2022-07-18 at 10 49 24 PM

- You can expect a response like this
Screenshot 2022-07-15 at 9 13 36 PM

PTAL : @GMishx @Shruti3004 

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2256"></a>

This closes #2257 .

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2262"></a>

### DIFF
--- a/src/www/ui/api/Helper/UserHelper.php
+++ b/src/www/ui/api/Helper/UserHelper.php
@@ -1,0 +1,182 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: © 2022 Krishna Mahato <krishhtrishh9304@gmail.com>
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+/**
+ * @user
+ * @brief Helper for User related queries
+ */
+
+namespace Fossology\UI\Api\Helper;
+
+use Fossology\Lib\Auth\Auth;
+use Fossology\Lib\Dao\UserDao;
+use Fossology\UI\Api\Models\Info;
+use Fossology\UI\Api\Models\InfoType;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * @class UserHelper
+ * @brief Handle user related queries
+ */
+class UserHelper
+{
+  /**
+   * @var $user_pk
+   */
+  private $user_pk;
+
+  /**
+   * Constructor for UserHelper
+   *
+   * @param $user_pk
+   */
+  public function __construct($user_pk)
+  {
+    $this->user_pk = $user_pk;
+  }
+
+  public function modifyUserDetails($reqBody)
+  {
+    global $container;
+    $restHelper = $container->get('helper.restHelper');
+    $userEditObj = $restHelper->getPlugin('user_edit');
+    /* Is the session owner an admin? */
+    $sessionOwnerUser_pk = $restHelper->getUserId();
+    $SessionUserRec = $userEditObj->GetUserRec($sessionOwnerUser_pk);
+    $SessionIsAdmin = $userEditObj->IsSessionAdmin($SessionUserRec);
+
+    $symReq = $this->createSymRequest($reqBody);
+    if (!$SessionIsAdmin) {
+      $returnVal = new Info(403, "The session owner is not an admin!", InfoType::INFO);
+    } else {
+      $userRec = $userEditObj->CreateUserRec($symReq);
+      $ErrMsgs = $userEditObj->UpdateUser($userRec, $SessionIsAdmin);
+
+      if ($ErrMsgs == null) {
+        $returnVal = new Info(200, "User updated succesfully!", InfoType::INFO);
+      } else {
+        $returnVal = new Info(400, $ErrMsgs, InfoType::INFO);
+      }
+    }
+    return $returnVal;
+  }
+
+  /**
+   * @param array $userDetails parsed from request body
+   * @return Request $symfonyRequest
+   */
+  public function createSymRequest($userDetails)
+  {
+    global $container;
+    $restHelper = $container->get('helper.restHelper');
+
+    /**
+     * @var UserDao $userDao
+     * User dao
+     */
+    $userDao = $restHelper->getUserDao();
+    $user = $userDao->getUserByPk($this->user_pk);
+
+    $symfonyRequest = new Request();
+    $symfonyRequest->request->set('user_pk', isset($userDetails['id']) ? $userDetails['id'] : $this->user_pk);
+    $symfonyRequest->request->set('user_name', isset($userDetails['name']) ? $userDetails['name'] : $user['user_name']);
+    $symfonyRequest->request->set('root_folder_fk', isset($userDetails['rootFolderId']) ? $userDetails['rootFolderId'] : $user['root_folder_fk']);
+    $symfonyRequest->request->set('default_group_fk', isset($userDetails['defaultGroup']) ? $userDetails['defaultGroup'] : $user['group_fk']);
+    $symfonyRequest->request->set('public', isset($userDetails['defaultVisibility']) ? $userDetails['defaultVisibility'] : $user['upload_visibility']);
+    $symfonyRequest->request->set('default_folder_fk', isset($userDetails['defaultFolderId']) ? $userDetails['defaultFolderId'] : $user['default_folder_fk']);
+    $symfonyRequest->request->set('user_desc', isset($userDetails['description']) ? $userDetails['description'] : $user['user_desc']);
+    $symfonyRequest->request->set('_pass1', isset($userDetails['user_pass']) ? $userDetails['user_pass'] : null);
+    $symfonyRequest->request->set('_pass2', isset($userDetails['user_pass']) ? $userDetails['user_pass'] : null);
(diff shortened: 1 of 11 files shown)


src/www/ui/api/documentation/openapi.yaml
```diff
@@ -757,6 +757,115 @@ paths:
...
+      summary: Edit user details by id
+      description: >
+        Edit user details by id
+      requestBody:
+        required: true
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              properties:
```
GMishx, line 774: Please reuse existing data models instead of creating new ones to keep consistent models.

https://github.com/fossology/fossology/blob/5eaca363ca51c212038939acc98c24672b6392e0/src/www/ui/api/documentation/openapi.yaml#L1669

```suggestion
            schema:
              $ref: '#/components/schemas/User'
```

GMishx (changes requested):
Please make sure all parameters are required to update the user.

For example, if someone just wants to update their password, they need not send their agent list and email.

src/www/ui/api/Helper/UserHelper.php
```diff
@@ -0,0 +1,121 @@
+<?php
+
+/**
+ * *************************************************************
+ * Copyright (C) 2022 Siemens AG
```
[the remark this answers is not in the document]
krishna9304: Okay!

src/www/ui/api/Helper/UserHelper.php
```diff
@@ -0,0 +1,121 @@
...
+    $this->user_pk = $user_pk;
+  }
+
+  public function modifyUserDetails($reqBody)
+  {
+    global $container;
+    $restHelper = $container->get('helper.restHelper');
+    $userEditObj = $restHelper->getPlugin('user_edit');
+    /* Is the session owner an admin? */
+    $sessionOwnerUser_pk = Auth::getUserId();
```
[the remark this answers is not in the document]
krishna9304: Okay

src/www/ui/api/documentation/openapi.yaml
krishna9304: I tried reusing that, but there are couple of new parameters that needs to be added  without which the API wont work.

krishna9304: [Update]: I learnt how we can add more parameters to an existing schema using `allOf`. Will do the required changes.

src/www/ui/api/Helper/UserHelper.php
```diff
@@ -0,0 +1,144 @@
...
+      }
+      $symfonyRequest->request->set('user_perm', $user_perm);
+    } else {
+      $symfonyRequest->request->set('user_perm', 0);
+    }
+
+    // setting the checked agents
+    $agents = array();
+    if (isset($userDetails['agents'])) {
+      $agents['Check_agent_mimetype'] = isset($userDetails['agents']['mime']) && $userDetails['agents']['mime'] ? 1 : 0;
```
GMishx, line 127:
```suggestion
      $agents['Check_agent_mimetype'] = isset($userDetails['agents']['mimetype']) && $userDetails['agents']['mimetype'] ? 1 : 0;
```

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -1687,6 +1742,7 @@ components:
             - none
             - read_only
             - read_write
+            - clearing_admin
             - admin
         rootFolderId:
```
GMishx, line 1747: Can you please add `default_group_fk` from `users` table as well to the model, preferable called as `defaultGroup`?

It denotes the default group selected by the user.

Doing so, it will be available from `GET /api/v1/users[/self|{id}]` endpoint as well and while `PUT /api/v1/users/{id}`, it can be read to update the default group id for the user.

GMishx (changes requested):

src/www/ui/api/documentation/openapi.yaml
krishna9304: You mean, adding a parameter to the `User` model which will be called `defaultGroup` so as to allow the admin to edit the default group of the user? right?

src/www/ui/api/Helper/UserHelper.php
krishna9304: sure will do this change!

krishna9304: I think this will break existing `/jobs` POST API for creating jobs, as `agents` model is also being used there. As of now, do you suggest to change this, as more code changes needs to done to adapt this literal change?

src/www/ui/api/documentation/openapi.yaml
krishna9304: I have done the change and updated the PR. Please have a look.

krishna9304:
@GMishx , I have realised, the current API in this PR will require all the parameters in order to update even a single parameter of user else all other fields will become empty. I am making changes so as to allow the user to send the only field which he/she wants to modify(edit).

For instance, If the user wants to update the `description` he will just give the userId in `params` in path, and `description` field in the request body. All other fields which are not given in the request body will be automatically set in the backend from the DB. I think this would be the ideal way to do it.

PTAL at the present changes.

GMishx (approved):
Changes look good. Needs test.

GMishx:
> @GMishx , I have realised, the current API in this PR will require all the parameters in order to update even a single parameter of user else all other fields will become empty. I am making changes so as to allow the user to send the only field which he/she wants to modify(edit).
> 
> For instance, If the user wants to update the `description` he will just give the userId in `params` in path, and `description` field in the request body. All other fields which are not given in the request body will be automatically set in the backend from the DB. I think this would be the ideal way to do it.

Makes perfect sense @krishna9304 and also reduce effort for REST API users :+1:

krishna9304:
> > @GMishx , I have realised, the current API in this PR will require all the parameters in order to update even a single parameter of user else all other fields will become empty. I am making changes so as to allow the user to send the only field which he/she wants to modify(edit).
> > For instance, If the user wants to update the `description` he will just give the userId in `params` in path, and `description` field in the request body. All other fields which are not given in the request body will be automatically set in the backend from the DB. I think this would be the ideal way to do it.
> 
> Makes perfect sense @krishna9304 and also reduce effort for REST API users 👍

great!

src/www/ui/api/Helper/UserHelper.php
```diff
@@ -0,0 +1,159 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ * *************************************************************
+ */
```
rohitpandey49, line 20:
```suggestion
/*
 SPDX-FileCopyrightText: © 2022 Krishna Mahato <krishhtrishh9304@gmail.com>
 
 SPDX-License-Identifier: GPL-2.0-only
*/
```

rohitpandey49, line 20:
```suggestion
/*
 SPDX-FileCopyrightText: © 2022 Krishna Mahato <krishhtrishh9304@gmail.com>
 
 SPDX-License-Identifier: GPL-2.0-only
*/
```

krishna9304: sure!

GMishx:
@krishna9304 please apply following patch file to the PR to fix the test cases and other bug

[rest-fixes.patch.txt](https://github.com/fossology/fossology/files/9281337/rest-fixes.patch.txt)

krishna9304:
> @krishna9304 please apply following patch file to the PR to fix the test cases and other bug
> 
> [rest-fixes.patch.txt](https://github.com/fossology/fossology/files/9281337/rest-fixes.patch.txt)

Done!

GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

PR #2269, opened 2022-07-18
feat(API): remove member from group.

### LINKED ISSUE #2267: Create a REST API to delete a user from a group.
### Description
There's a need for the API to delete a member from a particular group.

#### Steps to be followed

1. Create new route in the main page (index.php) as a part of Group routes.
2. Add a new controller function to handle the feature.
3. Test the API.
4. Add the new API's documentation  in [https://github.com/fossology/fossology/blob/master/src/www/ui/api/documentation/openapi.yaml](url)
5. Open a pull request.

#### Expected Results

1.  An endpoint to be accessed at ''/groups/{gid}/user/{uid}".
2. The API should return the object of `Info` type with status `202` as a successful deletion. 

cc: @shaheemazmalmmd @GMishx

dushimsam:
## Description

Remove a particular user from a specific group.

### Changes

1. Added a new function in `GroupController`. to implement the core functionality.
2. Updated  the main file(`index.php`) by adding a new route `/groups/{id}/user/{uid}`.
3. Updated the `openapi.yaml` file to introduce a new API.

## How to test

1. Open and run the project then visit  the url `http://localhost/repo/api/v1/groups/{groupId}/user/{userId}`
2. Response
If the groupId and the UserId exist and the user is part of the group, you should see a response as such.

Selection_014

Else you should see the message `Not a member !` if the user is not a member of the group.

eg:
Selection_015

### Related Issue:
Fixes #2267
    
cc: @shaheemazmalmmd @GMishx

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2269"></a>

### DIFF
--- a/src/www/ui/api/Controllers/GroupController.php
+++ b/src/www/ui/api/Controllers/GroupController.php
@@ -134,4 +134,47 @@ class GroupController extends RestController
 
     return $response->withJson($returnVal->getArray(), $returnVal->getCode());
   }
+
+   /**
+   * Delete a given group member
+   *
+   * @param ServerRequestInterface $request
+   * @param ResponseHelper $response
+   * @param array $args
+   * @return ResponseHelper
+   */
+  public function deleteGroupMember($request, $response, $args)
+  {
+    $returnVal = null;
+    $dbManager = $this->dbHelper->getDbManager();
+
+    $group_pk = intval($args['id']);
+    $user_pk = intval($args['uid']);
+
+    if (!$this->dbHelper->doesIdExist("groups", "group_pk", $group_pk)) {
+      $returnVal = new Info(404, "Group id not found!", InfoType::ERROR);
+    } else if (!$this->dbHelper->doesIdExist("users", "user_pk", $user_pk)) {
+      $returnVal = new Info(404, "User id not found!", InfoType::ERROR);
+    } else {
+      try {
+        $dbManager->prepare($stmt = __METHOD__ . ".getByGroupAndUser",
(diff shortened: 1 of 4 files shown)


GMishx (approved):

src/www/ui/api/Controllers/GroupController.php
```diff
@@ -134,4 +134,47 @@ public function deleteGroup($request, $response, $args)
...
+    } else {
+      try {
+        $dbManager->prepare($stmt = __METHOD__ . ".getByGroupAndUser",
+          "SELECT group_user_member_pk FROM group_user_member WHERE group_fk=$1 AND user_fk=$2");
+        $fetchResult = $dbManager->execute($stmt, array($group_pk, $user_pk));
+        $fetchResult = $dbManager->fetchAll($fetchResult);
+        $dbManager->freeResult($fetchResult);
+        if (!empty($fetchResult)) {
+          $group_user_member_pk = $fetchResult[0]['group_user_member_pk'];
+          $adminGroupUsers = new AdminGroupUsers();
```
GMishx, line 167: Please use `getPlugin()` to get class object.
```suggestion
          $adminGroupUsers = $this->restHelper->getPlugin('group_manage_users');
```

GMishx (changes requested):

src/www/ui/api/Controllers/GroupController.php
dushimsam: done

GMishx (approved):
Tested, working as expected.

PR #2278, opened 2022-07-29
feat(ui): close banner for a session

GMishx:
## Description

When a user closes the banner message, it should remain closed on all page loads for current session.

### Changes

1. New JS function `closeBannerForSession()` to close banner if session cookie exists.
2. Adding more files to `.gitignore`
3. Add `jquery.cookie.js` from `foot.html.twig`, eventually removing redundant imports from other places.

## How to test

1. Goto "Admin >> Customize" and add a new "Banner message".
2. Close the banner message and navigate through FOSSology.
  1. The banner should remain closed for all page loads.
3. Navigate other pages which uses cookies for stickiness, and make sure they still work.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2278"></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ cov.txt
 *Tests-Listing.xml
 *Tests-Results.xml
 VERSION
+VERSIONSTRING
 /.project
 /.cproject
 .settings
@@ -24,8 +25,10 @@ VERSION
 tags
 php.ini
 .cache
+.phan
 
 # compiled or derived files
+compile_commands.json
 install/db/dbcreate
 install/defconf/Db.conf
 install/defconf/fossology.conf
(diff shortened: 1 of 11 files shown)


shaheemazmalmmd:
@GMishx : Looks good. but once the banner is closed. User was not able to see the banner after logout and login. should be better to maintain close only for that session. once the user logs back he should see the banner again.

GMishx:
> @GMishx : Looks good. but once the banner is closed. User was not able to see the banner after logout and login. should be better to maintain close only for that session. once the user logs back he should see the banner again.

I have made the changes @shaheemazmalmmd . The banner closure is now handled by PHP and not the JS. Also, the cookie will be removed upon user login/logout. So they will always see the banner upon login.

shaheemazmalmmd (approved):

PR #2287, opened 2022-08-10
fix(API): merge multiple upload-job-apis calls into one.

dushimsam:
## Description
This implementation is merging the two processes for the upload feature into one request.
**The processes are :**
 1. Create the upload.
 2. Create the corresponding job for the new upload.

In  the previous implementation, the frontend would wait up to 10 calls for the agents `ununpack` and `adj2nest` to finish their role.

Now from a single upload request with corresponding scan options, we can post the upload and manage the rest of the work on the backend.

### Changes

   1. Make a reusable function `handleScheduleAnalysis` in `UploadHelper` to proceed with analysis job after upload has been 
       created.
   2. Update the `createJob` function from `JobController` to reuse `handleScheduleAnalysis`.
   3. Call the  `handleScheduleAnalysis` after creating the upload.
  
###  1. Upload Request
 
**Here's the expected upload request format.**

merge_uploads_request

merge_uploads

 ### 2. Response
   **You should see this response object with the ID of the corresponding job in the message.**
   
   
merge_uploads_response

## Related issue
closes : https://github.com/Shruti3004/FOSSology-REST-API/issues/2

cc: @shaheemazmalmmd @GMishx


src/www/ui/api/Controllers/UploadController.php
```diff
@@ -345,7 +350,14 @@ private function changeUpload($request, $response, $args, $isCopy)
    */
   public function postUpload($request, $response, $args)
   {
+
+    $reqBody = $this->getParsedBody($request);
+    if (empty($reqBody["scanOptions"])) {
+      $error = new Info(403, "No agents selected!", InfoType::ERROR);
```
GMishx, line 356: It is not required for the user to select any agent during the upload. This will be contradicting with the current UI.

src/www/ui/api/Controllers/UploadController.php
```diff
@@ -391,6 +403,9 @@ public function postUpload($request, $response, $args)
     }
   }
 
+
+
+
```
GMishx, line 408: Please remove irrelevant changes.

src/www/ui/api/Models/ScanOptions.php
```diff
@@ -83,21 +83,24 @@ public function getArray()
    * current settings.
    * @param integer $folderId Folder with the upload
    * @param integer $uploadId Upload to be scanned
-   * @return Fossology::UI::Api::Models::Info
+   * @return \Fossology\UI\Api\Models\Info::UI::Api::Models::Info
```
GMishx, line 86:
```suggestion
   * @return \Fossology\UI\Api\Models\Info
```

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -1973,6 +1973,17 @@ components:
         - token_scope
         - token_expire
     VcsUpload:
+      description: To create an upload from a version control system
```
GMishx, line 1976: Please add the same options for upload methods like upload from file as well.

GMishx (changes requested):

src/www/ui/api/Controllers/UploadController.php
dushimsam: Yes, this's because I want to create the corresponding job immediately after the upload-creation on the backend. So i want to receive all the required arguments onwards including those for the job I will do later; so as to execute the two processes simultaneously (one after the other).

So in order to continue with the job process, I'm required to specify  those  scanOptions eventhough it's properties might all be set to false.

In order words; I've followed the same flow for creating the job for the upload referring to this job controller's function.

Menu_002

dushimsam: let me know if I'm right sir

src/www/ui/api/Controllers/UploadController.php
dushimsam: sure done.

src/www/ui/api/Controllers/UploadController.php
GMishx: Not exactly. The other endpoint is built specifically to schedule agents on an upload. So it there are no agents selected by the user to schedule, the endpoint will not do any thing. Therefore, it send `403` if no agent is selected.

However, the upload endpoint still have some tasks to do even if user selects no agents to schedule. The same principle is followed in current UI. You can select no agent in section 7 (aptly names "Select **optional** analysis)
image

Where as while scheduling agents on an existing upload, the section is called "Select additional analysis". You can visit the UI for it by visiting [Jobs > Schedule Agents](http://localhost/repo/?mod=agent_add)
image

src/www/ui/api/documentation/openapi.yaml
GMishx: Reminder @dushimsam

shaheemazmalmmd:
@dushimsam Can you fix the test cases, i see that the testcases are failing for uploadController.php

dushimsam:
Everything working perfectly now.

GMishx (approved):
Changes looks good.
Tested, working as expected.

PR #2290, opened 2022-08-11
feat(API): REST-API to initiate FOSSology maintenance

### LINKED ISSUE #2289: INITIATE FOSsology maintenance operations via REST API
### Description
There's a need for the API to initiate fossology Mantainance operations according to the user options in the request..

#### Steps to be followed

1. Create a new route on the main page (index.php) as a part of the Maintenance routes.
2. Add a new function in MaintenanceController to handle the functionality.
3. Test the API.
4. Add the new API's documentation  in [https://github.com/fossology/fossology/blob/master/src/www/ui/api/documentation/openapi.yaml](url)
5. Open a pull request.

#### Expected Results

1.  A POST endpoint to be accessed at ` ''/maintenance" `.
2. The API should return a success message if the maintenance was initiated successfully else another relevant error message.

cc: @shaheemazmalmmd @GMishx

dushimsam:
Signed-off-by: dushimsam <dushsam100@gmail.com>

## Description
 A rest API to initiate FOSSology maintenance operations.

### Changes

1. Added a new controller file and the function to handle the logic in `MaintenanceController`.
2. Updated the file `maintagent.php` to be reusable by the outside callers.
3. Updated  the main file(`index.php`) by adding a new route `POST` `/maintenance`.
4. Updated the `openapi.yaml` file  to introduce a new API.

## How to test

1. Design a request as follows. 
2. The **options** property should contain the list of the keys whose operations are to be executed.
3. The **logsDate** property is the date from which to remove older gold files from repository (_Free to keep it as null_).
4. The **goldDate** property is the date from which to remove older log files from repository (_Free to keep it as null_).
3. Run the API.

maintenance_operations

### Related Issue:
Fixes #2289
    
cc: @shaheemazmalmmd @GMishx

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2290"></a>

### DIFF
--- a/src/www/ui/api/Controllers/MaintenanceController.php
+++ b/src/www/ui/api/Controllers/MaintenanceController.php
@@ -1,0 +1,87 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © 2022 Samuel Dushimimana <dushsam100@gmail.com>
+
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+
+namespace Fossology\UI\Api\Controllers;
+
+use Fossology\Lib\Auth\Auth;
+use Fossology\UI\Api\Helper\ResponseHelper;
+use Fossology\UI\Api\Models\Info;
+use Fossology\UI\Api\Models\InfoType;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Valid status inputs
+ */
+
+
+/**
+ * @class ReportController
+ * @brief Controller for Maintenance model
+ */
+class MaintenanceController extends RestController
+{
+  /**
+   * Create maintenance
+   *
+   * @param ServerRequestInterface $request
+   * @param ResponseHelper $response
+   * @param array $args
+   * @return ResponseHelper
+   */
+  public function createMaintenance($request, $response, $args)
+  {
+    // Check if the request comes from the admin.
+    if (!Auth::isAdmin()) {
+      $returnVal =  new Info(401, "Only admins can run this job.", InfoType::ERROR);
+    } else {
+
+      $body = $this->getParsedBody($request);
+
+      if ($body['options']) {
+
+        //Remove all duplicate options
+
+        if (!is_array($body['options'])) {
+
+          $returnVal = new Info(400, "Options property should be an array.", InfoType::ERROR);
+
+        } else if (in_array("o",$body["options"]) && empty($body["goldDate"])) {
+
+            $returnVal = new Info(400, "Please provide a gold date.", InfoType::ERROR);
+
+        } else if (in_array("l",$body["options"]) && empty($body["logsDate"])) {
+
+          $returnVal = new Info(400, "Please provide a log date.", InfoType::ERROR);
+
+        } else {
+
+          $body['options'] = array_unique($body['options']);
+          $alteredOptions = array();
+          $maintain = $this->restHelper->getPlugin('maintagent');
+          $existingOptions = $maintain->getOptions();
+
+          //Check if all the given keys exist in the known options array
+          foreach ($body['options'] as $key) {
+            if (!array_key_exists($key, $existingOptions)) {
+              $returnVal = new Info(404, "KEY '" . $key . "' NOT FOUND!", InfoType::ERROR);
+              return $response->withJson($returnVal->getArray(), $returnVal->getCode());
+            }
+            $alteredOptions[$key] = $key;
+          }
+
+          $body['options'] = $alteredOptions;
+          $mess = $maintain->handle($body);
+          $returnVal = new Info(201, $mess, InfoType::INFO);
+        }
+
+      } else {
+        $returnVal = new Info(400,"No options provided!", InfoType::ERROR);
+      }
+    }
+    return $response->withJson($returnVal->getArray(), $returnVal->getCode());
+  }
+}
--- a/src/www/ui/api/documentation/openapi.yaml
+++ b/src/www/ui/api/documentation/openapi.yaml
@@ -128,7 +128,48 @@ paths:
                 $ref: '#/components/schemas/ApiInfo'
         default:
           $ref: '#/components/responses/defaultResponse'
-
+  /maintenance:
+    post:
+      operationId: initiateMaintenance
+      tags:
+        - Maintenance
+      summary:  Initiate maintenance operations
+      description: > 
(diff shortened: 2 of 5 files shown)


src/www/ui/api/Controllers/MaintenanceController.php
```diff
@@ -0,0 +1,61 @@
...
+*/
+
+namespace Fossology\UI\Api\Controllers;
+
+
+use Fossology\Lib\Dao\UserDao;
+use Fossology\UI\Api\Helper\ResponseHelper;
+use Fossology\UI\Api\Models\Info;
+use Fossology\UI\Api\Models\InfoType;
+use maintagent;
```
GMishx, line 15: Please remove undefined namespace.

src/www/ui/api/Controllers/MaintenanceController.php
```diff
@@ -0,0 +1,61 @@
...
+    $body = $this->getParsedBody($request);
+    $maintain  = $this->restHelper->getPlugin('maintagent');
+
+    //Remove all duplicate options
+    $body['options'] = array_unique($body['options']);
+
+    //Check if all the given keys exist in the known options array
+    $alteredOptions = array();
+    foreach ($body['options'] as $key) {
+      if (!array_key_exists($key, $maintain->getOptions())) {
```
GMishx, line 48: Please store response from `$maintain->getOptions()` in a variable to reduce function calls.

src/www/ui/api/Controllers/MaintenanceController.php
```diff
@@ -0,0 +1,61 @@
...
+        $returnVal = new Info(404, "KEY NOT FOUND!", InfoType::ERROR);
```
GMishx, line 49: It will be better if user knows which key is not found :-)

src/www/ui/api/Controllers/MaintenanceController.php
```diff
@@ -0,0 +1,61 @@
...
+   * Create maintenance
+   *
+   * @param ServerRequestInterface $request
+   * @param ResponseHelper $response
+   * @param array $args
+   * @return ResponseHelper
+   */
+  public function createMaintenance($request, $response, $args)
+  {
+    $body = $this->getParsedBody($request);
```
GMishx, line 39: Please add a check here so only admins can run this job.

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -2294,6 +2329,26 @@ components:
...
+    CreateMaintenance:
+      description: Maintenance options
+      type: object
+      properties:
+        options:
+          type: array
+          description: options list
+          nullable: false
+          items:
+            type: string
```
GMishx, line 2341: Please provide all possible options as an `enum`.

GMishx (changes requested):

dushimsam:
Applied all chances. 
 
Let me know if there is any other thing to modify.

src/www/ui/api/Controllers/MaintenanceController.php
```diff
@@ -0,0 +1,65 @@
...
+   * @param ServerRequestInterface $request
+   * @param ResponseHelper $response
+   * @param array $args
+   * @return ResponseHelper
+   */
+  public function createMaintenance($request, $response, $args)
+  {
+    // Check if the request comes from the admin.
+    if (!Auth::isAdmin()) {
+      $returnVal =  new Info(400, "Only admins can run this job.", InfoType::INFO);
```
GMishx, line 39: Please use [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) unauthorized request code.

GMishx (changes requested):
A small change required.

src/www/ui/api/Controllers/MaintenanceController.php
dushimsam: yeap , done

GMishx (approved):
Changes look good.

GMishx (changes requested):
While testing:

1. Providing `options` as an empty array schedules the maintenance agent without any action to preform. Which is not expected. In case the `options` is empty, a `400` response should be sent saying "At least one option should be selected".
2. In case of using `o` as the option ("Remove older gold files from repository.") and providing date using `logsDate` cause maintenance agent to fail with following message:
    ```
    /usr/local/etc/fossology/mods-enabled/maintagent/agent/maintagent: option requires an argument -- 'o'
    ```

Please check following code for option `o`, `l` and `t`
https://github.com/fossology/fossology/blob/460e6985b66e5de3fe0b1bda13bd6d77b72c7cb2/src/maintagent/ui/maintagent.php#L40-L53

dushimsam:
> While testing:
> 
> 1. Providing `options` as an empty array schedules the maintenance agent without any action to preform. Which is not expected. In case the `options` is empty, a `400` response should be sent saying "At least one option should be selected".
> 2. In case of using `o` as the option ("Remove older gold files from repository.") and providing date using `logsDate` cause maintenance agent to fail with following message:
>    ```
>    /usr/local/etc/fossology/mods-enabled/maintagent/agent/maintagent: option requires an argument -- 'o'
>    ```
> 
> Please check following code for option `o`, `l` and `t`
> 
> https://github.com/fossology/fossology/blob/460e6985b66e5de3fe0b1bda13bd6d77b72c7cb2/src/maintagent/ui/maintagent.php#L40-L53

Ok, cool I've found what's causing the problem.
Thanks to let me know.
releasing a clean PR in time.

dushimsam:
> While testing:
> 
> 1. Providing `options` as an empty array schedules the maintenance agent without any action to preform. Which is not expected. In case the `options` is empty, a `400` response should be sent saying "At least one option should be selected".
> 2. In case of using `o` as the option ("Remove older gold files from repository.") and providing date using `logsDate` cause maintenance agent to fail with following message:
>    ```
>    /usr/local/etc/fossology/mods-enabled/maintagent/agent/maintagent: option requires an argument -- 'o'
>    ```
> 
> Please check following code for option `o`, `l` and `t`
> 
> https://github.com/fossology/fossology/blob/460e6985b66e5de3fe0b1bda13bd6d77b72c7cb2/src/maintagent/ui/maintagent.php#L40-L53

Done could you review sir .

GMishx (approved):
Changes looks good.
Tested, working as expected.

PR #2304, opened 2022-08-31
fix(reportImport): remove dual check for access and fix array warning

shaheemazmalmmd:
## Description

Remove code which checks iseditible two times and fix array warnings.

## How to test

* Upload findings using rdf file and check if the reportimport runs.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2304"></a>


GMishx (approved):
Changes looks good. Needs testing.

GMishx (approved):
Working as expected.

PR #2334, opened 2022-10-14
fix(oneShotMonk): convert value to int to fix php fatal

shaheemazmalmmd:
## Description

add intval function to avoid php fatal .

## How to test

* Go to -> admin -> license admin -> candidates.
* Edit some candidate license and compare.

Credits : @GMishx 

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2334"></a>


GMishx (approved):
Changes looks good.

PR #2337, opened 2022-10-19
feat(user-edit): make use of retention period to display expired tokens

shaheemazmalmmd:
## Description

Change query to get tokens above Max expired token retention period.

## How to test

* set Max expired token retention period from conf.
* go to user edit.
* user shouldn't see tokens before Max expired token retention period.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2337"></a>


GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected

PR #2347, opened 2022-11-30
fix(licenseRef): show only active licenses in bulk and user decisions

shaheemazmalmmd:
Signed-off-by: Shaheem Azmal M MD <shaheem.azmal@siemens.com>

## Description

Inactive licenses ate still been listed in user decisions and bulk license selection.

### Changes

Change query to select only active licenses.

## How to test

* Deactivate some license from license admin.
* It should not appear while adding user decisions and bulk license.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2347"></a>


GMishx (approved):

GMishx (approved):
Tested, working as expected.

PR #2349, opened 2022-12-02
fix(oauth): update username if oauth email matches

GMishx:
## Description

While using OIDC, if user is matched using `user_email` field, update the username from DB for following checks.

### Changes

In `getUserAndDefaultGroupByUserName()` make parameter `$userName` as pass-by-reference and update it to `user_name` from DB in case of OAuth.

## How to test

1. Setup OIDC.
2. Create a user with different username but keep email field to match oidc.
3. Login with the new user using OIDC. The login should succeed.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2349"></a>

### DIFF
--- a/src/lib/php/Dao/UserDao.php
+++ b/src/lib/php/Dao/UserDao.php
@@ -273,7 +273,7 @@ class UserDao
         array($userId, $groupId), __FUNCTION__);
   }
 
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):

PR #2351, opened 2022-12-04
fix(package): use prefix /usr for Debian packages

GMishx:
## Description

CMake by default uses `/usr/local` as install prefix. This works fine for local setup but fails for Debian package installations.

While creating release packages, fix the install prefix.

### Changes

Use `-DCMAKE_INSTALL_PREFIX=/usr` while creating release packages.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2351"></a>

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -89,7 +89,7 @@ jobs:
       run: |
         mkdir build
         cd build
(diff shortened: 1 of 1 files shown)


GMishx:
Closing as #2353 contains the required changes.

PR #2361, opened 2022-12-15
feat(eximporter): add file path for upload tree

GMishx:
## Description

While exporting dump, add "path" of file to uploadtree list. This allows matching files even if ununpack version changes over time. (for example, old version of ununpack did not extracted `.dsc` files)

### Changes

1. Add "path" to uploadtree list while creating export dump.
2. While importing, check if item path matches to get new item id. If path does not exist or does not match, then fall back to lft and rgt.


shaheemazmalmmd (approved):

PR #2373, opened 2023-01-25
fix(unifiedReport): fix table distortion for component link

shaheemazmalmmd:
## Description

When creating the unified report, the cell get extremely large which holds the component link.

### Changes

Added fixed width using existing table classes.

## How to test

* Add large URL into component link column from conf page.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2373"></a>


GMishx (approved):
Changes looks good. Needs test

GMishx:
Can you please keep the tables to be left aligned instead of center? As the tables in new report (on the right) looks very odd centered when they have different width.

image

GMishx (approved):
Tested, working as expected.

PR #2377, opened 2023-02-11
Reduce number of layer in docker development file

Adityasaini10012001:
Signed-off-by: Aditya Saini <72370777+Adityasaini10012001@users.noreply.github.com>

## Description

Please describe the changes in your pull request in few words here.

### Changes

## How to test

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2377"></a>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,9 +42,7 @@ COPY ./src/ununpack/mod_deps ./src/ununpack/
 COPY ./src/wget_agent/mod_deps ./src/wget_agent/
 
 RUN mkdir -p /fossology/dependencies-for-runtime \
- && cp -R /fossology/src /fossology/utils /fossology/dependencies-for-runtime/
-
-RUN DEBIAN_FRONTEND=noninteractive apt-get update \
+ && cp -R /fossology/src /fossology/utils /fossology/dependencies-for-runtime/ || DEBIAN_FRONTEND=noninteractive apt-get update \
  && DEBIAN_FRONTEND=noninteractive /fossology/utils/fo-installdeps --build -y \
  && DEBIAN_FRONTEND=noninteractive /fossology/install/fo-install-pythondeps --build -y \
  && rm -rf /var/lib/apt/lists/*


AtulRajput01 (approved):
Nice!!
aaa..
As a way to make it clearer, I think I'll move the wget test earlier so that I don't need to ask hard questions about the order of evaluation

Dockerfile
```diff
@@ -42,9 +42,7 @@ COPY ./src/ununpack/mod_deps ./src/ununpack/
 COPY ./src/wget_agent/mod_deps ./src/wget_agent/
 
 RUN mkdir -p /fossology/dependencies-for-runtime \
- && cp -R /fossology/src /fossology/utils /fossology/dependencies-for-runtime/
-
-RUN DEBIAN_FRONTEND=noninteractive apt-get update \
+ && cp -R /fossology/src /fossology/utils /fossology/dependencies-for-runtime/ || DEBIAN_FRONTEND=noninteractive apt-get update \
```
GMishx, line 45: As a standard best practice, the installation of dependencies should always be done in a separate layer. I do not see any advantage of this change here.

GMishx (changes requested):
Thank you for your contribution @Adityasaini10012001 .

Can you please elaborate on the changes more?

Dockerfile
Adityasaini10012001: > As a standard best practice, the installation of dependencies should always be done in a separate layer. I do not see any advantage of this change here.

@GMishx   I'm very concerned about the number of layers in container images that I always not build locally and push to a container repository, but no problem to simplify.

As a way to make it clearer, I think I'll move the wget test earlier so that I don't need to ask hard questions about the order of evaluation of &&

GMishx: Number of layers should not matter much, infact they help you caching the builds. Also, the images which are pulled from repositories are generally compressed.

I am not sure what you mean by "I think I'll move the wget test earlier"
Which tests you are referring to?

Adityasaini10012001:
I reduced the number of layers in the file in order to optimize build times and make the Dockerfile easier to understand and maintain.

While I understand that layers can help with caching, having fewer layers can also reduce the amount of data that needs to be transferred during the build process, which can help with overall build times. Additionally, having a simpler Dockerfile can make it easier for developers to understand and maintain the file.

Regarding my comment about moving the "wget test" earlier, I would appreciate some clarification on what this test is and what it does. Once I have a better understanding of the test, I can evaluate whether moving it to an earlier point in the Dockerfile would be beneficial.

I would be happy to discuss this further with you and come to a shared understanding about the best approach for our Dockerfile.

Adityasaini10012001:
i am referring a  "wget test" in the Dockerfile could include:
Ensuring that all necessary dependencies are downloaded and installed before building and running the application.
Saving time during the build process by downloading dependencies in parallel.
Allowing for offline builds by caching dependencies locally in the Docker image..

Dockerfile
Adityasaini10012001: I reduced the number of layers in the file in order to optimize build times and make the Dockerfile easier to understand and maintain.

While I understand that layers can help with caching, having fewer layers can also reduce the amount of data that needs to be transferred during the build process, which can help with overall build times. Additionally, having a simpler Dockerfile can make it easier for developers to understand and maintain the file.

Regarding my comment about moving the "wget test" earlier, I would appreciate some clarification on what this test is and what it does. Once I have a better understanding of the test, I can evaluate whether moving it to an earlier point in the Dockerfile would be beneficial.

I would be happy to discuss this further with you and come to a shared understanding about the best approach for our Dockerfile.

Adityasaini10012001: i am referring a "wget test" in the Dockerfile could include:
Ensuring that all necessary dependencies are downloaded and installed before building and running the application.
Saving time during the build process by downloading dependencies in parallel.
Allowing for offline builds by caching dependencies locally in the Docker image..

GMishx:
Unfortunately, I see no clear benefit of the changes done in the PR. Infact, I see some drawbacks related to best practices with package manager runs in Docker.
Also, I am still not clear which wget test you are talking about in the `Dockerfile` as it contains none.

Thank you for contributing, but closing the PR for now. Please create/re-open if there are more improvements.

PR #2398, opened 2023-03-19
fix(user-edit.php): Fixed editing emails allows for duplicate emails …

### LINKED ISSUE #2397: Editing emails allows for duplicate emails
### Description

When editing user information, Changing the email address allows for multiple users to have the same email address value

#### How to reproduce

- Add two new users
- Login to each user, go the "Edit User Account" page under the "Admin" panel
- Change each user's email to the same email

#### Screenshots

image

image

### Versions

* Last commit id on master: 1308485f8e8622bd539e64b9457b8eb7034fbf33
* Operating System (lsb_release -a): Ubuntu 20.04.5 LTS

Hero2323:
…for multiple users

Added a check to see if the new email is unique or not This fixes #2397

## Description

Users can't edit their email with an email that's not unique.

### Changes

Added a uniqueness check for the email field

## How to test

- Create two new users with different emails.
- Try to edit the email value of one of the users to the other's email value.
- Previously, this would have worked, now it displays an error message and the email isn't changed.

This fixes #2397

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2398"></a>

### DIFF
--- a/src/www/ui/user-edit.php
+++ b/src/www/ui/user-edit.php
@@ -265,6 +265,17 @@ class UserEditPage extends DefaultPlugin
       $Errors .= "<li>" . _("Invalid email address.") . "</li>";
     }
 
+    /* Make sure email is unique */
+    $email_count = 0;
+    if (!empty($UserRec['user_email'])) {
+      $email_count = $this->dbManager->getSingleRow(
+        "SELECT COUNT(*) as count FROM users WHERE user_email = $1 LIMIT 1;",
+        array($UserRec['user_email']))["count"];
+    }
+    if ($email_count > 0) {
+      $Errors .= "<li>" . _("Email address already exists.") . "</li>";
+    }
+
     /* Make sure user can't ask for blank password if policy is enabled */
     if (passwordPolicyEnabled() && !empty($UserRec['_blank_pass'])) {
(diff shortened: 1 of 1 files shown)


src/www/ui/user-edit.php
```diff
@@ -265,6 +265,16 @@ function UpdateUser($UserRec, $SessionIsAdmin)
       $Errors .= "<li>" . _("Invalid email address.") . "</li>";
     }
 
+    /* Make sure email is unique */
+    if (!empty($UserRec['user_email'])) {
+      $email_count = $this->dbManager->getSingleRow(
+        "SELECT COUNT(*) as count FROM users WHERE user_email = $1 LIMIT 1;",
+        array($UserRec['user_email']))["count"];
+    }
+    if ($email_count > 0) {
```
GMishx, line 274: `$email_count` can be undefined based on condition if email is empty or not.

GMishx (changes requested):

src/www/ui/user-edit.php
Hero2323: Fixed it by defining the `$email_count` variable beforehand.

GMishx (approved):
Changes looks good.
Tested, working as expected.

@Hero2323 , please rebase the branch and squash 3 commits into 1.

Hero2323:
Squashed the three commits.

Issue #542, opened 2015-11-03
cp2foss should have an option to set/change folder description


pguttmann:
Currently, the -d option allows the addition of a description field for uploaded file(s)? 
I'm not sure how this works for multiple file uploads, eg: directory upload.
There should also be a capability for "enhancing" the -f option to included a "folder description" to be set along with the FOSSology repository folder path and/or name.

shaheemazmalmmd:
Update if upload description can be done through REST API.

PR #2412, opened 2023-03-24
Master

Vaibhav992:
changing -d to -f(file description) in #542

### DIFF
--- a/src/cli/cp2foss.pod
+++ b/src/cli/cp2foss.pod
@@ -12,7 +12,7 @@ cp2foss - upload one or more files or archives into FOSSology and run scans.
 
  Usage: cp2foss [options] [archives]
  Options: -AQThv [-f path] [-AA num] [-q {agent,agent.. | all}] [-X path]
- [-n name] [-d description] [--username user name] [--password password]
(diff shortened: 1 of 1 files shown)


src/cli/cp2foss.pod
```diff
@@ -12,7 +12,7 @@ cp2foss - upload one or more files or archives into FOSSology and run scans.
 
  Usage: cp2foss [options] [archives]
  Options: -AQThv [-f path] [-AA num] [-q {agent,agent.. | all}] [-X path]
- [-n name] [-d description] [--username user name] [--password password]
+ [-n name] [-f description] [--username user name] [--password password]
```
GMishx, line 15: Code says
https://github.com/fossology/fossology/blob/c624c39837c5961d596283d3ceed814e5f74b742/src/cli/cp2foss.php#L46
https://github.com/fossology/fossology/blob/c624c39837c5961d596283d3ceed814e5f74b742/src/cli/cp2foss.php#L52

Please recheck your change.

GMishx (changes requested):

PR #2416, opened 2023-03-27
fix(adminLicensecandidate): replace while loop with foreach

shaheemazmalmmd:
## Description

Not able to load admin -> license admin -> candidates page

### Changes

* Change loop from while to foreach
* Swap variables as required. 

## How to test

* Check  admin -> license admin -> candidates
* Page should load without error.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2416"></a>


GMishx (approved):
Changes looks good. Tested, working as expected.

PR #2421, opened 2023-03-30
fix(api): Fixed filesearch request

### LINKED ISSUE #2186: REST API: endpoint /filesearch: data type description inaccurate
### Description

The OpenAPI description, version 1.4.3, says that the input for the /filesearch endpoint
is an array of `Hash` objects.

This is not totally correct:
* As soon as the property `size` is available in the JSON data, the call to /filesearch does not return results,
  but says `Invalid keys`.
* The example data provided, 
```
[
  {
    "sha1": "81fe8bfe87576c3ecb22426f8e57847382917acf",
    "md5": "e2fc714c4727ee9395f324cd2e7f331f",
    "sha256": "88d4266fd4e6338d13b845fcf289579d209c897823b9217da3e161936f031589",
    "size": 0
  }
]
```
will result in such an `Invalid keys` message.
* As soon as the `file` property is omitted, i.e.
```
[
  {
    "sha1": "81fe8bfe87576c3ecb22426f8e57847382917acf",
    "md5": "e2fc714c4727ee9395f324cd2e7f331f",
    "sha256": "88d4266fd4e6338d13b845fcf289579d209c897823b9217da3e161936f031589",
  }
]
```
a correct result will be returned.

### Versions

* Commit: [#fcf1de] 2022/03/09 08:26 +00:00 built @ 2022/03/09 08:32 +00:00
* Latest version of the containerized version of FOSSology

srideep-banerjee:
## Description

Fixed the issue of filesearch request with size field returning Invalid key response

### Changes

Added size to to the list of allowed keys in Hash.php

## How to test
Create any filesearch request with size field in Json data.

Screenshot from 2023-03-30 11-18-43

Fixed #2186

### DIFF
--- a/src/www/ui/api/Models/Hash.php
+++ b/src/www/ui/api/Models/Hash.php
@@ -21,7 +21,7 @@ class Hash
    * @var array $ALLOWED_KEYS
    * Allowed keys from user to parse
    */
-  const ALLOWED_KEYS = ['sha1', 'sha256', 'md5'];
+  const ALLOWED_KEYS = ['sha1', 'sha256', 'md5', 'size'];
 
   /**
    * @var string $id


GMishx (approved):
Tested, working as expected.

Issue #2404, opened 2023-03-21
REST API Improvements For FOSSology


shaheemazmalmmd:
## REST API Improvements

Following endpoints are required to support the new [FOSSologyUI](https://github.com/fossology/FOSSologyUI). The endpoints have been assigned priorities as well in the order they should be implemented.

If you are working on any endpoint from here, please create a corresponding new issue for the same and comment here with the issue ID.

TOPIC | FEATURE | PRIORITY | Issue type | NOTES
-- | -- | -- | -- | --
Open API | Open API Info -> YAML/JSON | 1 | New |
License Browser | Endpoint -> "uploads/{id}/license" | 2 | Modify | Add pagination
License Browser | Tree view endpoint With Folder & file {itemid + type}, Scanner licenses, edited licenses, count and filters. | 4 | New |
License Browser | Tree view: Reuse Summary, Scanner Details (GET & for filter) | 4 | New |
License File View | Get file content | 3 | New
License File View | File Details: Scanner results, edited results with text, ack, comment. Main license if any (GET & SET for all) Highlight Info with Type (Keyword, license), start and end   details. | 3 | New
License File View | Get Bulk History, clearing history, bulk scan schedule and right left with 3   filters. (GET & SET) | 3 | New |
Conf | Unified Report Settings (GET & SET), SPDX Report Settings (GET & SET), Obligations Settings (GET & SET), Decision Settings (GET & SET) | 6 | New |
Info | Get info based on ItemId (GET) | 6 | New | The "Info" page
Copyright Browser | Aggregated Copyrights, ECC, IPRA, Email/url/Author and Keyword with count, content, activated (FOSSology, Scancode, User) + Operations: edit, delete | 5 | New |
Copyright File-view | File Details (Copyright/email/url/Author, ECC, IPRA) Content, Count, finding, activated/deactivated, delete +  User findings with  description, text finding, comment for both activated and deactivated. With all decision options. | 5 | New |
OneShot | One short analysis for license, copyright, monk and OJO |   | New |
Customize | All the data inside the page with GET & SET | 7 | New |
Jobs | Modify my recent and all recent jobs to get all the required data for UI. | 8 | Modify |
Folder | Unlink folder |   | New |
License | Candidate License Merge Request and edit licenses | 8 | New |
Dashboard | Get all 3 page info for dashboard |   | New |
License Admin | Get & Set Acknowledgements & comments (Standard) | 10 | New |
[...]

PR #2430, opened 2023-04-06
feat(api): add pagination to license browser

mayank-pathakk:
# Description

As mentioned in the issue #2404, pagination feature needed to be added to the endpoint "uploads/{id}/license".
Hence, this PR aims to solve the same. 

## Changes

Additional fields ( i.e. limits and page ) can be passed to UploadController, which calls for UploadHelper. While receiving response from UploadHelper, the response array is modified using `array_slice` according to the needed page and limit fields. 

Also updated the `open-api.yaml` to display the changes.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2430"></a>

## Testing
1.) Using an auth token send a get request to `uploads/{id}/license` . This will display all the results.
2.) Include page and limit in the headers of the request. This will display the paginated results.

### Before :

Screenshot from 2023-04-06 23-31-05

### After : 

Screenshot from 2023-04-06 23-31-18

### DIFF
--- a/src/www/ui/api/Helper/UploadHelper.php
+++ b/src/www/ui/api/Helper/UploadHelper.php
@@ -737,7 +737,7 @@ class UploadHelper
    * @return array Array containing `filePath`, `agentFindings` and
    * `conclusions` for each upload tree item
    */
-  public function getUploadLicenseList($uploadId, $agents, $printContainers, $boolLicense, $boolCopyright)
+  public function getUploadLicenseList($uploadId, $agents, $printContainers, $boolLicense, $boolCopyright, $page = 0, $limit = 50)
   {
     global $container;
     $restHelper = $container->get('helper.restHelper');
@@ -821,6 +821,9 @@ class UploadHelper
         $responseList[] = $responseRow->getArray();
       }
     }
-    return $responseList;
+    $offset = $page * $limit;
+    $paginatedResponseList = array_slice($responseList, $offset, $limit);
+    $paginatedResponseListSize = sizeof($responseList);
+    return array($paginatedResponseList, $paginatedResponseListSize);
   }
 }
--- a/src/www/ui/api/documentation/openapi.yaml
+++ b/src/www/ui/api/documentation/openapi.yaml
@@ -654,6 +654,23 @@ paths:
         schema:
           type: boolean
           default: false
+      - name: page
+        description: Page number (starts from 1)
+        required: false
+        in: header
+        schema:
+          type: integer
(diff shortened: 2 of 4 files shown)


src/www/ui/api/documentation/openapi.yaml
```diff
@@ -654,6 +654,23 @@ paths:
...
+            minimum: 1
+      - name: limit
+          description: Limits of responses per request
+          required: false
+          in: header
+          schema:
+            type: integer
+            default: 50
+            minimum: 1
+            maximum: 100
```
GMishx, line 673: For this particular endpoint, I think the max limit can be increased to 1000

src/www/ui/api/Helper/UploadHelper.php
```diff
@@ -821,6 +821,9 @@ public function getUploadLicenseList($uploadId, $agents, $printContainers, $bool
         $responseList[] = $responseRow->getArray();
       }
     }
-    return $responseList;
+    $Offset = $Page * $Limit;
```
GMishx, line 824: Please use proper naming convention for variables. They should follow camel casing.

src/www/ui/api/Helper/UploadHelper.php
```diff
@@ -821,6 +821,9 @@ public function getUploadLicenseList($uploadId, $agents, $printContainers, $bool
...
+    $paginatedResponseList = array_slice($responseList, $Offset, $Limit);
+    $paginatedResponseListSize = sizeof($paginatedResponseList);
```
GMishx, line 826: `$paginatedResponseListSize` should represent the total size of array, not the size of response. Otherwise client will never know how many pages it can fetch.

GMishx (changes requested):

src/www/ui/api/documentation/openapi.yaml
mayank-pathakk: Corrected it

mayank-pathakk:
@GMishx sir I have made the requested changes, rebased my PR, and fixed the tests, which were causing 2 workflows to fail, now all the workflows have passed. :)

dushimsam:
These changes look good to me. Related issue [#2453 ](https://github.com/fossology/fossology/issues/2453)

@GMishx  @shaheemazmalmmd

mayank-pathakk:
Rebased

GMishx (approved):
Changes looks good.
Tested, working as expected.

Issue #2432, opened 2023-04-13
Broken Travis-CLI badge in README


mayank-pathakk:
### Description

The Travis svg in the readme does not render and instead a hyperlink is displayed 

#### How to reproduce

Visible on the home page of repository in the README.
I looked up and found that `travis.org` was been ceased and now it uses `travis.com`, but still I could not find fossology in travis.com.

Possible fix : Designing a svg from `image shield` just like the other svgs, and using the correct link to travis.

#### Screenshots
![travis-badge-bug](https://user-images.githubusercontent.com/97532378/231659063-7411f985-a2bc-4a09-b176-d30e8c7eed43.jpeg)

mayank-pathakk:
I would like to work on this issue, can someone provide the correct link

GMishx:
Solved by #2433

PR #2433, opened 2023-04-13
fix(README): Fix broken Travis SVG

mayank-pathakk:
## Description

This is a possible solution to #2432, the link where the badge should point to still needs to be edited, because currently it points to a page which does not exist

### Changes

The badge was rendered via an SVG, changed to a badge from `image shields`.

`Link still needs to be added`

## How to test

Badge can be viewed in the README.md of the project.

Might close issue #2432 .

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2433"></a>


GMishx:
FOSSology no longer uses Travis CI. The status link can be removed from Readme.

PR #2445, opened 2023-05-12
fix(build): remove jq

GMishx:
## Description

1. Fix the version in CMake for rc tags.
2. Remove unnecessary dependencies from scanoss.

### Changes

1. Fix the string append command in CMake.
2. Remove jq with json-c in scanoss
3. Remove not needed dependencies from scanoss

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2445"></a>


shaheemazmalmmd (approved):

PR #2448, opened 2023-05-18
fix(bulkReuse): concider folder level and package level bulk

shaheemazmalmmd:
## Description

Consider folder level bulk while reuse. 

### Changes

Search for folder uploadtreeID and schedule a bulk scan.

## How to test

* upload package x.
* schedule a bulk scan for package x for a folder.
* upload package x.1 and reuse bulk decisions from package x.
* bulk should only be applied to folder level.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2448"></a>

### DIFF
--- a/src/decider/agent/BulkReuser.php
+++ b/src/decider/agent/BulkReuser.php
@@ -51,34 +51,54 @@ class BulkReuser
      * UploadDao from container
      */
     $uploadDao = $GLOBALS['container']->get('dao.upload');
-    $topItem = $uploadDao->getUploadParent($uploadId);
-    /** @var DeciderJobAgentPlugin $deciderPlugin
-     * DeciderJobAgentPlugin object to add deciderjob
-     */
-    $deciderPlugin = plugin_find("agent_deciderjob");
-    $dependecies = array();
-    $sql = "INSERT INTO license_ref_bulk (user_fk,group_fk,rf_text,upload_fk,uploadtree_fk,ignore_irrelevant,bulk_delimiters) "
-            . "SELECT $1 AS user_fk, $2 AS group_fk,rf_text,$3 AS upload_fk, $4 as uploadtree_fk, ignore_irrelevant, bulk_delimiters
-              FROM license_ref_bulk WHERE lrb_pk=$5 RETURNING lrb_pk, $5 as lrb_origin";
-    $sqlLic = "INSERT INTO license_set_bulk (lrb_fk, rf_fk, removing, comment, reportinfo, acknowledgement) "
-            ."SELECT $1 as lrb_fk, rf_fk, removing, comment, reportinfo, acknowledgement FROM license_set_bulk WHERE lrb_fk=$2";
-    $this->dbManager->prepare($stmt=__METHOD__.'cloneBulk', $sql);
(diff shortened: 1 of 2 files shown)


src/decider/agent/BulkReuser.php
```diff
@@ -51,34 +51,52 @@ public function rerunBulkAndDeciderOnUpload($uploadId, $groupId, $userId, $bulkI
...
+    $pUTree = $this->dbManager->getSingleRow("SELECT uploadtree_fk FROM license_ref_bulk WHERE lrb_pk=$1",
+      array($bulkId), __METHOD__);
+    $uploadEntry = $uploadDao->getUploadEntry($pUTree['uploadtree_fk']);
+    $pUID = intval($uploadEntry['upload_fk']);
+    $pTopItem = $uploadDao->getUploadParent($pUID);
+    if ($pTopItem == $pUTree['uploadtree_fk']) {
+      $topItem = $nTopItem;
+    } else {
+      $pfUTree = $this->dbManager->getSingleRow("SELECT uploadtree_pk FROM uploadtree WHERE upload_fk=$1 AND ufile_name=$2 AND ufile_mode=$3",
+        array($pUID, $uploadEntry['ufile_name'], $uploadEntry['ufile_mode']), __METHOD__.'getRealUploadtreeEntry');
```
GMishx, line 64: Shouldn't the uploadID used here (`$pUID`) be of current upload, i.e. `$uploadId`? Otherwise, you are fetching the item ID of element from previous upload and it already exists as `$pUTree['uploadtree_fk']`

src/decider/agent/BulkReuser.php
```diff
@@ -51,34 +51,52 @@ public function rerunBulkAndDeciderOnUpload($uploadId, $groupId, $userId, $bulkI
...
+      if (!empty($pfUTree) && count($pfUTree) <= 1) {
```
GMishx, line 65: If the above comment is to be considered, the conditions here should flip as well because `$pfUTree` will not contain the corresponding item id in the new upload. This can then be used as `$topItem`.

src/decider/agent/BulkReuser.php
```diff
@@ -51,34 +51,52 @@ public function rerunBulkAndDeciderOnUpload($uploadId, $groupId, $userId, $bulkI
...
+        $topItem = $pfUTree['uploadtree_pk'];
+      } else {
+        $topItem = "";
+      }
+    }
+    if (!empty($topItem)) {
```
GMishx, line 71: There is no return in case this condition is false. And as per doc string, the function should always return an integer.

https://github.com/fossology/fossology/blob/ffb851da6a31b811ae53964190defca056a5c549/src/decider/agent/BulkReuser.php#L46

GMishx (changes requested):
Some questions for clarification.

GMishx:
@shaheemazmalmmd , looks like in test cases, no bulk is being scheduled due to changes. Please fix the test cases:

https://github.com/fossology/fossology/actions/runs/5176896849/jobs/9326307295?pr=2448#step:10:35

GMishx (approved):
Tested, working as expected.

Issue #2472, opened 2023-06-14
License File View - Bulk History


dushimsam:
Get Bulk History, clearing history, bulk scan schedule and right left with 3 filters. (GET & SET)

PR #2477, opened 2023-06-16
feat(api): Get clearing-history API for an item

dushimsam:
## Description

Added the API to a get a list of the clearing history for a particular item.

### Changes

1. Added a new method in  `UploadTreeController` to handle the logic.
2. Created the method in `UploadTreeController` for retrieve the list of the data.
3. Updated  the main file(`index.php`) by adding a new route `GET` `/uploads/{id}/items/{itemId}/clearing-history`.
4. Updated the `openapi.yaml` file  to write the new API's documentation.

## How to test

Make a GET request on the endpoint: `GET` `/uploads/{id}/items/{itemId}/clearing-history`.,

## Screenshots

image

### Related Issue:
Fixes [#2472](https://github.com/fossology/fossology/issues/2472)

    
cc: @shaheemazmalmmd @GMishx

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2477"></a>

### DIFF
--- a/src/www/ui/api/Controllers/UploadTreeController.php
+++ b/src/www/ui/api/Controllers/UploadTreeController.php
@@ -12,9 +12,11 @@
 
 namespace Fossology\UI\Api\Controllers;
 
+use Fossology\Lib\Data\DecisionScopes;
 use Fossology\Lib\Data\DecisionTypes;
 use Fossology\UI\Api\Helper\ResponseHelper;
 use Fossology\UI\Api\Models\BulkHistory;
+use Fossology\UI\Api\Models\ClearingHistory;
 use Fossology\UI\Api\Models\Info;
 use Fossology\UI\Api\Models\InfoType;
 use Psr\Http\Message\ServerRequestInterface;
@@ -228,5 +230,54 @@ class UploadTreeController extends RestController
       $updatedRes[] = $obj->getArray();
     }
     return $response->withJson($updatedRes, 200);
+  }
+
+  /**
+   * Get the clearing history for a particular upload-tree
+   *
+   * @param ServerRequestInterface $request
+   * @param ResponseHelper $response
+   * @param array $args
+   * @return ResponseHelper
+   */
+  public function getClearingHistory($request, $response, $args)
+  {
+    $itemId = intval($args['itemId']);
+    $uploadId = intval($args['id']);
+    $uploadDao = $this->restHelper->getUploadDao();
+    $clearingDao = $this->container->get('dao.clearing');
+
+    $returnVal = null;
+
+    if (!$this->dbHelper->doesIdExist("upload", "upload_pk", $uploadId)) {
+      $returnVal = new Info(404, "Upload does not exist", InfoType::ERROR);
+    } else if (!$this->dbHelper->doesIdExist($uploadDao->getUploadtreeTableName($uploadId), "uploadtree_pk", $itemId)) {
+      $returnVal = new Info(404, "Item does not exist", InfoType::ERROR);
+    }
+
+    if ($returnVal !== null) {
+      return $response->withJson($returnVal->getArray(), $returnVal->getCode());
+    }
+
+    $itemTreeBounds = $uploadDao->getItemTreeBoundsFromUploadId($itemId, $uploadId);
+    $clearingDecWithLicenses = $clearingDao->getFileClearings($itemTreeBounds, $this->restHelper->getGroupId(), false, true);
+
+    $data = [];
+    $scope = new DecisionScopes();
+
+    foreach ($clearingDecWithLicenses as $clearingDecision) {
+      $removedLicenses = [];
+      $addedLicenses = [];
+
+      foreach ($clearingDecision->getClearingLicenses() as $lic) {
+        $shortName = $lic->getShortName();
+        $lic->isRemoved() ? $removedLicenses[] = $shortName : $addedLicenses[] = $shortName;
+      }
+      ksort($removedLicenses, SORT_STRING);
+      ksort($addedLicenses, SORT_STRING);
+      $obj =  new ClearingHistory(date('Y-m-d', $clearingDecision->getTimeStamp()), $clearingDecision->getUserName(), $scope->getTypeName($clearingDecision->getScope()), $this->decisionTypes->getConstantNameFromKey($clearingDecision->getType()), $addedLicenses, $removedLicenses);
+      $data[] = $obj->getArray();
+    }
+    return $response->withJson($data, 200);
   }
 }
--- a/src/www/ui/api/Models/ClearingHistory.php
+++ b/src/www/ui/api/Models/ClearingHistory.php
@@ -1,0 +1,180 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © 2023 Samuel Dushimimana <dushsam100@gmail.com>
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+/**
+ * @file
+ * @brief Info model
+ */
+namespace Fossology\UI\Api\Models;
+
+/**
+ * @class Info
+ * @brief Info model to contain general error and return values
+ */
+class ClearingHistory
+{
+  /**
+   * @var string $date
+   * HTTP response code
+   */
+  private $date;
+    /**
+     * @var string $username
+     * Reponse message
+     */
+  private $username;
+    /**
+     * @var string $scope
+     * Response type
+     */
+  private $scope;
+    /**
+     * @var string $type
+     * Response type
+     */
+  private $type;
+
+  /**
+   * @var array $addedLicenses
+   * Response type
+   */
+  private $addedLicenses;
+  /**
+   * @var array $removedLicenses
+   * Response type
+   */
+  private $removedLicenses;
+
+  /**
+   * @param string $date
+   * @param string $username
+   * @param string $scope
+   * @param string $type
+   * @param array $addedLicenses
+   * @param array $removedLicenses
+   */
+  public function __construct($date, $username, $scope, $type, $addedLicenses, $removedLicenses)
+  {
+    $this->date = $date;
+    $this->username = $username;
+    $this->scope = $scope;
+    $this->type = $type;
+    $this->addedLicenses = $addedLicenses;
+    $this->removedLicenses = $removedLicenses;
+  }
+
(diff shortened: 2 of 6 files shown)


src/www/ui/api/Controllers/UploadTreeController.php
```diff
@@ -0,0 +1,99 @@
...
+  /** @var HighlightDao */
+  public $highlightDao;
+
+  /** @var DecisionTypes */
+  private $decisionTypes;
+
+  public function __construct($container)
+  {
+    parent::__construct($container);
+    $this->container = $container;
```
GMishx, line 48: Already done in parent class. Please remove.

src/www/ui/api/Controllers/UploadTreeController.php
```diff
@@ -0,0 +1,99 @@
...
+    $this->clearingDao = $this->container->get('dao.clearing');
+    $this->licenseDao = $this->container->get('dao.license');
+    $this->highlightDao = $container->get("dao.highlight");
+    $this->decisionTypes = $container->get('decision.types');
```
GMishx, line 52:
```suggestion
    $this->highlightDao = $this->container->get("dao.highlight");
    $this->decisionTypes = $this->container->get('decision.types');
```

src/www/ui/api/Controllers/UploadTreeController.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    if ($returnVal !== null) {
+      return $response->withJson($returnVal->getArray(), $returnVal->getCode());
+    }
+
+    $itemTreeBounds = $uploadDao->getItemTreeBoundsFromUploadId($itemId, $uploadId);
+    $clearingDecWithLicenses = $this->clearingDao->getFileClearings($itemTreeBounds, $this->restHelper->getGroupId(), false, true);
+
+    $data = array();
```
GMishx, line 75: Please use new format to create arrays:
```suggestion
    $data = [];
```

src/www/ui/api/Controllers/UploadTreeController.php
```diff
@@ -0,0 +1,99 @@
...
+    $scope = new DecisionScopes();
+
+    foreach ($clearingDecWithLicenses as $clearingDecision) {
+      $removedLicenses = array();
+      $addedLicenses = array();
```
GMishx, line 80: Please update the array creation syntax.

src/www/ui/api/Controllers/UploadTreeController.php
```diff
@@ -0,0 +1,99 @@
...
+      ksort($removedLicenses, SORT_STRING);
+      ksort($addedLicenses, SORT_STRING);
+      $data[] = array(
+        'date' => date('Y-m-d', $clearingDecision->getTimeStamp()),
+        'username' => $clearingDecision->getUserName(),
+        'scope' => $scope->getTypeName($clearingDecision->getScope()),
+        'type' => $this->decisionTypes->getTypeName($clearingDecision->getType()),
+        'addedLicenses' => $addedLicenses,
+        'removedLicenses' => $removedLicenses,
+      );;
```
GMishx, line 95: Would suggest to create a new model for this complex data. It will help to make sure the response format is consistent with help of test cases.

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -7,7 +7,7 @@ openapi: 3.0.2
 info:
   title: FOSSology API
   description: Automate your fossology instance using REST API
-  version: 1.5.1
+  version: 1.5.2
```
GMishx, line 10: As discussed, please avoid updating version in each PR.

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -3125,6 +3168,38 @@ components:
           type: string
           example:
             "Main group"
+    GetClearingHistory:
+      type: object
+      properties:
+        date:
+          type: string
+          format: date-time
```
GMishx, line 3176: Correct format should be `date`. `date-time` is `YYYY-mm-ddTHH:MM:SS`

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -3125,6 +3168,38 @@ components:
...
+          description: Date of the clearing history
+          example: "2023-06-15"
+        username:
+          type: string
+          description: Username of the user who cleared the package
```
GMishx, line 3181:
```suggestion
          description: Username of the user who created the decision.
```

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -3125,6 +3168,38 @@ components:
...
+          example: "fossy"
+        scope:
+          type: string
+          description: Scope of the clearing
+          example: "local"
```
GMishx, line 3186: Can add enum with "local", "package" and "global" values. (as per `\Fossology\Lib\Data\DecisionScopes::__construct()`)

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -3125,6 +3168,38 @@ components:
...
+        type:
+          type: string
+          description: Type of the clearing
+          example: "Identified"
```
GMishx, line 3190: Would recommend adding enum from `\Fossology\Lib\Data\DecisionTypes::__construct()` for possible values.

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -3125,6 +3168,38 @@ components:
...
+        addedLicenses:
+          type: array
+          description: List of licenses added to the package
```
GMishx, line 3193:
```suggestion
          description: List of license shortnames added to the decision
```

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -3125,6 +3168,38 @@ components:
...
+          items:
+            type: string
+            example: "MIT"
+        removedLicenses:
+          type: array
+          description: List of licenses removed from the package
```
GMishx, line 3199:
```suggestion
          description: List of license shortnames removed from the decision
```

GMishx (changes requested):

GMishx:
Tested, working as expected.

src/www/ui/api/documentation/openapi.yaml
dushimsam: Updated the changes.

src/www/ui/api/Controllers/UploadTreeController.php
```diff
@@ -138,4 +140,60 @@ public function setClearingDecision($request, $response, $args)
...
+    $clearingDao = $this->container->get('dao.clearing');
+
+    $returnVal = null;
+    $decisionKeysMap = [
+      3 => "TO_BE_DISCUSSED",
+      4 => "IRRELEVANT",
+      5 => "IDENTIFIED",
+      6 => "DO_NOT_USE",
+      7 => "NON_FUNCTIONAL"
+    ];
```
GMishx, line 166: Can the object of class `\Fossology\Lib\Data\DecisionTypes` be used instead and use it's function `getTypeName($type)` to get the display name? It will reduce the duplication of name storage arrays.

src/www/ui/api/Models/ClearingHistory.php
```diff
@@ -0,0 +1,182 @@
...
+ SPDX-FileCopyrightText: © 2023 Samuel Dushimimana <dushsam100@gmail.com>
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+/**
+ * @file
+ * @brief Info model
+ */
+namespace Fossology\UI\Api\Models;
+
+use EasyRdf\Literal\Date;
```
GMishx, line 12: This class is not needed here.

src/www/ui/api/Models/ClearingHistory.php
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * @class Info
+ * @brief Info model to contain general error and return values
+ */
+class ClearingHistory
+{
+  /**
+   * @var Date $date
```
GMishx, line 21: Seeing the constructor call and other usage, it seems like the data type of `$date` should be `string` instead. (`date()` returns a string).

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -3363,6 +3406,46 @@ components:
           type: string
           example:
             "Main group"
+    GetClearingHistory:
+      type: object
+      properties:
+        date:
+          type: string
+          format: YYYY-mm-ddTHH:MM:SS
```
GMishx, line 3414: That is not what I meant :-)
```suggestion
          format: date
```

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -3363,6 +3406,46 @@ components:
...
+            - global
+        type:
+          type: string
+          description: Type of the clearing
+          enum:
+            - TO_BE_DISCUSSED
+            - IRRELEVANT
+            - IDENTIFIED
+            - DO_NOT_USE
+            - NON_FUNCTIONAL
```
GMishx, line 3436: I think this will update as well according to

https://github.com/fossology/fossology/blob/b42626c170f7eac68de7e246683646b2aa00afbf/src/lib/php/Data/DecisionTypes.php#L23-L29

GMishx (changes requested):

src/www/ui/api/Controllers/UploadTreeController.php
dushimsam: I agree, but I have a different perspective. I think returning constants is a better option. This way, the UI can map them to the appropriate display texts for each type instead of deciding on the backend what the UI should display.  Also, the UI comparison process will be simpler, if we want to customize the display for a specific type, comparing with the value `TO_BE_DISCUSSED` is easier than when we have `To be discussed` as a value.

dushimsam: Let me know what you think about this ? @GMishx

src/www/ui/api/documentation/openapi.yaml
dushimsam: Updated, please let me know if there's anything remaining @GMishx

GMishx (approved):
Changes looks good.

Issue #2467, opened 2023-06-13
License Browser - Scanner Licenses


soham4abc:
Tree view endpoint With Folder & file {itemid + type}, Scanner licenses, edited licenses, count and filters.

soham4abc:
I want to work on this issue... can anyone please assign me..

Thank you

dushimsam:
Hello, I can take over this issue.

PR #2499, opened 2023-06-26
feat(api): Update Upload Summary API to add more info

dushimsam:
## Description

Update the current API to for getting the upload's summary to add more information related to licenses.

### Changes

1. Updated the method `getUploadSummary` in  `UploadController` to make the changes.
2. Updated the `openapi.yaml` file  to introduce a new changes for the API.

## How to test

Make a GET request on the endpoint: `/uploads/{id}/summary`.

## Screenshots

image

### Related Issue:

Fixes [#2467](https://github.com/fossology/fossology/issues/2467)

cc: @shaheemazmalmmd @GMishx

### DIFF
--- a/src/www/ui/api/Controllers/UploadController.php
+++ b/src/www/ui/api/Controllers/UploadController.php
@@ -305,17 +305,40 @@ class UploadController extends RestController
   public function getUploadSummary($request, $response, $args)
   {
     $id = intval($args['id']);
+    $query = $request->getQueryParams();
+    $selectedAgentId = $query['agentId'];
+    $agentDao = $this->container->get('dao.agent');
     $upload = $this->uploadAccessible($this->restHelper->getGroupId(), $id);
     if ($upload !== true) {
       return $response->withJson($upload->getArray(), $upload->getCode());
+    } else if ($selectedAgentId !== null && !$this->dbHelper->doesIdExist("agent", "agent_pk", $selectedAgentId)) {
+      $error = new Info(404, "Agent does not exist", InfoType::ERROR);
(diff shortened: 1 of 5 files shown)


GMishx:
Would recommend extending existing endpoint `/uploads/{id}/summary` as it already contains most of the required information.
https://github.com/fossology/fossology/blob/2c8522e142abea18c7db44013723276284612df9/src/www/ui/api/documentation/openapi.yaml#L592

Existing endpoint of left, new endpoint on the right.
image

dushimsam:
> Would recommend extending existing endpoint `/uploads/{id}/summary` as it already contains most of the required information.
> 
> https://github.com/fossology/fossology/blob/2c8522e142abea18c7db44013723276284612df9/src/www/ui/api/documentation/openapi.yaml#L592
> 
> Existing endpoint of left, new endpoint on the right. image

Yes, that looks better.

GMishx:
Oh, I didn't notice the endpoint is dependent on agent id. But still, they can be merged if the agent id is not provided, the current behavior can be followed. Otherwise, the values will change based on agent id.

It is better to send little more information from a single endpoint than creating one more which shares the same info. Creating new endpoint will also cause confusion if more than 1 endpoint can be used to get same info (and will also need multiple calls if info from both endpoint is needed :-))

dushimsam:
Awesome, I have applied the new changes to the current implementation and have revised the commit and the PR's description.
Please let me know if there is anything else I need to modify.

Thank you very much.

GMishx (approved):
Changes looks good.
Tested, working as expected.

PR #2521, opened 2023-07-05
fix(api): fix linter issues

GMishx:
## Description

Fix the linter issues in API. Check https://github.com/fossology/fossology/runs/14784207342

### Changes

Fix linter issues.

## How to test

Check GitHub Actions.


ag4ums (approved):
Chnages looks good.

PR #2567, opened 2023-08-10
refactor(browse): use join to fetch data from two tables

shaheemazmalmmd:
## Description

Refactor two queries.

### Changes

In browse view and in folder move or copy view.

## How to test

* Check if all the main licenses load on browse view .
* Delete / Add main licenses and check.
* Check if all the contents of a folder are visible in Organize > Folders > Move or Copy
* Move the contents and check if they are copied perfectly.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2567"></a>


GMishx (approved):
In Postgres, `SELECT (*) FROM tab1, tab2 WHERE tab1.col = tab2.col` implicitly means a `JOIN .. ON`.

PR #2569, opened 2023-08-11
perf(scancode):Improve scanning speed of scancode agent

its-sushant:
## Description

Merging these changes will improve the scanning speed of scancode agent

### Changes

- Scanning is performed all at once for every file.
- Once the scanning is complete, the results are stored in a temporary JSON file.
- Subsequently, these results are pushed to the database.

cc: @GMishx @shaheemazmalmmd @avinal 

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2569"></a>


shaheemazmalmmd (approved):
Tested works as expected.

src/scancode/agent/scancode_wrapper.cc
```diff
@@ -62,42 +62,28 @@ unsigned getFilePointer(const string &filename, size_t start_line,
...
 
   string command =
-      "PYTHONPATH='/home/" + projectUser + "/pythondeps/' " +
-      "SCANCODE_CACHE=" + cacheDir + "/scancode " + // Use fossology's cache
-      "/home/" + projectUser + "/pythondeps/bin/scancode -" +
-      state.getCliOptions() +
-      " --custom-output - --custom-template scancode_template.html " +
-      file.getFileName() + " --quiet " +
-      ((state.getCliOptions().find('l') != string::npos) ? " --license-text --license-score " +
-      to_string(MINSCORE): "");
```
GMishx, line 82: Please check if the API or `runscanonfile.py` script can make use of this `MINSCORE` which defines a minimum threshold of scancode confidence above which only the result will be stored in DB.

src/scancode/agent/scancode_wrapper.hpp
```diff
@@ -27,8 +27,7 @@
 
 using namespace std;
 
-string scanFileWithScancode(const State& state, const fo::File& file);
-string scanFileWithScancode(string filename);
+void scanFileWithScancode(const State& state, string file_location, string outputFile);
```
GMishx, line 30: What will happen if someone now tries to run scancode agent from CLI like `/usr/local/share/fossology/scancode/agent/scancode file_to_scan.py`? Have you covered it in `main()` somehow?

src/scancode/agent/CMakeLists.txt
```diff
@@ -58,3 +58,8 @@ install(FILES
     scancode_template.html
     DESTINATION ${FO_MODDIR}/${PROJECT_NAME}/agent
     COMPONENT scancode)
+
+install(FILES
+    runscanonfiles.py
+    DESTINATION ${FO_MODDIR}/${PROJECT_NAME}/agent
+    COMPONENT scancode)
```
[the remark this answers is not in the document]
GMishx: Now I see from the changes bellow, `scancode_template.html` is no longer needed and can be removed from the repository?

GMishx (changes requested):

src/scancode/agent/CMakeLists.txt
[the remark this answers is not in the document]
its-sushant: Yup, will remove it.

src/scancode/agent/scancode_wrapper.cc
its-sushant: Yes, we can also paas `MINSCORE` to API.

its-sushant: Will make changes accordingly.

src/scancode/agent/runscanonfiles.py
```diff
@@ -0,0 +1,193 @@
...
+
+if __name__ == "__main__":
+  parser = argparse.ArgumentParser(description="Process a file specified by its location.")
+  parser.add_argument("-c", "--scan-copyrights", action="store_true", help="Scan for copyrights")
+  parser.add_argument("-l", "--scan-licenses", action="store_true", help="Scan for licenses")
+  parser.add_argument("-e", "--scan-emails", action="store_true", help="Scan for emails")
+  parser.add_argument("-u", "--scan-urls", action="store_true", help="Scan for urls")
+  parser.add_argument("min_score", type=int, default=0, help="Minimum score for a license to be considered valid")
+  parser.add_argument('file_location', type=str, help='Path to the file you want to process')
+  parser.add_argument('outputFile', type=str, help='Path to the file you want save results to')
```
GMishx, line 182: Thanks for incorporating the changes. But making `min_score` a positional argument might be problematic if the user is not interested in license scan. In that case, FOSSology will call something like `runscanonfiles.py -ceu input.txt output.json`

In such case, `min_score` will become `input.txt`.

Is my assumption right or since you've provided `min_score` as `type=int`, it will get the default value of `0` ?

src/scancode/agent/scancode_wrapper.hpp
GMishx: I see now the function `string scanFileWithScancode(string filename);` was never implemented and the CLI scan was not supported. Thus, it is not in the scope of this PR.
If you are interested to work on it, I will request to open an issue and work on a separate PR.

src/scancode/agent/runscanonfiles.py
its-sushant: Yes, making it positional will be problematic if user is not interested in license scan.

its-sushant: For that we can have flag for min_score, then there should not be any problem.

GMishx (commented):
image

While scanning a package, I notice everything was done in a single thread. Do you think making the loop in `runscanonfiles.py::process_files()` a multi-threaded loop will improve the scan speed?

GMishx (approved):
Tested, working as expected.

chacabuco:
Hello! 
@its-sushant I've been running a scancode job REST POST request on a repo/folder that has already been analyzed. Thus far, I am able to see that the request gets received by the web-api, yet no job gets triggered. Is my understanding right, that with this change, if no file has been modified, and the analysis already exist, no job will be triggered?

PR #2576, opened 2023-08-22
feat(SETUP-V2): Support Multiple Versions (V1 & V2)

### LINKED ISSUE #2575: Enable Version 2 REST APIs for FOSSology
### Description

Currently, the FOSSology API is designed to support only version 1 of the REST APIs. However, in order to introduce and utilize the new features offered by version 2 of the API, we need to make the necessary modifications to allow the inclusion and usage of the API version 2.

### Objective

The main objective of this issue is to update the FOSSology API setup to accommodate the introduction of version 2 REST APIs. This will involve the following tasks:
 
1. **Allow Version 2 Routes:** Modify the API configuration to allow the routing of requests to version 2 of the APIs. This includes defining the appropriate routes and handling mechanisms.

2. **Version Handling:** Implement a mechanism that automatically serves version 1 APIs by default when the user has not explicitly specified an API version.

#### Related issue
[#2404](https://github.com/fossology/fossology/issues/2404#issuecomment-1671046653)

@GMishx @shaheemazmalmmd

dushimsam:
## Description

This pull request (PR) focuses on enhancing the functionality of the API configuration to accommodate the routes of Version 2.

#### 1. Dynamic Base URL Support:

The i`index.php` file has been revised to facilitate the dynamic base URL, capable of accommodating both V1 and V2 versions of the API. 

#### 2. OpenAPI Specification Update:

The `openapi.yml` file has been updated to incorporate the server URL specific to Version 2 of the API. 

## How to test

**1. Add this block of code in the `getInfo` method from `InfoController`:**
  
image

**2. Call the Version 1 API** 

image

**3. Call the Version 2 API**

image

**4. Other Cases**

image

### Related Issue:
Fixes #2575 
    
cc: @shaheemazmalmmd @GMishx

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2576"></a>

### DIFF
--- a/src/www/ui/api/Models/ApiVersion.php
+++ b/src/www/ui/api/Models/ApiVersion.php
@@ -1,0 +1,21 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © 2023 Samuel Dushimimana <dushsam100@gmail.com>
+
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+/**
+ * @file
+ * @brief ApiVersion enum
+ */
+namespace Fossology\UI\Api\Models;
+
+/**
+ * @class ApiVersion
+ * @brief ApiVersion enum
+ */
+class ApiVersion
+{
+  const V1 = 1;
+  const V2 = 2;
+}
--- a/src/www/ui/api/index.php
+++ b/src/www/ui/api/index.php
@@ -2,6 +2,7 @@
 /*
  SPDX-FileCopyrightText: © 2017-2018,2021 Siemens AG
  SPDX-FileCopyrightText: © 2021 Orange by Piotr Pszczola <piotr.pszczola@orange.com>
+ SPDX-FileCopyrightText: © 2023 Samuel Dushimimana <dushsam100@gmail.com>
 
  SPDX-License-Identifier: GPL-2.0-only
 */
@@ -43,6 +44,7 @@ use Fossology\UI\Api\Helper\ResponseFactoryHelper;
 use Fossology\UI\Api\Helper\ResponseHelper;
 use Fossology\UI\Api\Middlewares\FossologyInitMiddleware;
 use Fossology\UI\Api\Middlewares\RestAuthMiddleware;
+use Fossology\UI\Api\Models\ApiVersion;
 use Fossology\UI\Api\Models\Info;
 use Fossology\UI\Api\Models\InfoType;
 use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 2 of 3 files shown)


src/www/ui/api/index.php
```diff
@@ -95,7 +109,16 @@
...
+$app->setBasePath($BASE_PATH);
+
+// Custom middleware to set the API version as a request attribute
+$apiVersionMiddleware = function (Request $request, $handler) use ($apiVersion) {
+  $request = $request->withAttribute('apiVersion', $apiVersion);
+  return $handler->handle($request);
+};
+
+// Attach the middleware globally to be applied to all routes
+$app->add($apiVersionMiddleware);
```
GMishx, line 121: Please move the middleware addition after the comment bellow (just before the `$dbConnected` check). And also update the documentation in the comment to write about this new middleware.

src/www/ui/api/Models/ApiVersion.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Fossology\UI\Api\Models;
+
+/**
+ * @class ApiVersion
+ * @brief ApiVersion enum
+ */
+class ApiVersion
+{
+  const V1 = 'v1';
+  const V2 = 'v2';
```
GMishx, line 20: Would request to have the versions as integers. It will make the system future proof and allows checking versions like bellow
```php
if ($request->getAttribute('apiVersion') > ApiVersion::V1) {
  // do something new
}
```

src/www/ui/api/index.php
```diff
@@ -51,16 +53,28 @@
...
+use Slim\Psr7\Request;
 use Slim\Psr7\Response;
 use Throwable;
 
-const REST_VERSION = "1";
+// Extracts the version from the URL
+function getVersionFromUri ($uri)
+{
+  $matches = [];
+  preg_match('/\/repo\/api\/(v\d+)/', $uri, $matches);
```
GMishx, line 64: Can just extract the version number here to help making `ApiVersion` enum to use integers instead of strings
```suggestion
  preg_match('/\/repo\/api\/v(\d+)/', $uri, $matches);
```

GMishx (changes requested):

src/www/ui/api/Models/ApiVersion.php
dushimsam: Noted

src/www/ui/api/index.php
dushimsam: Done @GMishx

GMishx (approved):

GMishx (approved):
Tested, working as expected.

PR #2577, opened 2023-08-23
chore(UploadTreeProxy): optimize license file query

GMishx:
## Description

Replace the use of JOINs with `license_ref` and `license_candidate` by getting license ID of 'No_license_found' and 'Void' licenses and filter the IDs directly.

Also, no agent currently scans with candidate license and thus, there should be no candidate license ID in `license_file` table. Thus, we can remove join with `license_candidate` as well.

### Changes

Removing following join:
```
->  Nested Loop Left Join  (cost=0.85..5.82 rows=1 width=0) (actual time=0.008..0.008 rows=1 loops=3908)
      Filter: ((lr.rf_shortname <> ALL ('{No_license_found,Void}'::text[])) OR (lr.rf_pk IS NULL))
      Rows Removed by Filter: 1
      ->  Index Only Scan using lf_pfile_agent_rf_lf_idx on license_file lf  (cost=0.57..3.31 rows=1 width=8) (actual time=0.006..0.006 rows=1 loops=3908)
            Index Cond: (pfile_fk = ut.pfile_fk)
            Filter: (agent_fk = ANY ('{1897,1900}'::integer[]))
            Rows Removed by Filter: 19
            Heap Fetches: 35
      ->  Index Scan using rf_pkpk on license_ref lr  (cost=0.28..2.50 rows=1 width=25) (actual time=0.001..0.001 rows=1 loops=5711)
            Index Cond: (rf_pk = lf.rf_fk)
```

## How to test

Load the License Browser view for a large component and notice the difference in loading speed.

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2577"></a>

### DIFF
--- a/src/lib/php/Proxy/UploadTreeProxy.php
+++ b/src/lib/php/Proxy/UploadTreeProxy.php
@@ -8,6 +8,8 @@
 namespace Fossology\Lib\Proxy;
 
 use Fossology\Lib\BusinessRules\LicenseMap;
+use Fossology\Lib\Dao\LicenseDao;
+use Fossology\Lib\Dao\UploadDao;
 use Fossology\Lib\Data\AgentRef;
 use Fossology\Lib\Data\DecisionScopes;
 use Fossology\Lib\Data\DecisionTypes;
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (commented):

PR #2590, opened 2023-09-27
feat(schedule agent): add select2 to search for uploads with name

shaheemazmalmmd:
## Description

Use select2 to search for uploads in  jobs > schedule agents

## How to test

* Install fossology.
* go to   jobs > schedule agents
* select upload using select 2
* schedule agent and check.


GMishx (changes requested):
The select2 search goes away if the folder is changed and the list is refreshed.

GMishx (approved):
Working as expected.

PR #2591, opened 2023-09-27
feat(conf): support CLIXML conf for a upload

shaheemazmalmmd:
## Description

support clixml conf for a upload in fossology.

## How to test

* For a upload make some changes in conf page for CLIXML.
* Generate report and test if the confs reflects or not.

### DIFF
--- a/src/clixml/agent/template/clixml-file.xml.twig
+++ b/src/clixml/agent/template/clixml-file.xml.twig
@@ -82,18 +82,22 @@
     <Content><![CDATA[{{ ids.text }}
 ]]></Content>
     <Files><![CDATA[
-    {%- for id in ids.files %}
-      {%- if id %}
-        {{- id -}}
-      {%- endif ~%}
-    {%~ endfor %}
+    {%- if fileOperations.licensepath ~%}
+      {%- for id in ids.files %}
+        {%- if id %}
+          {{- id -}}
+        {%- endif ~%}
+      {%~ endfor %}
+    {%- endif ~%}
 ]]></Files>
     <FileHash><![CDATA[
-    {%- for id in ids.hash %}
-      {%- if id %}
-        {{- id -}}
-      {%- endif ~%}
-    {%~ endfor %}
+    {%- if fileOperations.licensehash ~%}
+      {%- for id in ids.hash %}
+        {%- if id %}
+          {{- id -}}
+        {%- endif ~%}
+      {%~ endfor %}
+    {%- endif ~%}
 ]]></FileHash>
         {%- if ids.acknowledgement is not empty ~%}
     <Acknowledgements><![CDATA[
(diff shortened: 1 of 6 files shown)


src/clixml/agent/template/clixml-file.xml.twig
```diff
@@ -82,18 +82,22 @@
     <Content><![CDATA[{{ ids.text }}
 ]]></Content>
     <Files><![CDATA[
-    {%- for id in ids.files %}
-      {%- if id %}
-        {{- id -}}
-      {%- endif ~%}
-    {%~ endfor %}
+    {%- if fileOperations.licensepath ~%}
```
GMishx, line 85: If `licensepath` is not provided, shouldn't the entire `<Files></Files>` tag should be removed?

GMishx (changes requested):
A question.
I checked, the specification does not define the minimum count for the tags `<Files>` and `<FileHash>`.

src/clixml/agent/template/clixml-file.xml.twig
shaheemazmalmmd: kept this just to indicate the specification flow,

GMishx: In my opinion, if the tag does not contain any value, it will be better to remove the tag as well to make the file even smaller. If the specification, however, does not allow it, then we can proceed with the same.

shaheemazmalmmd: Agreed let me do the changes.

GMishx (approved):
Changes looks good.

GMishx:
image

1. Since there are no headings in CLIXML reports, can the headings be simple text instead on a text box?
2. I cannot disable all the fields, see the screenshot. Is it expected?
3. Looks like I was wrong, the fields `<Files>` and `<FileHash>` in `<License>` and `<Acknowledgements>` are required.
4. As per the specification, there should be at least `<License>` and `<Obligation>` inside `<ComponentLicenseInformation>`. Thus, cannot disable all Obligations.

You can use following tool to validate the XML against the specification:
```shell
xmllint --schema ~/path/to/CLI_V1.6.xsd --noout ~/Downloads/CLIXML_FOSSology.xml
```

GMishx:
Also @shaheemazmalmmd , do you think it won't be confusing to add CLIXML settings in Unified Report settings tab?

snap

shaheemazmalmmd:
> Also @shaheemazmalmmd , do you think it won't be confusing to add CLIXML settings in Unified Report settings tab?
> 
> snap

Let me change it to a general name.

GMishx:
image

It is confusing to have "Main Licenses", "Licenses" in the list even though they cannot be disabled entirely, just the filepath and filehash.

GMishx (approved):
Tested, working as expected.

PR #2597, opened 2023-10-10
feat(bulk): checkbox to select scan for findings only

shaheemazmalmmd:
## Description

New option for bulk scan instead of scanning whole package you can choose to scan files where a finding exists.

### Changes

Both UI and backend has been changed.

## How to test

* scan a package and check if this option works.

### DIFF
--- a/src/monk/agent/database.c
+++ b/src/monk/agent/database.c
@@ -14,16 +14,33 @@
 
 PGresult* queryFileIdsForUploadAndLimits(fo_dbManager* dbManager, int uploadId,
                                          long left, long right, long groupId,
-                                         bool ignoreIrre) {
+                                         bool ignoreIrre, bool scanFindings) {
   char* tablename = getUploadTreeTableName(dbManager, uploadId);
   gchar* stmt;
   gchar* sql;
   PGresult* result;
-  if (!ignoreIrre)
+
+  char* distinctPfile = "SELECT DISTINCT pfile_fk FROM %s"
+                        " WHERE upload_fk = $1 AND (ufile_mode&x'3C000000'::int) = 0 "
+                        " AND (lft BETWEEN $2 AND $3) AND pfile_fk != 0";
+  char* distinctPfileNoDec = "SELECT DISTINCT pfile_fk FROM ("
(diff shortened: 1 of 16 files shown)


src/monk/agent/database.c
```diff
@@ -34,6 +34,48 @@ PGresult* queryFileIdsForUploadAndLimits(fo_dbManager* dbManager, int uploadId,
       uploadId, left, right
     );
   }
+  else if(!ignoreIrre && scanFindings)
+  {
+    sql = g_strdup_printf("WITH allPfileData AS (SELECT DISTINCT pfile_fk FROM %s"
+      " WHERE upload_fk = $1 AND (ufile_mode&x'3C000000'::int) = 0 "
+      " AND (lft BETWEEN $2 AND $3) AND pfile_fk != 0) SELECT pfile_fk FROM allPfileData"
+      " WHERE pfile_fk NOT IN (SELECT pfile_fk FROM license_file WHERE rf_fk IN"
+      " (SELECT rf_pk FROM license_ref WHERE rf_shortname IN ('No_license_found', 'Void')));", tablename);
```
GMishx, line 43: SELECT can be optimized here
```suggestion
      " (SELECT rf_pk FROM license_ref WHERE rf_shortname = ANY(VALUES('No_license_found'), ('Void'))));", tablename);
```

GMishx (changes requested):

GMishx (approved):
Changes looks good.

src/www/ui/template/ui-clearing-view_bulk.html.twig
```diff
@@ -135,6 +135,15 @@
                     </label>
                   </div>
                 </div>
+                <div class="col-sm-auto pr-0">
+                  <div class="form-check form-check-inline">
+                    <input type="checkbox" class="form-check-input" id="scanOnlyFindings"/>
```
GMishx, line 140: Missing the `name` attribute causes this flag to be not passed with the form request.
```suggestion
                    <input type="checkbox" class="form-check-input" id="scanOnlyFindings" name="scanOnlyFindings" />
```

GMishx (changes requested):
Other tests looks good.

GMishx (approved):
Tested, working as expected.

PR #2598, opened 2023-10-10
refactor(api): introduce error handling

GMishx:
## Description

Use custom exceptions for handling HTTP error status codes instead of writing long if-else ladders. This makes code cleaner and easier to maintain.

At the same time, make small changes to make code more robust.

### Changes

Create new exceptions for REST API and throw them in Controllers instead of writing long if-else ladders.

## How to test

Test REST endpoints.


shaheemazmalmmd (approved):
Code looks good. tested with few endpoints seems good.

PR #2599, opened 2023-10-11
fix(PSR-12): closing ?> tag MUST be omitted

vargenau:
## Description

PSR-12:

All PHP files MUST end with a non-blank line, terminated with a single LF.

The closing ?> tag MUST be omitted from files containing only PHP.

### Changes

Remove closing ?> tag


GMishx (approved):
LGTM.
Thank you for your contribution.

PR #2600, opened 2023-10-11
fix(php) Optional parameter declared before required parameter

vargenau:
## Description

You cannot have an optional parameter declared before a required parameter

### Changes

Remove default value


GMishx (approved):
Changes looks good.

PR #2601, opened 2023-10-12
fix(php): Using ${var} in strings is deprecated

vargenau:
## Description

Using ${var} in strings is deprecated

### Changes

Use {$var} instead

### DIFF
--- a/src/lib/php/common-active.php
+++ b/src/lib/php/common-active.php
@@ -98,7 +98,7 @@ function ActiveHTTPscript($RequestName,$IncludeScriptTags=1)
 
   $HTML .= "var $RequestName=null;\n";
   /* Check for browser support. */
-  $HTML .= "function ${RequestName}_Get(Url)\n";
+  $HTML .= "function {$RequestName}_Get(Url)\n";
   $HTML .= "{\n";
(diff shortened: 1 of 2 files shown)


GMishx:
@vargenau , can you please provide some documentation about this deprecation?

Also, there should ideally be a `:` after the scope in commit message :smiley:  (`fix(php): using...`)

vargenau:
See: https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated

```
php --version
PHP 8.2.11 (cli) (built: Sep 28 2023 16:20:53) (NTS)
```

```
php -l src/lib/php/common-active.php
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in src/lib/php/common-active.php on line 99

Deprecated: Using ${var} in strings is deprecated, use {$var} instead in src/lib/php/common-active.php on line 99
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in src/lib/php/common-active.php on line 113

Deprecated: Using ${var} in strings is deprecated, use {$var} instead in src/lib/php/common-active.php on line 113
```

GMishx (approved):
Changes looks good.

PR #2603, opened 2023-10-12
fix(php): string offset access with curly braces deprecated in PHP 7.4

vargenau:
## Description

Array and string offset access syntax with curly braces is deprecated in PHP 7.4. 

https://wiki.php.net/rfc/deprecate_curly_braces_array_access

### Changes

Use square brackets instead.

### DIFF
--- a/src/nomos/agent_tests/testdata/NomosTestfiles/AGPL/JsonMapper.php
+++ b/src/nomos/agent_tests/testdata/NomosTestfiles/AGPL/JsonMapper.php
@@ -106,7 +106,7 @@ class JsonMapper
         continue;
       }
 
-      if ($type{0} != '\\') {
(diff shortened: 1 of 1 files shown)


GMishx:
Thanks for your contribution, but the file `src/nomos/agent_tests/testdata/NomosTestfiles/AGPL/JsonMapper.php` is never executed. It is just a test file for nomos to check license.
Thus, it is not a good idea to modify the file as it can result in unnecessary identifications on nomos test.

You can make use of `src/testing/syntax/syntaxtest.sh` to run PHP syntax check on all source files.

If you agree with the points, can we close the PR?

vargenau:
OK to close the PR.

GMishx:
Closing this PR as discussed.

PR #2604, opened 2023-10-12
fix(php): add missing semicolon

vargenau:
## Description

PHP 8.2 complains about a syntax error (but PHP 7 does not).

### Changes

Add missing semicolon at end of file.


GMishx:
As discussed in https://github.com/fossology/fossology/pull/2603#issuecomment-1763706713, the modified file is never executed.

If you agree with the point, requesting you to close this PR.

vargenau:
@GMishx 
Understood, I will stop modifying the nomos files.
But for this one, I would merge the PR as this file was already modified by https://github.com/fossology/fossology/pull/2599

GMishx:
Unfortunately, this PR is also modifying the licensing statement which is problematic for nomos regression test case. Thus it cannot be merged as is.

If you remove the change on licensing and leave the semicolon change, we can proceed.

Issue #2584, opened 2023-09-14
Removed PHP functions


vargenau:
### Description

The documentation says "The PHP versions 7.x are supported to work for FOSSology".

However, the following functions are still present in the code:
- split: removed in PHP 7.0
- ereg, eregi: removed in PHP 7.0

GMishx:
We need to update the readme as we currently support PHP 7.3+

vargenau:
See https://github.com/fossology/fossology/pull/2628

shaheemazmalmmd:
https://github.com/fossology/fossology/pull/2628/files solves the issue. hence closing.

PR #2606, opened 2023-10-12
fix(php): replace deprecated split by explode

vargenau:
## Description

Function ```split``` was DEPRECATED in PHP 5.3.0, and REMOVED in PHP 7.0.0.

### Changes

Replace ```split``` by ```explode```.

Partial fix for #2584

### DIFF
--- a/src/testing/lib/TestRun.php
+++ b/src/testing/lib/TestRun.php
@@ -136,7 +136,7 @@ class TestRun
 		{
 		  return(NULL);
 		}
-		$parts = split(' ', $psLast);
+		$parts = explode(' ', $psLast);
 		//print "parts is:\n"; print_r($parts) . "\n";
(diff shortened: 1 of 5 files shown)


src/nomos/agent_tests/testdata/NomosTestfiles/UnclassifiedLicense/ldapbaseauthenticationprovider.inc.php
GMishx, line 1: As discussed in https://github.com/fossology/fossology/pull/2603#issuecomment-1763706713, the file is never executed and used as nomos test case.

Requesting to revert the changes in this file.

GMishx (changes requested):

GMishx (approved):
Changes looks good.

PR #2608, opened 2023-10-17
fix(spdx): de-duplicate licenses with same SPDX ID

GMishx:
## Description

If 2 or more licenses share the same SPDX ID, this commit de-duplicates them.
In the same process, change the data structure sent to twig files from arrays to `FileNode` and `SpdxLicenseInfo`.

### Changes

1. Create 2 new Data structures `FileNode` and `SpdxLicenseInfo`.
2. Use new data structures to send data to SPDX twig files instead of plain arrays.
3. De-Duplicate licenses to each get a unique IRI.
  1. First sort licenses by SPDX ID + shortname
  2. Check consecutive licenses if they have the same SPDX ID. If yes, then second license is duplicate.
  3. Change the license's shortname with `LicenseRef-<shortname>-<md5(text)>` to generate a unique ID.
  4. This new ID will then be used as IRI and `<licenseId>` will still point to the SPDX ID.

## How to test

1. Create variants of a license and give them all same SPDX ID (for example BSD-3-Clause).
2. Add these variants to different files.
3. In some of the files, update the license text at decision level.
4. Add other license clearing information like comments and acknowledgements.
5. Generate SPDX and DEP5 reports.
6. Validate the SPDX reports with https://tools.spdx.org/app/validate/ or https://github.com/spdx/tools-java

- [x] Todo: Fix CycloneDX report generator as common code was updated.

### DIFF
--- a/src/cyclonedx/agent/cyclonedx.php
+++ b/src/cyclonedx/agent/cyclonedx.php
@@ -22,12 +22,12 @@ namespace Fossology\CycloneDX;
 use Fossology\Lib\Agent\Agent;
 use Fossology\Lib\BusinessRules\LicenseMap;
 use Fossology\Lib\Dao\ClearingDao;
+use Fossology\Lib\Dao\LicenseDao;
 use Fossology\Lib\Dao\TreeDao;
 use Fossology\Lib\Dao\UploadDao;
 use Fossology\Lib\Data\AgentRef;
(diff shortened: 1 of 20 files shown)


shaheemazmalmmd (approved):
code looks good, needs test

shaheemazmalmmd (approved):

PR #2613, opened 2023-10-20
chore(ci): tag scanner image on release

nejch:
## Description

This adds an explicit tag to the `scanner` docker tag on releases, just like the main image, so that people can explicitly version it.

### Changes

Adds a release build similar to the main build:

https://github.com/fossology/fossology/blob/master/.github/workflows/docker-build.yml#L46-L66

The tag scheme is similar to kaniko's debug tags, which should be picked up by renovate and similar tools. See https://github.com/GoogleContainerTools/kaniko/releases/tag/v1.17.0 for example.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -148,3 +148,12 @@ jobs:
         labels: |
           org.opencontainers.image.version=${{ steps.get_release.outputs.tag_name }}.0
         context: .
+
+    - name: Build and push runner image
+      uses: docker/build-push-action@v3
+      with:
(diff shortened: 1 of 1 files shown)


GMishx (approved):

nejch:
Thanks for the merge @GMishx! I see the last release was in June. Are there plans to tag this in the near future so it gets deployed on Docker Hub? :bow: It'll give people a chance to version the scanner image in CI.

GMishx:
Yes, we are planning a release soon. A lot of new features have been integrated during this GSoC season.

PR #2617, opened 2023-10-26
feat(api): migrate `/tokens` endpoints to v2

GMishx:
## Description

Create a new `openapiv2.yaml` file to store the v2 related changes.
The idea is to keep using this file to store the changes related to only v2 migration like camelCasing changes, moving parameters from header to query, etc. Other changes like adding new endpoint need to be updated in both `openapi.yaml` and `openapiv2.yaml`.
Once every endpoint is migrated, the original file `openapi.yaml` can be replaced with this. This will allow to let everyone know v1 is deprecated from that point of time and v2 can be released with version `2.0.0`.

Update the `TokenRequest` component to use camelCasing and update endpoints accordingly.

## !Breaking
Remove the deprecated `/version` endpoint.

### Changes

1. Update `TokenRequest` component to use camelCasing.
2. Create new file `openapiv2.yaml`.
3. Remove deprecated `/version` endpoint.

## How to test

Test the `GET /api/tokens` and `POST /users/tokens` endpoints with both V1 and V2 to see if they are working as expected.

/cc @dushimsam

### DIFF
--- a/src/composer.json
+++ b/src/composer.json
@@ -18,6 +18,7 @@
         "ext-pgsql": "*",
         "ext-curl": "*",
         "ext-uuid": "*",
+        "ext-openssl": "*",
         "composer/spdx-licenses": "1.5.7",
         "firebase/php-jwt": "v6.3.0",
         "monolog/monolog" : "2.8.0",
--- a/src/composer.json.in
+++ b/src/composer.json.in
@@ -18,6 +18,7 @@
         "ext-pgsql": "*",
         "ext-curl": "*",
         "ext-uuid": "*",
+        "ext-openssl": "*",
         "composer/spdx-licenses": "1.5.7",
         "firebase/php-jwt": "v6.3.0",
         "monolog/monolog" : "2.8.0",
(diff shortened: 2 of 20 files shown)


GMishx:
/cc @deveaud-m @shaheemazmalmmd 

Does this strategy of migration from v1 to v2 makes sense or shall we look for an alternative?

dushsam2:
> /cc @deveaud-m @shaheemazmalmmd
> 
> Does this strategy of migration from v1 to v2 makes sense or shall we look for an alternative?

Thanks for introducing this change @GMishx I had been wondering how we can handle the two versions in the same file.

dushimsam:
Thanks for introducing this change @GMishx . I was also looking for a way to handle the two versions in the same file, but this is a better and more efficient solution.

shaheemazmalmmd:
> /cc @deveaud-m @shaheemazmalmmd
> 
> Does this strategy of migration from v1 to v2 makes sense or shall we look for an alternative?

Perfect. Idea looks great to me.

PR #2618, opened 2023-10-27
feat(os): add support for Debian Bookworm (12)

### LINKED ISSUE #2596: [BUG]: Error during setup: php-sqlite3 does not install as a dependency
### Description

When the steps for install as a dependency are followed during setup the build command gives error as php-sqlite3 extension is not installed and is a dependency.

#### How to reproduce

Follow the steps for installing from source given here: https://github.com/fossology/fossology/wiki/Install-from-Source
When `cmake --build ./build --parallel` command is run it gives the error attached as a screenshot below.

#### Screenshots

error

### Versions

* Operating System (lsb_release -a): Ubuntu 22.04.3

GMishx:
## Description

Add support for Debian Bookworm which was released on 2023-06-10.

Note: The package `upx-ucl` used by `ununpack` to extract `upx` files is no longer available on Debian package repositories. It appears to have failing a CI test https://ci.debian.net/packages/u/upx-ucl/
Keeping this PR as a WIP until the package is available. If it takes long, we would have to disable extraction of `upx` files from ununpack agent and go ahead with merge of this PR.

### Changes

Add packages for `bookworm` in dependency installer scripts.

## How to test

Build and install FOSSology on Debian Bookworm (12).

Closes #2596

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -23,7 +23,7 @@ jobs:
       DEBIAN_FRONTEND: "noninteractive"
     services:
       postgres:
-        image: postgres:13
+        image: postgres:15
         env:
           POSTGRES_DB: postgres
           POSTGRES_USER: postgres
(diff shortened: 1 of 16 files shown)


GMishx:
Since the issue with `upx-ucl` does not appear to be resolved soon, dropping the support for upx files with a CMake flag `UPX_INSTALLED`.

PR #2621, opened 2023-11-03
fix(user-edit): handle HttpForbiddenException

GMishx:
## Description

Fix `user-edit` page by handling exceptions.

### Changes

Handle exceptions with REST tokens.

## How to test

1. Edit user with expired tokens.
2. Edit user with new tokens.
3. The page should load without any errors.


shaheemazmalmmd (approved):

PR #2628, opened 2023-11-08
fix(documentation): update README.md with PHP version

vargenau:
## Description

Update README.md with supported PHP versions.


GMishx (approved):
Changes looks good.

PR #2646, opened 2023-12-11
fix(licenseRef): make dataype consistent

GMishx:
## Description

Make the datatype of `rf_detector_type` consistent to int type from string based on `src/cli/exportLicenseRefUsingSPDX.php`. This is helpful for https://github.com/fossology/LicenseDb and anyother project trying to read this file from typed programming language like Golang.

See: https://github.com/fossology/LicenseDb/pull/21

### Changes

Modify `licenseRef.json` to use `rf_detector_type` as only int, not mix of string and int type.

## How to test

1. Setup new FOSSology instance or update and old one.
2. Make sure the licenses are updated correctly by `fo-postinstall` script.
3. Make sure the licenses are updated in DB.


GMishx (commented):
Tested with fresh and existing upgrade.

PR #2649, opened 2023-12-13
fix(token): fix token generation for user

GMishx:
## Description

Fix token generation from user-edit page.

### Changes

1. Fix token generation for user page.
2. Fix error handling in user-edit.php
3. Fix the path `/users/token` to be authenticated.
4. Handle `user_edit` plugin exception in UserController.php

## How to test

Try to generate new tokens from UI (User edit page) and from rest API (`/users/tokens`).


ag4ums (approved):
LGTM ... working as expected

Issue #2648, opened 2023-12-11
getAllLicenseDataForUpload() SQL Syntax icorrect, export fails


uholeschak:
In AllDecisionsDao.php the SQL Syntax is incorrect, a space is missing, so export is not possible:
`"PHP Warning: pg_prepare(): Query failed: FEHLER: Müll folgt auf Parameter bei »$1U«"`
` " INNER JOIN uploadtree ut ON ce.uploadtree_fk=ut.uploadtree_pk WHERE ut.upload_fk=$1" .`
` "UNION DISTINCT " .`
To fix this add a space before UNION:
` " INNER JOIN uploadtree ut ON ce.uploadtree_fk=ut.uploadtree_pk WHERE ut.upload_fk=$1" .`
` " UNION DISTINCT " .`

GMishx:
Can you please open a PR with the fix @uholeschak ?

uholeschak:
Can't you simply add this space and done? Creating a pull request consumes to much time ...

Hemant2335:
Can I work on this isuue

PR #2655, opened 2023-12-16
fix: added missing space before UNION DISTINCT

reshma045:
fixes issue #2648 .
Added a missing space before `UNION DISTINCT`


GMishx:
@reshma045 thanks for your contribution. The issue is already fixed by #2656

PR #2659, opened 2023-12-20
chore(notice): update both notice and notice.spdx files to latest

shaheemazmalmmd:
## Description

Update Notice files.


GMishx (approved):
LGTM

PR #2673, opened 2024-01-17
fix(licenseExport): add all licenses to csv and fix cells

shaheemazmalmmd:
## Description

Add new export for all the licenses with candidates.

## How to test

* Create Candidate licenses in different groups with few clicking on merge request.
* Generate CSV using "CSV Export Marydone". -> All the main licenses and licenses with merge request should appear.
* Generate CSV using "CSV Export ALL". -> All the main licenses and all candidate licenses with or without merge request should appear.
* Import both with modifying some main and candidate licenses.
* Check the Changes in UI.

### DIFF
--- a/src/lib/php/Application/LicenseCsvExport.php
+++ b/src/lib/php/Application/LicenseCsvExport.php
@@ -65,12 +65,16 @@ class LicenseCsvExport
    * @param int $rf Set the license ID to get only one license, set 0 to get all
    * @return string csv
    */
-  public function createCsv($rf=0)
+  public function createCsv($rf=0, $allCandidates=false)
   {
+    $forAllCandidates = "WHERE marydone = true";
+    if ($allCandidates) {
+      $forAllCandidates = "";
+    }
     $forGroupBy = " GROUP BY rf.rf_shortname, rf.rf_fullname, rf.rf_spdx_id, rf.rf_text, rc.rf_shortname, rr.rf_shortname, rf.rf_url, rf.rf_notes, rf.rf_source, rf.rf_risk, gp.group_name";
     $sql = "WITH marydoneCand AS (
   SELECT * FROM license_candidate
-  WHERE marydone = true
+  $forAllCandidates
 ), allLicenses AS (
 SELECT DISTINCT ON(rf_pk) * FROM
   ONLY license_ref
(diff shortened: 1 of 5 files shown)


src/lib/php/Application/LicenseCsvImport.php
```diff
@@ -228,7 +229,7 @@ private function updateLicense($row, $rfPk)
       $extraParams[] = "rf_spdx_id=$" . count($param);
       $log .= ", updated SPDX ID";
     }
-    if (!empty($row['text']) && $row['text'] != $oldLicense['rf_text']) {
+    if (!empty($row['text']) && $row['text'] != $oldLicense['rf_text'] && strlen($oldLicense['rf_text']) < $maxAllowedChars) {
```
GMishx, line 232: Should check `"The maximum number of characters per cell in a CSV file(32,767) exceeded, Please edit this license text in UI"` string instead here.

src/lib/php/Application/LicenseCsvExport.php
```diff
@@ -65,12 +65,16 @@ public function setEnclosure($enclosure='"')
    * @param int $rf Set the license ID to get only one license, set 0 to get all
    * @return string csv
    */
-  public function createCsv($rf=0)
+  public function createCsv($rf=0, $allCandidates=0)
   {
+    $forAllCandidates = "WHERE marydone = true";
+    if (!empty($allCandidates)) {
```
GMishx, line 71: Can it be `boolean` instead? It will be easier to handle then.

GMishx (changes requested):

GMishx:
Please put `fputs($out, $bom =( chr(0xEF) . chr(0xBB) . chr(0xBF) ));` in the `Fossology\Lib\Application\LicenseCsvExportTest::testCreateCsv` to fix the test.

src/lib/php/Application/LicenseCsvImport.php
shaheemazmalmmd: was thinking that in case of edits to same text would modify the actual licenses text. As it is only checking the old license text/ previous text i have the same check ..

do you think there are any drawbacks ??

src/lib/php/Application/LicenseCsvExport.php
shaheemazmalmmd: Agreed

src/lib/php/Application/LicenseCsvImport.php
GMishx: Yes, in current setup, the import will update the license text to `"The maximum number of characters per cell in a CSV file(32,767) exceeded, Please edit this license text in UI"` wherever it sees it.
If the text is longer than the limit, and fossology can read it, then there is no issue right?

GMishx (approved):

GMishx:
During tests, I constantly see FOSSology could not find `shortname` column in CSV.
It is because of the BOM at the start of file being added and should be removed before reading.
@shaheemazmalmmd please add following lines of code to fix it.

```gitdiff
diff --git a/src/lib/php/Application/LicenseCsvImport.php b/src/lib/php/Application/LicenseCsvImport.php
index 2c3cf198d..ccb081991 100644
--- a/src/lib/php/Application/LicenseCsvImport.php
+++ b/src/lib/php/Application/LicenseCsvImport.php
@@ -158,6 +158,7 @@ class LicenseCsvImport
private function handleHeadCsv($row)
{
$headrow = array();
+    $row[0] = trim($row[0], "\xEF\xBB\xBF");  // Remove BOM
     foreach (array('shortname','fullname','text') as $needle) {
     $col = ArrayOperation::multiSearch($this->alias[$needle], $row);
     if (false === $col) {
```

PR #2686, opened 2024-02-20
feat(buildsystem): Introduce FORCE_CONF_OVERWRITE option to control configuration file overwrites

### LINKED ISSUE #2683: Custom Db.conf parameters get overwritten during upgrade, and fo-postinstall fails
### Description

In case of an upgrade of an existing installation, the Db.conf file gets overwritten with the default one. If db name, host, user or password are not the default ones, this will cause `fo-postinstall` to fail and fossology to stop working because of a db connection error. In such case, the correct parameters need to be put again into Db.conf, and `fo-postinstall` needs to be re-run.

This is a regression: in previous versions using make instead of cmake, this was handled correctly in [install/Makefile](https://github.com/fossology/fossology/blob/76450a21575d2a9ad3e5873b2f758fe408caf7a5/install/Makefile#L63C1-L65C68):

```bash
	@if [ ! -f $(DESTDIR)$(CONFPATH)/Db.conf -o "$(OVERWRITE)" ]; then \
		echo "NOTE: using default version for $(DESTDIR)$(CONFPATH)/Db.conf"; \
	 	$(INSTALL) -m 640 defconf/Db.conf $(DESTDIR)$(CONFPATH)/Db.conf; \
```

This is not the case anymore in [install/CMakeLists.txt](https://github.com/fossology/fossology/blob/8286d8aa0e02d22a54bf1010d65f20f9787f4bbb/install/CMakeLists.txt#L40):

```
    ${CMAKE_CURRENT_BINARY_DIR}/gen/Db.conf
```

#### How to reproduce

Upgrade an existing Fossology installation using make, with custom Db.conf parameters, to any new version using cmake.

### Versions

It should affect all versions since 4.3.0 (CMake)

dvjsharma:
## Description
In the previous repository state, configuration files were being overwritten during the installation process, potentially leading to unintended changes. To address this, a new option, `FORCE_CONF_OVERWRITE`, has been introduced. When set to `OFF` (default), the installation process will skip overwriting existing configuration files. Conversely, setting `FORCE_CONF_OVERWRITE` to `ON` will enable overwriting, allowing users to control whether configuration files should be replaced during installation.

### Changes
* Introduced a new option `FORCE_CONF_OVERWRITE` (default `OFF`) to manage weather user files should be overwritten or not during installation. 
* Implemented a general function to handle installs in CMake. It takes `FILES` (list of files to install with their path), `DEST_DIR` (installation path) and `COMPONENT_NAME` (value of `COMPONENT`) as parameters and based on the value of `FORCE_CONF_OVERWRITE` option, it handles if files should be overwritten or not.
* Currently excluded `Db.conf`, `fossology.conf`, `samplefooter.txt`, `sampleheader.txt`, `fossdash_metrices.yml`, `fo-apache.conf`, and `src-install-apache-example.conf`.

## How to test
Change anything in the above mentioned files (example, change the default password `fossy`) and then build the project using `cmake -S. -B./build -G Ninja` followed by installation as specified in [wiki](https://github.com/fossology/fossology/wiki/Install-from-Source). You will notice the files are not overwritten.

Screenshot from 2024-02-21 00-49-18

To test the option, use `cmake -S. -B./build -DFORCE_CONF_OVERWRITE=ON -G Ninja` and you will see the files are overwritten.

Screenshot from 2024-02-21 01-03-48

PS. If the above steps do not work, please use `rm -rf build` command before installation to delete build directory (reason- CMake caches stuff).

fixes #2683 
CC @GMishx @avinal

### DIFF
--- a/install/CMakeLists.txt
+++ b/install/CMakeLists.txt
@@ -3,6 +3,8 @@ SPDX-License-Identifier: GPL-2.0-only
 SPDX-FileCopyrightText: © 2021 Avinal Kumar <avinal.xlvii@gmail.com>
 #]=======================================================================]
 
+option(FORCE_CONF_OVERWRITE "Force overwrite of configuration files" OFF)
+
 set(FO_CWD ${CMAKE_CURRENT_SOURCE_DIR})
 
 file(GLOB_RECURSE FO_INSTALL_CONFS ${FO_CWD}/*.in)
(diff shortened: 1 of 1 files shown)


install/CMakeLists.txt
```diff
@@ -35,29 +37,44 @@ add_custom_target(BUILD_version ALL
...
-install(DIRECTORY
-    ${CMAKE_CURRENT_SOURCE_DIR}/defconf/
-    ${CMAKE_CURRENT_SOURCE_DIR}/fossdash/
-    COMPONENT common
-    DESTINATION ${FO_SYSCONFDIR}
-    FILES_MATCHING
-    PATTERN *.txt
-    PATTERN *.yml
+set(FO_DIR_FILES
+    ${CMAKE_CURRENT_SOURCE_DIR}/defconf/samplefooter.txt
```
GMishx, line 71: Would recommend using `file(GLOB)` function to retain the existing functionality to copy all `.txt` and `.yml` files from respective directories.
This will help adding more configurations files in future easy.

Check documentation at: https://cmake.org/cmake/help/v3.13/command/file.html#glob

GMishx (changes requested):
Minor changes needed.

install/CMakeLists.txt
```diff
@@ -35,29 +37,44 @@ add_custom_target(BUILD_version ALL
...
 
-install(FILES
-    "${CMAKE_CURRENT_BINARY_DIR}/VERSION"
+function(install_config_files FILES DEST_DIR)
+    foreach(FILE_NAME ${FILES})
+        get_filename_component(FILE_BASE ${FILE_NAME} NAME)
+        if (NOT EXISTS "${DEST_DIR}/${FILE_BASE}" OR FORCE_CONF_OVERWRITE)
+            message(STATUS "Installing ${FILE_BASE}")
+            install(FILES ${FILE_NAME}
+                    COMPONENT common
```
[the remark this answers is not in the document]
dvjsharma: Done

install/CMakeLists.txt
avinal: The first one looks good.

avinal (commented):
Changes look good to me. Will test over the weekend.

avinal (approved):
LGTM

GMishx (approved):
Tested, working as expected.

PR #2690, opened 2024-03-01
fix(ojo): ignore extra symbols at end

GMishx:
## Description

License names should end with either words, digits or a + symbol.

This prevents creation of licenses such as "MIT."

### Changes

1. Reduce the min length of licenses from 3 to 2 (there are "GD" and "SL" licenses now).
2. Make sure the license text ends with a word, digit or a "+".
3. Add test case to check scenarios with "MPL-1.1+" and "MIT."

## How to test

1. Create test files with following kind of license text:
    - "SPDX-License-Identifier: GPL-3.0-or-later."
    - "SPDX-License-Identifier: GPL-3.0+."
    - "SPDX-License-Identifier: GPL-3.0-or-later WITH GCC-exception-3.1."
2. Archive the file and scan with OJO.
3. OJO should find only "GPL-3.0-or-later", "GPL-3.0+" and "GCC-exception-3.1".

### DIFF
--- a/src/ojo/agent/ojoregex.hpp
+++ b/src/ojo/agent/ojoregex.hpp
@@ -27,10 +27,11 @@
  * @brief Regex to filter license names from list of license list
  *
  * -# License names will consist of words, digits, dots and hyphens.
- * -# Length of license names greater than 3 (based on
+ * -# Length of license names greater than 2 (based on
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (approved):

PR #2697, opened 2024-03-08
fix(RestAuthMiddleware): exclude from auth. only the right endpoints

### LINKED ISSUE #2693: REST API: /uploads/{id}/item/{itemId}/info does not work because it is incorrectly excluded from authentication
### Description

While trying to get /uploads/{id}/item/{itemId}/info via REST API I always get error 500. By looking at apache2 error logs, I see an "Unable to find plugin view_info" error message.

Info view in WebUI /?mod=view_info&upload={id}&item={itemId} works normally.  Also any other REST API calls I have tried works normally, only that one returns error 500.

> **EDIT**: solved!  in RestAuthMiddleware class, there is some code that excludes some endpoints from authentication ("/info", "/openapi", "/health") but the problem is that code simply searches for such strings inside the request path, so also paths like  `/uploads/{id}/item/{itemId}/info` are excluded from authentication. In such case, user is set to "Default User" even if the request is authenticated with a token, and thus all plugins that do require authentication (including view_info) are not loaded - so the above error occurs.  If one tries to force loading of view_info (by adding a call to PostInitialize() in the corresponding source file) and adds some debug code, they can verify that user is set to "Default User" and group id to 2, even if the request is authenticated. 
>
> The problem can be solved by modifying RestAuthMiddleware code, so that only endpoints that do not actually require authentication are excluded from authentication (see the solution in the comments below)

#### How to reproduce

Make a request via REST API: /uploads/{id}/item/{itemId}/info

#### Screenshots

n/a

### Versions

version 4.4.0, built from sources on Debian 11

### Logs

Any logs (if any) generated in

#### FOSSology logs

n/a

#### Apache logs

```
[Thu Mar 07 11:02:03.188837 2024] [php7:notice] [pid 517218] [client 100.97.73.63:50110] [2024-03-07T11:02:03.188732+00:00]
default.ERROR: Unable to find plugin view_info
```

json data follows, I formatted it for better readability:

<p><details>
<summary><b>click to see json data</b></summary>

```json
[
    {
        "file": "/usr/local/share/fossology/www/ui/api/Controllers/FileInfoController.php",
        "line": 33,
        "function": "getPlugin",
        "class": "Fossology\\\\UI\\\\Api\\\\Helper\\\\RestHelper",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/CallableResolver.php",
        "line": 154,
        "function": "__construct",
        "class": "Fossology\\\\UI\\\\Api\\\\Controllers\\\\FileInfoController",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/CallableResolver.php",
        "line": 103,
        "function": "resolveSlimNotation",
        "class": "Slim\\\\CallableResolver",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/CallableResolver.php",
        "line": 72,
        "function": "resolveByPredicate",
        "class": "Slim\\\\CallableResolver",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/Routing/Route.php",
        "line": 366,
        "function": "resolveRoute",
        "class": "Slim\\\\CallableResolver",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php",
        "line": 81,
        "function": "handle",
        "class": "Slim\\\\Routing\\\\Route",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php",
        "line": 81,
        "function": "handle",
        "class": "Slim\\\\MiddlewareDispatcher",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/Routing/Route.php",
        "line": 341,
        "function": "handle",
        "class": "Slim\\\\MiddlewareDispatcher",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/Routing/RouteRunner.php",
        "line": 84,
        "function": "run",
        "class": "Slim\\\\Routing\\\\Route",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/www/ui/api/Middlewares/FossologyInitMiddleware.php",
        "line": 43,
        "function": "handle",
        "class": "Slim\\\\Routing\\\\RouteRunner",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php",
        "line": 313,
        "function": "__invoke",
        "class": "Fossology\\\\UI\\\\Api\\\\Middlewares\\\\FossologyInitMiddleware",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/www/ui/api/Middlewares/RestAuthMiddleware.php",
        "line": 58,
        "function": "handle",
        "class": "class@anonymous\\u0000/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php:294$29",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php",
        "line": 313,
        "function": "__invoke",
        "class": "Fossology\\\\UI\\\\Api\\\\Middlewares\\\\RestAuthMiddleware",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/Middleware/ContentLengthMiddleware.php",
        "line": 27,
        "function": "handle",
        "class": "class@anonymous\\u0000/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php:294$29",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php",
        "line": 147,
        "function": "process",
        "class": "Slim\\\\Middleware\\\\ContentLengthMiddleware",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/www/ui/api/index.php",
        "line": 121,
        "function": "handle",
        "class": "class@anonymous\\u0000/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php:128$27",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php",
        "line": 313,
        "function": "Fossology\\\\UI\\\\Api\\\\{closure}",
        "class": "Closure",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/Middleware/ErrorMiddleware.php",
        "line": 107,
        "function": "handle",
        "class": "class@anonymous\\u0000/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php:294$29",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php",
        "line": 147,
        "function": "process",
        "class": "Slim\\\\Middleware\\\\ErrorMiddleware",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php",
        "line": 81,
        "function": "handle",
        "class": "class@anonymous\\u0000/usr/local/share/fossology/vendor/slim/slim/Slim/MiddlewareDispatcher.php:128$27",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/App.php",
        "line": 215,
        "function": "handle",
        "class": "Slim\\\\MiddlewareDispatcher",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/vendor/slim/slim/Slim/App.php",
        "line": 199,
        "function": "handle",
        "class": "Slim\\\\App",
        "type": "->"
    },
    {
        "file": "/usr/local/share/fossology/www/ui/api/index.php",
        "line": 480,
        "function": "run",
        "class": "Slim\\\\App",
        "type": "->"
    }
]
```
</details></p>

#### Job logs

n/a

alpianon:
## Description

In RestAuthMiddleware class, there is some code that excludes some endpoints from authentication ("/version", "/info", "/openapi", "/health") but the problem is that that code simply searches for the corresponing strings (i.e. "/version", "/info", "/openapi", "/health") inside the request path regardless of their position inside the path, so also paths like /uploads/{id}/item/{itemId}/info are excluded from authentication. This implies that when one makes an authenticated request with /uploads/{id}/item/{itemId}/info, authentication gets ignored, user is set to "Default User" and group id is set to 2. Therefore all plugins that do require authentication (including view_info, which is needed to respond to the above-mentioned API call) are not loaded, and this causes a 500 error - "Unable to find plugin view_info". 

To debug the above error, you may force the loading of view_info (by adding a call to $NewPlugin->PostInitialize() at the end of www/ui/ui-view-info.php), and you get a 403 error -  "Upload is not accessible" - when you try to perform the same request. Then, you may add some debug code to www/ui/api/Controllers/RestController.php to log $_SESSION['User'] and group ID values before the "Upload is not accessible" exception is thrown: you will see that user is set to "Default User" and group id to 2, even if the request is authenticated. So the problem lies in an authenticated request that gets transformed into an unauthenticated request by RestAuthMiddleware.

The problem can be solved by modifying RestAuthMiddleware code, so that only endpoints that do not actually require authentication are excluded from authentication. 

### Changes

RestAuthMiddleware should check if any of the authFreePaths exactly matches the endpoint of the request, and not simply search for the authFreePath string in the request, which may lead to the errors described above.

Moreover, the "/version" endpoint does not exist in the API, so it does not need to be included in the list of endpoints not requiring authentication.

## How to test

Perform an /uploads/{id}/item/{itemId}/info API call, and check that it returns a correct result (200).

Fix #2693

### DIFF
--- a/src/www/ui/api/Middlewares/RestAuthMiddleware.php
+++ b/src/www/ui/api/Middlewares/RestAuthMiddleware.php
@@ -43,11 +43,14 @@ class RestAuthMiddleware
   {
     $requestUri = $request->getUri();
     $requestPath = strtolower($requestUri->getPath());
-    $authFreePaths = ["/version", "/info", "/openapi", "/health"];
+    $authFreePaths = ["/info", "/openapi", "/health"];
 
     $isPassThroughPath = false;
+    // path is /repo/api/v2/<endpoint>, we need to get only the endpoint part
+    $parts = explode("/", $requestPath, 5);
+    $endpoint = "/".end($parts);
     foreach ($authFreePaths as $authFreePath) {
-      if (strpos($requestPath, $authFreePath) !== false) {
+      if ( $endpoint === $authFreePath ) {
         $isPassThroughPath = true;
         break;
       }


src/www/ui/api/Middlewares/RestAuthMiddleware.php
```diff
@@ -43,11 +43,14 @@ public function __invoke(Request $request, RequestHandler $handler) : ResponseIn
   {
     $requestUri = $request->getUri();
     $requestPath = strtolower($requestUri->getPath());
-    $authFreePaths = ["/version", "/info", "/openapi", "/health"];
+    $authFreePaths = ["/info", "/openapi", "/health"];
 
     $isPassThroughPath = false;
+    // path is /repo/api/v2/<endpoint>, we need to get only the endpoint part
+    $parts = explode("/", $requestPath, 5);
+    $endpoint = "/".end($parts);
```
GMishx, line 51: If someone have ditched the `/repo` path from Apache conf, then the `$requestPath` will be `/api/v2/<endpoint>`. Here, `$endpoint = "/".end($parts);` can result in something unexpected.

```php
preg_match("/\/?(?:(?:repo)?\/api\/v\d)(.*)/i", $requestPath, $matches);
$endpoint = "";
if (!empty($matches[1])) {
  $endpoint = $matches[1];
}
```

This will help keep the `/repo` optional and capture anything after `/v<digit>` in the `$endpoint`.

GMishx: However, with a quick check, the proposed solution will still work of `/info`, `/openapi` and `health` endpoints.
But for endpoints like `/api/v2/uploads/{id}/item/{itemId}/info`, the variable endpoint will contain `/{id}/item/{itemId}/info`. Which is wrong but sufficient for the usecase.

GMishx: What do you think @alpianon and @shaheemazmalmmd ?

alpianon: @GMishx as you correctly pointed out, the `/repo` part of the path could be changed in Apache conf, and actually one may remove it or change it to any other thing (eg. `/fossology`).

I guess the regex solution (without the `(?:repo)`) would be the cleanest option. However, I noticed that all the code in the API implementation assumes a path starting with `/repo`, see `src/www/ui/api/index.php`:

```php
/ Extracts the version from the URL
function getVersionFromUri ($uri)
{
  $matches = [];
  preg_match('/\/repo\/api\/v(\d+)/', $uri, $matches);
  return isset($matches[1]) ? intval($matches[1]) : null;
}

// Determine the API version based on the URL
$requestedVersion = isset($_SERVER['REQUEST_URI']) ? getVersionFromUri($_SERVER['REQUEST_URI']) : null;
$apiVersion = in_array($requestedVersion, [ApiVersion::V1, ApiVersion::V2]) ? $requestedVersion : ApiVersion::V1; // Default to "1"

// Construct the base path
$BASE_PATH = "/repo/api/v" .$apiVersion;
```

So either this gets solved globally by expanding the scope of this PR, or IMHO this specific issue may be temporarily solved by using explode+end, and another issue may be opened to globally handle API calls with paths not starting with `/repo` because of a non-standard Apache conf.

alpianon: BTW if you fear that regex would be too slow, this might be an alternative solution that would produce the same result (even if I just assume but I've not tested if it's actually faster):

```php
$pos = strpos($requestPath, "/api/v") + 6;
$pos = strpos($requestPath, "/", $pos);
$endpoint = substr($requestPath, $pos);
```

GMishx (approved):
Tested, working as expected.

PR #2702, opened 2024-03-19
fix(scancode): Use spdx_license_key instead of key

JanAltenberg:
## Description

For scancode findings currently only the ScanCode LicenseDB key is being extracted and reported. I think, this should be changed to use the SPDX ID. The comment before `scanFileWithScancode` (`src/scancode/agent/scancode_wrapper.cc`) already mentions the SPDX ID:
```
  * In licenses array:
  * key-> license spdx key
[...]
```
but the ScanCode LicenseDB key is being used (key instead of spdx_license_key):
```
keys_to_extract_from_licenses = ['key', 'score', 'name', 'text_url', 'start_line',, 'matched_text']
```
### Changes

This change extracts and uses `spdx_license_key` from the scancode results instead of `key`.

## How to test

Upload a package and activate: "ScancodeToolkit, scan for License". The scancode results should be shown with their SPDX ID (e.g. GPL-2.0-or-later instead of gpl-2.0-plus).

### DIFF
--- a/src/scancode/agent/runscanonfiles.py
+++ b/src/scancode/agent/runscanonfiles.py
@@ -25,7 +25,7 @@ def update_license(licenses):
     list: A list of dictionaries containing relevant license information.
   """
   updated_licenses = []
-  keys_to_extract_from_licenses = ['key', 'score', 'name', 'text_url', 'start_line', 'matched_text']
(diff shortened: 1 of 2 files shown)


GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

Issue #2709, opened 2024-04-03
Upgrade the User & Group APIs to Version 2


dushimsam:
# Upgrade the /users & /groups APIs to Version 2

**In accordance with our new policy documented [here](https://github.com/fossology/fossology/issues/2404#issuecomment-1671046653), we want to enhance our REST APIs to Version 2 migrating all `/users` APIs. The setup for the Version 2 is already in place.**

#### 1.  **/users**,  POST
<img width="926" alt="Screenshot 2024-04-03 at 05 43 58" src="https://github.com/fossology/fossology/assets/66276301/ecd29585-9848-41a1-b42a-5e289625e2b2">

- Change these fields to camelCase: `copyright_email_author`, `user_pass`

#### 2.  **/users** & **/users/{id}**,  GET

<img width="654" alt="Screenshot 2024-04-03 at 05 46 23" src="https://github.com/fossology/fossology/assets/66276301/3fbfa693-b0d8-472c-ba59-36aa332f4780">

- Change these fields to camelCase: `copyright_email_author`

#### 3.  **/users/{id}**,  PUT
<img width="926" alt="Screenshot 2024-04-03 at 05 43 58" src="https://github.com/fossology/fossology/assets/66276301/75eedf74-689c-42b5-a496-49be0f510af7">

- Change these fields to camelCase: `copyright_email_author`, `user_pass`

####  4. **/users/self**, GET
<img width="799" alt="Screenshot 2024-04-03 at 05 51 49" src="https://github.com/fossology/fossology/assets/66276301/7a317d74-0c1f-438f-a3fc-d2a3ea407e5a">

- Change these fields to camelCase: `copyright_email_author`, `default_group`

#### 5. **/users/tokens**, POST

<img width="946" alt="Screenshot 2024-04-03 at 05 57 04" src="https://github.com/fossology/fossology/assets/66276301/10c7eb0e-44f1-4060-95cf-18ef4221b409">

- Change these fields in requestBody to camelCase: `token_name`, `token_scope`, `token_expire`

#### 6. **/users/tokens/{type}**, GET

<img width="782" alt="Screenshot 2024-04-03 at 05 58 58" src="https://github.com/fossology/fossology/assets/66276301/73326cac-e2f4-415c-873d-aafa2c3c38cb">

- Change these fields to camelCase: `active_tokens`, `token_name`, `token_expire`

#### 7.  **/groups**,  POST

[...]

PR #2711, opened 2024-04-03
feat(api): Upgrade User & Group APIs to Version 2

dvjsharma:
## Description

This PR aims to upgrade the `/users` and `/group` endpoints to v2 in accordance with the new upgrade guidelines.

### Changes

* Completed all tasks mentioned in #2709.
* Updated `openapiv2.yaml` to have all documentation changes.
* Modified models & unit tests to make them version aware.

### How to Test

* Open the `openapiv2.yaml` file in Swagger Editor and test all the updated endpoints for v2 specific responses. All function as intended.
* Execute unit tests, all pass without any issues.

CC @GMishx @shaheemazmalmmd @dushimsam

### DIFF
--- a/src/www/ui/api/Controllers/UserController.php
+++ b/src/www/ui/api/Controllers/UserController.php
@@ -46,9 +46,10 @@ class UserController extends RestController
    */
   public function getUsers($request, $response, $args)
   {
+    $apiVersion = ApiVersion::getVersion($request);
     $id = null;
-    if (isset($args['id'])) {
-      $id = intval($args['id']);
+    if (isset($args['pathParam'])) {
+      $id = $apiVersion == ApiVersion::V2 ? intval($this->restHelper->getUserDao()->getUserByName($args['pathParam'])['user_pk']) : intval($args['pathParam']);
       if (! $this->dbHelper->doesIdExist("users", "user_pk", $id)) {
         throw new HttpNotFoundException("UserId doesn't exist");
       }
@@ -57,7 +58,7 @@ class UserController extends RestController
 
     $allUsers = array();
     foreach ($users as $user) {
-      $allUsers[] = $user->getArray();
+      $allUsers[] = $user->getArray($apiVersion);
     }
     if ($id !== null) {
       $allUsers = $allUsers[0];
@@ -76,13 +77,14 @@ class UserController extends RestController
    */
   public function addUser($request, $response, $args)
   {
+    $apiVersion = ApiVersion::getVersion($request);
     $userDetails = $this->getParsedBody($request);
     $userHelper = new UserHelper();
     // creating symphony request
     $symfonyRequest = new \Symfony\Component\HttpFoundation\Request();
     $symfonyRequest->request->set('username', $userDetails['name']);
-    $symfonyRequest->request->set('pass1', $userDetails['user_pass']);
-    $symfonyRequest->request->set('pass2', $userDetails['user_pass']);
+    $symfonyRequest->request->set('pass1', $userDetails[$apiVersion == ApiVersion::V2 ? 'userPass' : 'user_pass']);
+    $symfonyRequest->request->set('pass2', $userDetails[$apiVersion == ApiVersion::V2 ? 'userPass' : 'user_pass']);
     $symfonyRequest->request->set('description', $userDetails['description']);
     $symfonyRequest->request->set('permission', $userHelper->getEquivalentValueForPermission($userDetails['accessLevel']));
     $symfonyRequest->request->set('folder', $userDetails['rootFolderId']);
@@ -100,7 +102,7 @@ class UserController extends RestController
       $agents['Check_agent_monk'] = isset($userDetails['agents']['monk']) && $userDetails['agents']['monk'] ? 1 : 0;
       $agents['Check_agent_ojo'] = isset($userDetails['agents']['ojo']) && $userDetails['agents']['ojo'] ? 1 : 0;
       $agents['Check_agent_bucket'] = isset($userDetails['agents']['bucket']) && $userDetails['agents']['bucket'] ? 1 : 0 ;
-      $agents['Check_agent_copyright'] = isset($userDetails['agents']['copyright_email_author']) && $userDetails['agents']['copyright_email_author'] ? 1 : 0;
+      $agents['Check_agent_copyright'] = isset($userDetails['agents'][$apiVersion == ApiVersion::V2 ? 'copyrightEmailAuthor' : 'copyright_email_author']) && $userDetails['agents'][$apiVersion == ApiVersion::V2 ? 'copyrightEmailAuthor' : 'copyright_email_author'] ? 1 : 0;
       $agents['Check_agent_ecc'] = isset($userDetails['agents']['ecc']) && $userDetails['agents']['ecc'] ? 1 : 0;
       $agents['Check_agent_keyword'] = isset($userDetails['agents']['keyword']) && $userDetails['agents']['keyword'] ? 1 : 0;
       $agents['Check_agent_nomos'] = isset($userDetails['agents']['nomos']) && $userDetails['agents']['nomos'] ? 1 : 0;
@@ -138,7 +140,8 @@ class UserController extends RestController
    */
   public function deleteUser($request, $response, $args)
   {
-    $id = intval($args['id']);
+    $apiVersion = ApiVersion::getVersion($request);
+    $id = $apiVersion == ApiVersion::V2 ? intval($this->restHelper->getUserDao()->getUserByName($args['pathParam'])['user_pk']) : intval($args['pathParam']);
     if (!$this->dbHelper->doesIdExist("users","user_pk", $id)) {
       throw new HttpNotFoundException("UserId doesn't exist");
     }
@@ -158,10 +161,14 @@ class UserController extends RestController
    */
   public function getCurrentUser($request, $response, $args)
   {
-    $user = $this->dbHelper->getUsers($this->restHelper->getUserId())[0]->getArray();
+    $apiVersion = ApiVersion::getVersion($request);
+    $user = $this->dbHelper->getUsers($this->restHelper->getUserId())[0]->getArray($apiVersion);
+    if ($apiVersion == ApiVersion::V2) {
+      return $response->withJson($user, 200);
+    }
     $userDao = $this->restHelper->getUserDao();
     $defaultGroup = $userDao->getUserAndDefaultGroupByUserName($user["name"])["group_name"];
-    $user["default_group"] = $defaultGroup;
+    $user['default_group'] = $defaultGroup;
     return $response->withJson($user, 200);
   }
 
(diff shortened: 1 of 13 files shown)


dvjsharma:
Few points to keep in mind regarding some upgrades: 

- `/users`, POST:  Code changes are fine, but in `openapiv2.yaml`, to change `copyright_email_author` to camelCase, I modified the `Analysis` component. However, this component is used by `ScanOptions`, which is in turn used by `UploadOptions` and the `/jobs` endpoint. So this change might break down other endpoints. Need opinion on how to handle this.

- `/users/self`, GET: Changing `default_group` to camelCase matches with an already existing key named `defaultGroup`, thus overwriting it. Therefore, I modified `default_group` to `default_group_name` and it's camelCase for v2. I also updated the documentation for both v1 and v2 APIs to reflect this change, as well as updated the unit tests to accommodate this new name.

- `/users/tokens/{type}`, GET: `token_name` and `token_expire` keys were already renamed to `name` and `expire` during the token endpoint upgrade. I have updated the v1 API docs accordingly. Also, changed `active_tokens` key to camelCase as requested.

- For the unit tests, all modal unit tests concerning users and group endpoints are now version aware and verify the data format for both v1 and v2 specifications (enhanced coverage). However, I focused solely on resolving failing controller tests without making them version aware since many of them do not specifically assess the semantics of request or response objects, but rather the functionality, which remains largely consistent between v1 and v2. Need thoughts on this.

Apart from these points, everything functions as anticipated.

src/www/ui/api/documentation/openapiv2.yaml
```diff
@@ -2482,7 +2482,7 @@ paths:
                   - $ref: '#/components/schemas/User'
                   - type: object
                     properties:
-                      default_group:
+                      defaultGroupName:
```
GMishx, line 2485: Would rather recommend refactoring `defaultGroup` in `#/components/schemas/User` as a string field which holds the group name rather than group id as there are no other endpoints which uses group ids. And perhaps, move `/groups/{id}` and endpoints under it to use the group names instead.

If you agree, we can document the changes in `/groups/{id}` as part of #2709.
/cc @dushimsam

GMishx:
> Few points to keep in mind regarding some upgrades:
> 
>     * `/users`, POST:  Code changes are fine, but in `openapiv2.yaml`, to change `copyright_email_author` to camelCase, I modified the `Analysis` component. However, this component is used by `ScanOptions`, which is in turn used by `UploadOptions` and the `/jobs` endpoint. So this change might break down other endpoints. Need opinion on how to handle this.

Ultimately the changes should be done in all endpoints. Following the principle of atomic change, changes required in different endpoints can be opened as an issue and dealt in a separate PR.

>     * `/users/self`, GET: Changing `default_group` to camelCase matches with an already existing key named `defaultGroup`, thus overwriting it. Therefore, I modified `default_group` to `default_group_name` and it's camelCase for v2. I also updated the documentation for both v1 and v2 APIs to reflect this change, as well as updated the unit tests to accommodate this new name.

As mentioned earlier, users/developers should not deal with raw IDs from application. Thus, I will recommend removing the group id, user id, etc. from API and replace with names (which are also unique).

>     * `/users/tokens/{type}`, GET: `token_name` and `token_expire` keys were already renamed to `name` and `expire` during the token endpoint upgrade. I have updated the v1 API docs accordingly. Also, changed `active_tokens` key to camelCase as requested.

Thanks :+1: I will test it again.

>     * For the unit tests, all modal unit tests concerning users and group endpoints are now version aware and verify the data format for both v1 and v2 specifications (enhanced coverage). However, I focused solely on resolving failing controller tests without making them version aware since many of them do not specifically assess the semantics of request or response objects, but rather the functionality, which remains largely consistent between v1 and v2. Need thoughts on this.

That makes sense. We shall cover other aspects when we start working on test cases for REST API.

src/www/ui/api/documentation/openapiv2.yaml
dvjsharma: Agreed, I'll push the changes soon.

dvjsharma:
As suggested, I have updated the `/users` and `/groups` endpoints to use unique names instead of group IDs or user IDs. The current modification is made with the consideration that v1 API calls also need to be supported.  I have also updated all model and controller unit tests, as well as documentation to reflect these changes. I made a few decisions, please take a look and comment:
* Since the request URL needed to be changed to use names instead of IDs (data type change as well), I adjusted the regular expression to include both names and IDs rather than creating 2 separate endpoints based on version. I managed the additional differentiation logic in the corresponding controllers. Does this sound good?
* For the `user` schema, I updated `defaultGroup` to display the group name instead of the Id by adjusting the model's get array function. Another (maybe better) option is to make this change universally in the `UserDao` to return name instead ID, but will require changes in core models as well as tests. Need thoughts on this.

src/www/ui/api/index.php
```diff
@@ -246,11 +246,11 @@ function (\Slim\Routing\RouteCollectorProxy $app) {
...
   function (\Slim\Routing\RouteCollectorProxy $app) {
-    $app->get('[/{id:\\d+}]', UserController::class . ':getUsers');
-    $app->put('[/{id:\\d+}]', UserController::class . ':updateUser');
-    $app->post('', UserController::class . ':addUser');
-    $app->delete('/{id:\\d+}', UserController::class . ':deleteUser');
     $app->get('/self', UserController::class . ':getCurrentUser');
+    $app->get('[/{pathParam:[\\w\\- \\.\\d]+}]', UserController::class . ':getUsers');
+    $app->put('[/{pathParam:[\\w\\- \\.\\d]+}]', UserController::class . ':updateUser');
+    $app->post('', UserController::class . ':addUser');
+    $app->delete('/{pathParam:[\\w\\- \\.\\d]+}', UserController::class . ':deleteUser');
```
GMishx, line 253: I would rather not allow whitespace in the group/user name. Please remove the whitespace from the regex.

src/www/ui/api/index.php
```diff
@@ -273,12 +273,12 @@ function (\Slim\Routing\RouteCollectorProxy $app) {
...
-    $app->delete('/{id:\\d+}', GroupController::class . ':deleteGroup');
-    $app->delete('/{id:\\d+}/user/{userId:\\d+}', GroupController::class . ':deleteGroupMember');
+    $app->post('/{pathParam:[\\w\\- \\.\\d]+}/user/{userPathParam:[\\w\\- \\.\\d]+}', GroupController::class . ':addMember');
+    $app->delete('/{pathParam:[\\w\\- \\.\\d]+}', GroupController::class . ':deleteGroup');
+    $app->delete('/{pathParam:[\\w\\- \\.\\d]+}/user/{userPathParam:[\\w\\- \\.\\d]+}', GroupController::class . ':deleteGroupMember');
     $app->get('/deletable', GroupController::class . ':getDeletableGroups');
-    $app->get('/{id:\\d+}/members', GroupController::class . ':getGroupMembers');
-    $app->put('/{id:\\d+}/user/{userId:\\d+}', GroupController::class . ':changeUserPermission');
+    $app->get('/{pathParam:[\\w\\- \\.\\d]+}/members', GroupController::class . ':getGroupMembers');
+    $app->put('/{pathParam:[\\w\\- \\.\\d]+}/user/{userPathParam:[\\w\\- \\.\\d]+}', GroupController::class . ':changeUserPermission');
```
GMishx, line 281: I would rather not allow whitespace in the group/user name. Please remove the whitespace from the regex.

src/www/ui/api/Controllers/UserController.php
```diff
@@ -28,6 +28,7 @@
 use Fossology\UI\Api\Models\InfoType;
 use Fossology\UI\Api\Models\TokenRequest;
 use Psr\Http\Message\ServerRequestInterface;
+use SebastianBergmann\CodeCoverage\Driver\Xdebug;
```
GMishx, line 31: Please remove unused import.

GMishx (changes requested):
Minor changes needed. Rest looks good.

src/www/ui/api/index.php
dvjsharma: By doing this, we won't be able to access `Default User` via requests.

dvjsharma: Hey @GMishx. Please guide on the above.

GMishx: Humm... We can rename the user to `defaultUser` or something like that. Allowing spaces will only cause problems.

dvjsharma: Done

src/www/ui/api/index.php
dvjsharma: Done

src/www/ui/api/index.php
```diff
@@ -246,11 +246,11 @@ function (\Slim\Routing\RouteCollectorProxy $app) {
 ////////////////////////////ADMIN-USERS/////////////////////
 $app->group('/users',
   function (\Slim\Routing\RouteCollectorProxy $app) {
-    $app->get('[/{id:\\d+}]', UserController::class . ':getUsers');
-    $app->put('[/{id:\\d+}]', UserController::class . ':updateUser');
-    $app->post('', UserController::class . ':addUser');
-    $app->delete('/{id:\\d+}', UserController::class . ':deleteUser');
     $app->get('/self', UserController::class . ':getCurrentUser');
+    $app->get('[/{pathParam:[\\w\\-\\.]+}]', UserController::class . ':getUsers');
+    $app->put('[/{pathParam:[\\w\\-\\.]+}]', UserController::class . ':updateUser');
```
GMishx, line 251: However, this regex will not work for API V1 which expects the path variable to be a digit. Also, will fail for organizations which uses email id as username (does not allow `@`).

How about using `[\w\d\-\.@_]+` as regex? It will allow every word, digit and other special characters commonly used in emails (`-.@_`). Please feel free to expand and experiment.
Also, once you've settled on the pattern, put it in a variable as it is used at multiple locations and keeping track to update pattern in future will be difficult.

dvjsharma: Hm..You are right. Also I think `[\w\d\-\.@_]+` looks good enough. I was searching for some more additions but those are very uncommon and might not be useful for our usecase. Although I need thoughts on finalising this. Additions I found were adding ``!#$%&'*+-/=?^_`{|}~`` and `"(),:;<>@[\]` characters for local part of the email. Domain part is already covered by the expression you suggested. WDYT?

GMishx: Lets keep it simple and not do a complete validation.

src/www/ui/api/Controllers/UserController.php
```diff
@@ -176,13 +183,14 @@ public function getCurrentUser($request, $response, $args)
    */
   public function updateUser($request, $response, $args)
   {
-    $id = intval($args['id']);
+    $apiVersion = ApiVersion::getVersion($request);
+    $id = $apiVersion == ApiVersion::V2 ? intval($this->restHelper->getUserDao()->getUserByName($args['pathParam'])['user_pk']) : intval($args['pathParam']);
     if ($this->dbHelper->doesIdExist("users","user_pk", $id)) {
```
GMishx, line 188: Oops, think we never tested it. This condition needs to be inverted (user not exists) to throw the exception. Can you please do that?

```suggestion
    if (!$this->dbHelper->doesIdExist("users","user_pk", $id)) {
```

src/www/ui/api/index.php
```diff
@@ -245,12 +248,12 @@ function (\Slim\Routing\RouteCollectorProxy $app) {
...
 $app->group('/users',
-  function (\Slim\Routing\RouteCollectorProxy $app) {
-    $app->get('[/{id:\\d+}]', UserController::class . ':getUsers');
-    $app->put('[/{id:\\d+}]', UserController::class . ':updateUser');
-    $app->post('', UserController::class . ':addUser');
-    $app->delete('/{id:\\d+}', UserController::class . ':deleteUser');
+  function (\Slim\Routing\RouteCollectorProxy $app) use ($pattern) {
     $app->get('/self', UserController::class . ':getCurrentUser');
+    $app->get("[/{pathParam:$pattern]", UserController::class . ':getUsers');
+    $app->put("[/{pathParam:$pattern]", UserController::class . ':updateUser');
```
GMishx, line 254: And if you are working on the previous change, please remove the `[]` from this path as the user information is a must for this endpoint unlike the `:getUsers`.

```suggestion
    $app->put("/{pathParam:$pattern", UserController::class . ':updateUser');
```

GMishx (changes requested):
@dvjsharma final 2 changes needed in the PR. Then I can merge it.

src/www/ui/api/Controllers/UserController.php
dvjsharma: Ah, I believe I overlooked that detail during testing as well. Fixed

Issue #2710, opened 2024-04-03
Upgrade the Folder & License APIs to Version 2


dushimsam:
# Upgrade the /folders , /obligations & /license APIs to Version 2

**In accordance with our new policy documented [here](https://github.com/fossology/fossology/issues/2404#issuecomment-1671046653), we want to enhance our REST APIs to Version 2 migrating all `/folders` & `/license` APIs. The setup for the Version 2 is already in place.**

#### 1.  **/folders**,  POST
<img width="743" alt="Screenshot 2024-04-03 at 06 35 29" src="https://github.com/fossology/fossology/assets/66276301/64d7cce4-f7a1-4944-933d-01db4de5a262">

- Move `parentFolder`, `folderName`, `folderDescription`, `groupName` from the headers to query parameters.

#### 2.  **/folders/{id}**,  DELETE
<img width="738" alt="Screenshot 2024-04-03 at 06 37 29" src="https://github.com/fossology/fossology/assets/66276301/ab0eb68c-03ca-4442-bb61-3cf72fd51f19">

- Move `groupName` from the headers to query parameters.

#### 3.  **/folders/{id}**,  PATCH
<img width="787" alt="Screenshot 2024-04-03 at 06 38 33" src="https://github.com/fossology/fossology/assets/66276301/ab30fe30-956c-461a-b9ee-ba930747ad6d">

- Move `name`, `description`, `groupName` from the headers to query parameters.

#### 4.  **/folders/{id}**,  PUT
<img width="780" alt="Screenshot 2024-04-03 at 06 40 40" src="https://github.com/fossology/fossology/assets/66276301/db8442b9-e85a-4719-9a35-78b8806e2922">

- Move `parent`, `action`, `groupName` from the headers to query parameters.

#### 5.  **/folders**,  GET
<img width="494" alt="Screenshot 2024-04-03 at 06 42 13" src="https://github.com/fossology/fossology/assets/66276301/5cdff416-2325-4d9c-8f3d-381e66c89acb">

- Move `groupName` from the headers to query parameters.

#### 6. **/folders/{id}**, GET
<img width="774" alt="Screenshot 2024-04-03 at 06 43 43" src="https://github.com/fossology/fossology/assets/66276301/31d70216-1e0b-4287-a11d-371ce92fb72f">

- Move `groupName` from the headers to query parameters.

#### 6. **/license**, GET

[...]

PR #2633, opened 2023-11-13
feat(api): (1) Upgrade Upload APIs to Version 2 

dvjsharma:
## Description

This PR addresses tasks 1, 3, 4, 5, 6, 9, 10, and 11 outlined in issue #2623 

### Changes

**Common for all: Moved `groupName` variable from headers to query parameters by making a change in  RestAuthMiddleware**

**Task 1: /updoads/{id}, GET:**
 - moved `groupName` variable  from headers to query params
 - changed the response keys to camelCase
 - modified the unit test to accommodate changes  

**Task 3: /updoads/{id}, DELETE:**

-  moved `groupName` variable  from headers to query params

**Task 4: /updoads/{id}, PATCH:**

-  moved `groupName` variable  from headers to query params

**Task 5: /updoads/{id}, PUT:**
-  moved `groupName`, `folderId`, `action` variable  from headers to query params
-  modified the unit test to accommodate changes  

**Task 6: /updoads, GET:** 
- moved `groupName`, `page`, `limit` variable  from headers to query params

**Task 9: /uploads/{id}/item/{itemId}/info, GET:**
- changed the response keys to camelCase 

**Task 10: /uploads/{id}/licenses, GET:**
-  moved `groupName`, `page`, `limit` variable  from headers to query params 
-  changed the response keys to camelCase 

**Task 11: /uploads/{id}/perm-groups, GET:**
-  updated the API Url to `/uploads/{id}/groups/permission`
-  changed the response keys to camelCase 

## How to test

- Changes pass all the unit tests
- Can be tested manually by making requests to the mentioned endpoints using swagger or postman

/cc @dushimsam @GMishx @shaheemazmalmmd

PR #2712, opened 2024-04-04
feat(api): Upgrade Folder, License & Obligation APIs to Version 2

dvjsharma:
## Description

This PR aims to upgrade the `/folder` and `/license` endpoints to v2 in accordance with the new upgrade guidelines.

### Changes

* Completed all tasks mentioned in #2710.
* Updated `openapiv2.yaml` to have all documentation changes.
* Modified models to make them version aware.

### How to Test

* Open the `openapiv2.yaml` file in Swagger Editor and test all the updated endpoints for v2 specific responses. All function as intended.
* Execute unit tests, all pass without any issues.

CC @GMishx @shaheemazmalmmd @dushimsam

### DIFF
--- a/src/www/ui/api/Controllers/LicenseController.php
+++ b/src/www/ui/api/Controllers/LicenseController.php
@@ -27,9 +27,12 @@ use Fossology\UI\Api\Exceptions\HttpNotFoundException;
 use Fossology\UI\Api\Helper\ResponseHelper;
 use Fossology\UI\Api\Models\Info;
 use Fossology\UI\Api\Models\InfoType;
+use Fossology\UI\Api\Models\ApiVersion;
 use Fossology\UI\Api\Models\License;
 use Fossology\UI\Api\Models\LicenseCandidate;
 use Fossology\UI\Api\Models\Obligation;
+use Fossology\UI\Api\Models\AdminAcknowledgement;
+use Fossology\UI\Api\Models\LicenseStandardComment;
 use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ServerRequestInterface as Request;
 use Slim\Psr7\Factory\StreamFactory;
@@ -151,8 +154,17 @@ class LicenseController extends RestController
    */
   public function getAllLicenses($request, $response, $args)
   {
+    $apiVersion = ApiVersion::getVersion($request);
     $query = $request->getQueryParams();
-    $limit = $request->getHeaderLine(self::LIMIT_PARAM);
+    if ($apiVersion == ApiVersion::V2) {
+      $page = $query[self::PAGE_PARAM] ?? "";
+      $limit = $query[self::LIMIT_PARAM] ?? "";
+      $onlyActive = $query[self::ACTIVE_PARAM] ?? "";
+    } else {
+      $page = $request->getHeaderLine(self::PAGE_PARAM);
+      $limit = $request->getHeaderLine(self::LIMIT_PARAM);
+      $onlyActive = $request->getHeaderLine(self::ACTIVE_PARAM);
+    }
     if (! empty($limit)) {
       $limit = filter_var($limit, FILTER_VALIDATE_INT);
       if ($limit < 1) {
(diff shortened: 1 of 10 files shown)


dvjsharma:
Few points to keep in mind regarding some upgrades: 

- I focused solely on resolving failing controller tests without making them version aware since many of them do not specifically assess the semantics of request or response objects, but rather the functionality, which remains largely consistent between v1 and v2. We can expand this coverage to accommodate v2 tests as well. But, creating separate tests with only a difference in parameter acceptance would result in a significant amount of duplicated code. Need thoughts on this. (refer [this](https://github.com/fossology/fossology/blob/master/src/www/ui_tests/api/Models/UploadTest.php), makes sense for models but not controllers. In-line integration might be helpful).

- Changing `file_input` to camelCase might break a lot of code. Will update more on this. (a thing left in this PR, will do)

- Migration of `groupName` from the headers to query parameters has already been globally completed (via Middleware) in #2633, so didn't touch it.

- for `/license/adminacknowledgements` GET & `/license/stdcomments` GET, response models do not exist. Although I did the change but need advice on should I create new models for them. (these are specific endpoints and new model won't be used anywhere, but to keep things consistent, IMO new model should be made).

Apart from these points, everything functions as anticipated.

GMishx:
Sorry for such a late response.

> Few points to keep in mind regarding some upgrades:
> 
>     * for `/license/adminacknowledgements` GET & `/license/stdcomments` GET, response models do not exist. Although I did the change but need advice on should I create new models for them. (these are specific endpoints and new model won't be used anywhere, but to keep things consistent, IMO new model should be made).

Having models for them as well, it is good keeping things consistent.

>     * Should I also modify `/folder` endpoints to use unique names instead folder IDs?

Folder names are not unique, thus they can operate on IDs. Also, the endpoints expose he folder ids in the response fields.

src/www/ui/api/Controllers/LicenseController.php
```diff
@@ -473,7 +483,8 @@ public function getAllAdminAcknowledgements($request, $response, $args)
     foreach ($res as $key => $ack) {
       $res[$key]['id'] = intval($ack['la_pk']);
       unset($res[$key]['la_pk']);
-      $res[$key]['is_enabled'] = $ack['is_enabled'] == "t";
+      $res[$key]['isEnabled'] = $ack['is_enabled'] == "t";
+      unset($res[$key]['is_enabled']);
```
GMishx, line 487: The field `is_enabled` is still required for v1 response, right?

src/www/ui/api/Controllers/LicenseController.php
```diff
@@ -578,8 +589,9 @@ public function getAllLicenseStandardComments($request, $response, $args)
     $res = $this->licenseStdCommentDao->getAllComments();
     foreach ($res as $key => $ack) {
       $res[$key]['id'] = intval($ack['lsc_pk']);
-      $res[$key]['is_enabled'] = $ack['is_enabled'] == "t";
+      $res[$key]['isEnabled'] = $ack['is_enabled'] == "t";
       unset($res[$key]['lsc_pk']);
+      unset($res[$key]['is_enabled']);
```
GMishx, line 594: The field `is_enabled` is still required for v1 response, right?

GMishx (changes requested):
Minor changes needed.

src/www/ui/api/Controllers/LicenseController.php
dvjsharma: Yes, I will create a new model for this response and address it there.

dvjsharma:
Hey @GMishx, I have added the following based on the last review:
1. The `/license/adminacknowledgements` GET and `/license/stdcomments` GET endpoints now each have their own models.
2. The `is_enabled` version bug is fixed.
3. Completed additional tasks recently added to #2710 

CC @dushimsam

GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

PR #2720, opened 2024-04-18
fix(reports): remove additional licenses in obligation section

shaheemazmalmmd:
## Description

Remove added licenses which are not identified from obligations.
Make default checked for scan only findings to improve the performance. 

## How to test

* Upload a package add dummy bulk findings
* ex: choose a license which doesnt exist in that package and add with dummy text.
* same license should not appear in reports


GMishx (approved):
Tested, working as expected.

GMishx (approved):
Changes looks good.

PR #2721, opened 2024-04-18
fix(conf): unify license names to match in reports

shaheemazmalmmd:
## Description

Deactivated licenses are still appearing in report. as the license names that are stored in report info and license names that we are comparing are different.

### Changes

## How to test

* Create a license name "gpl-2.0-with-classpath-exception" and add some obligations.
* Go to conf section and deactivate few obligations.
* Deactivated obligations should not appear in report.


GMishx (approved):
Changes looks good.

GMishx (approved):
Tested, working as expected.

Issue #2701, opened 2024-03-18
Reimport of an exported SPDX RDF report is not working (since version 4.2.1.45)


JanAltenberg:
### Description

Starting from 4.2.1.45 (7bd7ecba621d25741c6b08e53f6dd9a8328be173) reimporting an exported RDF report does not work properly. The reportImport job returns success, but nothing is being imported. A `git bisect` returns the following as the offending commit:

```
commit 7bd7ecba621d25741c6b08e53f6dd9a8328be173
Author: Gaurav Mishra <mishra.gaurav@siemens.com>
Date:   Mon Jan 23 22:39:12 2023 +0530

    fix(spdx): add license text for valid RDF
    
    Signed-off-by: Gaurav Mishra <mishra.gaurav@siemens.com>

 src/spdx2/agent/spdx2.php                          | 101 +++++++++++++++++++--
 .../agent/template/dep5-copyright-document.twig    |  12 +--
 src/spdx2/agent/template/spdx2-document.xml.twig   |  33 +++++--
 src/spdx2/agent/template/spdx2-file.xml.twig       |  55 ++++++++++-
 src/spdx2/agent/template/spdx2tv-document.twig     |  12 +--
 5 files changed, 182 insertions(+), 31 deletions(-)
 ```
 
This commit introduces `<spdx:ListedLicense>`, which is not being parsed when importing an RDF report.

#### How to reproduce

1. Take any FOSSology version >= 4.2.1.45
2. Upload and scan a package
3. Put your license decisions
4. Generate an SPDX RDF report
5. Upload the same package without any action
6. Do "Import Report" for the new upload and use the previously exported RDF report

### Versions

* Any version starting with commit 7bd7ecba621d25741c6b08e53f6dd9a8328be173

#### Job logs

The logs of the reportImport job show the following entries:

```
[...]
reportImport [0] :: JOB[96].reportImport[2613423.localhost]: "ERROR: can not handle license=[Array] of type=[array]"
reportImport [0] :: JOB[96].reportImport[2613423.localhost]: "ERROR: can not handle license=[Array] of type=[array]"
reportImport [0] :: JOB[96].reportImport[2613423.localhost]: "ERROR: can not handle license=[Array] of type=[array]"
reportImport [0] :: JOB[96].reportImport[2613423.localhost]: "ERROR: can not handle license=[Array] of type=[array]"
[...]

PR #2722, opened 2024-04-22
fix(reportImport): Parse for ListedLicense when importing an RDF report

JanAltenberg:
## Description

Since 4.2.1.45 the ListedLicense tag is being used for exported reports, but is not being parsed when importing a report. This prevents created RDF reports from being re-imported. This patch takes ListedLicense into account when importing an RDF report and treats it the same way as "License".

### Changes

Take "ListedLicense" into account when importing an RDF report and treat it the same way as "License".

## How to test

1. Take any FOSSology version >= 4.2.1.45
2. Upload and scan a package
3. Put your license decisions
4. Generate an SPDX RDF report
5. Upload the same package without any action
6. Do "Import Report" for the new upload and use the previously exported RDF report

With the change in this PR the import will work. Without the change the import agent will return success but nothing is being imported.

Fixes: #2701

### DIFF
--- a/src/reportImport/agent/SpdxTwoImportSource.php
+++ b/src/reportImport/agent/SpdxTwoImportSource.php
@@ -302,7 +302,8 @@ class SpdxTwoImportSource implements ImportSource
         return array($item);
       }
     }
-    elseif ($this->isPropertyOfType($license, 'License'))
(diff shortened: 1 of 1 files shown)


GMishx (approved):
Tested, working as expected.

Issue #2703, opened 2024-03-20
Importing an RDF report can fail with "shortname already in use" when matching with SPDX ID


JanAltenberg:
### Description
Commit d1e1b6e0dde4e37aa62e63b3246fd6b313e3d4d4 introduced an option to select the method for matching licenses when importing an RDF report (either by "SPDX ID" or by "Shortname").

```
commit d1e1b6e0dde4e37aa62e63b3246fd6b313e3d4d4
Merge: c669ce4f7 c173a05ce
Author: Shaheem Azmal M MD <shaheem.azmal@gmail.com>
Date:   Thu Jan 19 14:33:20 2023 +0530

    Merge pull request #2363 from siemens/feat/spdx/spdx-2-id
    
    feat(license): use spdx identifiers for reporting licenses
    
    Reviewed-by: shaheem.azmal@siemens.com
    Tested-by: shaheem.azmal@siemens.com
```

When selecting "SPDX ID", import of an RDF report can fail completely, if the report contains at least one license statement without an SPDX ID, that can be matched.

#### How to reproduce

1. Take any FOSSology version starting from commit d1e1b6e0dde4e37aa62e63b3246fd6b313e3d4d4
2. Choose a package for which you have an existing report
   * To reproduce the problem, the report has to contain at least one license statement that cannot be matched by SPDX ID (examples, that failed in my tests are: Public-domain, Dual-license, ...)
   * I tested with busybox-1.36.1 and took an existing report from the OSSelot project (https://www.osselot.org)
3. Upload the same package without any action
4. Do "Import Report" with the existing SPDX RDF report (and select "SPDX ID" under "Match license using")

The import will fail at the point, where the license statement can not be matched with an SPDX ID.

### Versions

Any version starting with commit d1e1b6e0dde4e37aa62e63b3246fd6b313e3d4d4

#### Job logs
```
reportImport [0] :: JOB[81].reportImport[2377839.localhost]: "Caught exception while processing uploadId=14: shortname already in use#0 /usr/local/share/fossology/reportImport/agent/ReportImportSink.php(105): Fossology\ReportImport\ReportImportSink->getIdForDataItemOrCreateLicense()"
reportImport [0] :: JOB[81].reportImport[2377839.localhost]: "#1 /usr/local/share/fossology/reportImport/agent/ReportImportAgent.php(291): Fossology\ReportImport\ReportImportSink->handleData()"
[...]

PR #2723, opened 2024-04-24
fix(reportImport): Adding a fallback to ShortName for RDF imports.

JanAltenberg:
## Description

When matching licenses by SPDX ID, an RDF import can fail completely if at least one license can not be matched by SPDX ID. Let's fall back to the ShortName in this case.

### Changes

The change adds a fallback to ShortName when a license can not be matched by SPDX ID.

## How to test

1. Take any FOSSology version starting from commit https://github.com/fossology/fossology/commit/d1e1b6e0dde4e37aa62e63b3246fd6b313e3d4d4
2. Choose a package for which you have an existing report
3. To test this change, the report has to contain at least one license statement that cannot be matched by SPDX ID (examples, that failed in my tests are: Public-domain, Dual license, ...)
4. I tested with busybox-1.36.1 and took an existing report from the OSSelot project (https://www.osselot.org/)
5. Upload the same package without any action
6. Do "Import Report" with the existing SPDX RDF report (and select "SPDX ID" under "Match license using")

With the change in this PR the import will work. Without the change the import agent will fail with "shortname already in use".

Fixes: #2703

### DIFF
--- a/src/reportImport/agent/ReportImportSink.php
+++ b/src/reportImport/agent/ReportImportSink.php
@@ -139,6 +139,10 @@ class ReportImportSink
     $licenseShortName = $dataItem->getLicenseId();
     if ($this->configuration->shouldMatchLicenseNameWithSPDX()) {
       $license = $this->licenseDao->getLicenseBySpdxId($licenseShortName, $groupId);
+      if ($license == null) {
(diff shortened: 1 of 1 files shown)


GMishx (approved):
Changes looks good. Tested, working as expected.

PR #2729, opened 2024-05-02
refactor(reportImport): use EasyRdf to read report

GMishx:
## Description

1. Use EasyRdf to navigate through RDF graph instead of reading it as an associative PHP array. This allows for more flexibility between SPDX versions.
2. Support reading Turtle files as EasyRdf already supports it.

### Changes

1. Move code in `SpdxTwoImportSource.php` from associative array to EasyRdf object.

## How to test

1. Test if reports generated from FOSSology can be imported.
2. Test if reports generated from other tools can also be imported.

### DIFF
--- a/src/lib/php/Util/StringOperation.php
+++ b/src/lib/php/Util/StringOperation.php
@@ -51,4 +51,16 @@ class StringOperation
   {
     return strncmp($haystack, $needle, strlen($needle)) === 0;
   }
+
+  /**
+   * Polyfill for PHP8's str_ends_with
+   * https://www.php.net/manual/en/function.str-ends-with.php
+   * @param string $haystack String to search in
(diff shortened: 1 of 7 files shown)


JanAltenberg:
Thanks @GMishx for taking this effort! :)
I started testing with a few reports from the OSSelot project and the imports looked good a first glance. Let me know, if I can support with some further testing.

GMishx:
Thanks @JanAltenberg , just wanted to test imports from other tools as there are different ways of relating files to a package in SPDX. Also, the file path naming used by FOSSology is different from other tools I've seen.

If this PR works for few most common tools then it will be a big plus for FOSSology.

shaheemazmalmmd (approved):
Code looks Good. Tested Works as expected.

Issue #2732, opened 2024-05-03
DB upgrade from < 4.3 to >= 4.3: Possible loss of edited results or main license


JanAltenberg:
### Description
When upgrading from version < 4.3 to 4.3 or later, obsolete identifiers will be renamed (e.g. GPL-2.0 -> GPL-2.0 only). If a license candidate exists with a shortname identical to the one to which we want to rename (e.g. we have a candidate with the name GPL-2.0-only), edited results and the main license can be lost for existing uploads.

#### How to reproduce

1. Install any FOSSology version < 4.3 (I tested with 4.1)
2. Create a license candidate with the name GPL-2.0-only
3. Upload a package with GPL-2.0 licensed files (I tested with busybox 1.36.1) and scan with nomos, monk and ojo for licenses
4. Clear a few files with a GPL-2.0 finding and set "Identified" for "Clearing decision type"
6. Select GPL-2.0 as main license
7. Upgrade your installation to >= 4.3. The fo-postinstall step will rename old identifiers (e.g. GPL-2.0 -> GPL-2.0-only)
8. Review the previously cleared package: The main license and the "Edited results" for all files, which had GPL-2.0 as a clearing decision, will be gone

### Versions

* The issue can be triggered when updating from < 4.3 to any version >= 4.3

### Further analysis

During the `fo-postinstall` step some functions in `fo_mapping_license.php` will be called. This will replace all obsolete license identifiers with their recent versions. Usually this step will just call `change_license_name()` for every obsolete identifier. But if the target name already exists (e.g. if there is a license candidate with that name), `update_license()` is being used. This will do the following steps:
1. Update rf_fk in the license_file table from the old to the new license
2. Remove the obsolete license from the license_ref table

Problem: `upload_clearing_license`, `clearing_event` and `license_set_bulk` wont' be updated at a later point.  `mergeCandidateLicense` (called from `initLicenseReftable`) won't update any of these tables, since no license with that name is existing, it will just call `insertNewLicense`.

As a result, e.g. the main license and / or edited results might be missing for existing uploads (the tables are still having references to the "old" license, which has been removed from the `license_ref` table)

Let me know if I can support with any further information.

Many thanks
Jan

PR #2743, opened 2024-05-22
fix(fo_mapping_license.php): Fix update_license()

JanAltenberg:
## Description

When upgrading from version < 4.3 to 4.3 or later, obsolete identifiers will be renamed (e.g. GPL-2.0 -> GPL-2.0 only). If a license candidate exists with a shortname identical to the one to which we want to rename (e.g. we have a candidate with the name GPL-2.0-only), edited results and the main license can be lost for existing uploads. This Pull request addresses this issue.

### Changes

The problem happens, because update_license() is not taking all relevant tables into account (only the license_file table is touched). This can be fixed by updating clearing_event, license_set_bulk and upload_clearing_license as well.

## How to test

1. Install any FOSSology version < 4.3 (I tested with 4.1)
2. Create a license candidate with the name GPL-2.0-only
3. Upload a package with GPL-2.0 licensed files (I tested with busybox 1.36.1) and scan with nomos, monk and ojo for licenses
4. Clear a few files with a GPL-2.0 finding and set "Identified" for "Clearing decision type"
5. Select GPL-2.0 as main license
6. Upgrade your installation to >= 4.3. The fo-postinstall step will rename old identifiers (e.g. GPL-2.0 -> GPL-2.0-only)
7. Review the previously cleared package

Without this fix the main license and the "Edited results" for all files, which had GPL-2.0 as a clearing decision, will be gone. With this fix applied, the update works as expected.

Fixes: #2732

### DIFF
--- a/install/db/fo_mapping_license.php
+++ b/install/db/fo_mapping_license.php
@@ -526,17 +526,26 @@ function update_license($old_rf_pk, $new_rf_pk)
 {
   global $PG_CONN;
 
+  $updateTables = array(
+    "clearing_event",
+    "license_file",
+    "license_set_bulk",
+    "upload_clearing_license"
+  );
+
   /** transaction begin */
   $sql = "BEGIN;";
   $result_begin = pg_query($PG_CONN, $sql);
   DBCheckResult($result_begin, $sql, __FILE__, __LINE__);
   pg_free_result($result_begin);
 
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):
Code looks good. Needs test.

PR #2745, opened 2024-05-24
fix(ReportUtils): Include files with 'No_license_found' as well

JanAltenberg:
## Description

Since the changes for de-duplication of licenses with the same SPDX ID in commit e01006b21646b4129b61bc72de7015e9714f0c71 , exported reports do not contain entries for files fith 'No_license_found'. These have to be included (and were included before the above mentioned commit).

### Changes

This commit adjusts the behaviour of addScannerResults() in a way, that files with 'No_license_found' will be reported.

## How to test

1. Take any FOSSology version starting with commit e01006b21646b4129b61bc72de7015e9714f0c71
2. Upload and clear a package (I've uploaded busybox-1.36.1.tar.bz2 and imported an existing RDF report from the OSSelot project)
3. Create a SPDX TV or RDF export

Without these changes the reports won't contain any entries for all files with 'No_license_found'  as scanner result (for busybox, Makefile.custom in the top level directory is such a file). With these changes applied, the reports will contain the relevant entries.

### DIFF
--- a/src/lib/php/Report/ReportUtils.php
+++ b/src/lib/php/Report/ReportUtils.php
@@ -134,8 +134,7 @@ class ReportUtils
     foreach ($rows as $row) {
       $reportedLicenseId = $this->licenseMap->getProjectedId($row['rf_fk']);
       $foundLicense = $this->licenseDao->getLicenseById($reportedLicenseId);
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):

PR #2747, opened 2024-05-28
fix(Report): Check the return value of getLicenseById()

JanAltenberg:
## Description

getCachedLicenseRisk() and getCachedLicenseText() do call getRisk() / getText() on $this->licenseCache[$licenseId] without checking the value before.

```
if (!array_key_exists($licenseId, $this->licenseCache)) {
  $this->licenseCache[$licenseId] = $this->licenseDao->getLicenseById($licenseId, $groupId);
  }
return $this->licenseCache[$licenseId]->getText();
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

In an error condition, this will make a report generation fail with a null pointer dereference.

### Changes
I propose to check $this->licenseCache[$licenseId] for null before calling getRisk() or getText(). In case $this->licenseCache[$licenseId] is null, we return null to the callers of getCachedLicenseRisk() and getCachedLicenseText(), so they can deal with it.

## How to test

Only the behaviour in an unexpected error condition is changed. No functional changes are expected. No visible changes for the users.

### DIFF
--- a/src/lib/php/Report/LicenseClearedGetter.php
+++ b/src/lib/php/Report/LicenseClearedGetter.php
@@ -199,7 +199,12 @@ class LicenseClearedGetter extends ClearedGetterCommon
     if (!array_key_exists($licenseId, $this->licenseCache)) {
       $this->licenseCache[$licenseId] = $this->licenseDao->getLicenseById($licenseId, $groupId);
     }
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):

PR #2749, opened 2024-05-30
feat(Installation Script): Enhance Cross-Platform Compatibility and Line Encoding Issue

Aaditya-Singh78:
## Description
This PR aims to update `fo-installdeps`, `install_composer` & `fo-cleanold` installation script to improve line encoding handling across windows. The goal is to ensure that these script function runs seamlessly across unix-like systems, [Windows Subsystem for Linux(WSL)](https://learn.microsoft.com/en-us/windows/wsl/), & other environments by managing line encoding & utilising MacOs, package manager, [Homebrew](https://brew.sh/) for installing project dependencies efficiently. 

### Changes

 - Implemented automatic detection and normalization of line endings using `dos2unix` to ensure scripts execute without errors across different platforms.
 - Added checks for the presence of necessary utilities (`curl`, `dos2unix`), with automated installation prompts if not found.
 - Adjusted script behavior based on user execution privileges, modifying default installation paths and permissions to suit different user levels.

## How to test
 - Ensure that the necessary utilities and configurations are correctly in place. Use the following commands to check the status of the installed services:
 
   #### For systemd-based systems:
         systemctl status fossology 
   #### For SysV-based systems:
        /etc/init.d/fossology status 

 - Confirm that the FOSSology Apache HTTPD configuration is properly loaded by the Apache server.Use this command to test the configuration:
   <pre> <code> apachectl configtest </code> </pre> 

CC:  [Gaurav Mishra](https://github.com/GMishx), [Kaushlendra Pratap](https://github.com/Kaushl2208)

### DIFF
--- a/utils/fo-cleanold
+++ b/utils/fo-cleanold
@@ -19,6 +19,28 @@
 # over or for testing.
 # See the help text for options.
 
+# Line Encoding
+required_utils=("curl" "dos2unix")
+for util in "${required_utils[@]}"; do
+    if ! command -v $util &> /dev/null; then
+        echo "$util is not installed. Attempting to install..."
+        if [[ $EUID -ne 0 ]]; then
+            echo "Unable to install $util Aborting..."
+            exit 1
+        else
+            case $(uname) in
+                Linux) apt-get install -y $util ;;
+                Darwin) brew install $util ;;
+                *) echo "Unsuported OS." ; exit 1 ;;
+            esac
+        fi
+    fi
+done
+
+if grep -qEi "(Microsoft|WSL)" /proc/version; then 
+    find . -type f -print0 | xargs -0 dos2unix
+fi
+
 ## Options parsing and setup
 # parse options
 OPTS=$(getopt -o h --long delete-conffiles,delete-database,delete-repository,delete-user,delete-everything,help -n 'fo-cleanold' -- "$@")
--- a/utils/fo-cleanold
+++ b/utils/fo-cleanold
@@ -19,8 +19,9 @@
 # over or for testing.
 # See the help text for options.
 
-# Line Encoding
-required_utils=("curl" "dos2unix")
+# Installing Package Manager for different Enviornment 
+required_utils=("dos2unix")
+
 for util in "${required_utils[@]}"; do
     if ! command -v $util &> /dev/null; then
         echo "$util is not installed. Attempting to install..."
@@ -28,18 +29,41 @@ for util in "${required_utils[@]}"; do
             echo "Unable to install $util Aborting..."
             exit 1
         else
-            case $(uname) in
-                Linux) apt-get install -y $util ;;
-                Darwin) brew install $util ;;
-                *) echo "Unsuported OS." ; exit 1 ;;
+            case $(uname -s) in
+                Linux)
+                    # Detecting Package Manager based on Enviornment
+                    if command -v apt-get &> /dev/null; then
+                        sudo apt-get install -y $util
+                    elif command -v yum &> /dev/null; then
+                        sudo yum install -y $util
+                    elif command -v dnf &> /dev/null; then
+                        sudo dnf install -y $util
+                    elif command -v pacman &> /dev/null; then
+                        sudo pacman -S --noconfirm $util
+                    else
+                        echo "No compatible package manager found. Aborting..."
+                        exit 1
+                    fi
+                    ;;
+                Darwin)
+                    if command -v brew &> /dev/null; then
+                        brew install $util
+                    else
+                        echo "Homebrew is not installed. Please install Homebrew to proceed."
+                        exit 1
+                    fi
+                    ;;
+                *)
+                    echo "Unsupported OS."
+                    exit 1
+                    ;;
             esac
         fi
     fi
 done
+# Line Encoding
+find . -type f -print0 | parallel -0 'grep -Iql $'\r'' {} && dos2unix {}'
 
-if grep -qEi "(Microsoft|WSL)" /proc/version; then 
-    find . -type f -print0 | xargs -0 dos2unix
-fi
 
 ## Options parsing and setup
 # parse options
--- a/utils/fo-installdeps
+++ b/utils/fo-installdeps
@@ -12,6 +12,28 @@
 # canonical location for such info, then fix it there first and also
 # update this file and the INSTALL document.
 
+# Line Encoding 
+required_utils=("curl" "dos2unix")
+for util in "${required_utils[@]}"; do
+    if ! command -v $util &> /dev/null; then
+        echo "$util is not installed. Attempting to install..."
+        if [[ $EUID -ne 0 ]]; then
+            echo "Unable to install $util Aborting..."
+            exit 1
+        else
+            case $(uname) in
+                Linux) apt-get install -y $util ;;
+                Darwin) brew install $util ;;
+                *) echo "Unsuported OS." ; exit 1 ;;
+            esac
+        fi
+    fi
+done
+
+if grep -qEi "(Microsoft|WSL)" /proc/version; then
+    find . -type f -print0 | xargs -0 dos2unix
+fi
+
 show_help() {
   cat <<EOF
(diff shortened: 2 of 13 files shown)


utils/fo-cleanold
```diff
@@ -19,6 +19,28 @@
...
+                Linux) apt-get install -y $util ;;
+                Darwin) brew install $util ;;
+                *) echo "Unsuported OS." ; exit 1 ;;
+            esac
+        fi
+    fi
+done
+
+if grep -qEi "(Microsoft|WSL)" /proc/version; then 
+    find . -type f -print0 | xargs -0 dos2unix
```
GMishx, line 41: This operation will be very costly. I will recommend to remove it.

utils/fo-cleanold
```diff
@@ -19,6 +19,28 @@
...
+for util in "${required_utils[@]}"; do
+    if ! command -v $util &> /dev/null; then
+        echo "$util is not installed. Attempting to install..."
+        if [[ $EUID -ne 0 ]]; then
+            echo "Unable to install $util Aborting..."
+            exit 1
+        else
+            case $(uname) in
+                Linux) apt-get install -y $util ;;
+                Darwin) brew install $util ;;
```
GMishx, line 33: Installation is not required for removal of tool??????

utils/fo-installdeps
```diff
@@ -12,6 +12,28 @@
...
+required_utils=("curl" "dos2unix")
+for util in "${required_utils[@]}"; do
+    if ! command -v $util &> /dev/null; then
+        echo "$util is not installed. Attempting to install..."
+        if [[ $EUID -ne 0 ]]; then
+            echo "Unable to install $util Aborting..."
+            exit 1
+        else
+            case $(uname) in
+                Linux) apt-get install -y $util ;;
```
GMishx, line 25: Would recommend checking underlying OS as well to get the correct package manager (as done for other dependencies.)

utils/fo-installdeps
```diff
@@ -12,6 +12,28 @@
...
+                Darwin) brew install $util ;;
+                *) echo "Unsuported OS." ; exit 1 ;;
+            esac
+        fi
+    fi
+done
+
+if grep -qEi "(Microsoft|WSL)" /proc/version; then
+    find . -type f -print0 | xargs -0 dos2unix
```
GMishx, line 34: I will recommend removing from here as well.

utils/fo-installdeps
GMishx: Also, curl will be installed as part of buildtime dependency. Any reason to install it separately here at the beginning of the script?

utils/fo-installdeps
```diff
@@ -85,6 +107,31 @@ CODENAME=$(lsb_release --codename --short)
 
 echo "install core dependencies"
 
+# Darwin
+if [[ $BUILDTIME && "$(uname)" == "Darwin" ]]; then
```
GMishx, line 111: Would recommend creating a new variable like `CODENAME` and `DISTRO` for `uname`. Something like `ARCH` sounds good?

utils/fo-installdeps
```diff
@@ -85,6 +107,31 @@ CODENAME=$(lsb_release --codename --short)
...
+   echo "*** Installing buildtime dependencies ***";
+   brew install curl libxml2 icu4c cmake ninja boost postgresql composer
+   brew install mxml cunit libmagic glib libpq cppunit
+
+   case "$CODENAME" in
+       stretch|buster|bullseye|bookworm|sid|focal|jammy)
```
GMishx, line 117: Should be replaced with correct codenames for Darwin?

GMishx: However, should not effect for homebrew?????

utils/fo-installdeps
GMishx: In mac, is there any equivalent for `lsb_release`? If not, the script will break at line 103 and 104.

utils/fo-installdeps
```diff
@@ -85,6 +107,31 @@ CODENAME=$(lsb_release --codename --short)
...
+
+   # Check if PostgreSQL development headers are needed
+   if ! brew list | grep -q postgresql; then
+       brew install postgresql
+   fi
+
+fi
+
+# wsl & linux 
 if [[ $BUILDTIME ]]; then
```
GMishx, line 135: Should check for **non-Darwin** here, currently runs everytime.

utils/fo-installdeps
```diff
@@ -155,6 +202,34 @@ fi
...
 ########################################################################
 
+# Darwin
+if [[ $RUNTIME && "$(uname)" == "Darwin" ]]; then
+    echo "*** Installing runtime dependencies ***"
+    # Basic Homebrew installations
+    brew install httpd php@7.4 postgresql libxml2 p7zip wget git svn unrar libboost
+
+    case "$CODENAME" in
+        stretch|buster|bullseye|bookworm|sid|focal|jammy)
```
GMishx, line 212: Same points as above.

utils/fo-installdeps
```diff
@@ -155,6 +202,34 @@ fi
...
+    brew services start httpd
+    brew services start postgresql
+
+    # Optional: Install Mail Transport Agent for email notifications
+    echo "*** Try 'brew install msmtp' or another preferred MTA ***"
+
+fi
+
+# wsl & linux 
 if [[ $RUNTIME ]]; then
```
GMishx, line 233: Should check **non-Darwin** here as well.

utils/install_composer.sh
```diff
@@ -21,6 +21,28 @@ version="2.2.6"
...
+required_utils=("curl" "dos2unix")
+for util in "${required_utils[@]}"; do
+    if ! command -v $util &> /dev/null; then
+        echo "$util is not installed. Attempting to install..."
+        if [[ $EUID -ne 0 ]]; then
+            echo "Unable to install $util Aborting..."
+            exit 1
+        else
+            case $(uname) in
+                Linux) apt-get install -y $util ;;
```
GMishx, line 34: Same point as above.

utils/install_composer.sh
```diff
@@ -21,6 +21,28 @@ version="2.2.6"
...
+                Darwin) brew install $util ;;
+                *) echo "Unsuported OS." ; exit 1 ;;
+            esac
+        fi
+    fi
+done
+
+if grep -qEi "(Microsoft|WSL)" /proc/version; then 
+    find . -type f -print0 | xargs -0 dos2unix
```
GMishx, line 43: Same as above.

GMishx (changes requested):
Other agent specific `mod_deps` files also needs to be updated.

You can find time with `find ./src -type f -name mod_deps`

utils/fo-installdeps
Aaditya-Singh78: Since, `curl` is installed during the build-time dependencies, there’s no need to include it in this script. I wasn't aware of this before I started modifying the script. I will make this modification in a new commit.

utils/fo-installdeps
Aaditya-Singh78: No, Lsb_release is not available in macOs. instead I'm trying to opt for `sw_vers -productVersion`, while creating a new varible like `codename`, `Distro` & `Arch`

utils/fo-installdeps
Aaditya-Singh78: No, according to the current code, it would not affect `Homebrew`. The `codenames` are used to handle specific Linux distributions like `Debian` and `Ubuntu`. Replacing `codenames` for `Darwin` would involve identifying version-specific actions for different macOS versions.

Therefore, I decided to proceed without adding complexity to the script, as the package manager can automatically detect dependencies.

utils/fo-installdeps
Aaditya-Singh78: I've suggested a new approach to writing that can identify files without line encoding and utilizes parallel processing. This is because the only obstacle to cross-platform compatibility between WSL and other Linux distributions is this issue alone.

PR #2760, opened 2024-06-13
feat(ubuntu): support ubuntu noble 24.04 for fossology

shaheemazmalmmd:
## Description

Add support for ubuntu jammy 22.04 for fossology.

### Changes

added noble support in installdeps file and many other.

## How to test

* Test every functionality of fossology using source install as well as debian install in Ubuntu 24.04

<a href="https://gitpod.io/#https://github.com/fossology/fossology/pull/2760"></a>


debian/control
GMishx, line 1: Please update the changes in `cmake/FoPackaging.cmake` as well.

GMishx (changes requested):

debian/control
shaheemazmalmmd: Done

GMishx (approved):
Changes looks good.

PR #2762, opened 2024-06-16
refactor(automation): Upgrade spdx_tools package from 0.8.0a2 to 0.8.…

rajuljha:
## Description

Upgraded `spdx-tools` from version 0.8.0a2 to 0.8.2

### Changes

Change spdx_tools version in requirements.txt in utils/automation

## How to test

Rebuild the image with the new version.

### DIFF
--- a/utils/automation/requirements.txt
+++ b/utils/automation/requirements.txt
@@ -3,4 +3,4 @@
 
 # SPDX-License-Identifier: FSFAP
 # Description: Dependencies used by the CI scanner script
-spdx-tools==0.8.0a2
+spdx-tools==0.8.2


rajuljha:
> Looks good but Why do we need to update to a new version? Let me know please? @rajuljha

Hey @Kaushl2208 
I was thinking along these lines:
* The current version that we are using is a pre-release version. The latest versions `0.8.2` is more [stable](https://github.com/spdx/tools-python/releases) and has many more bug fixes. 
* During the project call last week, @GMishx also mentioned that we can upgrade to the latest spdx_tools version.
* Also, I tested the same scan diff's by using both versions seperately and the latest version 0.8.2 is significantly fast in spitting out spdx_results. With Version [0.8.0a2](https://drive.google.com/file/d/1s8Yubvl-G5KKU52R0I3KCi3BP5Dtd1ph/view?usp=sharing), whereas with Version [0.8.2](https://drive.google.com/file/d/1knBnET0eicXNW-_8crhK1nnOcXGfBS5i/view?usp=sharing).
PS: Not able to add screenshot here for some reason 😢

GMishx (approved):
Changes looks good.

PR #2785, opened 2024-07-05
feat(nomos): Add start, end, len to nomos JSON output.

rajuljha:
## Description

* Add start, end and len information about scan results to nomos's json output.
* Need to give `-S` flag along with `-J` flag to populate the output.

### Changes

Add new function in `json_writer.c` to populate info in json output.

## How to test

Rebuild nomos scanner

### DIFF
--- a/src/nomos/agent/json_writer.c
+++ b/src/nomos/agent/json_writer.c
@@ -10,23 +10,48 @@
 #include "nomos_utils.h"
 #include <json-c/json.h>
 
+static void writeHighlightInfoToJson(GArray* theMatches, json_object *resultsArray) {
+    int currentLicence;
+    for (currentLicence = 0; currentLicence < theMatches->len; ++currentLicence) {
+      LicenceAndMatchPositions* theLicence = getLicenceAndMatchPositions(theMatches, currentLicence);
+
+      int highl;
+      for (highl = 0; highl < theLicence->matchPositions->len; ++highl) {
+        MatchPositionAndType* ourMatchv = getMatchfromHighlightInfo(theLicence->matchPositions, highl);
+        json_object *result = json_object_new_object();
+        json_object_object_add(result, "license", json_object_new_string(theLicence->licenceName));
+        json_object_object_add(result, "start", json_object_new_int(ourMatchv->start));
+        json_object_object_add(result, "end", json_object_new_int(ourMatchv->end));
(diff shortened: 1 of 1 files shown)


src/nomos/agent/json_writer.c
```diff
@@ -37,7 +47,8 @@ void writeJson()
     fileLocation = json_object_new_string(basename(cur.targetFile));
   }
   json_object_object_add(result, "file", fileLocation);
-  json_object_object_add(result, "licenses", licenses);
+  json_object_object_add(result, "results", resultsArray);
```
GMishx, line 50: Please do not change the key name as it will break existing parsers which read nomos output.

src/nomos/agent/json_writer.c
```diff
@@ -10,23 +10,33 @@
...
+static void writeHighlightInfoToJson(GArray* theMatches, json_object *resultsArray) {
+    int currentLicence;
+    for (currentLicence = 0; currentLicence < theMatches->len; ++currentLicence) {
+      LicenceAndMatchPositions* theLicence = getLicenceAndMatchPositions(theMatches, currentLicence);
+
+      int highl;
+      for (highl = 0; highl < theLicence->matchPositions->len; ++highl) {
+        MatchPositionAndType* ourMatchv = getMatchfromHighlightInfo(theLicence->matchPositions, highl);
+        json_object *result = json_object_new_object();
+        json_object_object_add(result, "license", json_object_new_string(theLicence->licenceName));
```
GMishx, line 22: If the `-S` flag for highlight is not passed, the output does not contain the name of licenses found in the file. This breaks backward compatibility. Maybe check if the `OPTS_HIGHLIGHT_STDOUT` is set and branch to use new code or existing code.

GMishx (changes requested):

src/nomos/agent/json_writer.c
rajuljha: Corrected.

src/nomos/agent/json_writer.c
rajuljha: I'll look into it.

rajuljha:
Hello @GMishx. Made the necessary changes. Here are the relevant Screenshots:

Screenshot 2024-07-16 at 1 47 25 PM
Screenshot 2024-07-16 at 1 47 45 PM
Screenshot 2024-07-16 at 1 48 04 PM

GMishx (approved):
Changes looks good.
Tested, working as expected.

PR #2815, opened 2024-08-28
fix(import): Do not require an agent if the report has no related data

martyone:
## Description

This fixes an error that happened when importing a FOSSology dump from other machine. On the receiving side data was uploaded from a Git repository without running any agents. The dump contained no copyright/ECC/IPRA data. Yet it kept failing with the "No agent 'copyright' exists on server." and similar errors until I manually ran the three agents on the upload.

### Changes

Bail out in the nothing-to-do case.

## How to test

1. Have two FOSSsology instances - sending and receiving
2. On the sending side have an upload with decisions but no Copyright/ECC/IPRA data. Export a FOSSology dump of it.
3. On the receiving side upload the same data using the From Version Control System option. Select NO optional analysis and no options below that
4. On the receiving side import the previously created dump

Original result:

Import fails with the error "No agent 'copyright' exists on server.". After running the "copyright agent" it goes further and complains about ECC, then about IPRA.

Fixed result:

It does not fail. (Though it would still fail if the sending side had any of those data - hence the FIXME note as a suggestion for further improvement.)

### DIFF
--- a/src/decisionimporter/agent/DecisionImporterDataCreator.php
+++ b/src/decisionimporter/agent/DecisionImporterDataCreator.php
@@ -276,12 +276,6 @@ class DecisionImporterDataCreator
                                       DecisionImporterAgent &$agentObj,
                                       string $agentName, int $jobId): void
   {
-    if (!$this->agentDao->arsTableExists($agentName)) {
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):

PR #2843, opened 2024-10-03
fix(created_by): Add created by column to candidate license table

deo002:
## Description

Add column "Created By" to Candidate Licenses Table.


GMishx (approved):
Changes looks good.

PR #2854, opened 2024-10-10
feat(scanoss): Adds proxy support

scanoss-qg:
## Description

This PR adds  proxy support for scanoss agent. The agent sets environment variable https_proxy with any value defined on http_proxy or https_proxy variables defined on fossology.conf file. The following configurations are supported:
- http_proxy = http://\<host>:\<port>
- http_proxy = http://\<user>:\<password>@\<host>:\<port>

### Changes

Followed guidelines from wget agent and added routines to set the environment variables

## How to test
1. Open fossology.conf file
2. Uncomment and set the right proxy addeess:port on http_proxy or https_proxy. Depending on the proxy configuration, potentially need to set up the credentials.
3. Reload scheduler configuration

### DIFF
--- a/src/scanoss/agent/main.c
+++ b/src/scanoss/agent/main.c
@@ -235,6 +235,20 @@ int main(int argc, char *argv[])
       if (RebuildUpload(upload_pk,tempFolder) != 0) /* process the upload_pk code */{
           LOG_ERROR("Error processing upload\n");
       } else {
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
@scanoss-qg : The agent still fails please have a look.

screenShot146

shaheemazmalmmd (approved):

PR #2856, opened 2024-10-15
chore(package): add compatibility agent deb package

GMishx:
## Description

Add packaging info for compatibility agent.

### Changes

Added variables for compatibility agent in cmake.

## How to test

Create and install debian packages.


shaheemazmalmmd (approved):
Code looks good.

PR #2857, opened 2024-10-15
chore(cmake): ignore anaconda paths

GMishx:
## Description

Ignore Anaconda paths as Python is not used in the build process and if defined, interferes with package finding.

### Changes

Ignore conda paths from search path if defined.

## How to test

1. Setup anaconda environment.
2. Setup CMake and build project.


avinal (approved):
Tested, working as expected.

PR #2871, opened 2024-11-25
Fix(ui): Fixed wrong pagination for 'Show Jobs' page & missing upload accessible check

andreas-menzl:
### Wrong pagination
Fixed a wrong pagination for the page `Show Jobs (/repo/?mod=showjobs)` if the currently selected group does not have access to all the uploads of the previous jobs.

This resulted in the following problem:

#### Page 2
grafik
#### Page 3
grafik
#### Page 4
grafik

The issue is that the `totalPages` are always calculated for all `jobs` from last month, but the jobs itself can be much less if the currently selected group does not have access to the uploads of the jobs. Also if the uploads are not accessible, it can result to pages displaying not `maxJobsPerPage` (currently 10), but instead one page can display 4 jobs, the other 10, the next 8 and so on. This is also caused by the fact the upload accessible check is made after getting 10 results from the database.

### Jobs for upload visible, even if job is not accessible
Also fixed a bug, which caused the jobs are always visible if a specific upload is given `(e.g. /repo/?mod=showjobs&upload=1234)`. Here the check for the accessibility of the group for the given upload was missing.

### Changes

`ShowJobsDao.php`:
* `uploads2Jobs function`: Check if the currently selected group has access to the given uploads
* `myJobs function`: Changed so it now receives first all jobs of the last `nHours (currently 4 weeks)` and afterwards check the uploads to be accessible. The `totalPages` and jobs per page are now calculated with only the accessible jobs.

`ShowJobsDaoTest.php`:
* Adjusted unit tests

**Note:**
It would also be possible to adjust the database statement to directly check the permissions and use the offset and limit to improve the performance. But since it currently used the `$this->uploadDao->isAccessible` i didnt change it, since i'm not sure if this is wanted. I thought you probably always want to use the `isAccessible` function to check for this.

I tested my solution of this PR with 10,000 jobs (1000 pages) and it was still quite performant (~1,3 seconds instead of 300ms the old).

## How to test

Change to a group which does not have access to all uploads of the previous jobs and it should still display the correct total pages for the pagination and 10 jobs per page.

### DIFF
--- a/src/lib/php/Dao/ShowJobsDao.php
+++ b/src/lib/php/Dao/ShowJobsDao.php
@@ -43,6 +43,13 @@ class ShowJobsDao
   function uploads2Jobs($upload_pks, $page = 0)
   {
     $jobArray = array();
+
+    // only use the uploads the user / group has access to
+    $upload_pks = array_filter($upload_pks, function($upload_pk) {
+      return $upload_pk !== null && $this->uploadDao->isAccessible($upload_pk, Auth::getGroupId());
+    });
+
+    // get count of upload pks, return empty array if count equals 0
     $jobCount = count($upload_pks);
     if ($jobCount == 0) {
       return $jobArray;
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd:
@andreas-menzl : can you please squash your commits to one ?? also change the commit message according to our contributing guidelines.

Kaushl2208 (approved):
Looks good to me. Functionally working as expected (bit slow) but accurate pages and jobs :)

PR #2877, opened 2024-12-13
Fix/2779 duplicate licence entry

### LINKED ISSUE #2779: Duplicate License-Entry for CMU-Mach
### Description

During review of our SPDX-Files generated using FOSSology, we noticed that a `CMU-Mach` license was referenced as `CMU` (without `-Mach`). We believe that this is incorrect. Currently the SPDX license list contains only licenses for the IDs `MIT-CMU` and `CMU-Mach` license, but not `CMU`. 

When I tried to understand how FOSSology came to that conclusion, I found that the [licenseRef.json](https://github.com/fossology/fossology/blob/master/install/db/licenseRef.json) file contains two entries with almost the same license text: one for `CMU-Mach` (starting at line 15909) and one for `CMU` (starting at line 1675).
Removing differences in line breaks and lower/upper case (see attached:
[FOSSology_CMU.txt](https://github.com/user-attachments/files/16081903/FOSSology_CMU.txt), [FOSSology_CMU-Mach.txt](https://github.com/user-attachments/files/16081904/FOSSology_CMU-Mach.txt)), gave this diff.
```diff
diff --git "a/.\\FOSSology_CMU-Mach.txt" "b/.\\FOSSology_CMU.txt"
index bafc084..49d4acb 100644
--- "a/.\\FOSSology_CMU-Mach.txt"
+++ "b/.\\FOSSology_CMU.txt"
@@ -1,5 +1,8 @@
-Copyright (c) 1991,1990,1989 Carnegie Mellon University
-All Rights Reserved.
+CMU License
+
+          Mach Operating System
+          Copyright \u00a9 1991,1990,1989 Carnegie Mellon University
+          All Rights Reserved.
 
 Permission to use, copy, modify and distribute this software and its
 documentation is hereby granted, provided that both the copyright
@@ -13,10 +16,9 @@ ANY DAMAGES WHATSOEVER RESULTING FROM THE USE OF THIS SOFTWARE.
 
 Carnegie Mellon requests users of this software to return to
 
- Software Distribution Coordinator  or  Software.Distribution@CS.CMU.EDU
- School of Computer Science
- Carnegie Mellon University
- Pittsburgh PA 15213-3890
-
-any improvements or extensions that they make and grant Carnegie Mellon
+           Software Distribution Coordinator
+           School of Computer Science
+           Carnegie Mellon University
+           Pittsburgh PA 15213-3890
+or Software.Distribution@CS.CMU.EDU any improvements or extensions that they make and grant Carnegie Mellon
 the rights to redistribute these changes.
```

I think that both entries should probably be merged or one removed in favor of the other.

laky241:
Fixes #2779  

A CMU-Mach license was incorrectly listed as CMU in the SPDX license reference. This caused confusion since the SPDX license list contains separate entries for MIT-CMU and CMU-Mach, but not for a general CMU license.

The entry labeled as CMU was updated to CMU-Mach in order to align with the correct license reference. We made sure that the license text, including the detailed description and conditions, was consistent with the CMU-Mach license.

updated the URLs for the licenses to ensure they correctly point to the corresponding license documentation pages. The CMU-Mach entry was linked to the official URL for the Mach Operating System license, while the CMU license now correctly references its own URL.

CC: @shaheemazmalmmd @GMishx


shaheemazmalmmd:
@laky241 : please check that the permission changes are also there in this branch. and also when you directly rename a license it will add a extra entry as it is not associated with any ID. 

also i remember that license names are directly getting stored in report_info. where we would need a migration script.(also needs a check for other places where the license name is directly getting stored.)

laky241:
@shaheemazmalmmd sir i am working on the changes you suggested from past 1 week and i am not able to fix that can you guide me a bit more on the changes, it will be a great help.

shaheemazmalmmd:
> @shaheemazmalmmd sir i am working on the changes you suggested from past 1 week and i am not able to fix that can you guide me a bit more on the changes, it will be a great help.

@laky241 : 

1) The changes you have created for permissions are here in this branch. 
2) A simple rename will create a new entry in database, can you please confrim that there is no duplicate entry for CMU and CMU-Mach
3) In report_info table we store the license names directly not the licenseid. a small script is needed to replace the license CMU with CMU-Mach in this case.

PR #2914, opened 2025-01-10
feat(NPMParser): Add a Parser to parse NPM packages and download them…

rajuljha:
## Description

Add NPM Parser for parsing and downloading NPM Packages

### Changes

- Add class NPMParser with a few helper methods.
- Used [package-url](https://github.com/package-url/packageurl-python) package to get download urls from purls

## How to test

Run  `fossologyscanner.py nomos ojo copyright keyword scan-only-deps --sbom-path <test_sbom.json>`
It will currently download all the packages of the type NPM and PyPI, It'll classify the rest as unsupported for now.


utils/automation/ScanDeps/Parsers.py
```diff
@@ -131,3 +132,40 @@ def parse_components(self, components: list[Dict]) -> Union[list[tuple[str,str]]
...
+        """
+        Get download url from purl for NPM Packages
+        Args:
+            purl: str
+        Return:
+            download_url: str
+        """
+        return purl2url.get_download_url(purl)
+    
+    def parse_components(self, components: list[Dict]) -> Union[list[tuple[str,str]],None]:
```
Kaushl2208, line 153: Can this function be used as a util function? NPM and Python both are using the same logic.

utils/automation/fossologyscanner.py
```diff
@@ -362,16 +362,18 @@ def main(parsed_args):
...
-    if len(python_comps) != 0:
+    if len(parser.python_components) != 0:
       python_parser = PythonParser()
-      python_list = python_parser.parse_components(python_comps)
+      python_list = python_parser.parse_components(parser.python_components)
       download_list += python_list
 
-    if len(unsupported_comps) != 0:
-      for comp in unsupported_comps:
+    if len(parser.npm_components) != 0:
```
Kaushl2208, line 370: python takes empty lists as `false`
`if parser.python_components:` will also do the same job.

Kaushl2208 (changes requested):
Also rebase it with latest master.

utils/automation/ScanDeps/Parsers.py
rajuljha: At first, I implemented it via a Mixin class, but it led to unnecessary inheritance, so I thought maybe we implement the method with the same name without inheritance itself.

shaheemazmalmmd:
@rajuljha : Please check the failing test cases.

Kaushl2208 (approved):
Looks good, Tested! Working as expected.

PR #2921, opened 2025-01-15
fix(api): Jobs no longer displayed wrongly as queued

andreas-menzl:
## Description

Fixed a bug of the previous PR #2870 from me. This caused jobs to be displayed as "queued" if the do not have any item inside the `jobqueue` table.

### Changes
* `DbHelper.php`: Changed the `join statement` from `left` to `right` to make sure only jobs with at least one queue item are displayed as queued.


shaheemazmalmmd (approved):
Changes looks good.

PR #2940, opened 2025-01-24
fix(ci): setup php before syntax check

GMishx:
## Description

Setup PHP to minimum supported PHP version before running syntax check.

### Changes

Setup PHP to minimum supported PHP version before running syntax check.

## How to test

Check CI


Kaushl2208 (approved):
Looks good!

Issue #1299, opened 2019-03-09
nomos CLI tries to read Db.conf file in, despite no DB Connection is neccessary


maxhbr:
### Description

If one starts the latest vagrant images with https://github.com/fossology/fossology/commit/69629cc1f45f2cbde7e1318dba8cf96a3ac3de8a one can observe the following behaviour:
```
$ cd /fossology
$ ./src/nomos/agent/nomos -h
FATAL libfossscheduler.c.279: unable to open database connection: Database conf file: /usr/local/etc/fossology/Db.conf, Permission denied
```

### Solution
The CLI tools should not try to open the conf file, if no DB connection is necessary.

mcjaeger:
note that a first step has been done: https://github.com/fossology/fossology/pull/1293

SalmanDeveloperz:
Hello,

I’ve submitted a pull request to resolve this issue: [#1299](https://github.com/fossology/fossology/issues/1299).

The PR updates the `nomos` CLI to avoid unnecessary database connections when they are not required, such as when using the help command (`-h`). This fixes the permission error and enhances overall CLI performance.

PR Link: [#2947](https://github.com/fossology/fossology/pull/2947)

Thank you, and I look forward to your feedback.

PR #2947, opened 2025-02-01
fix: prevent nomos CLI from connecting to DB when not required (#1299)

SalmanDeveloperz:
## Description

This pull request addresses **issue #1299** by preventing the `nomos` CLI from attempting to read the `Db.conf` file when a database connection is not required. Specifically, the CLI no longer tries to open a DB connection when displaying the help message (`-h` flag), which previously caused a permission error.

### Changes

- Introduced a global flag `should_connect_to_db` in `libfossscheduler.c` to control whether a database connection should be initiated.
- Modified the `Usage()` function in `nomos.utils.c` to set `should_connect_to_db = 0` when the help command (`-h`) is used.
- Updated the `fo_scheduler_connect_conf()` function to respect this flag, ensuring the database connection is only attempted when necessary.

---

## How to test

1. **Display Help Message (`-h`):**  
   Run the following command to verify that the help message is displayed without any DB connection attempt or permission error:
   ```bash
   ./src/nomos/agent/nomos -h
![Screenshot from 2025-02-01 12-09-39](https://github.com/user-attachments/assets/e6a5d1c5-92a6-4ce7-a0ba-d61899fb1d5a)

### DIFF
--- a/src/lib/c/libfossscheduler.c
+++ b/src/lib/c/libfossscheduler.c
@@ -34,6 +34,7 @@ int userID;          ///< The id of the user that created the job
 int groupID;         ///< The id of the group of the user that created the job
 int jobId;           ///< The id of the job
 char* module_name = NULL;   ///< The name of the agent
+int should_connect_to_db = 1; /// Default: connect to DB unless explicitly disabled
 
 /** Check for an agent in DB */
 const static char* sql_check = "\
@@ -252,8 +253,8 @@ void fo_scheduler_connect_conf(int* argc, char** argv, PGconn** db_conn, char** 
     fo_config_free(version);
   }
 
-  /* create the database connection */
-  if (db_conn)
+  /* create the database connection only if needed */
+  if (db_conn && should_connect_to_db)  // Add a condition
   {
     db_config = g_strdup_printf("%s/Db.conf", sysconfigdir);
     (*db_conn) = fo_dbconnect(db_config, &db_error);
--- a/src/nomos/agent/nomos_utils.c
+++ b/src/nomos/agent/nomos_utils.c
@@ -11,6 +11,8 @@
 #include "nomos_utils.h"
 #include "nomos.h"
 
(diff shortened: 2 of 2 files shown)


src/lib/c/libfossscheduler.c
```diff
@@ -34,6 +34,7 @@ int userID;          ///< The id of the user that created the job
 int groupID;         ///< The id of the group of the user that created the job
 int jobId;           ///< The id of the job
 char* module_name = NULL;   ///< The name of the agent
+int should_connect_to_db = 1; /// Default: connect to DB unless explicitly disabled
```
dvjsharma, line 37: Shouldn't this be a `bool` variable instead?

src/nomos/agent/nomos_utils.c
```diff
@@ -11,6 +11,8 @@
 #include "nomos_utils.h"
 #include "nomos.h"
 
+extern int should_connect_to_db;  /* Global variable to control DB connection */
```
dvjsharma, line 14: Ditto

SalmanDeveloperz: Thanks for the feedback! Just to confirm, should I update the `should_connect_to_db` variable to a `bool` in `nomos_utils.c` as well, and include the `<stdbool.h>` header if necessary? Please let me know if that's the intended change.

src/lib/c/libfossscheduler.c
dvjsharma: Hmm I thought that was already there, just saw `valid` is also used as type `int`. The change looks good then, no need to modify the type.

Kaushl2208 (approved):
Changes looks good, Needs test :)

dvjsharma (approved):
Changes looks good.

shaheemazmalmmd:
@SalmanDeveloperz : please change the commit message according to our [contributing guidelines](https://github.com/fossology/fossology/blob/master/CONTRIBUTING.md)

PR #2951, opened 2025-02-03
fix(copyright): preventing 'é' from being detected as ©

### LINKED ISSUE #1813: Copyright agent: many false positives with 'é' characters
### Description

I'd like to share a fix to reduce many false positives, but introduces a light risk.

The Copyright agent scans the data byte by byte to detect words as well as specific '©' (followed by space) characters.
This, without knowing if it's scanning UTF8, Unicode, or whatever.

The code is here: https://github.com/fossology/fossology/blob/b7a846ca391b091407b8173ec0f74a1dbab5bc46/src/copyright/agent/copyright.conf#L30

**Now:**
- [UTF8](https://www.utf8-chartable.de/): '©' is `C2 A9`
- [Unicode](https://www.codetable.net/unicodecharacters): '©' is `A9`

So the agents detects all `A9` followed by space characters (to get Unicode symbol).

**But:**
- UTF8 `é` is `c3 a9`:
Therefore the agents also matches all UTF8 `é` == `c3 a9` followed by a space (that is *very* common in French)

To prevent this, a [negative look behind](https://www.regular-expressions.info/lookaround.html) can be added to `A9` in the regex: `(?<!\xC3)\xA9`). This prevents selecting `A9` byte if `C3` is preceding it.

**However:**
- Unicode: `Ã` is `c3`:
- So the Unicode sequence `Ã©` is the same as the UTF8  `é` : `c3 a9`
- So with the hack above, we would miss the Unicode sequence `Ã© + space` (that I don't think should be a usual copyright statement)

**Hence the question:** is it acceptable to improve false positive (not detect French `é + space`) while risking missing `Ã©` sequence ?

#### How to reproduce
- Scanning a text with: *Votre compte a été suspendu*
- Gives:
image

laky241:
Fixes #1813 

## Description
 the issue where the Copyright agent mistakenly detects  **"é + space"** as the copyright symbol **"©"**.

### Changes
Added a **negative lookbehind** (`(?<!\xC3)\xA9`) to ensure that **A9 is not matched when come after C3** 

## How to test
 `"Votre compte a été suspendu"`  **Should NOT be detected as copyright**

### DIFF
--- a/src/copyright/agent/copyright.conf
+++ b/src/copyright/agent/copyright.conf
@@ -25,7 +25,7 @@ author=(?:__author__)[:]?
 author=__author____SPACESALL____NAMESLIST__\.?
 author=__author__|(?<=<author>)(.*?)(?=<\/author>)
 #
-COPYSYM=(?:\(c\)|&copy;|\xA9|\xC2\xA9|\$\xB8|\xE2\x92\xB8|\$\xD2|\xE2\x93\x92|\$\x9E|\xE2\x92\x9E)
+COPYSYM=(?:\(c\)|&copy;|(?<!\xC3)\xA9|\xC2\xA9|\$\xB8|\xE2\x92\xB8|\$\xD2|\xE2\x93\x92|\$\x9E|\xE2\x92\x9E)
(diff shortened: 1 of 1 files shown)


GMishx:
Interesting companion for #2770

/cc @shaheemazmalmmd

laky241:
@shaheemazmalmmd made the changes

shaheemazmalmmd (approved):

PR #2953, opened 2025-02-11
refactor(ui): refactor copyright and conclude license page

its-sushant:
## Description

Refactor copyright and conclude license page

### Changes

- Made the delete and checkbox column clickable in copyright page.
- Hide not frequently used options in conclude license page


src/copyright/ui/template/histTable.js.twig
```diff
@@ -141,6 +141,32 @@ function createTable{{ table.type }}() {
     }
   });
 
+  $('#copyright{{ table.type }} tbody').on('click', 'tr td:nth-child(3)', function (event) {
+    if ($(this).find('img:visible').length > 0) {
+      if ($(event.target).is('img')) {
+        return;
+      }
+      var currentValue = $(this).closest('tr').attr('id');
+      var currUploadData = currentValue.split(',');
```
Kaushl2208, line 150: Assuming `currentValue` contains values with `,` can lead to unexpected errors. A defensive check would be better.

Kaushl2208 (changes requested):
Just a small change sushant. Everything else looks good.
Testing the working

src/copyright/ui/template/histTable.js.twig
its-sushant: Thank you for the review @Kaushl2208. Yes, we can add a defensive check, but `currentValue `only contains structured values i.e. `upload_id, item_id, item_hash, item_kind`. Since it doesn’t include arbitrary text or unexpected characters, I’m not sure if an additional check is necessary. Let me know if my understanding is correct.

PR #2959, opened 2025-02-14
enabled partial file searching

### LINKED ISSUE #2954: File searching works with Exact Match not with Partial Matching
**Issue Description:**

The search currently requires an exact filename match, making it difficult to find files efficiently.

**Expected Behavior:**

The search should support partial matching (e.g., searching "licen" should return "license.txt") to improve usability.

Steps to Reproduce:

- Navigate to the File Browser in Fossology.

- Try searching for a file using only a partial filename (e.g., searching "licen" for "license.txt").

- Notice that the search does not return results unless the exact filename is type

laky241:
Fixes #2954 

## Description
Fossology only finds folders if you type the exact name.
This PR enables the file search with partial name of the file 
### Changes

Fossology used = in the search.
I changed it to LIKE, which allows partial matching.
Added (%) This makes search for folders even if you type only part of the name

### DIFF
--- a/src/lib/php/Dao/FolderDao.php
+++ b/src/lib/php/Dao/FolderDao.php
@@ -69,19 +69,19 @@ class FolderDao
   }
 
   public function getFolderId($folderName, $parentFolderId = self::TOP_LEVEL)
-  {
-    $statementName = __METHOD__;
-    $this->dbManager->prepare($statementName,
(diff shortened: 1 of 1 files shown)


its-sushant:
Can we close this @shaheemazmalmmd

PR #2970, opened 2025-02-25
Enhance Error Logging in Nomos License Detection

laky241:
Nomos agent scans a file for license detection, failures with messages like "License match detection failed." this lack context,

 updated the logging in **AgentLicenseEventProcessor.php** to provide more details when a license is not detected.

Changes made:

-  Added file name to logs.

- Included a text snippet of the scanned content


shaheemazmalmmd:
@laky241 : these changes are not required as the file changes only concentrates on license checking not detection. so logging is not required here.

PR #2990, opened 2025-03-09
Removed unnecessary log

### LINKED ISSUE #2027: Remove unnecessary log using maintenance agent
### Description

Whenever we use re-use agent, Some unnecessary  logs gets created. which is not required though.
also we can remove the logs which are older then 1 year or 6 months.

* Upload a package using some older package through re-user agent.
* After the scan gets completed click on job id and check the log.

#### Version

On latest master

laky241:
Fixes #2027 

removed redundant log outputs in the re-use agent
echo statement resulted in excessive logs


shaheemazmalmmd:
@laky241 : The provided fix doesn't satisfy the requirement. the log removed here is required as this gets created only one time per upload and that too in the case of selected reuse config and data doesn't.

Issue #2988, opened 2025-03-06
error : 500 error fetching the uploaded files  at  browse page


ChayanDass:
error in fetching the uploads , 

![Image](https://github.com/user-attachments/assets/1b45b870-82b0-4c65-a483-6fb89e6caef3)

Chayan8837:
@shaheemazmalmmd ,   can you check it once?

shaheemazmalmmd:
@Chayan8837 : Can you paste the Apache and fossology logs ?? with this minimal information i will not be able to figure out what is happening at your end ?

Chayan8837:
sudo tail -f /var/log/apache2/error.log
[Thu Mar 06 12:44:43.357197 2025] [php:warn] [pid 20987] [client ::1:36260] PHP Warning:  Trying to access array offset on false in /usr/local/share/fossology/scanoss/ui/agent-scanoss.php on line 73, referer: http://localhost/repo/?mod=home
[Thu Mar 06 12:44:43.420745 2025] [php:warn] [pid 20987] [client ::1:36260] PHP Warning:  Trying to access array offset on false in /usr/local/share/fossology/scanoss/ui/agent-scanoss.php on line 73, referer: http://localhost/repo/?mod=browse
[Thu Mar 06 12:44:44.972736 2025] [php:warn] [pid 20987] [client ::1:36260] PHP Warning:  Trying to access array offset on false in /usr/local/share/fossology/scanoss/ui/agent-scanoss.php on line 73, referer: http://localhost/repo/?mod=browse
[Thu Mar 06 12:44:45.708373 2025] [php:warn] [pid 20987] [client ::1:36260] PHP Warning:  Trying to access array offset on false in /usr/local/share/fossology/scanoss/ui/agent-scanoss.php on line 73, referer: http://localhost/repo/?mod=browse
[Thu Mar 06 12:44:45.897755 2025] [php:warn] [pid 20987] [client ::1:36260] PHP Warning:  Trying to access array offset on false in /usr/local/share/fossology/scanoss/ui/agent-scanoss.php on line 73, referer: http://localhost/repo/?mod=browse
[Thu Mar 06 12:44:46.078603 2025] [php:warn] [pid 20987] [client ::1:36260] PHP Warning:  Trying to access array offset on false in /usr/local/share/fossology/scanoss/ui/agent-scanoss.php on line 73, referer: http://localhost/repo/?mod=browse
[...]

PR #2991, opened 2025-03-09
fix(Convert2BrowserTime): improve error handling and validation

Vaibhavsahu2810:
## Description
This pull request fixes issue #2988, which caused a 500 error when fetching uploaded files on the browse page. The issue was due to the Convert2BrowserTime function lacking proper validation and error handling.

## Changes

- Added validation for $server_time to prevent crashes when an empty or invalid value is passed.
- Ensured session status is checked before accessing $_SESSION["timezone"] to avoid undefined index errors.
- Verified the timezone against timezone_identifiers_list() to prevent exceptions from invalid timezones.
- Added error logging to improve debugging.

## How to test

- Ensure the application is running and accessible.
- Upload a file and navigate to the browse page.
- Verify that the uploaded files are displayed without triggering a 500 error.

### DIFF
--- a/src/lib/php/common-ui.php
+++ b/src/lib/php/common-ui.php
@@ -309,13 +309,35 @@ function Get1stUploadtreeID($upload)
  * \brief Convert the server time to browser time
  * \param time $server_time to be converted
  */
-function Convert2BrowserTime($server_time)
-{
-  $server_timezone = date_default_timezone_get();
-  $browser_time = new \DateTime($server_time, new \DateTimeZone($server_timezone));
-  if (array_key_exists("timezone", $_SESSION)) {
-    $tz = $_SESSION["timezone"];
-    $browser_time->setTimeZone(new \DateTimeZone($tz));
-  }
-  return $browser_time->format('Y-m-d H:i:s');
+function Convert2BrowserTime($server_time) {
+    try {
+        if (empty($server_time)) {
+            throw new \InvalidArgumentException('Server time cannot be empty');
+        }
+        $server_timezone = date_default_timezone_get();
+
+        try {
+            $browser_time = new \DateTime($server_time, new \DateTimeZone($server_timezone));
+        } catch (\Exception $e) {
+            error_log("Invalid server time format: $server_time");
+            return false;
+        }
+
+        if (session_status() === PHP_SESSION_ACTIVE && 
+            isset($_SESSION["timezone"])) {
+            $tz = $_SESSION["timezone"];
+
+            if (in_array($tz, timezone_identifiers_list())) {
+                $browser_time->setTimeZone(new \DateTimeZone($tz));
+            } else {
+                error_log("Invalid timezone in session: $tz. Falling back to server timezone.");
+                $browser_time->setTimeZone(new \DateTimeZone($server_timezone));
+            }
+        }
+
+        return $browser_time->format('Y-m-d H:i:s');
+    } catch (\Exception $e) {
+        error_log("Error in Convert2BrowserTime: " . $e->getMessage());
+        return false;
+    }
 }


Chayan8837:
nice work

shaheemazmalmmd:
@Vaibhavsahu2810 : check the failing test cases.

Vaibhavsahu2810:
> @Vaibhavsahu2810 : check the failing test cases.
Done.

shaheemazmalmmd:
@Vaibhavsahu2810 : can you please squash the commits to one ?

src/lib/php/common-ui.php
```diff
@@ -311,11 +311,34 @@ function Get1stUploadtreeID($upload)
...
+    if (empty($server_time)) {
+      throw new \InvalidArgumentException('Server time cannot be empty');
+    }
+    $server_timezone = date_default_timezone_get();
+
+    try {
+      $browser_time = new \DateTime($server_time, new \DateTimeZone($server_timezone));
+    } catch (\Exception $e) {
+      error_log("Invalid server time format: $server_time");
+      return false;
```
Kaushl2208, line 324: If a failure occurs, throw an exception instead of returning `false`. This ensures consistent return values and shifts error handling to the caller.

Or Better :
If there is an exception `return $server_time` with the format: `format('Y-m-d H:i:s')`

This also satisfies the goal of, browser time into server timezone.

src/lib/php/common-ui.php
```diff
@@ -311,11 +311,34 @@ function Get1stUploadtreeID($upload)
...
+      } else {
+        error_log("Invalid timezone in session: $tz. Falling back to server timezone.");
+        $browser_time->setTimeZone(new \DateTimeZone($server_timezone));
+      }
+    }
+
+    return $browser_time->format('Y-m-d H:i:s');
+  } catch (\Exception $e) {
+    error_log("Error in Convert2BrowserTime: " . $e->getMessage());
+    return false;
```
Kaushl2208, line 342: Same here.

src/lib/php/common-ui.php
```diff
@@ -311,11 +311,34 @@ function Get1stUploadtreeID($upload)
...
+  try {
+    if (empty($server_time)) {
+      throw new \InvalidArgumentException('Server time cannot be empty');
+    }
+    $server_timezone = date_default_timezone_get();
+
+    try {
+      $browser_time = new \DateTime($server_time, new \DateTimeZone($server_timezone));
+    } catch (\Exception $e) {
+      error_log("Invalid server time format: $server_time");
```
Kaushl2208, line 323: You have added multi layer of `try catch` with general catches, without any specific Exception defined. The exception message says, `Invalid Time Format`. It doesn't align and bit erratic, can lead to wild chase of thinking only time format is wrong.

Also these two `try-catch` can be flattened, You are already checking 
`if (empty($server_time)) {` and also throwing a `InvalidArgumentException`, Just a simple if condition to check  format as well. Will lead to flatten the `try-catch` structures.

Kaushl2208 (changes requested):
The function `Convert2BrowserTime` is a dependency on a lot of other functions. Where the change in return style is not expected. Please take a look into this.

Kaushl2208:
Also, Update the feature branch with latest master.
Squash multiple commits and follow the commit guidelines

src/lib/php/common-ui.php
Vaibhavsahu2810: removed all false returns. now it wil return formatted date('Y-m-d H:i:s'), ensuring consistent return values.

src/lib/php/common-ui.php
Vaibhavsahu2810: done

src/lib/php/common-ui.php
Vaibhavsahu2810: flattened the try-catch by handling the empty check and wrapping only the DateTime logic in a single try block. used InvalidArgumentException and UnexpectedValueException with clear messages.

Vaibhavsahu2810:
Done made all fixes :)

shaheemazmalmmd (approved):
Tested works as expected.

Issue #1704, opened 2020-04-27
pkgagent printing HTML encoded characters


GMishx:
### Description

Checking the info of a Debian package parsed by `pkgagent` prints HTML encoded characters.

#### How to reproduce

Upload a package with Debian source package (DSC) and check the `Info` page for the source file.

#### Screenshots

![snap](https://user-images.githubusercontent.com/18077542/80333647-658ff480-886c-11ea-89d2-e1d60140f50d.png)

### Versions

* Last commit id on master: 9cfd4ecb15a694e007dff37c5b5dab6c718ee49d

ritankarsaha:
@GMishx could you provide with a package having with debian source packaging for testing. I have tried solving this issue and also provided with a solution

PR #2998, opened 2025-03-12
Fix(pkgagenterror): pkgagent printing html encoded characters

ritankarsaha:
This PR fixes the issue #1704 

## Description
This PR addresses the issue where package metadata, particularly in Debian source packages, is being displayed with double-encoded HTML entities. For example, angle brackets were showing as `&amp;lt;` and `&amp;gt;` instead of their proper representation.

### Changes
- Removed all `htmlentities()` calls in the `ShowPackageInfo()` function
- this prevents double-encoding of metadata that's already encoded when stored or retrieved from the database
- Affects display of RPM, Debian binary, and Debian source package metadata

### DIFF
--- a/src/www/ui/ui-view-info.php
+++ b/src/www/ui/ui-view-info.php
@@ -432,7 +432,6 @@ class ui_view_info extends FO_Plugin
           $Count++;
           $vars['packageEntries'][] = $entry;
         }
-        pg_free_result($result);
 
         $sql = "SELECT * FROM pkg_deb_req WHERE pkg_fk = $1;";
         $this->dbManager->prepare(__METHOD__ . "getPkg_rpm_req", $sql);
@@ -469,7 +468,6 @@ class ui_view_info extends FO_Plugin
           $Count++;
           $vars['packageEntries'][] = $entry;
         }
-        pg_free_result($result);
 
         $sql = "SELECT * FROM pkg_deb_req WHERE pkg_fk = $1;";
         $this->dbManager->prepare(__METHOD__ . "getPkg_rpm_req", $sql);
--- a/src/www/ui/ui-view-info.php
+++ b/src/www/ui/ui-view-info.php
@@ -26,7 +26,6 @@ class ui_view_info extends FO_Plugin
    * User DAO to use
    */
   private $userDao;
-
   function __construct()
   {
     $this->Name    = "view_info";
@@ -394,7 +393,7 @@ class ui_view_info extends FO_Plugin
           $entry = [];
           $entry['count'] = $Count;
           $entry['type'] = _($key);
-          $entry['value'] = htmlentities($R["$value"]);
+          $entry['value'] = htmlspecialchars($R["$value"], ENT_QUOTES|ENT_HTML5, 'UTF-8');
           $Count++;
           $vars['packageEntries'][] = $entry;
         }
@@ -407,7 +406,7 @@ class ui_view_info extends FO_Plugin
           $entry = [];
           $entry['count'] = $Count;
           $entry['type'] = _("Requires");
-          $entry['value'] = htmlentities($R['req_value']);
+          $entry['value'] = htmlspecialchars($R["$value"], ENT_QUOTES|ENT_HTML5, 'UTF-8');
           $Count++;
           $vars['packageRequires'][] = $entry;
         }
@@ -429,7 +428,7 @@ class ui_view_info extends FO_Plugin
           $entry = [];
           $entry['count'] = $Count;
           $entry['type'] = _($key);
-          $entry['value'] = htmlentities($R["$value"]);
+          $entry['value'] = htmlspecialchars($R["$value"], ENT_QUOTES|ENT_HTML5, 'UTF-8');
           $Count++;
           $vars['packageEntries'][] = $entry;
         }
@@ -443,7 +442,7 @@ class ui_view_info extends FO_Plugin
           $entry = [];
           $entry['count'] = $Count;
           $entry['type'] = _("Depends");
-          $entry['value'] = htmlentities($R['req_value']);
+          $entry['value'] = htmlspecialchars($R["$value"], ENT_QUOTES|ENT_HTML5, 'UTF-8');
           $Count++;
           $vars['packageRequires'][] = $entry;
         }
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd:
@ritankarsaha Can you please squash the commits to one ??

src/www/ui/ui-view-info.php
```diff
@@ -394,7 +394,7 @@ function ShowPackageInfo($Upload, $Item, $ShowMenu = 0)
           $entry = [];
           $entry['count'] = $Count;
           $entry['type'] = _($key);
-          $entry['value'] = htmlentities($R["$value"]);
+          $entry['value'] = $R["$value"];
```
Kaushl2208, line 397: I dont think removing the sanitization will solve the issue. We introduced it because we want to preserve the html entities coming in `$R`. There is no other process defined which makes sure that the data is properly sanitized and will definitely create issues in escaping characters like `<` `>` and `&`.

If there is any other functionality in php packages that can solve it would be a better option. Something like: `htmlspecialchars`


Also please squash the commits :) and update the branch from base.

ritankarsaha:
Sure @Kaushl2208 @shaheemazmalmmd will do the squash and update with the new approach !

ritankarsaha:
Done @Kaushl2208 @shaheemazmalmmd . Squashed the commits also made the adjustments as you asked :)

shaheemazmalmmd:
@ritankarsaha : check the failing test cases

ritankarsaha:
> @ritankarsaha : check the failing test cases

Done! Fixed the error, signed the commits and squashed the commits into one !

Kaushl2208 (changes requested):
Hey @ritankarsaha , I am a bit confused. I referenced `htmlspecialchars` because it encodes very selective html characters.

Difference is in how much they encode:
`htmlspecialchars` encodes only:

- & → &amp;
- < → &lt;
- `>` → &gt;
- " → &quot; (with ENT_QUOTES)
- ' → &#039; (with ENT_QUOTES)

on the other hand `htmlentities` encodes all applicable characters into HTML entities (like ©, ®, etc.)

Now if I look at the issue: **Frontend rendering HTML encoded characters**, Isnt that the same situation now? Its just the number of characters encoded is reduced. 

There is a need of decode html type functionality during the render process. This can be done either through `javascript` or `twig` as well. Most template engines provide the filters.

NOW, Please look into a solution that prevents you from `HTML INJECTION`. 

Let me know if I am missing something?

ritankarsaha:
> Hey @ritankarsaha , I am a bit confused. I referenced `htmlspecialchars` because it encodes very selective html characters.
> 
> Difference is in how much they encode: `htmlspecialchars` encodes only:
> 
> * & → &
> * < → <
> * `>` → >
> * " → " (with ENT_QUOTES)
> * ' → ' (with ENT_QUOTES)
> 
> on the other hand `htmlentities` encodes all applicable characters into HTML entities (like ©, ®, etc.)
> 
> Now if I look at the issue: **Frontend rendering HTML encoded characters**, Isnt that the same situation now? Its just the number of characters encoded is reduced.
> 
> There is a need of decode html type functionality during the render process. This can be done either through `javascript` or `twig` as well. Most template engines provide the filters.
> 
> NOW, Please look into a solution that prevents you from `HTML INJECTION`.
> 
> Let me know if I am missing something?

Hello, @Kaushl2208 I have now added an `escape_pkg_value` function at the top that converts all the entities like all the HTML5 entities back into their original UTF-8 characters and then re-escapes certain characters which are the only characters browser will treat as HTML or script delimiters, rest everything remains untouched.
Is this approach correct now? @Kaushl2208

ritankarsaha:
> > Hello, @Kaushl2208 I have now added an `escape_pkg_value` function at the top that converts all the entities like all the HTML5 entities back into their original UTF-8 characters and then re-escapes certain characters which are the only characters browser will treat as HTML or script delimiters, rest everything remains untouched. Is this approach correct now? @Kaushl2208
> 
> Hey, @ritankarsaha
> 
> I dont think this is the right way to do it. `htmlspecialchars` was good enough to sanitize the data and store in DB. The Decoding should be done at UI side (while fetching the data from database). If you encode and then decode at the same time what good it is doing if the string is being stored with HTML tags? (Risking HTML Injection).
> 
> So the idea is you store the encoded string, when you retrieve it for the UI, Then decode!! Either use JavaScript methods to decode. Look into [Twig Raw Filter](https://twig.symfony.com/doc/3.x/filters/raw.html) using this we just have to make sure that we trust the string that is coming from the user (to avoid XSS vulnerabilities).
> 
> Also, This can easily be done with javascript with something like this (But on UI side)
> 
> ```js
> function decodeHtmlEntities(str) {
>   const textarea = document.createElement('textarea');
>   textarea.innerHTML = str;
>   return textarea.value;
> }
> const decoded = decodeHtmlEntities(encodedString);
> ```

Sure @Kaushl2208 will do this and push a new commit !!

ritankarsaha:
Hello, @Kaushl2208 made the changes and pushed a new commit into the PR. :)) Is this approach now correct and in accordance ? I have used `Javascript methods` here as you had suggested, but I think using `Twig Raw Filter` would have been okaay as well as the data is already going through `htmlspecialchars()` at write-time. However I preferred this to be completely sure, is this okaay now or should I do the Twig Raw Filter approach ? 
Also, apologies for the delay, I have had my end semester exams going on so got a bit delayed ;)

ritankarsaha:
> Hello, @Kaushl2208 made the changes and pushed a new commit into the PR. :)) Is this approach now correct and in accordance ? I have used `Javascript methods` here as you had suggested, but I think using `Twig Raw Filter` would have been okaay as well as the data is already going through `htmlspecialchars()` at write-time. However I preferred this to be completely sure, is this okaay now or should I do the Twig Raw Filter approach ? Also, apologies for the delay, I have had my end semester exams going on so got a bit delayed, sorry ;)

Hello @Kaushl2208 is the implementation of the PR now alright and in accordance or are there some other modifications that I need to make as well?

Kaushl2208 (commented):
Changes looks good. Needs Test!

Kaushl2208 (approved):
Tested, working as expected.

ritankarsaha:
> Tested, working as expected.

Sure thanks @Kaushl2208 !!

Issue #2986, opened 2025-03-05
Compatibility Agent Hangs


jkelso662:
When using the prebuilt image, I have never seen the compatibility agent complete. It always remains on yellow with status 'Restarted' or blank. Is this expected behaviour?

![Image](https://github.com/user-attachments/assets/91b75ea2-b16e-4ef8-bc9b-d20cfb5dd62a)

There is no log output.

I am running the container via `docker run -p 8081:80 fossology/fossology` on macbook pro 2020.

I have confirmed the same thing is happening via the docker-compose build. 

Steps to reproduce:
* Clone fossology
* Update postgres database path to local postgres (tried postgresql 17/16/14)
* Upload something to scan

shaheemazmalmmd:
@jkelso662 : It works on our end perfectly. can you give us the more information ?? like pasting the Logs ??

jkelso662:
@shaheemazmalmmd No logs showing when I click on the job / dependency links 

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/eb4844ff-0cf3-4716-8cee-9a8ffdf871d3" />

jkelso662:
I have confirmed the same thing is happening via the docker-compose build. 

Steps to reproduce:
* Clone fossology
* Update postgres database path to local postgres
* Upload something to scan.

shaheemazmalmmd:
@jkelso662 : Thank you for reporting this issue. we have fixed this. please check.

Feel free to reopen if the issue persists.

PR #2999, opened 2025-03-12
fix(docker): Runtime dependency not found for compatibility

Kaushl2208:
## Description
`libjsoncpp24` is the package that was not supported in `debian-bookworm`. Fix the correct name of the package in different debian codenames.

### Changes

1. Fix correct version of a dependency in `debian-bookworm`
2. Fix flag `--build` should be `--buildtime`
3. Remove the explicit installation of the runtime dependency `libyaml-cpp0.7`

## How to test

- Build and Run the Docker Image (Master Branch).
- Upload a Component and Run the Compatibility Agent. The agent should fail.
- Switch to the Feature Branch.
- Rebuild and Run the Updated Docker Image.
- Upload a Component and Run the Compatibility Agent. The agent should pass.

cc @shaheemazmalmmd @its-sushant @GMishx 

CLOSES: #2986

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@
 # SPDX-FileCopyrightText: © mishra.gaurav@siemens.com
 # SPDX-FileCopyrightText: © 2016-2017 TNG Technology Consulting GmbH
 # SPDX-FileCopyrightText: © maximilian.huber@tngtech.com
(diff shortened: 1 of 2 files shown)


its-sushant:
Thank you @Kaushl2208. I have reviewed and tested the branch. Looking good to me.

CC: @shaheemazmalmmd

shaheemazmalmmd (approved):
Tested works as expected.

PR #3009, opened 2025-03-17
Update Dockerfile

Jotthecode:
Description:
Missing CMD Instruction
The CMD instruction provides a default command that runs when the container starts. Right now, your Dockerfile only has an ENTRYPOINT. Without CMD, users must provide a command explicitly when running the container. If they don’t, the container will execute only the entrypoint script and then exit if it doesn’t start a long-running process. 
Changes made:
I added a cmd function to define default process after the entrypoint. 
How to test:
The entrypoint script runs first (for setup).
The container remains running with Apache by default. Users can override the default command when running the container


GMishx:
The entrypoint already starts the apache in foreground and run many other things required to setup the container.

Please run the container first before making such critical changes.

PR #3010, opened 2025-03-17
Update agent-compatibility.php

Jotthecode:
## **Description**

This PR replaces direct `$_POST` access in `AgentAdd` with Symfony's `Request` object for better encapsulation and testability.

### **Changes**
- **Replaced** direct `$_POST` access with `$request->request->all()`, ensuring proper request handling.
- **Improved** compatibility with Symfony-based request processing.
- **Enhanced** maintainability and unit testability.

### **Before:**
```php
if ($request == null) {
  $request = $_POST;
}
 
###**After**
'''php
if ($request instanceof Request) {
  $request = $request->request->all();
} elseif (!is_array($request)) {
  $request = [];
}

### DIFF
--- a/src/compatibility/ui/agent-compatibility.php
+++ b/src/compatibility/ui/agent-compatibility.php
@@ -37,12 +37,7 @@ class CompatibilityAgentPlugin extends AgentPlugin
     }
 
     $compatibilityDependencies = array("agent_adj2nest");
-
-    if ($request instanceof Request) {
-       $request = $request->request->all();
-    } elseif (!is_array($request)) {
-       $request = [];
-    }
+    $request = $this->normalizeRequest($request);
 
     $compatibilityDependencies = array_merge($compatibilityDependencies,
         $this->getCompatibilityDependencies($request));
--- a/src/compatibility/ui/agent-compatibility.php
+++ b/src/compatibility/ui/agent-compatibility.php
@@ -38,9 +38,12 @@ class CompatibilityAgentPlugin extends AgentPlugin
 
     $compatibilityDependencies = array("agent_adj2nest");
 
-    if ($request == null) {
-      $request = $_POST;
+    if ($request instanceof Request) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/compatibility/ui/agent-compatibility.php [hunk 1]
```diff
@@ -38,9 +38,12 @@ public function AgentAdd($jobId, $uploadId, &$errorMsg, $dependencies = [],
 
     $compatibilityDependencies = array("agent_adj2nest");
 
-    if ($request == null) {
-      $request = $_POST;
+    if ($request instanceof Request) {
+       $request = $request->request->all();
+    } elseif (!is_array($request)) {
+       $request = [];
```
Kaushl2208, line 44: Change looks decent. Can we make it general for all the agent and add this change in Parent Class?

Kaushl2208 (changes requested):

src/compatibility/ui/agent-compatibility.php
Jotthecode: sir do you mean adding it into compatibilityagentplugin class?

Kaushl2208: What I mean is, Other agents like `copyright, decider, ecc, etc.` have `AgentAdd()` function and Class will inherit properties from common class. 
Expectation is that the change should be in such a way that all these agents can utilise it. Rather than changing each and every agent.

Jotthecode:
Changes are done in agentplugin that is the parent class with a new pr and are taken forward in compatibility.php class

Jotthecode:
https://github.com/fossology/fossology/pull/3016
Above is the new pr for the change done.

shaheemazmalmmd:
@Jotthecode : You may push this changes in #3016 also please follow the [commit convention(https://github.com/fossology/fossology/blob/master/CONTRIBUTING.md#git-commit)

PR #3011, opened 2025-03-17
fix(db): Check on drop view if exists

Kaushl2208:
## Description

Drop view SQL query might fail if the view doesnt exist.

### Changes

dropViews() --> Function runs a drop query, which fails if the view doesnt exist.

CC: @shaheemazmalmmd


shaheemazmalmmd (approved):
Code Looks good. Tested works as expected.

PR #3014, opened 2025-03-18
feat(workflow): Commit Linting Stage in Static Checks

Kaushl2208:
## Description

Added a commit linter, which can follow certain commit guidelines from the community.

### Changes

1. Added a Lint commit stage
2. Added a config file, which follows few of the requirements with commit guidelines. (more can be added in future)
3. Added DCO check 

cc: @GMishx @shaheemazmalmmd

### DIFF
--- a/.github/workflows/static-checks.yml
+++ b/.github/workflows/static-checks.yml
@@ -1,4 +1,5 @@
 # SPDX-FileCopyrightText: © 2021 Avinal Kumar <avinal.xlvii@gmail.com>
+# SPDX-FileCopyrightText: © 2025 kaushlendra-pratap.singh@siemens.com
 
 # SPDX-License-Identifier: GPL-2.0-only AND LGPL-2.1-only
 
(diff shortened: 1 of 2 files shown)


.github/workflows/static-checks.yml
```diff
@@ -115,3 +116,33 @@ jobs:
...
+        with:
+          fetch-depth: 0
+
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
+
+      - name: Install Commit-Lint
+        run: npm install --save-dev @commitlint/cli @commitlint/config-conventional
```
GMishx, line 134: Better to use actions from marketplace than installing tools manually...

See https://github.com/marketplace/actions/semantic-pull-request for example.

Kaushl2208: Sure, Will take a look into this :)

GMishx (approved):
LGTM

PR #3019, opened 2025-03-22
Fix(decisionImporter): FOSSology dump import failure due to improper argument passing

### LINKED ISSUE #3018: decisionimporter fails with Uncaught TypeError (userId is missing)
# Description
Importing a FOSSology dump fails with an Uncaught TypeError with the current master branch:

```
"PHP Fatal error:  Uncaught TypeError: Fossology\DecisionImporter\DecisionImporterDataCreator::setUserId(): Argument #1 ($userId) must be of type int, null given, called in /usr/local/share/fossology/decisionimporter/agent/DecisionImporterAgent.php on line 131 and defined in /usr/local/share/fossology/decisionimporter/agent/DecisionImporterDataCreator.php:108"
"Stack trace:"
"#0 /usr/local/share/fossology/decisionimporter/agent/DecisionImporterAgent.php(131): Fossology\DecisionImporter\DecisionImporterDataCreator->setUserId()"
"#1 /usr/local/share/fossology/lib/php/Agent/Agent.php(328): Fossology\DecisionImporter\DecisionImporterAgent->processUploadId()"
"#2 /usr/local/share/fossology/decisionimporter/agent/decisionimporter.php(18): Fossology\Lib\Agent\Agent->run_scheduler_event_loop()"
"#3 /usr/local/lib/fossology/fo_wrapper(65): require('...')"
"#4 {main}"
"  thrown in /usr/local/share/fossology/decisionimporter/agent/DecisionImporterDataCreator.php on line 108"
agent failed, code: 255

```

The argument for the userId seems to be missing. The problem occurs for me starting with this commit:

```
commit 928502b25c7891d4460c3a8e31e93786538a85bc
Author: Sushant Kumar <sushant.kumar@siemens-healthineers.com>
Date:   Wed Jan 8 14:26:42 2025 +0530

    fix(decider): parse multiple arguments correctly

```

# Steps to reproduce
1. Take any version starting from 928502b25c7891d4460c3a8e31e93786538a85bc
2. Import a FOSSology dump for an existing upload
3. The decisionImporter will fail with the previously described error
4. Any version before will work

As usual, please let me know if I can support with any further information.

Thanks,
Jan

abhishekrj02:
## Description

This PR fixes an issue where FOSSology dump imports fail due to an incorrect variable being passed to setUserId(). The function expected an integer userId, but $this->SetUserId was used instead of $this->userId, leading to a TypeError.

### Changes

- Changed `$this->decisionImporterDataCreator->setUserId($this->setUserId) ` to 
`$this->decisionImporterDataCreator->setUserId($this->userId)`

- Ensures that the correct user ID is passed to prevent a TypeError.

## How to test

- Update version to this commit
- Reattempt importing a FOSSology dump.
- decisionimport job should complete successfully without errors.

test

fixes #3018


GMishx (approved):
Changes looks good. Nice catch.

abhishekrj02:
> Changes looks good. Nice catch.

Thank you sir!

src/decisionimporter/agent/DecisionImporterAgent.php
```diff
@@ -128,7 +128,7 @@ function processUploadId($uploadId)
 
     $this->decisionImporterUtils->getOrCreateIds($this->reportData, $this);
 
-    $this->decisionImporterDataCreator->setUserId($this->setUserId);
+    $this->decisionImporterDataCreator->setUserId($this->userId);
```
its-sushant, line 131:
```suggestion
    $this->decisionImporterDataCreator->setUserId($userId);
```
Just for consistency as `$userId = $this->userId;` is already defined.

its-sushant (changes requested):
One Minor suggestion. Can you have a look?

src/decisionimporter/agent/DecisionImporterAgent.php
abhishekrj02: Done the changes and committed.

JanAltenberg:
Hi,

@abhishekrj02: First of all, thanks a lot for looking into the issue and for the PR!

With the change I am able to import FOSSology dumps, but I think the setting for the user for which we want to import ("Import decision as") is ignored with this change. @GMishx, @its-sushant, @abhishekrj02: Or am I missing something?

Greetings and thanks again
Jan

its-sushant:
> Hi,
> 
> @abhishekrj02: First of all, thanks a lot for looking into the issue and for the PR!
> 
> With the change I am able to import FOSSology dumps, but I think the setting for the user for which we want to import ("Import decision as") is ignored with this change. @GMishx, @its-sushant, @abhishekrj02: Or am I missing something?
> 
> Greetings and thanks again Jan

Correct. It needs to be come from the argument and that is why the `$this->setUserId` was used. Using the `$userId` from the session would ignore `import decision as`. Will  create a PR.

JanAltenberg:
> > Hi,
> > @abhishekrj02: First of all, thanks a lot for looking into the issue and for the PR!
> > With the change I am able to import FOSSology dumps, but I think the setting for the user for which we want to import ("Import decision as") is ignored with this change. @GMishx, @its-sushant, @abhishekrj02: Or am I missing something?
> > Greetings and thanks again Jan
> 
> Correct. It needs to be come from the argument and that is why the `$this->setUserId` was used. Using the `$userId` from the session would ignore `import decision as`. Will create a PR.

Thanks a lot! :)

JanAltenberg:
Since the issue has already been addressed with #3021 this PR can be dropped.

PR #3022, opened 2025-03-24
fix(ci): dco check not running on push event on main

Kaushl2208:
## Description

Action for DCO check earlier was limited to work on Pull Requests Only. The check was also needed for push to `main` branch as well.

### Changes

Updated the older action `tisonkun/actions-dco@v1.1`  ---> `christophebedard/dco-check@0.5.0`
Which allows the check on both PRs and Merge Commits.

## How to test

1. Static Checks Stage on the main branch should not fail

CC: @GMishx

### DIFF
--- a/.github/workflows/static-checks.yml
+++ b/.github/workflows/static-checks.yml
@@ -137,5 +137,7 @@ jobs:
           configFile: ./commitlint.config.mjs
 
       - name: DCO Check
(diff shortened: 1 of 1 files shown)


GMishx (approved):
LGTM

shaheemazmalmmd (approved):

PR #3029, opened 2025-04-06
fix(api): fix /tree/view endpoint parameter check

tiyasakundu:
## Description
Fix the list of allowed query parameters in `/uploads/{id}/item/{itemId}/tree/view`

### Changes
1. Added missing `page` and `limit` to the list of allowed query parameters.
2. Rearranged the check for parameters before usage.

## How to Test
Fire `GET /uploads/{id}/item/{itemId}/tree/view?page=2&limit=3&flatten=true`
and verify that all query parameters are processed correctly without errors.

### DIFF
--- a/src/www/ui/api/Controllers/UploadTreeController.php
+++ b/src/www/ui/api/Controllers/UploadTreeController.php
@@ -353,6 +353,14 @@ class UploadTreeController extends RestController
     $uploadTreeId = intval($args['itemId']);
     $uploadId = intval($args['id']);
     $query = $request->getQueryParams();
+
+    $queryKeys = array_keys($query);
+    $allowedKeys = ['showQuick', 'agentId', 'flatten', 'scanLicenseFilter', 'editedLicenseFilter', 'sort', 'page', 'limit', 'tagId', 'search', 'filterOpen'];
+    $diff = array_diff($queryKeys, $allowedKeys);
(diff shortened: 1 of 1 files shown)


src/www/ui/api/Controllers/UploadTreeController.php
```diff
@@ -353,6 +353,14 @@ public function getTreeView($request, $response, $args)
     $uploadTreeId = intval($args['itemId']);
     $uploadId = intval($args['id']);
     $query = $request->getQueryParams();
+
+    $queryKeys = array_keys($query);
+    $allowedKeys = ['showQuick', 'agentId', 'flatten', 'scanLicenseFilter', 'editedLicenseFilter', 'sort', 'page', 'limit', 'tagId', 'search', 'filterOpen', 'showQuick'];
```
its-sushant, line 358:
```suggestion
    $allowedKeys = ['showQuick', 'agentId', 'flatten', 'scanLicenseFilter', 'editedLicenseFilter', 'sort', 'page', 'limit', 'tagId', 'search', 'filterOpen'];
```

its-sushant (changes requested):
@tiyasakundu Thank you for the PR. One minor suggestion. Can you take a look?

its-sushant:
Tested. LGTM!

CC: @shaheemazmalmmd

PR #3034, opened 2025-04-17
Fix: LicenseDB Pull mechanism Error Handling Improvement

Kaushl2208:
## Description

License and Obligation Pull feature from LicenseDB was not handling the exceptions better:
In case of anything other than `status=200`, it was showing generic mesg which is difficult to debug and handle.

### Changes

Introduce switch cases for multiple `status_codes`.
In files:
1. AdminObligationFromCSV.php
2. AdminLicenseFromCSV.php

## How to test

1. Setup LicenseDB server.
2. Give details of the server in `Admin > Customize`, Fille these customize information: image
3. Play around these information, To get several Response codes from server like 404, 500, 401 etc.
4. Go to `Admin > License Admin > License Import` Check various Outputs given.

cc: @shaheemazmalmmd @GMishx

### DIFF
--- a/src/lib/php/Util/HttpUtils.php
+++ b/src/lib/php/Util/HttpUtils.php
@@ -8,6 +8,7 @@
 
 namespace Fossology\Lib\Util;
 
+use Fossology\Lib\Exceptions\HttpClientException;
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\GuzzleException;
 use GuzzleHttp\Exception\RequestException;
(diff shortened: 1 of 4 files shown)


src/lib/php/Util/HttpUtils.php
```diff
@@ -48,6 +48,7 @@ public static function getGuzzleClient(array $SysConf, string $baseUri, string $
       'proxy' => $proxy,
       'base_uri' => $baseUri,
       'headers' => $headers,
+      'verify' => false,
```
GMishx: Vulnerable code.

src/lib/php/Util/HttpUtils.php
```diff
@@ -71,4 +71,45 @@ public static function checkLicenseDBHealth(string $getHealth, $guzzleClient)
 
     return false;
   }
+  /**
```
GMishx, line 74:
```suggestion

  /**
```

GMishx (changes requested):

GMishx (approved):
LGTM

PR #3044, opened 2025-05-02
feat(group): add option to edit groupname

shaheemazmalmmd:
## Description

add option to edit group-name

## How to test

* Create new groups.
* Edit/rename them using admin > Groups > edit group.

### DIFF
--- a/src/lib/php/Dao/UserDao.php
+++ b/src/lib/php/Dao/UserDao.php
@@ -443,4 +443,14 @@ class UserDao
   {
     return $this->dbManager->getRows("SELECT * FROM users ORDER BY user_name;");
   }
+
+  /**
+   * @param int $groupId
+   * @param string $newGroupName
+   */
+  function editGroup($groupId, $newGroupName)
+  {
+    $this->dbManager->getSingleRow('UPDATE groups SET group_name=$2 WHERE group_pk=$1;',
+            array($groupId, $newGroupName),__METHOD__.'.UpdateEditGroup');
+  }
 }
--- a/src/www/ui/page/AdminGroupEdit.php
+++ b/src/www/ui/page/AdminGroupEdit.php
@@ -1,0 +1,97 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © 2025 Siemens AG
+
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+
+namespace Fossology\UI\Page;
+
+use Fossology\Lib\Auth\Auth;
+use Fossology\Lib\Dao\UserDao;
+use Fossology\Lib\Plugin\DefaultPlugin;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * \class AdminGroupEdit extends DefaultPlugin
+ * \brief edit group name
+ */
+class AdminGroupEdit extends DefaultPlugin
+{
+
+  const NAME = 'group_edit';
+
+  function __construct()
+  {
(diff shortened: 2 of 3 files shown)


src/lib/php/Dao/UserDao.php
```diff
@@ -443,4 +443,14 @@ public function getAllUsers()
   {
     return $this->dbManager->getRows("SELECT * FROM users ORDER BY user_name;");
   }
+
+  /**
+   * @param int $groupId
+   * @param char $newGroupName
```
Kaushl2208, line 449:
```suggestion
   * @param string $newGroupName
```

src/www/ui/page/AdminGroupEdit.php
```diff
@@ -0,0 +1,71 @@
...
+
+    $groupId = $request->get('grouppk');
+    $newGroupName = $request->get('newgroupname');
+    if (! empty($groupId) && ! empty($newGroupName)) {
+      try {
+        $userDao->editGroup($groupId, $newGroupName);
+        $vars['message'] = _("Group") . ' ' . $groupMap[$groupId] . ' ' . _("edited to ".$newGroupName ) . '.';
+        $groupMap[$groupId] = $newGroupName;
+      } catch (\Exception $e) {
+        $vars['message'] = $e->getMessage();
```
Kaushl2208, line 57: We have a logger that can also used with different log levels, Its defined in `DefaultPlugin` so can be easily inherited. We can log the entire trace maybe? Helpful for development?

https://github.com/fossology/fossology/blob/da57da7adf118f05d8c7edeeb6a708949fab91a2/src/www/ui/page/AdminObligationFromCSV.php#L93

Kaushl2208 (changes requested):
Left a comment regarding a small issue. Everything else looks good. Needs Test!!

Also, as we are also looking into API first approaches, We should also create an API functionality to `Edit Group` as well?

shaheemazmalmmd:
> Left a comment regarding a small issue. Everything else looks good. Needs Test!!
> 
> Also, as we are also looking into API first approaches, We should also create an API functionality to `Edit Group` as well?

Yes we also need API for this. Lets do it in separate PR.

src/www/ui/page/AdminGroupEdit.php
```diff
@@ -0,0 +1,71 @@
...
+    $vars = array();
+
+    /** @var UserDao $userDao */
+    $userDao = $this->getObject('dao.user');
+    $groupMap = $userDao->getDeletableAdminGroupMap($userId,
+      $_SESSION[Auth::USER_LEVEL]);
+
+    $groupId = $request->get('grouppk');
+    $newGroupName = $request->get('newgroupname');
+    if (! empty($groupId) && ! empty($newGroupName)) {
```
Kaushl2208, line 51: I think we need to add a check here for validity of group name:

Currently it is editing group name: " " (an empty space), 
 Such cases involve: 
 1. Whitespace-only group name
 2. String beginning with a special character
 3. Numeric-only group name
 4. Cross-Site Scripting (XSS)
 5. HTML Injection

 
 XSS sanitization
 
image

HTML Injection:

image

Kaushl2208 (changes requested):
Found some vulnerabilities that we might have missed. We should update the code to handle scenarios, I mentioned in-review comment.

src/www/ui/page/AdminGroupEdit.php
shaheemazmalmmd: Done!

Kaushl2208 (approved):
Changes looks good!! Tested working as expected :)

PR #3046, opened 2025-05-06
fix(treeview): exclude scancode, reportimport and spasht from schedul…

shaheemazmalmmd:
## How to test

* Upload a component without scheduling scancode, spasht.
* You should not be able to see schedule from tree-view.


Kaushl2208 (approved):
Changes looks good, Tested! Working as expected.
We can just update the branch with latest master and merge!

Issue #3047, opened 2025-05-06
Reuser agent not being scheduled


JanAltenberg:
# Description
When selecting the reuse option for an upload and specifying another upload for reuse, the reuse agent is not being scheduled. This problem occurs since commit 151edd62f923538af6c3de7027f0a8b859cab963

```
git bisect bad
151edd62f923538af6c3de7027f0a8b859cab963 is the first bad commit
commit 151edd62f923538af6c3de7027f0a8b859cab963
Author: Kaushlendra Pratap <kaushlendra-pratap.singh@siemens.com>
Date:   Mon Apr 21 20:23:05 2025 +0530

    fix(ui): jquery deferred exception fix
    
    Signed-off-by: Kaushlendra Pratap <kaushlendra-pratap.singh@siemens.com>

 src/reuser/ui/template/agent_reuser.js.twig | 88 ++++++++++++++++-------------
 1 file changed, 48 insertions(+), 40 deletions(-)
 
```

and is still present.

# How to reproduce

1. Upload and scan [busybox-1.36.1.tar.bz2](https://busybox.net/downloads/busybox-1.36.1.tar.bz2).
2. Upload / scan [busybox-1.37.0.tar.bz2](https://busybox.net/downloads/busybox-1.37.0.tar.bz2) and select reuse options (Select an already uploaded package for reuse in specific folder, Enhanced reuse (slower), Reuse main license/s)
3. Select busybox-1.36.1.tar.bz2 as the upload to reuse
4. Start the upload
5. Check the scheduled jobs

## Problem:
1. When activating "Select an already uploaded package for reuse in specific folder" the relevant button for the upload selection is not getting highlighted
2. The upload for reuse can still be selected
3. But the reuser agent is not being scheduled at all

As always, please let me know if I can support with further information or with testing.

Thanks
Jan

PR #3038, opened 2025-04-21
Fix(ui): Schedule Agent Page JS Loading Delayed

Kaushl2208:
## Description

In `agent_reuser.js.twig`,  `tools.js` is being loaded asynchronously and `reloadUploads` function (which calls failed) is being defined or executed before `tools.js` finishes loading, even though  `getScript` callback logs ` "Failure handler loaded."`

It also led to `Upload to Reuse` element not re-loading.

### Changes

Introduced: `$.holdReady(hold)`,  which Holds or releases the execution of jQuery's `ready` event.

## How to test

1. Jobs > Schedule Agents
2. Choose a folder, Upload, Analysis.
3. In reuser: Check the first checkbox, Change the Repository to any other repo.
4. Upload to reuse should update.

Also: Big Red warning from console should go away.

Before: 
image

After:
image

cc: @shaheemazmalmmd @its-sushant

PR #3048, opened 2025-05-06
fix(jsload) : Reuser agent not being scheduled

Kaushl2208:
## Description

A wait callback function was added to load the `tools.js` file in the one of the bug fix : #3038 , Due to the wait function some other functionalities required by Upload Schedule Reuse, the folder loading feature was disabled. As mentioned in this issue: #3047 

### Changes

1. Introduced a Script loading function which loads the required `tools.js` before the rendering of footer.
2. Removed the wait and loading of the `tools.js` directly into `agent_adder.html.twig`

## How to test
1. Upload a component
2. Select reuse options (Select an already uploaded package for reuse in specific folder, Enhanced reuse (slower), Reuse main license/s) (The folderselector should not be disabled now)
3. Select another component as the upload to reuse
4. Start the upload and Check the scheduled jobs
5. Also Check for `Jobs > Schedule Agents` , everything there should also work for Reuse.

Closes: #3047

### DIFF
--- a/src/decider/ui/DeciderAgentPlugin.php
+++ b/src/decider/ui/DeciderAgentPlugin.php
@@ -60,6 +60,11 @@ class DeciderAgentPlugin extends AgentPlugin
     return "";
   }
 
+  public function getScriptIncludes(&$vars)
+  {
+    return "";
+  }
+
   /**
    * @brief Schedule decider agent
    * @param int $jobId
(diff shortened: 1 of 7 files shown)


JanAltenberg:
Thanks! I gave it a quick test and the issue described in #3047 seems to be fixed for me with this pull request.

shaheemazmalmmd (approved):

PR #3077, opened 2025-06-18
fix: correct YAML syntax error in static-checks workflow

### LINKED ISSUE #3075: Error In workflow
In stratic-checks.yml 
Image

IshaanAggrawal:
## Description

This pull request fixes a YAML syntax error in the `.github/workflows/static-checks.yml` GitHub Actions workflow that caused the workflow to fail.  
The issue was with the `args` value for the DCO Check step, where improper quoting and unnecessary escape characters were used in the `--exclude-pattern` argument.  
The line was corrected to use proper single quoting and unescaped characters, making the workflow file valid and allowing the DCO check to run as intended.

### Changes

- Fixed the `args` value in the DCO Check step within `.github/workflows/static-checks.yml`:
  - Changed from:
    ```yaml
    args: '--exclude-pattern ''.*dependabot\[bot\]@users.noreply.github.com'''
    ```
  - To:
    ```yaml
    args: '--exclude-pattern ".*dependabot\[bot\]@users\.noreply\.github\.com"'
    ```
- Ensured consistency with YAML syntax and GitHub Actions requirements.
- The workflow should now execute without YAML syntax errors.

image

## How to test
If all checks pass, the workflow syntax is fixed.

Fixes #3075


.github/workflows/static-checks.yml
```diff
@@ -141,5 +141,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          args: "--exclude-pattern '.*dependabot\[bot\]@users.noreply.github.com'"
+        args: '--exclude-pattern ".*dependabot\[bot\]@users\.noreply\.github\.com"'
```
Kaushl2208: - You are passing double quotes inside single quotes. This means the shell/command will treat the regex as quoted literally, and the regex engine may not parse it correctly. Might make inner double quotes unnecessary
- The backslashes are correct for escaping, but the extra inner quotes are not.

```suggestion
        args: --exclude-pattern '.*dependabot\[bot\]@users\.noreply\.github\.com'
```

Kaushl2208 (changes requested):

.github/workflows/static-checks.yml
Kaushl2208:
```suggestion
          args: '--exclude-pattern ".*dependabot\[bot\]@users\.noreply\.github\.com"'
```
Lint issue

Kaushl2208 (changes requested):

Kaushl2208:
@IshaanAggrawal , why are there unwanted changes in this commit. Also please resolve the merge conflicts.

PR #3081, opened 2025-07-01
feat(oidc): Add client credentials flow for m2m communication between licensedb and fossology

deo002:
## Description

Implemented Machine to machine communication between licensedb and fossology. 
If there's a licensedb token in config, then that will be used. If it's not there and there's a oidc instance set up both in fossology and licensedb, then that will be used. If even that is not there, then license/obligation import takes place via file upload.

## How to test

In Admin>Customize, set up the licensedb config and oidc config.
In LicenseDB, register a client via the api /oidcClients.
In Admin>License Admin>License Import, click on Import.
In Admin>Obligation Admin>Obligation Import, click on Import.

### DIFF
--- a/src/lib/php/common-sysconfig.php
+++ b/src/lib/php/common-sysconfig.php
@@ -279,6 +279,12 @@ function Populate_sysconfig()
   $valueArray[$variable] = array("'$variable'", "null", "'$oidcPrompt'",
     strval(CONFIG_TYPE_TEXT), "'OauthSupport'", "15", "'$oidcDesc'", "null", "null");
 
+  $variable = "OidcScope";
+  $prompt = _('OIDC Scope');
+  $desc = _('Scope of OIDC for client_credential grant. Used with LicenseDB.');
+  $valueArray[$variable] = array("'$variable'", "''", "'$prompt'",
+    strval(CONFIG_TYPE_TEXT), "'OauthSupport'", "16", "'$oidcDesc'", "null", "null");
+
   /*  Banner Message */
   $variable = "BannerMsg";
   $bannerMsgPrompt = _('Banner message');
@@ -582,19 +588,18 @@ function Populate_sysconfig()
     strval(CONFIG_TYPE_TEXT), "'SSS'", "2", "'$desc'", "null", "null");
 
   /* LicenseDB config */
-  $variable = "LicenseDBURL";
-  $prompt = _('LicenseDB URL');
-  $desc = _('URL to LicenseDB Server');
-  $valueArray[$variable] = array("'$variable'",
-    "''", "'$prompt'",
-    strval(CONFIG_TYPE_TEXT), "'LicenseDB'", "1", "'$desc'", "'check_url'", "null");
-
   $variable = "LicenseDBBaseURL";
   $prompt = _('LicenseDB API base URI');
   $desc = _('Base URI for API calls e.g. /api/v1');
-  $valueArray[$variable] = array("'$variable'", "'/api/v1'",
-    "'$prompt'", strval(CONFIG_TYPE_TEXT), "'LicenseDB'", "2", "'$desc'", "null",
+  $valueArray[$variable] = array("'$variable'", "'http://localhost:8080/api/v1'",
+    "'$prompt'", strval(CONFIG_TYPE_TEXT), "'LicenseDB'", "1", "'$desc'", "null",
     "null");
+
+  $variable = "LicenseDBHealth";
+  $prompt = _('Health check');
+  $desc = _('Endpoint to check health of LicenseDB service');
+  $valueArray[$variable] = array("'$variable'", "'/health'", "'$prompt'",
+    strval(CONFIG_TYPE_TEXT), "'LicenseDB'", "2", "'$desc'", "null", "null");
 
   $variable = "LicenseDBContent";
   $prompt = _('Export endpoint Licenses');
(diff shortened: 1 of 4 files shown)


src/lib/php/Util/HttpUtils.php
```diff
@@ -22,8 +22,9 @@ class HttpUtils
    * @param string $token Authentication token. Leave empty if not needed.
    * @return Client Guzzle Client configured with proxy
    */
-  public static function getGuzzleClient(array $SysConf, string $baseUri, string $token = "")
+  public static function getGuzzleClient(string $baseUri, string $token = "")
   {
+    global $SysConf;
```
GMishx: Oh, why not take the `SysConf` as an argument?

src/lib/php/common-sysconfig.php
```diff
@@ -279,6 +279,12 @@ function Populate_sysconfig()
   $valueArray[$variable] = array("'$variable'", "null", "'$oidcPrompt'",
     strval(CONFIG_TYPE_TEXT), "'OauthSupport'", "15", "'$oidcDesc'", "null", "null");
 
+  $variable = "OidcScope";
+  $prompt = _('OIDC Scope');
+  $desc = _('Scope of OIDC for communication with LicenseDB');
```
GMishx: I would suggest a different wording:
```suggestion
  $desc = _('Scope of OIDC for client_credential grant. Used with LicenseDB.');
```

src/lib/php/common-sysconfig.php
```diff
@@ -610,7 +615,7 @@ function Populate_sysconfig()
 
   $variable = "LicenseDBToken";
   $prompt = _('Auth token For LicenseDB');
-  $desc = _('');
+  $desc = _('Token from LicenseDB. No need to set if oidc is set up in both LicenseDB and Fossology.');
```
GMishx:
```suggestion
  $desc = _("Token from LicenseDB. Don't set if using OIDC for LicenseDB and FOSSology communication.");
```

src/www/ui/template/admin_license_from_licensedb.html.twig
```diff
@@ -19,8 +19,8 @@
     {% if not exportEndpoint %}
       {{ _self.renderMissingConfigItem('Export Endpoint is missing.') }}
     {% endif %}
-    {% if not authToken %}
-      {{ _self.renderMissingConfigItem('Auth Token for LicenseDB is missing.') }}
+    {% if not tokenConfig %}
+      {{ _self.renderMissingConfigItem('Auth Token for LicenseDB and Oidc Config is missing.') }}
```
GMishx:
```suggestion
      {{ _self.renderMissingConfigItem('Auth Token for LicenseDB and OIDC Config is missing. At least one is required.') }}
```

src/www/ui/page/AdminLicenseFromCSV.php
```diff
@@ -32,9 +33,10 @@ class AdminLicenseFromCSV extends DefaultPlugin
   const FILE_INPUT_NAME_V2 = 'fileInput';
 
   /**
-   * @var Client $guzzleClient
+   * @var GenericProvider $oidcProvider
    */
-  private $guzzleClient;
```
GMishx, line 37: Why the client is removed from the class property?

src/www/ui/page/AdminLicenseFromCSV.php
```diff
@@ -51,15 +53,30 @@ function __construct()
     ));
     /** @var LicenseCsvImport $licenseCsvImport */
     $this->licenseCsvImport = $GLOBALS['container']->get('app.license_csv_import');
-    $this->sysconfig = $SysConf['SYSCONFIG'];
```
GMishx, line 54: Do you require the entire array?

src/www/ui/page/AdminObligationFromCSV.php
```diff
@@ -32,9 +33,10 @@ class AdminObligationFromCSV extends DefaultPlugin
   const FILE_INPUT_NAME_V2 = 'fileInput';
 
   /**
-   * @var Client $guzzleClient
+   * @var GenericProvider $oidcProvider
    */
-  private $guzzleClient;
```
GMishx, line 37: Same here..

src/www/ui/page/AdminObligationFromCSV.php
```diff
@@ -51,15 +53,30 @@ function __construct()
     ));
     /** @var ObligationCsvImport $obligationsCsvImport */
     $this->obligationsCsvImport = $GLOBALS['container']->get('app.obligation_csv_import');
-    $this->sysconfig = $SysConf['SYSCONFIG'];
```
GMishx, line 54: And here as well.

src/www/ui/page/AdminLicenseFromCSV.php
deo002: I am initializing the guzzle client  with a new oidc token every time I make a request to licensedb. So, making it a local variable seemed more fitting.

The other approach would have been making a request for the oidc token and initializing the guzzle client in the constructor once and use that everywhere. But, with this approach, we would also have to reinitialize the guzzle client every time the token expires. Making a get request to a third party in constructor also didnt sit right with me. All this would have added more complexity to the code.

src/www/ui/page/AdminLicenseFromCSV.php
deo002: Yes, need to pass the entire sysconf to initialize guzzleclient

Kaushl2208:
Tested. Ready to merge.
Thanks @deo002 , For this contribution :)

PR #3088, opened 2025-07-11
fix(copyright): add copyright cleanup rules

its-sushant:
## Description

fix(copyright): add copyright cleanup rules

### DIFF
--- a/src/copyright/agent/copyright.conf
+++ b/src/copyright/agent/copyright.conf
@@ -27,7 +27,13 @@ author=__author__|(?<=<author>)(.*?)(?=<\/author>)
 #
 COPYSYM=(?:\(c\)|&copy;|(?<!\xC3)\xA9|\xC2\xA9|\$\xB8|\xE2\x92\xB8|\$\xD2|\xE2\x93\x92|\$\x9E|\xE2\x92\x9E)
 REG_COPYRIGHT=(?:spdx-filecopyrighttext|copyright)(?:ed|s)?[[:space:]:]*|__COPYSYM__[ \t]+([[:alnum:] ][^\0]{0,2}){5,}
-REG_EXCEPTION=\bcopyrights?(?:[ \t/\\\*\+#"\.-]+)(?:licen[cs]es?|notices?|holders?|and|statements?|owners?)[ \t\.,][^\0]*
+REG_EXCEPTION=\bcopyrights?(?:[ \t/\\\*\+#"\.-]+)(?:licen[cs]es?'?|notices?'?|files?'?|laws?'?|authors?'?|holders?\(?s\)?'?|and|by|neither|including|grant|from|help|as|redistributions?|not|sources?|each|any|forums?|softwares?|individuals?|for|materials?|years?'?|patents?'?|statements?'?|owners?'?)[ \t\.,][^\0]*
 REG_NON_BLANK=.*(?:[[:alpha:]][[:alpha:]]|[[:digit:]][[:digit:]]).*
 REG_SIMPLE_COPYRIGHT=\bcopyright\b|__COPYSYM__
 REG_SPDX_COPYRIGHT=spdx-filecopyrighttext:(.*)
+
+# Cleanup
+REG_REMOVE_FILE_STATEMENT=^\s*copyright\s+file\s*=.*
+REG_STRIP_LICENSE_TRAIL=(?i)(License:,|Licensed under|may be freely distributed under).*$
+REG_STRIP_TRADEMARK_TRAIL=\)\]\s*\[assembly: AssemblyTrademark\(.*$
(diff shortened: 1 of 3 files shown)


src/copyright/agent/copyscan.cc
```diff
@@ -105,3 +121,99 @@ void hCopyrightScanner::ScanString(const string& s, list<match>& out) const
...
+    {"JCOPYRIGHT", "Copyright"},
+    {"COPYRIGHT Copyright", "Copyright"},
+    {"Copyright Copyright", "Copyright"},
+    {"Copyright copyright", "Copyright"},
+    {"copyright copyright", "Copyright"},
+    {"copyright Copyright", "Copyright"},
+    {"copyright\"Copyright", "Copyright"},
+    {"copyright\" Copyright", "Copyright"}
+  };
+
```
Kaushl2208, line 198: Umm, is it a good idea to improve Mixed-case variants into a single case first and then replace?? That might even reduce a whole lot of replacement vectors? Or maybe instead of strings, if we already are using regex then why not that?? using `icase` maybe?

src/copyright/agent/copyscan.cc
```diff
@@ -105,3 +121,99 @@ void hCopyrightScanner::ScanString(const string& s, list<match>& out) const
...
+  NormalizeCopyright(cleaned);
+  StripSuffixes(cleaned);
+
+  return cleaned;
+}
+
+void hCopyrightScanner::TrimPunctuation(string &text) const{
+  const string trimCharsAll = ",\'\"-:;&@!";
+  const string trimStartOnly = ".>)]\\/";
+  const string trimEndOnly = "<([\\/";
```
Kaushl2208, line 149: Okay this prevents ASCII but do we also face issues with Unicode? Are there any cases?

src/copyright/agent/copyscan.cc
```diff
@@ -105,3 +121,99 @@ void hCopyrightScanner::ScanString(const string& s, list<match>& out) const
...
+  while (!text.empty() && trimStartOnly.find(text.front()) != string::npos) {
+    text.erase(0, 1);
+  }
+
+  while (!text.empty() && trimEndOnly.find(text.back()) != string::npos) {
+    text.pop_back();
+  }
+}
+
+void hCopyrightScanner::RemoveNoiseWords(string& text) const{
```
Kaushl2208:
```suggestion
void hCopyrightScanner::RemoveNoisePatterns(string& text) const{
```

src/copyright/agent/copyright.conf
```diff
@@ -27,7 +27,12 @@ author=__author__|(?<=<author>)(.*?)(?=<\/author>)
...
 REG_NON_BLANK=.*(?:[[:alpha:]][[:alpha:]]|[[:digit:]][[:digit:]]).*
 REG_SIMPLE_COPYRIGHT=\bcopyright\b|__COPYSYM__
-REG_SPDX_COPYRIGHT=spdx-filecopyrighttext:(.*)
\ No newline at end of file
+REG_SPDX_COPYRIGHT=spdx-filecopyrighttext:(.*)
+# Cleanup
+REG_REMOVE_FILE_STATEMENT=^\s*copyright\s+file\s*=.*
+REG_STRIP_LICENSE_TRAIL=(?i)(License:,|Licensed under|may be freely distributed under).*$
+REG_STRIP_TRADEMARK_TRAIL=\)\]\s*\[assembly: AssemblyTrademark\(.*$
+REG_ALL_RIGHT_RESERVE_TRAIL=(?i)(all right reserve).*$
```
Kaushl2208: Add a new line at the end?

Kaushl2208 (changes requested):
Hey @its-sushant , I have left some comments. Rather than changes they are more like questions.

Kaushl2208:
Hey @its-sushant , Any update on these changes?

src/copyright/agent/copyscan.cc
its-sushant: For handling the unicode cases, We will need to work on https://github.com/fossology/fossology/pull/2770

src/copyright/agent/copyscan.cc
its-sushant: I can look into this!

its-sushant:
> Hey @its-sushant , Any update on these changes?

Thank you @Kaushl2208 for the review. I have made the suggested changes!

shaheemazmalmmd:
Tested the changes with numpy-1.18.1.zip, 12 false positive copyrights are getting ignored. 

Can we also include "permission to use" in strip of copyrights ? 
image

also in REG_EXCEPTION_COPY we can make owners? instead of owners

image

its-sushant:
Thank you @shaheemazmalmmd I have made the suggested changes.

shaheemazmalmmd:
@its-sushant : Thanks for the changes. they work as expected.

PR #3093, opened 2025-07-22
perf(bulk): add upload_fk value to avoid query aggregation

shaheemazmalmmd:
Test steps are provided in https://github.com/fossology/fossology/pull/3049


its-sushant:
Tested: Worked as expected.

CC: @shaheemazmalmmd

GMishx (approved):
LGTM

PR #3097, opened 2025-08-02
fix(automation): do not download diff for sbom

GMishx:
## Description

Fix the `cli_options.diff_dir` for different scan modes. The tool still tries to download the diff even if scanning in dependency scan mode or dir scan mode.

### Changes

Make the `cli_options.diff_dir` assignment as `if-elif` ladder.


Kaushl2208:
Looks good to me!!

Kaushl2208 (approved):

PR #3099, opened 2025-08-04
fix(obligations): Import Obligations From CSV failed due bad param passed

Kaushl2208:
## Description

Import Obligations From CSV functionality was failing due to passing an `array` when a `string` is expected.  Which was creating a issue with internal `nl2br` function (which expects string).  Hence the functionality failed with `Unexpected Type`

### Changes

`src/www/ui/page/AdminObligationsFromCSV.php`

- Changed handling mechanism of `handleFileUpload` response. 
- Included a basic conditional statement: If return type `array` then use the correct param from the array if `string` then directly assign to `vars["message"]`

## How to test
1. Use a valid Obligations csv to import
2. Go to Admin > Obligations Admin >Obligations Import
3. Press Upload
4. Obligations should be handled and the Response should be visible.

cc: @shaheemazmalmmd @GMishx


GMishx (approved):
LGTM

PR #3101, opened 2025-08-05
feat(scancode): Added multiprocessing, heartbeat, and memory/resource limits to ScanCode agent

Vaibhavsahu2810:
## Description

This PR adds multiprocessing support to the ScanCode agent in FOSSology, along with a heartbeat mechanism to prevent job timeouts. It also introduces memory limits, CPU priority control, and better resource cleanup to ensure stable and efficient parallel scanning.

### Changes

* Introduced parallel file scanning using `multiprocessing.Pool`
* Added heartbeat signal (`SIGALRM`) to periodically report progress 
* Set memory limit per worker process using `resource.setrlimit()`
* Lowered worker CPU priority using `os.nice()`
* Added `maxtasksperchild` to avoid memory leaks during long runs
* Added config options for:

  * `SCANCODE_PARALLEL_PROCESSES`
  * `SCANCODE_NICE_LEVEL`
  * `SCANCODE_MAX_TASKS_PER_WORKER`
  * `SCANCODE_HEARTBEAT_INTERVAL`
* Maintained compatibility with sequential mode

## How to test

1. Configure values in customize section of Fossology

   ```
   PARALLEL_PROCESSES = 4
   NICE_LEVEL = 15
   MAX_TASKS_PER_WORKER = 1000
   HEARTBEAT_INTERVAL = 60
   ```
2. Run the ScanCode agent from FOSSology UI.
3. Monitor console logs for:

   * Heartbeat messages every minute
   * Per-file JSON output streamed continuously
4. Try canceling the job from UI and ensure that worker processes terminate gracefully.

### DIFF
--- a/install/fo-install-pythondeps
+++ b/install/fo-install-pythondeps
@@ -161,14 +161,14 @@ if [[ $RUNTIME ]]; then
       # Include experimental dependencies
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m pip install \
         --target=$HOME/pythondeps --no-input --upgrade \
-        scancode-toolkit==32.4.1 scanoss==1.5.1 safaa==0.0.3'
+        scancode-toolkit==32.4.1 scanoss==1.5.1 psutil==5.9.5 safaa==0.0.3'
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m spacy \
         download en_core_web_sm --target=$HOME/pythondeps'
     else
       # Only non-experimental dependencies
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m pip install\
         --target=$HOME/pythondeps --no-input --upgrade \
-        scancode-toolkit==32.4.1 scanoss==1.5.1'
+        scancode-toolkit==32.4.1 scanoss==1.5.1 psutil==5.9.5'
     fi
     ###########################################################################
   else
--- a/src/scancode/agent/runscanonfiles.py
+++ b/src/scancode/agent/runscanonfiles.py
@@ -9,12 +9,85 @@ SPDX-License-Identifier: GPL-2.0-only
 import os
 import json
 import argparse
+import signal
+import sys
+import time
+import logging
+from multiprocessing import Pool, Value, Manager
+import threading
+import psutil
+import queue
+import select
 
-# Set SCANCODE_CACHE environment variable
 script_directory = os.path.dirname(os.path.abspath(__file__))
 os.environ["SCANCODE_CACHE"] = os.path.join(script_directory, '.cache')
 
 from scancode import api
+
+SCANCODE_PARALLEL = 1
+SCANCODE_NICE = 10
+SCANCODE_MIN_MEMORY_PER_PROCESS = 1024
+SCANCODE_MAX_TASKS = 1000
+HEARTBEAT_INTERVAL = 60
+
+BYTES_TO_MB = 1024 * 1024 
+KB_TO_MB = 1024 
+USABLE_MEMORY_RATIO = 0.8  
+DEFAULT_FALLBACK_MEMORY_MB = 2048  
+TERMINATION_CHECK_INTERVAL_MS = 50  
+DEFAULT_SCAN_TIMEOUT_SECONDS = 15  
+STDIN_POLL_TIMEOUT_SECONDS = 0.5  
+WORKER_WATCHDOG_INTERVAL_SECONDS = 2  
+POOL_STARTUP_DELAY_SECONDS = 0.5  
+
+logger = logging.getLogger('scancode_processor')
+log_file_handler = None 
+
+def setup_logging(log_file_path=None, verbose=False):
+    """Setup comprehensive logging for the scancode agent"""
+    global logger, log_file_handler
+    
+    for handler in logger.handlers[:]:
+        logger.removeHandler(handler)
+        handler.close()
+    
+    log_level = logging.DEBUG if verbose else logging.INFO
+    logger.setLevel(log_level)
+    
+    console_handler = logging.StreamHandler(sys.stdout)
+    console_handler.setLevel(logging.INFO)
+    console_formatter = logging.Formatter('%(levelname)s: %(message)s')
+    console_handler.setFormatter(console_formatter)
+    logger.addHandler(console_handler)
+    
+    if log_file_path:
+        try:
+            log_file_handler = logging.FileHandler(log_file_path, mode='a')
+            log_file_handler.setLevel(logging.DEBUG) 
+            file_formatter = logging.Formatter(
+                '%(asctime)s - %(name)s - %(levelname)s - %(funcName)s:%(lineno)d - %(message)s'
+            )
+            log_file_handler.setFormatter(file_formatter)
+            logger.addHandler(log_file_handler)
+            logger.info(f"Logging to file: {log_file_path}")
+        except Exception as e:
+            logger.warning(f"Could not setup file logging to {log_file_path}: {e}")
+    
+    logger.propagate = False 
+    return logger
+
+files_processed = Value('i', 0)
+files_total = Value('i', 0)
+heartbeat_active = False
+
+active_pool = None
+parent_pid = os.getpid()
+manager = None
+shared_data = None
+termination_requested = False
+termination_event = threading.Event()
+stdin_closed = False  
+stdin_monitor_thread = None 
 
 def update_license(licenses):
   """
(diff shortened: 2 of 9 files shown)


install/fo-install-pythondeps
```diff
@@ -159,14 +159,14 @@ if [[ $RUNTIME ]]; then
...
-        scancode-toolkit==32.4.1 scanoss==1.5.1 safaa==0.0.2'
+        scancode-toolkit==32.4.1 scanoss==1.5.1 psutil safaa==0.0.2'
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m spacy \
         download en_core_web_sm --target=$HOME/pythondeps'
     else
       # Only non-experimental dependencies
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m pip install\
         --target=$HOME/pythondeps --no-input --upgrade \
-        scancode-toolkit==32.4.1 scanoss==1.5.1'
+        scancode-toolkit==32.4.1 scanoss==1.5.1 psutil'
```
Kaushl2208: Can we pin `psutils` version?

src/lib/php/common-sysconfig.php
```diff
@@ -285,6 +285,30 @@ function Populate_sysconfig()
...
+  $scancodeTasksDesc = _('Maximum number of files a worker process handles before recycling (100-10000). Helps prevent memory leaks. Default: 1000');
+  $valueArray[$variable] = array("'$variable'", "'1000'", "'$scancodeTasksPrompt'",
+    strval(CONFIG_TYPE_INT), "'ScanCode'", "4", "'$scancodeTasksDesc'", "null", "null");
+
+  $variable = "ScancodeHeartbeatInterval";
+  $scancodeHeartbeatPrompt = _('ScanCode Heartbeat Interval (seconds)');
+  $scancodeHeartbeatDesc = _('How often ScanCode sends progress updates to scheduler (10-300 seconds). Default: 60');
+  $valueArray[$variable] = array("'$variable'", "'60'", "'$scancodeHeartbeatPrompt'",
+    strval(CONFIG_TYPE_INT), "'ScanCode'", "5", "'$scancodeHeartbeatDesc'", "null", "null");
+
```
Kaushl2208: These changes are very much linked to a specific agent. In fossology, all agents have their own `.conf` file which takes care of specific confgurations linked to the agent. If there is any configuration that is related to entire instance, can be put here.

Also gives you a way if someone builds scancode agent standalone, they can still configure the agent.

Link to scancode conf file: https://github.com/fossology/fossology/blob/master/src/scancode/scancode.conf

src/lib/php/common-sysconfig.php
```diff
@@ -869,4 +893,4 @@ function get_system_load_average()
   }
 
   return '<button type="button" aria-disabled="true" disabled class="btn '.$class.'">System Load</button>';
-}
+}
```
Kaushl2208: New line?

src/scancode/agent/runscanonfiles.py
```diff
@@ -147,44 +174,584 @@ def scan(line, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score
...
+        available_mb = mem.available / (1024 * 1024)
+        return int(available_mb)
+    except Exception:
+        try:
+            with open('/proc/meminfo', 'r') as f:
+                for line in f:
+                    if line.startswith('MemAvailable:'):
+                        return int(line.split()[1]) // 1024
+        except:
+            return 2048
```
Kaushl2208: umm, Magic number!! Can we assign some constants for all the divisions and multiplications maybe? Pretty difficult to understand `2048` as response.

src/scancode/agent/runscanonfiles.py
```diff
@@ -147,44 +174,584 @@ def scan(line, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score
...
+    max_processes_by_memory = max(1, usable_memory // min_memory_per_process)
+    optimal_processes = min(requested_processes, max_processes_by_memory)
+    memory_per_process = usable_memory // optimal_processes
+    
+    print(f"Memory Analysis:")
+    print(f"  Available system memory: {available_memory} MB")
+    print(f"  Usable memory (80%): {usable_memory} MB")
+    print(f"  Requested processes: {requested_processes}")
+    print(f"  Optimal processes: {optimal_processes}")
+    print(f"  Memory per process: {memory_per_process} MB")
```
Kaushl2208: Please look into logging mechanism. Debugging through `print` statements will be a pain. Through a good logging mechanism we can also collect logs for errors and exceptions as well.

src/scancode/agent/runscanonfiles.py
```diff
@@ -147,44 +174,584 @@ def scan(line, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score
...
+    
+    if shared_data:
+        shared_data['terminate'] = True
+    
+    if active_pool:
+        try:
+            active_pool.terminate()
+            active_pool.join(timeout=1)
+        except:
+            pass
```
Kaushl2208, line 386: Dont forget to put logs here!!

src/scancode/agent/runscanonfiles.py
```diff
@@ -147,44 +174,584 @@ def scan(line, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score
...
+    if termination_requested:
+        return True
+    
+    if shared_data and shared_data.get('terminate', False):
+        return True
+    
+    return False
+
+class TimeoutError(Exception):
+    pass
```
Kaushl2208: Custom exception??? You can declare it into some `exceptions.py` file. (If we are using it)

src/scancode/agent/runscanonfiles.py
```diff
@@ -147,44 +174,584 @@ def scan(line, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score
...
+    Processes a single file and returns the results.
+    Enhanced with frequent termination checks for sequential processing
+    """
+    line, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score, parent_pid = args
+    
+    if not check_parent_alive(parent_pid):
+        print(f"Worker: Parent process died, exiting", flush=True)
+        sys.exit(1)
+    
+    if is_termination_requested():
```
Kaushl2208: Just a thought: `is_termination_requested()` function is called repetitively and creates unnecessary `if..else` ladder. Would it be a better way to centralize it using `decorator`?? Maybe runs at a time interval?? Or defined intervals? We can create a custom `Exception` to handle and exiting the loop gracefully??

src/scancode/agent/runscanonfiles.py
```diff
@@ -147,44 +174,584 @@ def scan(line, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score
...
+                return None
+            
+            print(f"Scanning copyrights for {result['file'][:50]}...", flush=True)
+            copyrights = run_with_timeout_and_termination_check(
+                api.get_copyrights, (result['file'],), timeout=30
+            )
+            if copyrights is None:
+                if is_termination_requested():
+                    return None
+                print(f"Timeout scanning copyrights for {result['file']}")
```
Kaushl2208: Will it make more sense to log this from `run_with_timeout_and_termination_check()` ?? Maybe an exception and handling it here???? This ladder makes it too much repetition. We are already doing: `is_termination_requested():` in the main function and doing it again. Needs more optimizations here.

src/scancode/agent/runscanonfiles.py
```diff
@@ -196,5 +763,18 @@ def process_files(file_location, outputFile, scan_copyrights, scan_licenses, sca
   min_score = args.min_score
   file_location = args.file_location
   outputFile = args.outputFile
+  
+  SCANCODE_PARALLEL = args.parallel
+  SCANCODE_NICE = args.nice_level
+  SCANCODE_MIN_MEMORY_PER_PROCESS = 1024
+  SCANCODE_MAX_TASKS = args.max_tasks
+  SCANCODE_HEARTBEAT_INTERVAL = args.heartbeat_interval
```
Kaushl2208: These are assigned through command line and I am hoping the Same constants declared above are a fallback. Can we use conditional operator here and assign hard values too?
Like:
`SCANCODE_PARALLEL = args.parallel if args.parallel else 1` 
Something like this will improve readability and remove redundancy???

And I am guessing we are using these values from `common-sysconfig.php` right? You can shift these to conf file as mentioned into another comment above.

src/scancode/agent/runscanonfiles.py
```diff
@@ -196,5 +763,18 @@ def process_files(file_location, outputFile, scan_copyrights, scan_licenses, sca
...
 
-  process_files(file_location, outputFile, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score)
+  try:
+    process_files(file_location, outputFile, scan_copyrights, scan_licenses, scan_emails, scan_urls, min_score)
+    print("Scan completed successfully", flush=True)
+  except SystemExit:
+    raise
+  except Exception as e:
+    print(f"Fatal error: {e}", flush=True)
+    sys.exit(1)
```
Kaushl2208: Newline...

src/scancode/ui/agent-scancode.php
```diff
@@ -69,23 +69,77 @@ public function scheduleAgent($jobId, $uploadId, &$errorMsg, $request)
...
+    $params = array();
+
+    $defaultParallel = isset($SysConf['SYSCONFIG']['ScancodeParallelProcesses']) ?
+                       $SysConf['SYSCONFIG']['ScancodeParallelProcesses'] : 4;
+    $defaultNice = isset($SysConf['SYSCONFIG']['ScancodeNiceLevel']) ?
+                   $SysConf['SYSCONFIG']['ScancodeNiceLevel'] : 15;
+    $defaultMaxTasks = isset($SysConf['SYSCONFIG']['ScancodeMaxTasks']) ?
+                       $SysConf['SYSCONFIG']['ScancodeMaxTasks'] : 1000;
+    $defaultHeartbeat = isset($SysConf['SYSCONFIG']['ScancodeHeartbeatInterval']) ?
+                        $SysConf['SYSCONFIG']['ScancodeHeartbeatInterval'] : 60;
```
Kaushl2208: Should be shifted to `conf` file

UMMMM....I am a bit confused now, Why are these parameters read through `$request`?  Is it an attempt to extend this support to REST ??? Even then this should be configured and read through one single source. Not every request. (Needs Clarification)

src/scancode/ui/agent-scancode.php
```diff
@@ -111,7 +165,25 @@ public function getScanCodeArgs($flags, $unpackArgs)
...
+      if (isset($parallelParams['nice_level'])) {
+        $args .= ' --nice-level=' . $parallelParams['nice_level'];
+      }
+      if (isset($parallelParams['max_tasks'])) {
+        $args .= ' --max-tasks=' . $parallelParams['max_tasks'];
+      }
+      if (isset($parallelParams['heartbeat'])) {
+        $args .= ' --heartbeat-interval=' . $parallelParams['heartbeat'];
+      }
+    }
```
Kaushl2208: Putting this in agent conf will reduce this complexity. Conf params can be directly read into CPP code.
This is already passing through : `src/scancode/agent/scancode_wrapper.cc`

Kaushl2208 (changes requested):
There is a lot of implementation design gaps. Needs discussion.

install/fo-install-pythondeps
Vaibhavsahu2810: Added version

src/lib/php/common-sysconfig.php
Vaibhavsahu2810: Removed from sysconfig and created new variables in .conf file , maintaining the FOSSology structure

src/lib/php/common-sysconfig.php
Vaibhavsahu2810: Removed

src/scancode/agent/runscanonfiles.py
Vaibhavsahu2810: Created the new constants and used it !!

src/scancode/agent/runscanonfiles.py
Vaibhavsahu2810: Added proper logging mechanism , with proper file

src/scancode/agent/runscanonfiles.py
Vaibhavsahu2810: Should i create a new exceptions.py file? , for now i have created in it

src/scancode/agent/runscanonfiles.py
Vaibhavsahu2810: Created ScanTimeoutError and TerminationRequestedError with meaningful messages including operation type and filename. The run_with_timeout_and_termination_check() function now:
Raises specific exceptions instead of returning None , Takes operation_name and filename parameters for better error context ,Logs errors at the appropriate level 

Eliminated repetitive is_termination_requested() checks , Removed the ladder of if copyrights is None: statements

src/scancode/agent/runscanonfiles.py
Vaibhavsahu2810: Made some optimization , explained above.

src/scancode/agent/runscanonfiles.py
Vaibhavsahu2810: Yup improved !!

src/scancode/agent/runscanonfiles.py
Vaibhavsahu2810: Removed

src/scancode/ui/agent-scancode.php
Vaibhavsahu2810: Actually i thought request directly will call the command , but it actually calls the agent , so it will only create redundancy here so removed

src/scancode/ui/agent-scancode.php
Vaibhavsahu2810: Removed redundancy!!

src/scancode/agent/runscanonfiles.py
Vaibhavsahu2810: Also ,  FOSSology closes stdin as the primary termination signal and earlier i added 30-second timeouts which exceeded 20-second kill window , so created monitor_stdin(): This continuously monitors stdin and immediately sets termination flags when it closes. reduce the timeout from 30 seconds to 15 seconds for scan operations

PR #3115, opened 2025-08-20
fix(decisionimporter): use content-based file matching instead of path-based

Vaibhavsahu2810:
## Description
The Decision Importer currently fails when importing decisions between different versions of software where files have been moved, renamed, or directory structures have changed, even if the file content remains identical.

## Solution
- Replace path-dependent matching with content-based matching using pfile hashes
- Files are now matched by content hash rather than path location
## Changes

- Modified `updateUploadTreeIds()` to use content-centric matching

## How to test 
-  Try to import FOSSology dump of package of different version to some different version


Vaibhavsahu2810:
Hey @shaheemazmalmmd , For now i didnt handle the edgecase if multiple files have identical content , have some ideas for that

Vaibhavsahu2810:
Update: When multiple files have identical content, decisions are applied to ALL matching files

PR #3133, opened 2025-09-24
feat(upload): Add ignore folder for specific uploads

Kaushl2208:
## Description

To extend the existing folder exclude functionality to accept the additional folder names per upload.
Idea is to also have a text(input) box to accept the folder name. which needs to be excluded.

### Changes

1. `upload.html.twig` : Add a text box field for user input (to exclude folder for specific upload
2. `UploadPageBase.php`: handle user input(if specific ignore folder is there) extend on top instance level ignore functionality

## How to test

1. Go to upload page
2. At point 7, a text box is placed for user inputs, User can type dynamic folders name which are not there for instance level ignore.
3. Select analysis
4. Upload
5. Check Browse page, mentioned folder should be ignored.

cc: @shaheemazmalmmd 

Screenshot:

image

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -174,7 +174,7 @@ jobs:
           libjson-c-dev librpm-dev libboost-all-dev libxslt1-dev
 
       - name: Setup CMake
-        uses: lukka/get-cmake@v4.0.3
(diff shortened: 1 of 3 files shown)


shaheemazmalmmd:
@Kaushl2208 : i have tried to add a folder in text input. it fails with following error.  

image

PR #3136, opened 2025-10-05
feat(trixie): add support for Debian 13 trixie

its-sushant:
## Description

Add support for Debian 13 trixie

CC: @GMishx @shaheemazmalmmd @Kaushl2208

### DIFF
--- a/src/pkgagent/agent/CMakeLists.txt
+++ b/src/pkgagent/agent/CMakeLists.txt
@@ -24,6 +24,11 @@ target_compile_definitions(pkgagent
     _FILE_OFFSET_BITS=64
     VERSION_S="${FO_VERSION}"
     COMMIT_HASH_S="${FO_COMMIT_HASH}")
+
+if(FO_RPM_CRYPTO_H_HEADER)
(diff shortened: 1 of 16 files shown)


src/pkgagent/agent/CMakeLists.txt
```diff
@@ -24,6 +31,11 @@ target_compile_definitions(pkgagent
     _FILE_OFFSET_BITS=64
     VERSION_S="${FO_VERSION}"
     COMMIT_HASH_S="${FO_COMMIT_HASH}")
+
+if(HAVE_RPM_CRYPTO_H_HEADER)
+    target_compile_definitions(pkgagent PRIVATE HAVE_RPM_CRYPTO_H=1)
+endif()
+
```
Kaushl2208, line 31: `cmake -S. -B./build -DTESTING=ON -G Ninja` think about this process too. When the tests are generated for `pkgagent` rpm libraries are not linked to compile definition and it breaks the build. Can we move this process to `SetDefaults.cmake`?

its-sushant: Sure @Kaushl2208 We can move the header check to `SetDefaults.cmake`

PR #3138, opened 2025-10-08
feat(upload): add chipsbased UI for ignore folders

Kaushl2208:
## Description

This pull request updates the "Exclude Folders" functionality in the upload UI and backend logic to use a unified chips-based mechanism. Users can now add or remove folders to be excluded from scanning using a single chips input. The backend processes the list only if the "Ignore Configured Folders" checkbox is checked, and all folder entries are sanitized before being passed to the unpack agent.

### Changes

- `upload.html.twig`:
    - The chips UI and input for excluding folders are now combined into a single, compact flex container.
    - The input and chips are visually grouped, and the input does not stretch across the screen.
    - The "Ignore Configured Folders" label, chips container, and tooltip are aligned in a single row.
    - The "Folders to ignore for this upload" section was removed; only one unified input is used.

- `UploadPageBase.php`:
    - The backend now checks if the "Ignore Configured Folders" checkbox (excludefolder) is checked.
    - If checked, it retrieves the list of folders from excludefolderSpecific, sanitizes them, and passes them as the -E argument to the unpack agent.
    - The logic for merging configured and user folders was removed; only the user-managed list from the chips UI is used.

## How to test
1. Go to upload page
2. At point 6, a chips based UI with pre-cofigured folders loaded is there, User can add new or remove folders.
3. Checkbox should be checked
4. Select analysis
5. Upload
6. Check Browse page, mentioned folder should be ignored.

Screenshot:
image

### DIFF
--- a/src/lib/php/UI/template/include/upload.html.twig
+++ b/src/lib/php/UI/template/include/upload.html.twig
@@ -53,9 +53,17 @@
       </li>
       <li>
         <div class="form-group">
-          <input type="checkbox" class="browse-upload-checkbox view-license-rc-size" name="excludefolder" value="1"/>
-          {{ 'Ignore Configured Folders:'|trans }} {{ configureExcludeFolders|join(', ') }}
-          <img src="images/info_16.png" data-toggle="tooltip" title="{{'Configure folders from Admin-Customize-Exclude-Files from scanning'|trans}}" alt="" class="info-bullet"/>
+          <div style="display: flex; align-items: center;flex-wrap: wrap;">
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd:
image

Suggestions for improvements after tests:

1) Please remove the extra space maked.
2) if i add a new folder in the exclude list no folders configured should be removed automatically.
3) if i select option 6 without adding any folder the ununpack agent will fail.
4) if i add at least 1 new folder to the list checkbox of option 6 should be automatically toggled.

@Kaushl2208 please take a look.

Kaushl2208:
Hey @shaheemazmalmmd , i have implemented the required changes.

PR #3145, opened 2025-10-25
fix(wget_agent): add FOSSOLOGY_SKIP_CHOWN for NFSv4 compatibility

### LINKED ISSUE #3129: [BUG] Upload fails with "chown failed" on NFSv4 volume, even with no_root_squash and 777 permissions
### Environment:

- Fossology official Docker image (version: latest)
- Deployment via Docker Swarm managed by Portainer.
- Volumes for /srv/fossology/repository and /srv/fossology/uploads are on an NFSv4 share.

### Problem:
The wget_agent fails during file upload with the error ERROR ... chown failed on /srv/fossology/uploads/..., error: Operation not permitted, followed by a FATAL ... Failed to import ... error.

### Troubleshooting Steps Taken:

- Confirmed no_root_squash is set on the NFS server export for the Docker Swarm nodes.
- Configured the Apache process inside the container to run as the same user as the backend agent (fossy) using APACHE_RUN_USER.
- Attempted to force correct ownership by using uid and gid options in the Docker volume's NFS driver options, matching the UID/GID of the fossy user on the NFS server.
- Set the target uploads directory permissions to 777 (drwxrwxrwx) on the NFS server itself.
- The chown failed error persists in all cases.

### Conclusion:
The hard-coded chown call in wget_agent.c (around line 118) appears to be incompatible with standard NFSv4 security configurations, as the operation is denied by the NFS server regardless of client-side user alignment or filesystem permissions. This prevents the use of Fossology with NFS volumes in a Docker Swarm environment.

sarika-03:
## Problem

Upload fails on NFSv4 volumes where `chown()` returns `Operation not permitted` due to NFSv4 identity mapping restrictions.

**Error:**
```
ERROR: chown failed on /srv/fossology/uploads/..., error: Operation not permitted
FATAL: Failed to import ...
```
This prevents FOSSology deployment in Docker Swarm environments using NFS storage.

## Solution

Added optional `FOSSOLOGY_SKIP_CHOWN` environment variable to allow uploads to continue when `chown()` fails.

### Changes Made

1. src/wget_agent/agent/wget_agent.c
Modified 3 chown() call sites in DBLoadGold() function:

2. **`docker-compose.yml`**: Added environment variable to scheduler and web services

## Behavior
- Without `FOSSOLOGY_SKIP_CHOWN`** (default): Original behavior, backward compatible
- With `FOSSOLOGY_SKIP_CHOWN=true`**: `chown()` failure logs WARNING, upload continues

##Testing
1 . Restart containers
     docker-compose down
     docker-compose up -d
2 . Verify env variable
     docker exec fossology_scheduler_1 env | grep FOSSOLOGY_SKIP_CHOWN
     # FOSSOLOGY_SKIP_CHOWN=true
3. Create & upload test file
    echo "NFSv4 Fix Test - $(date)" > ~/Desktop/test.txt
4.Check logs after upload
    docker logs fossology_scheduler_1 2>&1 | grep -i chown | tail -20
5.Result - Upload succeeds 

Fixes #3129

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,16 +7,16 @@
 # SPDX-License-Identifier: FSFAP
 #
 # Description: Recipe for setting up a multi container FOSSology
-#              Docker setup with separate Database instance
+# Docker setup with separate Database instance
 services:
   scheduler:
     build:
       context: .
       dockerfile: Dockerfile
       args:
-      - http_proxy
-      - https_proxy
-      - no_proxy
+        - http_proxy
+        - https_proxy
+        - no_proxy
     image: fossology
     restart: unless-stopped
     environment:
@@ -24,6 +24,7 @@ services:
       - FOSSOLOGY_DB_NAME=fossology
       - FOSSOLOGY_DB_USER=fossy
       - FOSSOLOGY_DB_PASSWORD=fossy
+      - FOSSOLOGY_SKIP_CHOWN=true
     command: scheduler
     depends_on:
       - db
@@ -39,9 +40,9 @@ services:
       context: .
       dockerfile: Dockerfile
       args:
-      - http_proxy
-      - https_proxy
-      - no_proxy
+        - http_proxy
+        - https_proxy
+        - no_proxy
     image: fossology
     restart: unless-stopped
     environment:
@@ -50,6 +51,7 @@ services:
       - FOSSOLOGY_DB_USER=fossy
       - FOSSOLOGY_DB_PASSWORD=fossy
       - FOSSOLOGY_SCHEDULER_HOST=scheduler
+      - FOSSOLOGY_SKIP_CHOWN=true
     command: web
     ports:
       - "8081:80"
@@ -80,6 +82,9 @@ services:
       retries: 5
       # start-period: 10s
 
+
 volumes:
   database:
   repository:
+
+  
--- a/src/wget_agent/agent/wget_agent.c
+++ b/src/wget_agent/agent/wget_agent.c
@@ -112,10 +112,22 @@ void DBLoadGold()
     SafeExit(56);
   }
 
+  // MODIFIED: Added FOSSOLOGY_SKIP_CHOWN environment variable check
   if ((int)ForceGroup > 0)
   {
     rc = chown(GlobalTempFile,-1,ForceGroup);
-    if (rc) LOG_ERROR("chown failed on %s, error: %s", GlobalTempFile, strerror(errno));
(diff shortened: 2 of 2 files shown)


sarika-03:
Hey @shaheemazmalmmd 
Any update on this ?

Kaushl2208:
Hey @sarika-03,

This change looks like a pragmatic workaround for environments where mounted volumes disallow chown (NFS, some PVs, bind mounts with UID/GID mismatches), but it’s important to be explicit about what it does and what it does not:
 
- **This is not a fix**: It simply suppresses the error path and downgrades chown failures to a warning when FOSSOLOGY_SKIP_CHOWN=true. Ownership on the filesystem is unchanged, so behavior and permissions still depend on the existing ownership on the mount.
- This behavior varies by deployment. In environments where containers can and should set correct ownership (host-managed volumes, Docker volumes, privileged init steps), skipping chown hides a real misconfiguration that should be fixed at the platform level.

**Risks**: silently ignoring chown failures can mask real permission problems later at runtime and can create inconsistent behaviour between deployments.

So if I take this as an workaround: There are better ways to solve this without adding suppressors in the code. If you have such requirement for your deployment, Maybe we can discuss on better solutions?

sarika-03:
Hi @Kaushl2208 .

I agree we need a better approach. Would you prefer:

1. Filesystem detection(automatic NFS detection instead of env var)
2. Make chown truly optional (bigger refactor, but cleaner)
3. Document "NFS not supported" for now

Happy to revise based on your preference. What direction would you recommend?

Thanks for the guidance! 🙏

shaheemazmalmmd:
@sarika-03 : Please join the community call next time for further discussion regarding this. You can find the [.ICS](https://lists.fossology.org/g/fossology-devel/attachment/3181/0/) for the meeting.

sarika-03:
Noted. I’ll join the next community call.

krrish175-byte:
Hi,
I wanted to provide a quick update on this. I recently reached out to @sarika-03 regarding the status of her previous contribution (PR #3145 ). She graciously gave me the green light to carry the torch forward on this issue.

I’d like to thank her for the initial investigation and groundwork, which helped identify the root cause regarding NFSv4 permission conflicts.

I have been working on a revised solution that implements automatic filesystem detection (removing the need for manual environment variables) to resolve this. I will be raising a new PR shortly with these changes for review.

@shaheemazmalmmd @Kaushl2208

shaheemazmalmmd:
Closing this as New PR is raised on this topic.

PR #3160, opened 2025-11-06
fix(postinstall): update postinstall with fix dependency graph

Kaushl2208:
## Description

Update python dependencies for multi OS support. 

### Changes

Update `fo-install-pythondeps` and `fo-postinstall` with pkg-config and libicu-dev.
Update version of SAFAA to accomodate Bullseye and Ubuntu Jammy.

## How to test

Install the package in any of the supported OS.


shaheemazmalmmd (approved):
Code looks good.

PR #3161, opened 2025-11-06
fix(spdx): propagate custom license to package-level concluded license

### LINKED ISSUE #3106: Declaring a license with custom text as main license
# Description
When declaring a license with a custom text as main license, the main license is not reflected correctly, neither in the UI, nor in the reports:

Image

The example above sets a custom text for a finding with the BSD-2-Clause license, which is being marked as the main license. In the corresponding SPDX2 TV export the main license is then stated as follows:
```
PackageLicenseConcluded: BSD-2-Clause
PackageLicenseDeclared: BSD-2-Clause
```
whereas the finding for the cleared file does refer correctly to the individual license:
```
LicenseConcluded: LicenseRef-fossology-BSD-2-Clause-a74867c030bca5bfabe0ab283782f619
```
The UI states the main license as follows for this example:

Image

# Problem
The current implementation for setting a main license does not take licenses with a custom text into account. Instead of the individual license the "generic license" is set as the main license (which is not correct).

As always, please let me know if I can support with any further information.

PixelKnightDev:
## Description

This PR fixes the issue where custom license text (reportinfo) for main licenses was not being reflected in the package-level licenseConcluded field of SPDX exports. Previously, only file-level licenses showed custom text, while package-level licenses always showed generic SPDX identifiers.

### Changes

Added getMainLicenseReportInfos() method to ClearingDao.php: Retrieves custom license texts (reportinfo) for main licenses from the database

Updated spdx.php main license processing logic:

- Fetches custom texts for main licenses using the new DAO method
- Checks if custom text exists for each main license
- Generates license IDs using hash of custom text instead of standard text when custom text is present
- Creates License objects with LicenseRef-fossology- prefixed shortNames for custom licenses
- Sets customText flag to true for custom licenses
- Always updates licensesInDocument when custom text exists, ensuring the custom license object is used

The fix ensures both package-level and file-level licenses reference the same custom license when custom text is present, maintaining consistency throughout the SPDX document.

## How to test

1. Upload a file with a detectable license (e.g., BSD-2-Clause)
2. Clear the license and mark it as the main license (star it)
3. Add custom text in the "Report Info" field of the main license
4. Export SPDX2 RDF report
5. Verify the XML output:

  - Package-level <spdx:licenseConcluded> should reference #LicenseRef-fossology- [LICENSE]-[HASH]
  - <spdx:hasExtractedLicensingInfo> should contain your custom text
  - File-level <spdx:licenseConcluded> should reference the same custom license
  - Before this fix, package-level would show generic http://spdx.org/licenses/[LICENSE]

Example output after fix:
```
<spdx:Package>
  <spdx:licenseConcluded rdf:resource="#LicenseRef-fossology-BSD-2-Clause-39e98c9285f40da205855538abfd81d3"/>
</spdx:Package>

<spdx:hasExtractedLicensingInfo>
  <spdx:ExtractedLicensingInfo rdf:about="#LicenseRef-fossology-BSD-2-Clause-39e98c9285f40da205855538abfd81d3">
    <spdx:extractedText>custom bsd-2-clauseiiii</spdx:extractedText>
  </spdx:ExtractedLicensingInfo>
</spdx:hasExtractedLicensingInfo>
```

Closes #3106

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -899,6 +899,38 @@ INSERT INTO clearing_decision (
   }
 
   /**
+   * Fetch custom main license texts (reportinfo) for an upload.
+   *
+   * @param int $uploadId
+   * @param int $groupId
+   * @return array<int,string> license_id => reportinfo text
+   */
+  public function getMainLicenseReportInfos($uploadId, $groupId)
+  {
+    $uploadTreeTableName = $this->uploadDao->getUploadtreeTableName($uploadId);
+    $statementName = __METHOD__;
+    $sql = "SELECT DISTINCT ON (ce.rf_fk)
+              ce.rf_fk AS license_id,
+              ce.reportinfo AS reportinfo
+            FROM $uploadTreeTableName ut
+            INNER JOIN clearing_event ce ON ce.uploadtree_fk = ut.uploadtree_pk
+            WHERE ut.upload_fk = $1
+              AND ce.group_fk = $2
+              AND NOT ce.removed
+              AND ce.reportinfo IS NOT NULL
+              AND ce.reportinfo <> ''
+            ORDER BY ce.rf_fk, ce.date_added DESC, ce.clearing_event_pk DESC";
+    $this->dbManager->prepare($statementName, $sql);
+    $result = $this->dbManager->execute($statementName, array($uploadId, $groupId));
+    $reportInfos = array();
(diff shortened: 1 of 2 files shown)


AnujRewar:
Hi @PixelKnightDev 
Please follow official commit syntax and remove trailing whitespace (unnecessary spaces or tabs at the very end of a line) in src/spdx/agent/spdx.php.

PR #3166, opened 2025-11-13
chore(ci): remove EOL image and update dependency installation

Kaushl2208:
## Description

Ubuntu focal is way ahead of its EOL. Removed the support for the image. There is an update in GitHub runner image dependency for Ubuntu 22.04 and Ubuntu 24.04, update the `build-test` pipeline to handle that.

### Changes

1. Removed Ubuntu Focal Support for FOSSology
2. Update the correct Cmake  (name+version)
3. Update gcc installation with exact version number.

## How to test

All CI stages should pass. There should be no Ubuntu Focal artifacts at release stage.

cc: @GMishx @shaheemazmalmmd

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -21,7 +21,6 @@ jobs:
           - 'debian:bullseye'
           - 'debian:bookworm'
           - 'debian:trixie'
-          - 'ubuntu:focal'
           - 'ubuntu:jammy'
           - 'ubuntu:noble'
 
@@ -63,7 +62,7 @@ jobs:
         ./install/scripts/install-spdx-tools.sh
         rm -rf src/vendor
 
-    - name: Get CMake v3.23.0
+    - name: Get CMake
       uses: lukka/get-cmake@v4.1.2
 
     - name: Fetch tags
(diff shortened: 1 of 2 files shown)


.github/workflows/release-publish.yml
```diff
@@ -63,7 +62,7 @@ jobs:
         ./install/scripts/install-spdx-tools.sh
         rm -rf src/vendor
 
-    - name: Get CMake v3.23.0
+    - name: Get CMake v4.1.2
       uses: lukka/get-cmake@v4.1.2
```
shaheemazmalmmd, line 66: As this is a only name can we make it more general like Get CMake ??

.github/workflows/release-publish.yml
```diff
@@ -169,7 +168,7 @@ jobs:
           sudo apt-get update && sudo apt-get install -y build-essential pkg-config libglib2.0-dev \
           libjson-c-dev librpm-dev libboost-all-dev libxslt1-dev
 
-      - name: Setup CMake
+      - name: Setup CMake v4.1.2
         uses: lukka/get-cmake@v4.1.2
```
shaheemazmalmmd, line 172: Same here.

shaheemazmalmmd (approved):
code looks good.

PR #3170, opened 2025-11-19
fix(copyright): advance search behaving erratically due to newline

Kaushl2208:
## Description

Replace text functionality in copyright was behaving erratically because of "zero-or-more-matches" in javascript regex. 

### Changes

Changed the wildcard to "one-or-more-match" 

## How to test

1. Upload a component
2. Run copyright on the upload, Move to the copyright page
3. Select a copyright from the table, Create replacement text and `Test Replacement`

image


shaheemazmalmmd (approved):

PR #3182, opened 2025-12-01
fix(nomos): update lgpl-2.1-or-later license identification

its-sushant:
## Description

fix(nomos): update lgpl-2.1-or-later license identification

## Changes

Updated regex for lgpl-2.1-or-later in `STRING.in` file to handle cases where there is no space between **public** and **license** in case of lgpl-2.1-or-later license like text.

image


shaheemazmalmmd (approved):
Code looks good.

Issue #3152, opened 2025-10-31
Restored DB data gets wiped out when docker compose down and then up while upgrading the pg postgres db from 15 to 16


dineshr93:
I was attempting to upgrade the fossology version from Version: [4.4.0.63], Branch: [master], Commit: [#4e8e87] to 4.5.1 (current stable release)

w.r.t https://github.com/fossology/fossology/commit/76ebce114882c1cfe626b7280c29d77ee80a1ea8  I tried to upgrade the db from 15 to 16
The restored data gets wiped out after the docker compose down and up -d

Here are the steps I followed
old container:
Dump_all the data
```
#backup
docker exec -t fossology_db_1 pg_dumpall -U fossy > /usr/local/src/pg15_dump.sql
```

delete and get new version of fossology

update the named volume
```
  db:
    image: postgres:16
    restart: unless-stopped
    environment:
      - POSTGRES_DB=fossology
      - POSTGRES_USER=fossy
      - POSTGRES_PASSWORD=fossy
      - POSTGRES_INITDB_ARGS='-E SQL_ASCII'
    volumes:
      - database16:/var/lib/postgresql/16/data  <------
```

docker-compose up -d db

Restore the data

```
#restore & check size
cat /usr/local/src/pg15_backup.sql | docker exec -i fossology_db_1 psql -U fossy postgres  or
cat /usr/local/src/pg15_backup.sql | docker exec -i fossology_db_1 psql -U fossy fossology
```

```
#check size
docker exec -it fossology_db_1 psql -U fossy -d fossology -c "SELECT pg_size_pretty(pg_database_size('fossology'));"
 pg_size_pretty
----------------
 2937 MB
(1 row)
```

docker-compose down && docker-compose up -d
The old data in the UI is not visible and when I check the size again it shows only 16mb. All data gets wiped out
```
#check size
docker exec -it fossology_db_1 psql -U fossy -d fossology -c "SELECT pg_size_pretty(pg_database_size('fossology'));"
 pg_size_pretty
----------------
 16 MB
(1 row)
```

dineshr93:
I completely wiped out the all old images and containes and rebuilt from scratch. After which the data is not getting wiped out. But however the web container is unhealthy

#size is not getting deleted after `docker compose up`
 docker exec -it fossology_db_1 psql -U fossy -d fossology -c "select pg_database.datname as \"database name
[...]

PR #3187, opened 2025-12-05
docs(db): clarify PostgreSQL volume mount usage in Docker

krrish175-byte:
## Description

This PR adds a short and clear documentation note under the Docker section to
clarify how PostgreSQL data volumes should be mounted when running Fossology
via Docker. During upgrades (such as PostgreSQL 15→16), users may
accidentally mount version-specific internal Postgres directories
(e.g., `/var/lib/postgresql/15/data` or `/var/lib/postgresql/16/data`),
which causes data not to persist after container recreation.

To avoid this, the README now recommends using the standard `PGDATA` location
(`/var/lib/postgresql/data`), as documented by the official Postgres Docker
image. This removes ambiguity and prevents accidental data loss for users
performing upgrades or recreating containers.

## Changes

- Added a new subsection: **"Notes on PostgreSQL volume mounts (Docker)"**
- Explained why version-specific Postgres paths should be avoided.
- Referenced the official Postgres container documentation.
- Removed any outdated/internal Postgres directory examples.

## How to test

No functional code changes.  
To verify the documentation:

1. Start Fossology with Docker using a volume mapped to `/var/lib/postgresql/data`.
2. Load data into the DB.
3. Run `docker compose down && docker compose up -d`.
4. Confirm that the database persists without data loss.

## Closing issue

Closes: #3152

@dineshr93 @Kaushl2208 @shaheemazmalmmd


shaheemazmalmmd (approved):
Changes looks good.

PR #3207, opened 2025-12-27
fix(scancode): skip license text for large files to reduce memory usage

### LINKED ISSUE #3204: Scancode scan big tar ball file would consume huge memory - without params --max-in-memory=-1
When start scan big file(5-6GB) which took very long time on scancode agent working and consume huge memory.

Totla 27GB memory used for the processing. Also the docker container consume 80GB memory(memory+ cache file)

```
:/home/fossy/fossology# docker stats fossy --no-stream
CONTAINER ID   NAME      CPU %     MEM USAGE / LIMIT     MEM %     NET I/O         BLOCK I/O        PIDS
cbf262291319   fossy     198.88%   80.24GiB / 251.6GiB   31.90%    27.5GB / 14GB   44.2GB / 734GB   22


fossy    3530134  0.0  0.2 584540 536896 ?       S    Dec24   0:23 /usr/local/etc/fossology/mods-enabled/scancode/agent/scancode --jobId=171 --config=/usr/local/etc/fossology --userID=3 --groupID=3 --scheduler_start -l
fossy    3530135  0.0  0.0   2584   924 ?        S    Dec24   0:00 sh -c PYTHONPATH='/home/fossy/pythondeps/' python3 runscanonfiles.py -l -m 50 /tmp/filemkqFAc /tmp/filebK6eJG
fossy    3530136 99.9 11.0 29147348 29102196 ?   R    Dec24 1061:01 python3 runscanonfiles.py -l -m 50 /tmp/filemkqFAc /tmp/filebK6eJG
root     3533024  0.0  0.0   4068  1856 ?        S    06:25   0:00 /usr/sbin/CRON
```

There have max-in-memory option for scanconde cli entrance, but no idea if fossolog runscancodeonfile.py could have this option passed?
https://scancode-toolkit.readthedocs.io/en/stable/cli-reference/core-options.html#max-in-memory-option

krrish175-byte:
## Description
This PR addresses the critical memory exhaustion issue reported in #3204, where the ScanCode agent consumes excessive RAM (up to ~29GB RSS) when processing large archives (5-6GB+), often leading to OOM kills or container crashes.

**The Problem:**
Currently, `runscanonfiles.py` calls `api.get_licenses()` with `include_text=True`. For massive files with thousands of detected licenses, loading the full license text for every match into a Python dictionary causes a massive memory spike.

**The Fix:**
I have implemented a conditional check based on file size:
* **Files < 1GB:** Behavior remains unchanged (`include_text=True`).
* **Files ≥ 1GB:** The agent switches to `include_text=False` to prevent the memory balloon.

This drastically reduces the memory footprint for edge cases, allowing the scan to complete successfully without crashing the container.

## Technical Details & Trade-offs
* **Threshold:** I chose **1GB** as a safe limit. Most standard source files are well below this and will retain full functionality.
* **Impact:** For files larger than 1GB, the "Matched License Text" field in the UI will be empty to save RAM. This is a necessary trade-off to ensure the scan actually finishes.

## How Has This Been Tested?
- [x] **Small Files (<1MB):** Verified that `include_text` remains `True` and license text appears in the UI as expected.
- [x] **Logic Check:** Verified the conditional logic correctly identifies file sizes using `os.path.getsize()`.
- [ ] **Large File Benchmark:** Pending community verification on a >1GB dataset to quantify RSS reduction.

> **Note:** This is intended as an immediate stability fix. A permanent architectural solution involves migrating the agent to use the **ScanCode CLI** (which supports streaming and `--max-in-memory`), but that is a larger change requiring discussion with @shaheemazmalmmd and the maintainers regarding the implementation strategy.

## Related Issue
Fixes #3204 

@cw1427 @AnujRewar @lakshyaberia1

### DIFF
--- a/src/scancode/agent/runscanonfiles.py
+++ b/src/scancode/agent/runscanonfiles.py
@@ -9,6 +9,7 @@ SPDX-License-Identifier: GPL-2.0-only
 import os
 import json
 import argparse
+import sys
 
 # Set SCANCODE_CACHE environment variable
 script_directory = os.path.dirname(os.path.abspath(__file__))
@@ -16,11 +17,16 @@ os.environ["SCANCODE_CACHE"] = os.path.join(script_directory, '.cache')
 
 from scancode import api
 
-def update_license(licenses):
+# File size threshold (1GB) for switching to memory-efficient scanning mode
+# Files larger than this will skip loading full license text to prevent memory bloat
+FILE_SIZE_THRESHOLD_BYTES = 1073741824  # 1 GB
+
+def update_license(licenses, include_matched_text=True):
   """
   Extracts relevant information from the 'licenses' data.
   Parameters:
     licenses (dict): A dictionary containing license information.
+    include_matched_text (bool): Whether matched_text should be included in output.
   Returns:
     list: A list of dictionaries containing relevant license information.
   """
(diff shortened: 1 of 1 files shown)


Kaushl2208:
Discussed it with contributor, There is already a PR for similar use case. Closing

PR #3210, opened 2025-12-27
fix(wget_agent): Add automatic nfs detection to skip chown on nfs mounts

### LINKED ISSUE #3129: [BUG] Upload fails with "chown failed" on NFSv4 volume, even with no_root_squash and 777 permissions

krrish175-byte:
## Context & Acknowledgements
**This PR picks up the torch from PR #3145.**

A huge thanks to **@sarika-03** for her initial investigation and work on this issue. She correctly identified the root cause regarding NFSv4 and `chown` conflicts and submitted the initial fix. I reached out to her, and she graciously gave me the green light to continue working on this.

I have taken inspiration from her findings but pivoted the implementation from a manual environment variable to an **automatic detection** mechanism to make the fix seamless for end-users.

---

## Description
This PR addresses **Issue #3129**, where file uploads fail with `chown failed: Operation not permitted` when FOSSology is deployed using NFSv4 volumes.

**The Problem:**
* The `wget_agent` attempts to `chown()` uploaded files to the repository group.
* On NFSv4 (especially with UID/GID mapping), `chown` operations are often restricted by the server, even if the user has full Read/Write permissions.
* This causes uploads to fail unnecessarily, blocking deployment in enterprise NFS environments.

**The Solution:**
Instead of requiring manual configuration or environment variables (which mask underlying permission issues), this PR implements **automatic NFS detection**:
1.  **Detects** if the repository path is on an NFS mount using `statfs()`.
2.  **Skips** the `chown` operation specifically for NFS mounts.
3.  **Validates** that the file is actually Readable and Writable (`R_OK | W_OK`) by the current user.
4.  **Fails Fast** with clear, actionable error messages if R/W permissions are missing.

## Type of Change
Bug fix (non-breaking change which fixes an issue)

## Technical Implementation
Modified `src/wget_agent/agent/wget_agent.c` to include three new helper functions:
* **`IsNFSMount(const char *path)`**: Checks if the filesystem type is `NFS_SUPER_MAGIC` (0x6969).
* **`ValidateFileAccess(const char *path)`**: Verifies R/W access and provides troubleshooting logs if it fails.
* **`SmartChown(const char *path, gid_t group)`**: Routes the logic. If Local → `chown()`. If NFS → Validate Access.

**Why this approach?**
The initial proposal in PR #3145 suggested using an environment variable (`FOSSOLOGY_SKIP_CHOWN`). While valid, it required manual configuration from admins. This new approach builds on that concept but makes it **zero-config** and **safer** by programmatically detecting the filesystem and explicitly validating file access before proceeding.

## Testing:
I have verified this behavior in three scenarios:

- [x] **Test 1: Local Filesystem (Backward Compatibility)**
    * *Scenario:* Standard installation on local disk.
    * *Result:* `chown` is performed normally. Upload succeeds.
- [x] **Test 2: NFS Mount (Proper Permissions)**
    * *Scenario:* Repository mounted via NFSv4 with R/W access.
    * *Result:* Log shows: `Skipping chown on NFS mount... Validating permissions instead`. Validation passes. Upload succeeds.
- [x] **Test 3: NFS Mount (Restricted Permissions)**
    * *Scenario:* Repository mounted read-only or with wrong UID mapping.
    * *Result:* Upload fails immediately with: `FATAL: File is not readable/writable by current user`.

## Related Issue
Fixes #3129
(Supersedes #3145)

@shaheemazmalmmd @Kaushl2208

### DIFF
--- a/src/wget_agent/agent/wget_agent.c
+++ b/src/wget_agent/agent/wget_agent.c
@@ -16,6 +16,11 @@
 #define ASPRINTF_MEM_ERROR_LOG LOG_FATAL("Not enough memory for asprintf before line %d", __LINE__)
 
 #include "wget_agent.h"
+#include <sys/vfs.h>
+
+#ifndef NFS_SUPER_MAGIC
+#define NFS_SUPER_MAGIC 0x6969
+#endif
 
 char SQL[STRMAX];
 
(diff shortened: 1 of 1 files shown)


Kaushl2208:
Hi @krrish175-byte , Thanks for the contribution but as we discussed in the community call. Its a permission issue from the administrator, toolchain should not take responsibility for such. Hence closing!

PR #3215, opened 2025-12-29
fix(install): add Linux Mint support to dependency installer

aadithyaa9:
## Description

The dependency installer script (utils/fo-installdeps) failed on Linux Mint 22 (Wilma) because the distribution ID and codename were not recognized by the OS detection logic.

Since Linux Mint is based on Ubuntu, I added 'Linuxmint' and its current codenames (Wilma/Zara) to the existing Ubuntu case block. This allows the script to correctly identify the OS and install the necessary dependencies using apt.

### Changes

- Modified utils/fo-installdeps:

     - -   Added linuxmint to the distribution ID check.

     - - Added wilma and zara to the release codename check alongside Ubuntu codenames

## How to test

Run the dependency installer on a Linux Mint machine:
        ./utils/fo-installdeps -e

Verify that the script identifies the distro correctly (e.g., "distro is Ubuntu/Mint") instead of exiting with "distro not recognised".

Verify that apt-get install commands run successfully.

### DIFF
--- a/utils/fo-installdeps
+++ b/utils/fo-installdeps
@@ -88,7 +88,7 @@ echo "install core dependencies"
 if [[ $BUILDTIME ]]; then
    echo "*** Installing $DISTRO buildtime dependencies ***";
    case "$DISTRO" in
-      Debian|Ubuntu|Tuxedo)
+      Debian|Ubuntu|Tuxedo|Linuxmint)
          echo "DB: Installing build essential....."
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
@aadithyaa9: FOSSology does not currently support Linux Mint. Adding support would require a discussion to understand the need and justification for it.

PR #3217, opened 2026-01-02
fix: CORS error when CorsOrigins config not set

### LINKED ISSUE #2188: CORS error for the FossologyUI react client
### Description

[**FossologyUI*](https://github.com/fossology/FOSSologyUI)* encouters CORS issue while connecting from origin `localhost:3000`.

#### How to reproduce

1. Start the FossologyUI react app.
2. Set up the fossology instance through docker.
3. Check your browser console. You'll see an error message mentioning a CORS issue like below. 

#### Screenshots

Screenshot 2022-03-18 at 12 07 39 PM

### Versions

* Last commit id on master: [c3255b8](https://github.com/krishna9304/fossology/commit/c3255b8bdb7d140aef356647271047a1a1b44aba)
* Operating System (lMacintosh; Intel Mac OS X 10_15_7):

### Possible solution
Configuring the CORS to accept requests locally through port 3000.

Nakshatra480:
Fixes #2188

## Problem
FossologyUI React client gets CORS errors when connecting to the API because CorsOrigins config might not be set in the database on fresh installs.

## Solution
added a check to see if CorsOrigins config exists before trying to use it. now it only sets the CORS header when the config is explicitly set by an admin.

## Changes
- checks if `$SysConf['SYSCONFIG']['CorsOrigins']` is set before using it
- only adds `Access-Control-Allow-Origin` header when config exists
- prevents undefined index warnings

## Testing
tested locally - no more warnings on fresh install. admins need to configure CorsOrigins to enable CORS.

### DIFF
--- a/src/www/ui/api/Helper/CorsHelper.php
+++ b/src/www/ui/api/Helper/CorsHelper.php
@@ -21,8 +21,14 @@ class CorsHelper
   public static function addCorsHeaders(ResponseInterface $response): ResponseInterface
   {
     global $SysConf;
+
+    $corsOrigin = $SysConf['SYSCONFIG']['CorsOrigins'] ?? '';
+
+    if (!empty($corsOrigin)) {
+      $response = $response->withHeader('Access-Control-Allow-Origin', $corsOrigin);
+    }
+
     return $response
-      ->withHeader('Access-Control-Allow-Origin', $SysConf['SYSCONFIG']['CorsOrigins'])
       ->withHeader('Access-Control-Expose-Headers', 'Look-at, X-Total-Pages, Retry-After')
       ->withHeader('Access-Control-Allow-Headers', 'X-Requested-With, Content-Type, Accept, Origin, Authorization, action, accesslevel, active, copyright, Content-Type, description, filename, filesizemax, filesizemin, folderDescription, folderId, folderName, groupName, ignoreScm, applyGlobal, license, limit, name, page, parent, parentFolder, public, reportFormat, searchType, tag, upload, uploadDescription, uploadId, uploadType, excludefolder')
       ->withHeader('Access-Control-Allow-Methods', 'GET, POST, PUT, DELETE, PATCH, OPTIONS')
--- a/src/nomos/agent/generator/STRINGS.in
+++ b/src/nomos/agent/generator/STRINGS.in
@@ -1632,7 +1632,7 @@
 #
 %ENTRY% _LT_DUAL_LICENSE_30
 %KEY% "licen[cs]"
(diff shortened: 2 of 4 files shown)


src/www/ui/api/Helper/CorsHelper.php
```diff
@@ -21,8 +21,15 @@ class CorsHelper
   public static function addCorsHeaders(ResponseInterface $response): ResponseInterface
   {
     global $SysConf;
+    
+    // Use configured value or default to allow all origins
+    $corsOrigin = $SysConf['SYSCONFIG']['CorsOrigins'] ?? '*';
+    if (empty($corsOrigin)) {
+      $corsOrigin = '*';
```
shaheemazmalmmd: Using "*" (allowing all origins) can introduce serious security risks in productiongGrade systems. Therefore, we need to consider an alternative or more secure approach to handle this CORS issue.

src/lib/php/common-parm.php
```diff
@@ -160,19 +160,40 @@ function Traceback_dir()
...
+
+  // Check direct HTTPS
+  if (!empty($_SERVER['HTTPS']) &&
+      $_SERVER['HTTPS'] !== 'off' &&
+      $_SERVER['HTTPS'] === 'on') {
+    return 'https://';
+  }
+
+  return 'http://';
+}
```
shaheemazmalmmd: Seems like This code is not relevant to PR. can be removed.

shaheemazmalmmd:
@Nakshatra480 : We accept only 1 commit per PR. please squash your commits to one.

src/www/ui/api/Helper/CorsHelper.php
Nakshatra480: thanks for catching that! i've already removed the '*' default in the latest commit. now it only sets the header if CorsOrigins is explicitly configured.

src/lib/php/common-parm.php
Nakshatra480: yeah, accidentally included that. removed now along with core-auth.php. should only be CorsHelper.php in the PR.

Nakshatra480:
> @Nakshatra480 : We accept only 1 commit per PR. please squash your commits to one.

done! squashed to 1 commit now.

shaheemazmalmmd:
@Nakshatra480 :  I have just observed that there is already a default value set in code. so this changes are not relevant.

PFL: https://github.com/fossology/fossology/blob/64212ffb029eb34eaec8622caea2d9d25e540c2f/src/lib/php/common-sysconfig.php#L167

PR #3232, opened 2026-01-09
Fix license browser 500 error (#3231)

### LINKED ISSUE #3231: License Browser button click with error 500 shows blank page
When after scancode and license scan, want to check the whole license browser, but shows 500 error without any useful logs backgroud.

Image

Image

ayushmaan-16:
Fixes #3231

- Replace broken `getUploadRootId()` fallback with proper validation
- Add clear error message for invalid uploadtree ID
- Prevents 500/505 crash after fresh scancode runs

## Description

Fix license browser crash when `item` parameter is invalid (≤ 0) after fresh scancode runs. The original code attempted to call `getUploadRootId()` on invalid item IDs, causing downstream `itemTreeBounds` lookup to fail and crash with 500/505 errors.

## Changes

- **Critical Fix**: Replace broken fallback `$item = getUploadRootId($upload)` with proper validation
- **User-Friendly**: Return clear error `"Invalid item parameter: item='X'. License Browser requires valid uploadtree ID."`
- **Defensive Programming**: Early return prevents invalid `itemTreeBounds` lookup crash

## How to test

1. Fresh FOSSology Docker install: `docker compose up -d`
2. Wait for unpack jobs to complete (2-3 mins)
3. Upload any file → Browse → **License** tab
4. **Before**: 500/505 crash
5. **After**: Clear error message (no crash)

**Also test valid flow**: Upload → Unpack → Browse → License → Shows histogram correctly.

## Why this fixes #3231

Issue occurs when License Browser receives `item=0` after scancode (common race condition). Original fallback created invalid `uploadtree_id` → `getItemTreeBounds(0)` → NULL → crash. This validation handles the root cause properly.

### DIFF
--- a/src/www/ui/page/BrowseLicense.php
+++ b/src/www/ui/page/BrowseLicense.php
@@ -120,6 +120,9 @@ class BrowseLicense extends DefaultPlugin
     }
 
     $item = intval($request->get("item"));
+    if ($item <= 0) {
+      return $this->flushContent("Invalid item parameter: item='$item'. License Browser requires valid uploadtree ID.");
+    }
     $this->uploadtree_tablename = $this->uploadDao->getUploadtreeTableName($upload);
     $itemTreeBounds = $this->uploadDao->getItemTreeBounds($item, $this->uploadtree_tablename);
     $left = $itemTreeBounds->getLeft();


AnujRewar:
Hi @ayushmaan-16 your code is only preventing the error , not  showing the correct information. I think you should take a look to the item in link bar as  no value is getting assigned to it . Maybe that part is triggering the issue and assigning  value to it can solve this issue.
I have verified that this error is occuring due to ScanCode agent only as when we are selecting Scancode then this error is occuring but when we are selecting  other agents  except ScanCode then it is not occuring.
image

shaheemazmalmmd:
Issue #3231  has been closed by creator. changes are not needed now, @ayushmaan-16 thanks for your efforts.

PR #3244, opened 2026-01-13
Fix/scanoptions multipart array

sakirr05:
## Description

Fix HTTP 500 error in `POST /uploads` API endpoint when `scanOptions` is sent as multipart nested form fields (e.g., `scanOptions[analysis][nomos]=true`). The endpoint was incorrectly attempting to `json_decode()` an array, causing a PHP 8 TypeError. This fix makes `scanOptions` parsing polymorphic to handle both array (multipart) and JSON string formats.

### Changes

1. **Added request body validation** in `UploadController::postUpload()`:
   - Ensures `$reqBody` is always an array to prevent `array_key_exists()` errors on null

2. **Updated `scanOptions` parsing logic** to be polymorphic:
   - If `scanOptions` is already an **array** (from multipart nested form fields), use it directly
   - If `scanOptions` is a **JSON string** (for backward compatibility), decode it with `json_decode()`
   - This resolves the `TypeError: json_decode(): Argument #1 ($json) must be of type string, array given`

3. **Added unit test coverage**:
   - Updated existing `testPostUpload` tests to use proper `ScanOptions` format
   - Added new test case `testPostUploadFileWithScanOptionsArray` to cover multipart array format

**Files Modified:**
- `src/www/ui/api/Controllers/UploadController.php` - Fixed scanOptions parsing logic (lines 442-483)
- `src/www/ui_tests/api/Controllers/UploadControllerTest.php` - Added test coverage for multipart array format

### Manual Testing

1. **Reproduce the original bug (before fix):**
   ```bash
   curl -X POST 'http://localhost:8081/repo/api/v1/uploads' \
     -H "Authorization: Bearer <token>" \
     -H 'folderId: 1' \
     -H 'uploadType: file' \
     -F 'fileInput=@sample.tar.gz' \
     -F 'scanOptions[analysis][nomos]=true' \
     -F 'scanOptions[analysis][monk]=true'
   ```
   **Expected (before fix):** HTTP 500 Internal Server Error with `TypeError`  
   **Expected (after fix):** HTTP 201 Created with upload ID

2. **Test backward compatibility (JSON string format):**
   ```bash
   curl -X POST 'http://localhost:8081/repo/api/v1/uploads' \
     -H "Authorization: Bearer <token>" \
     -H 'folderId: 1' \
     -H 'uploadType: file' \
     -F 'fileInput=@sample.tar.gz' \
     -F 'scanOptions={"analysis":{"nomos":true,"monk":true}}'
   ```
   **Expected:** HTTP 201 Created (should continue to work)

3. **Verify upload creation:**
   - Check that upload record is created in database
   - Verify jobs are scheduled correctly (ununpack, adj2nest, nomos, monk)

### Unit Testing

Run PHPUnit tests for `UploadController`:
```bash
cd /path/to/fossology
phpunit src/www/ui_tests/api/Controllers/UploadControllerTest.php
```

**Test Cases:**
- `testPostUploadV1()` - Updated to use proper scanOptions format
- `testPostUploadV2()` - Updated to use proper scanOptions format  
- `testPostUploadFileWithScanOptionsArray()` - New test for multipart array format

All tests should pass.

### Root Cause

The `UploadController::postUpload()` method unconditionally called `json_decode()` on `scanOptions`, assuming it would always be a JSON string. However, when using `multipart/form-data` (which is the natural encoding for file uploads), the Slim framework automatically parses nested form fields like `scanOptions[analysis][nomos]=true` into PHP arrays. This causes PHP 8's strict type checking to throw a `TypeError` when `json_decode()` receives an array instead of a string.

### Impact

- **Improves API usability** by supporting standard multipart form encoding used by many HTTP clients (including OpenAPI generators)
- **Backward compatible** - existing clients that send `scanOptions` as a JSON string will continue to work
- **No breaking changes** - minimal, focused fix that only touches the parsing logic

### DIFF
--- a/src/www/ui/api/Controllers/UploadController.php
+++ b/src/www/ui/api/Controllers/UploadController.php
@@ -439,6 +439,9 @@ class UploadController extends RestController
   public function postUpload($request, $response, $args)
   {
     $reqBody = $this->getParsedBody($request);
+    if (! is_array($reqBody)) {
+      $reqBody = [];
+    }
     if (ApiVersion::getVersion($request) == ApiVersion::V2) {
       $uploadType = $reqBody['uploadType'] ?? null;
       $folderId = $reqBody['folderId'] ?? null;
@@ -465,10 +468,17 @@ class UploadController extends RestController
     }
     $scanOptions = [];
     if (array_key_exists('scanOptions', $reqBody)) {
-      if ($uploadType == 'file') {
-        $scanOptions = json_decode($reqBody['scanOptions'], true);
-      } else {
-        $scanOptions = $reqBody['scanOptions'];
+      $scanOptionsRaw = $reqBody['scanOptions'];
+      if (is_array($scanOptionsRaw)) {
+        // multipart/form-data can provide scanOptions as nested arrays:
+        // e.g. scanOptions[analysis][nomos]=true
+        $scanOptions = $scanOptionsRaw;
+      } elseif (is_string($scanOptionsRaw)) {
+        // allow JSON string as well (some clients send scanOptions as JSON)
+        $decoded = json_decode($scanOptionsRaw, true);
+        if (is_array($decoded)) {
+          $scanOptions = $decoded;
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd:
@sakirr05 : Can you please explain why this changes are needed ?? why can we use the existing method to upload and choose the scan options ??

PR #3248, opened 2026-01-14
fix(clearing): link new events to existing decisions

### LINKED ISSUE #2059: Adding a comment resets the clearing status
### Description

1. The starting point is a file whose clearing status is "Identified". No comment was added to the assigned license:
image
The clearing status is displayed in green in the tree view:
image

2. A comment is now added to the clearing decision:
image
Whereupon the previous clearing decision is automatically canceled:
image
and the clearing status changes to red:
image

3. In order to determine the clearing decision again, the submit button must be pressed again. However, this also means that another entry appears in the clearing history:
image

### Question

Why does a comment on a clearing decision that has already been made change the clearing status? Is this behavior wanted that way?

### Versions

* Release 3.10.0 (commit 07f8e50592add625560b9a5c146ba59cbfc426da)

Nakshatra480:
## Description
Fixes #2059
When adding comments to licenses in a clearing decision, the system was resetting the status from "Identified" back to "Not Cleared". This happened because new clearing events weren't being linked to the existing decision.
This PR fixes the issue by linking new clearing events to existing decisions when comments or report info are updated.

## Changes
- Modified `src/lib/php/Dao/ClearingDao.php`
- Added logic to check for existing clearing decisions
- Links new events to preserve decision status

## How It Works
When a comment is added to a license:
1. Capture the new clearing event ID
2. Check if there's an existing non-WIP clearing decision
3. Verify the license is part of that decision
4. Link the new event via `clearing_decision_event` table
5. Status is preserved while comment is saved

## Testing
Tested locally using Docker:
- Created clearing decision with "Identified" status (green)
- Added a comment to a license in the decision
- ✅ Status remained "Identified" (green) - fix confirmed!
- ✅ Comment was saved correctly

## Impact
- Users can now add comments without losing clearing status
- No changes to WIP decisions or new license additions
- Maintains backward compatibility
before creating the pull request to make sure you follow all the standards. -->

## Testing Evidence
Tested locally - the fix works as expected:

image

**Test performed:**
1. Created clearing decision with "Identified" status
2. Added comment to a license in the decision
3. ✅ Status remained green (Cleared: 1/1)
4. ✅ Comment was saved correctly

**Before this fix:** Status would reset to red, requiring resubmission.
**After this fix:** Status is preserved, no resubmission needed.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -479,7 +479,32 @@ INSERT INTO clearing_decision (
       $comment = $row['comment'];
       $acknowledgement = $changeTo;
     }
-    $this->insertClearingEvent($uploadTreeId, $userId, $groupId, $licenseId, false, $row['type_fk'], $reportInfo, $comment, $acknowledgement);
+    $newEventId = $this->insertClearingEvent($uploadTreeId, $userId, $groupId, $licenseId, false, $row['type_fk'], $reportInfo, $comment, $acknowledgement);
+
+    // Check if this license is part of an existing clearing decision and link the new event to it
+    $itemTreeBounds = $this->uploadDao->getItemTreeBounds($uploadTreeId);
+    $existingDecision = $this->getRelevantClearingDecision($itemTreeBounds, $groupId);
+    
+    if ($existingDecision !== null && $existingDecision->getType() != DecisionTypes::WIP) {
+      // Check if this license is part of the existing decision
+      $licenseIsInDecision = false;
+      foreach ($existingDecision->getClearingEvents() as $event) {
+        if ($event->getLicenseId() == $licenseId) {
+          $licenseIsInDecision = true;
+          break;
+        }
+      }
+      
+      // If the license is part of the decision, link the new event to preserve the decision status
+      if ($licenseIsInDecision) {
+        $stmt = __METHOD__ . '.linkToDecision';
+        $sql = "INSERT INTO clearing_decision_event (clearing_decision_fk, clearing_event_fk) VALUES($1, $2)";
+        $this->dbManager->prepare($stmt, $sql);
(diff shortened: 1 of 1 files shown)


Kaushl2208:
I can do the same without the changes. Please try and reproduce the issue first. Instead of fixing 1000+ lines of changes. I think we have given enough slack in these regards. Its all good and great that AI Agents are there and making life easy but are we now completely dependent on the changes produced by it?? A slight glance on the number of changes, A slight glance to what all is getting committed, Is it too much to ask for? 

Try and put your best foot forward. :)

Nakshatra480:
Hi @Kaushl2208 , thanks for taking a look at this!

I completely understand your concern about the line count. Looking at the diff, it does show 1000+ lines changed, but I think something went wrong with how the file was committed - possibly line ending changes (CRLF/LF) or whitespace formatting that affected the entire file.

The actual functional change I made is quite small - just adding logic to link new clearing events to existing decisions. Let me explain what I was trying to fix:

The Issue: When I was testing FOSSology locally, I noticed that if you create a clearing decision with status "Identified" (shows green), and then add a comment to one of the licenses in that decision, the status would reset back to red/"Not Cleared". Users then had to click Submit again just to restore the green status.

What I found: Looking at the code in updateClearingEvent(), when a comment is added, it creates a new clearing event but doesn't link it to the existing clearing decision in the clearing_decision_event table. So the system treats it as an incomplete decision.

My fix: I added a check after creating the new event to see if there's already a finalized decision for that file. If the license is part of that decision, I link the new event to it. That way the status stays preserved.

Testing: I tested this locally with Docker:

Created a decision with type "Identified"
Added a comment to a license
Status stayed green (before my fix, it would turn red)
I'm happy to provide more details or clean up the commit if the file reformatting is causing issues. Would it help if I create a fresh commit with just the logical changes? I want to make sure this is easy to review.

PR #3253, opened 2026-01-15
fix(ui): avoid crash when job queue failure callback is undefined

### LINKED ISSUE #1595: Issue with License Browser [Edit]
### Description

From https://github.com/fossology/fossology/pull/1560#issuecomment-580641114

    - snap1

If the job finish successfully, the functionality works fine.

#### How to reproduce

1. Upload a package.
1. Install some different commit of FOSSology, restart scheduler.
1. Install master branch and **do not** restart scheduler.
    - It will cause an version check fail at scheduler and job will fail (required for this issue).
1. Browse the package and click `[Edit]` on a folder.
    1. Select some licenses and choose add.
1. The call will happen, the agent will be scheduled, the agent will fail but the page will not reload due to the bug.

### Versions

* Last commit id on master: 2ea434afe588eb9e4c7cce43b40a09d35a00f34e

SooryanshSingh:
### Summary

The job queue polling code assumes that a failure callback is always present and
invokes it unconditionally when a job finishes with a failure state.

In some cases, only a success callback is provided. When such a job fails, this
leads to a runtime JavaScript error (`callbackfail is not a function`) and
disrupts the UI.

### Fix

This change verifies that a failure callback exists before calling it. If no
failure callback is provided, the failure is handled gracefully without causing
a JavaScript error or breaking the UI.

### Related Issue

Since this is first PR, according to new rules , proof of installation->
image
image

Fixes #1595

### DIFF
--- a/src/www/ui/scripts/job-queue-poll.js
+++ b/src/www/ui/scripts/job-queue-poll.js
@@ -20,9 +20,15 @@ function updateCheckSuccess(data) {
       return null;
     }
     else if ((data[jqId]) && (data[jqId].end_bits > 1)) {
+    if (typeof callbackfail === "function") {
       callbackfail(jqId);
+      } 
+    else {
+      console.warn("Job failed but no failure callback provided. Job ID: ", jqId);
+      }
       return null;
     }
+    
     else {
       return val;
     }


shaheemazmalmmd:
@SooryanshSingh : can you please squash the commits to one ?? also add the screenshots of tests you have done after fixing this ??

SooryanshSingh:
@shaheemazmalmmd  
Thanks for your time.

## Reproduction note (current setup)

The reproduction steps described in the original issue rely on running the **web UI** and **scheduler** on **different code versions**:

1. Install one commit and restart the scheduler  
2. Install `master` and do **not** restart the scheduler  

This version mismatch causes the scheduler to fail its version check, which leads to a job failure and triggers the UI failure path.

### Why this is no longer reproducible

With the current FOSSology Docker setup, this configuration is no longer possible:

- The **web** and **scheduler** services are built from the **same Docker image**
- Both services are rebuilt together via `docker-compose build`
- There is no supported way (within a single compose project) to run web and scheduler on different code revisions

As a result, following the original steps today consistently leads to **scheduler connectivity/version errors**, rather than the JavaScript exception described in the issue.

---

## Validation approach used

Since this is a **UI safety issue** (unconditional invocation of `callbackfail`), validation was performed directly at the JavaScript level.

The failure branch was exercised by **injecting a broken queue entry** and simulating a failed job response:

js
jobPollQueue.push({
  jqId: 99999,
  callbacksuccess: function () {},
  callbackfail: "not a function"
});

var fakeData = {};
fakeData[99999] = { end_bits: 2 };

updateCheckSuccess(fakeData);


### Before fix 


image




### After fix

image


Js exception raised as invoked in the PR.

UI remains responsive

Job failure is handled safely even when callbackfail is missing or invalid

shaheemazmalmmd:
@SooryanshSingh : thank you for the contributions the changes doesn't work for me. it is still crashing and there is no use for actual users with consol log error messages as they do see them. if you still want to fix the issues then please join the community call for further discussions.

PR #3259, opened 2026-01-17
fix(api): align POST /uploads response with OpenAPI schema

### LINKED ISSUE #1683: Bug in openapi.yaml
### Description

When making a POST request on `/uploads` the value for the key `message` in the response is not of type `str` as listed in the [openapi.yaml](https://github.com/fossology/fossology/blob/master/src/www/ui/api/documentation/openapi.yaml), but of type `int`
#### How to reproduce

Make an upload to fossology via the API, query the content `message` of the response and check its type.

### Versions

* Last commit id on master: 56684de65e76cb832c12a783237310069a595976
* Operating System (lsb_release -a): Description:	Ubuntu 18.04.4 LTS

SooryanshSingh:
## OpenAPI mismatch for `POST /uploads` response
Fixes #1683
### 🐞 Problem
When making a `POST` request to `/uploads`, the API returns the upload ID in the `message` field as an **integer**.  
However, according to `openapi.yaml`, the `message` field in the `Info` object is defined as a **string**.

This causes a mismatch between the documented API schema and the actual runtime behavior.

### 📍 Root Cause
In `UploadController::postUpload`, the upload ID was passed as an integer when creating the `Info` response object.

### ✅ Solution
Ensure the upload ID is **always returned as a string** by explicitly casting it before creating the `Info` object.

This aligns the runtime response with the OpenAPI specification and keeps ID types consistent across endpoints.
image

### DIFF
--- a/src/www/ui/api/Controllers/UploadController.php
+++ b/src/www/ui/api/Controllers/UploadController.php
@@ -526,10 +526,10 @@ class UploadController extends RestController
       $info =  $uploadHelper->handleScheduleAnalysis(intval($uploadId),
         intval($folderId), $scanOptions, true);
       if ($info->getCode() == 201) {
-        $info = new Info($info->getCode(), intval($uploadId), $info->getType());
+        $info = new Info($info->getCode(), (string)($uploadId), $info->getType());
       }
     } else {
-      $info = new Info(201, intval($uploadId), InfoType::INFO);
+      $info = new Info(201, (string)($uploadId), InfoType::INFO);
     }
     if (array_key_exists("mainLicense", $reqBody) &&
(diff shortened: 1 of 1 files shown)


SooryanshSingh:
This PR follows the conclusion from the earlier discussion on this issue.

The POST /uploads endpoint now returns the upload ID as a string in the message field of the Info response, aligning the runtime behavior with the OpenAPI schema and keeping ID types consistent across the API.

Kaushl2208 (approved):
Changes looks good. Needs test

Kaushl2208:
Working as expected. Thanks for your contribution @SooryanshSingh

PR #3263, opened 2026-01-20
fix(util): make StringOperation::getCommonHead UTF-8 multibyte-safe

SooryanshSingh:
### Problem

`StringOperation::getCommonHead()` previously compared strings **byte-by-byte**, which could lead to incorrect prefix detection for UTF-8 multibyte strings.
---

### Changes

- Make `StringOperation::getCommonHead()` UTF-8 multibyte-safe using `mb_*`
  functions when available
- Preserve existing byte-based behavior as a fallback for compatibility
- Add a regression test covering UTF-8 input

### Testing

- To validate the behavior difference between the old byte-based implementation
and the new multibyte-safe implementation, I used an emoji (`😀`) and a
truncated version of it

- In real-world scenarios (e.g., partial file reads, interrupted uploads, or corrupted input), a string may end inside a multibyte sequence. In such cases, byte-wise comparison can return a prefix that ends within a UTF-8 codepoint, producing invalid UTF-8 output.

- The old byte-based implementation compares raw bytes and can therefore
return a prefix that ends inside a multibyte character, producing
invalid UTF-8 output.

- The new implementation compares complete characters using `mb_*`
functions and avoids splitting inside character boundaries.

image

### With old code->
image

### With the changes in PR->
image

### DIFF
--- a/src/lib/php/Util/StringOperation.php
+++ b/src/lib/php/Util/StringOperation.php
@@ -17,15 +17,37 @@ class StringOperation
    */
   public static function getCommonHead($a, $b)
   {
+
+    if (function_exists('mb_strlen')) {
+      $encoding = 'UTF-8';
+      $headLength = 0;
+
+      $maxNumberOfCharsToCompare = min(
+        mb_strlen($a, $encoding),
+        mb_strlen($b, $encoding)
+      );
+
+      while ($headLength < $maxNumberOfCharsToCompare &&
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
@SooryanshSingh thank you for your efforts. Function getCommonHead is only used once in reuse to compare the upload names which is not handling any complex task, for now these changes are not needed.

SooryanshSingh:
Thank you so much for the feedback, sir.

PR #3245, opened 2026-01-13
fix/folderid validation clean

sakirr05:
## Description

Fix broken `folderId` validation logic in `POST /uploads` endpoint. The validation condition incorrectly used `&&` (AND) instead of `||` (OR), causing invalid `folderId` values (non-numeric, zero, negative) to pass validation and fail later with confusing errors.

### Changes

1. **Fixed `folderId` validation logic** in `UploadController::postUpload()`:
   - Changed `!is_numeric($folderId) && $folderId > 0` to `!is_numeric($folderId) || intval($folderId) <= 0`
   - Now properly catches non-numeric strings, zero, and negative values
   - Aligns with validation patterns used in other controllers (e.g., `FolderController`, `ReportController`)

2. **Added comprehensive unit test coverage**:
   - 6 new test cases covering invalid `folderId` scenarios (non-numeric, zero, negative)
   - Tests for both API V1 and V2
   - Verifies `HttpBadRequestException` is thrown with correct error message

**Files Modified:**
- `src/www/ui/api/Controllers/UploadController.php` - Fixed validation logic (line 496-498)
- `src/www/ui_tests/api/Controllers/UploadControllerTest.php` - Added test coverage

### Manual Testing

1. **Test with non-numeric folderId:**
   ```bash
   curl -X POST 'http://localhost:8081/repo/api/v1/uploads' \
     -H "Authorization: Bearer <token>" \
     -H 'folderId: abc' \
     -H 'uploadType: file' \
     -F 'fileInput=@sample.tar.gz'
   ```
   **Expected:** HTTP 400 Bad Request with message `"folderId must be a positive integer!"`

2. **Test with zero folderId:**
   ```bash
   curl -X POST 'http://localhost:8081/repo/api/v1/uploads' \
     -H "Authorization: Bearer <token>" \
     -H 'folderId: 0' \
     -H 'uploadType: file' \
     -F 'fileInput=@sample.tar.gz'
   ```
   **Expected:** HTTP 400 Bad Request with message `"folderId must be a positive integer!"`

3. **Test with negative folderId:**
   ```bash
   curl -X POST 'http://localhost:8081/repo/api/v1/uploads' \
     -H "Authorization: Bearer <token>" \
     -H 'folderId: -1' \
     -H 'uploadType: file' \
     -F 'fileInput=@sample.tar.gz'
   ```
   **Expected:** HTTP 400 Bad Request with message `"folderId must be a positive integer!"`

4. **Test with valid folderId (should still work):**
   ```bash
[...]

PR #3268, opened 2026-01-23
fix folderid validation

sakirr05:
## Description

Fix `folderId` validation bug in `POST /uploads` endpoint. The validation had wrong operator precedence - `!is_numeric($folderId) && $folderId > 0` would never be true. This re-implements the fix from PR #3245 with clean commits and proper tests.

### Changes

- Fixed validation logic to properly check for empty, non-numeric, or <= 0 values
- Added handling for array case (multipart/form-data can send folderId as array)
- Added tests for null, empty, non-numeric, zero, and array folderId cases

**Files Modified:**
- `src/www/ui/api/Controllers/UploadController.php`
- `src/www/ui_tests/api/Controllers/UploadControllerTest.php`

## How to test

Run the unit tests:
```bash
phpunit src/www/ui_tests/api/Controllers/UploadControllerTest.php --filter testPostUpload
```

Manual test with invalid folderId should return 400:
```bash
curl -X POST 'http://localhost:8081/repo/api/v1/uploads' \
  -H "Authorization: Bearer <token>" \
  -H 'folderId: abc' \
  -H 'uploadType: vcs' \
  -d '{"location": "vcsData"}'
```

## Related PR

Addresses issues from PR #3245 which was closed due to mixed commits and test issues. This PR has clean commits and proper test coverage.

### DIFF
--- a/src/www/ui_tests/api/Controllers/UploadControllerTest.php
+++ b/src/www/ui_tests/api/Controllers/UploadControllerTest.php
@@ -1039,6 +1039,286 @@ class UploadControllerTest extends \PHPUnit\Framework\TestCase
    * @runInSeparateProcess
    * @preserveGlobalState disabled
    * @test
+   * -# Test for UploadController::postUpload() with null folderId with V1 parameters
+   * -# Check if response status is 400
+   */
+  public function testPostUploadNullFolderIdV1()
+  {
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
@sakirr05 : I still see there are test case failures 
image

also the extra commits 

image


This PR IS not in a acceptable state.

PR #3274, opened 2026-01-27
fix(ui): rename Copy to Alias and add tooltip

### LINKED ISSUE #613: Fix(content_move): the action **Copy** should be renamed to something like **Alias**
On the page `?mod=content_move` one has the option to **Move** or to **Copy** an upload or a folder.

### Situation
The command `./delagent -n fossy  -p fossy -f` gives the following output
```
 1 :: Software Repository
      2 :: folderA
      3 :: folderB
```

If one creates a "copy" of a folder **folderA** to **folderB/folderA** on obtains the following situation
<pre>
 1 :: Software Repository
      <b>2 :: folderA</b>
      3 :: folderB
         <b>2 :: folderA</b>
</pre>
where both folder have the same id and are the same folder.

### The good part
In the situation where you have two groups with disjoint root folders you can simply copy one folder to the other root folder and you have a shared folder between the two groups.

### Problem
The name **Copy** is confusing, since
* If one uploads a file to one of the folders, it also appears in the other.
* One is not able to delete only one of these folders using the delagent, since both have the same id. (One could use `?mod=content_delete`⁽¹⁾)

### More information:
This functionality of **Copy** is **actively used for aliasing folders** and **should not be removed**, instead one might
1) rename **Copy** to **Alias** and
2) implement a new **Copy**.

--
⁽¹⁾ the mod-name of this plugin is also confusing, since it only unlinks, i.e. deletes duplicates without removing the "real thing". In contrast to `?mod=admin_folder_delete` and `?mod=admin_upload_delete`.

Kartikharitwal:
## Description

Renames the "Copy" action to "Alias" on the content_move page and adds a tooltip explaining the alias behavior.

## Changes

- Renamed UI label from "Copy" to "Alias".
- Added tooltip explaining that Alias creates a shared folder and changes in one location appear in all aliases.
- No backend logic changes.

## How to test

1. Open Fossology Web UI.
2. Navigate to Organize → Folders → Move or Copy.
3. Verify the button label shows "Alias".
4. Hover over the Alias button and confirm tooltip text is displayed.

Fixes #613

### DIFF
--- a/src/www/ui/template/admin_content_move.html.twig
+++ b/src/www/ui/template/admin_content_move.html.twig
@@ -34,7 +34,8 @@
         <br/>
         <br/>
 
-        <input type="submit" name="move" value="{{ 'Move'|trans }}"/> &nbsp; <input type="submit" name="copy" value="{{ 'Copy'|trans }}"/>
+        <span title="{{ 'Move: Transfers folder or upload to another location.\nAlias: Creates a shared folder reference. Changes in one appear everywhere.'|trans }}">ℹ</span>
+        <input type="submit" name="move" value="{{ 'Move'|trans }}"/> &nbsp; <input type="submit" name="copy" value="{{ 'Alias'|trans }}"/>
       </form>
     </td>
   </tr>


Kaushl2208:
Hello @Kartikharitwal , can you please present a working demo for the changes done in the UI?

Kartikharitwal:
Screenshot 2026-01-29 184959
this is the proof of the button that i have changed @Kaushl2208

Kartikharitwal:
Hi @Kaushl2208,

I have built and run FOSSology locally from my branch and verified the UI change.

On the "Organize → Folders → Move or Copy" page, the action button is now labeled "Alias" and works as expected.  
i have attached the screenshot in above comment 

Let me know if anything else is required.

src/www/ui/template/admin_content_move.html.twig
```diff
@@ -34,7 +34,7 @@
         <br/>
         <br/>
 
-        <input type="submit" name="move" value="{{ 'Move'|trans }}"/> &nbsp; <input type="submit" name="copy" value="{{ 'Copy'|trans }}"/>
+        <input type="submit" name="move" value="{{ 'Move'|trans }}"/> &nbsp; <input type="submit" name="copy" value="{{ 'Alias'|trans }}" title="{{ 'Alias creates a shared folder. Changes in one location appear in all aliases.'|trans }}"/>
```
Kaushl2208: 1. The title approach is good, But do you think that provides the user information right away? Unless the hover? And there are no precedence to hover on it as well. Would it make sense to either add a (i) tooltip? Or maybe add a small `Note` kinda line where it talks about the buttons?
2. If we have updated the Alias, Can we also add info about `Move` button??

Maybe a single tooltip, That explains both buttons?

Kaushl2208 (changes requested):

src/www/ui/template/admin_content_move.html.twig
Kartikharitwal: Thanks for the feedback!
I agree that relying only on hover over the Alias button may not make the behavior obvious.
Proposed approach:
Keep the rename Copy → Alias.
Add a small info (ℹ️) icon near the Move/Alias buttons.
The tooltip on this icon will explain both actions:
Move: Moves the folder/upload to the selected destination.
Alias: Creates a shared folder reference; changes in one location appear in all aliases.
This makes the behavior discoverable without cluttering the UI.
If this approach is acceptable, I’ll implement it in the same PR.

Kartikharitwal: Updated PR:
- Kept changes limited to UI template only
- Added tooltip explaining Move and Alias buttons
ready for review

src/www/ui/template/admin_content_move.html.twig
```diff
@@ -1,58 +1,63 @@
...
-  <script type="text/javascript">{% include 'admin_content_move.js.twig' %}</script>
-{% endblock %}
+{# SPDX-FileCopyrightText: © 2015 Siemens AG
+
+   SPDX-License-Identifier: FSFAP
+#}
+{% extends "include/base.html.twig" %}
+{% block styles %}
+  {{ parent() }}
+  <link rel="stylesheet" href="css/jquery.treeview.css"/>
```
Kaushl2208: Sir, Please check the changes before you push. Diff shows the entire file as your change. Can you please update it???

Kaushl2208 (changes requested):
Fix the changes you are planning to commit. Also dont forget to rebase and squash

src/www/ui/template/admin_content_move.html.twig
Kartikharitwal: Thanks for the feedback.
I’ve rebased on master, cleaned the diff so only the intended lines are changed, and squashed everything into a single commit.
Please review again.

src/www/ui/template/admin_content_move.html.twig
```diff
@@ -34,7 +34,8 @@
         <br/>
         <br/>
 
-        <input type="submit" name="move" value="{{ 'Move'|trans }}"/> &nbsp; <input type="submit" name="copy" value="{{ 'Copy'|trans }}"/>
+        <span title="{{ 'Move: Transfers folder or upload to another location.\nAlias: Creates a shared folder reference. Changes in one appear everywhere.'|trans }}">ℹ</span>
```
Kaushl2208, line 37: Image


Do you think this (i) button looks good? Or is it the ideal place it should be at??

Look at the other places where we have utilized tooltips. One such example is below

https://github.com/fossology/fossology/blob/6154b3d7218eb0a0829affcade1ee96b34e28f38/src/lib/php/UI/template/include/upload.html.twig#L44

Kaushl2208 (changes requested):
PTAL

Kaushl2208:
No response from the contributor for Two weeks. Closing this now.

PR #3279, opened 2026-01-28
feat(reportImport): replace EasyRdf with SweetRDF (PHP 8 compatibility)

cooldark:
### Description

 - Replace easyrdf/easyrdf 1.1.1 with sweetrdf/easyrdf 1.18.2
 - Set platform PHP to 8.20 to fix XML handler failure on PHP 8
  
 Why
 - easyrdf/easyrdf has no recent releases (since 2020) and caused runtime errors under newer PHP
 - sweetrdf/easyrdf is maintained and supports PHP >= 8.0
 
 
###  Changes

**Breaking change as PHP 8.0 is minimum** - not sure how to work with that

### DIFF
--- a/src/composer.json
+++ b/src/composer.json
@@ -10,7 +10,7 @@
         "without any warranty."
     ],
     "require" : {
-        "easyrdf/easyrdf" : "1.1.1",
+        "sweetrdf/easyrdf" : "1.18.2",
         "ext-gettext": "*",
         "ext-mbstring": "*",
         "ext-json" : "*",
@@ -71,7 +71,7 @@
     },
     "config" : {
         "platform" : {
-            "php" : "7.4.3"
+            "php" : "8.0.30"
         }
     },
     "homepage" : "https://www.fossology.org/",
--- a/src/composer.json.in
+++ b/src/composer.json.in
@@ -10,7 +10,7 @@
         "without any warranty."
     ],
     "require" : {
-        "easyrdf/easyrdf" : "1.1.1",
+        "sweetrdf/easyrdf" : "1.18.2",
         "ext-gettext": "*",
         "ext-mbstring": "*",
         "ext-json" : "*",
@@ -71,7 +71,7 @@
     },
     "config" : {
         "platform" : {
-            "php" : "7.4.3"
+            "php" : "8.0.30"
         }
     },
     "homepage" : "https://www.fossology.org/",
--- a/src/composer.lock
+++ b/src/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "bf3f06c77e971498cca3dc6d971edc0d",
+    "content-hash": "4148442393b241a5312aaa65d9ff11fc",
     "packages": [
         {
             "name": "composer/spdx-licenses",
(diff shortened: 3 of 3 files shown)


src/composer.json
```diff
@@ -71,7 +71,7 @@
     },
     "config" : {
         "platform" : {
-            "php" : "7.4.3"
+            "php" : "8.2.0"
```
GMishx: Ummm..... Why?

cooldark: - sweetrdf/easyrdf  dropped PHP 7.x support with version [1.9.0](https://github.com/sweetrdf/easyrdf/releases/tag/1.9.0)
 - PHP 8.4 support was added with version [1.15.0](https://github.com/sweetrdf/easyrdf/releases/tag/1.15.0)

So a minimum of PHP 8.x is needed - Latest version is 8.0.30 then. I'll update.

Kaushl2208:
Hey @cooldark ,
> easyrdf/easyrdf has no recent releases (since 2020) and caused runtime errors under newer PHP

It was working perfectly under our `Build & Test` for the php versions that we test for. There is PHP `8.1` test available. My question is regarding the above statement to clarify --> Which specific PHP version the dependency was not behaving properly?
If you have tested under the version and it is failing, Can you share some Screenshot for such? 

Tho, I have gone through the `easyrdf/easyrdf` repo and I could see they were facing some issues with new PHP versions, (Still a log or a screenshot with fossology usecase would be great)

Now about changing the PHP version: I think we can go ahead with that but try to keep in mind the distributions support we provide.
for debian: trixie, bookworm and bullseye(EOL, can be removed)
for ubuntu: noble and jammy.

So find a minimal version that supports all these, remove any older version support from (code and CI). Then we can go ahead with the changes.

cooldark:
Hi @Kaushl2208 [Kaushl2208](https://github.com/Kaushl2208)

we updated our Fossology instance to Debian Trixie (from Bookworm), which brings PHP 8.4.16 (also named in utils/fo-installdeps). And the PHP Fatal error we get is with 'ReportImport' as described here: 
https://github.com/fossology/fossology/discussions/3275

And the underlying error is in the easyrdf/easyrdf - It could be fixable there i am sure, but contributing a fix to a no longer maintained repo? 

To be more precise: In bookworm with PHP 8.2 there was no error.

As there is PHP 8.2 / 8.3 and 8.4 officially supported when it comes to the supported OS maybe the Build & Test needs to be extended. Ubuntu Jammy brings 8.1 as far as i know. Dropping Debian Bullseye will allow to set a minimum of PHP 8.1 for future releases, no?

Kaushl2208:
@cooldark , Oh no, I didnt mean to contribute to the `easyrdf/easyrdf` project. I was just asking the version details so I can replicate the same locally and check where the issue is. 

And yes you are right, dropping bullseye, will fix the updated version. But this needs to handled step by step. 
1. We should drop the debian bullseye support
2. We should then update the `php` version according to least required by the supported OS.
3.  Then finally the `easyrdf/easyrdf` package should be updated to --> ` sweetrdf/easyrdf` with the test cases for the new package and required changes in the code (if any)

That would be the approach i would suggest, In three different PRs (easy to maintain and roll back), If you would like to contribute to any of it, please do.

cooldark:
Hi @Kaushl2208

thanks a lot for your answer and I am absolutely fine with that proposal. So nothing gets lost or broken by accident. I probably will be able to contribute on the steps. BTW: I also saw in a lot of files still Debian stretch and buster are named as well. 

Will you create a issue on that mentioning the step-by-step proposal where it can be tracked later?
And it seems it's more a mid term change as Bullsyse is on "Long Term Support" state lasting till end of august 2026 - maybe focus and 'cleanup' the older OSes first?

For the time being this PR seems to be on hold.

Kaushl2208:
Hey @cooldark , Sure, I will create an issue for tracking this. Lets close this PR for now. We will do it step by step.

As for bullseye, We track for the EOL. And I think we have passed that very long ago. And by the time we do a new release we would have crossed the LTS too. 

Hopefully there would not be a lot left to clean for older OS but if there is, Lets clean that too.

Issue #3290, opened 2026-02-01
Utility scripts fail due to hardcoded `pathinclude.php` require


Aditya30ag:
## Problem
Several Fossology utility scripts fail at runtime because they use a hardcoded, invalid include path:

```php
require_once("FIXMETOBERELATIVE/pathinclude.php");
```

This affects the following utility scripts:
    - get-projects.php
    - mk_fmdirs.php
    - diffm.php
    - mktop1k.php
    - get-fsrc.php

Aditya30ag:
@Kaushl2208 I wanted to confirm whether this change would be useful before proceeding further.

Kaushl2208:
No they arent, I have mentioned it on the PR. 
I will do it again here: they are old utility scripts which are not in use at all.

Aditya30ag:
@Kaushl2208 
Ohh ok thanks for clarification

PR #3291, opened 2026-02-01
Fix utility scripts to locate pathinclude.php robustly

Aditya30ag:
## Description

Fix utility scripts failing at runtime due to a hardcoded and invalid `pathinclude.php` require path.

Several scripts referenced `FIXMETOBERELATIVE/pathinclude.php`, which caused fatal errors when running
the utilities outside a very specific directory layout (e.g. system installs, WSL, or CI).

## Changes

- Replaced the hardcoded `require_once("FIXMETOBERELATIVE/pathinclude.php")` with a robust lookup
  mechanism.
- Added checks for `pathinclude.php` in common repository-relative and system install locations.
- Improved error handling by emitting a clear, actionable message when `pathinclude.php` cannot be found.

Updated scripts:
- `get-projects.php`
- `mk_fmdirs.php`
- `diffm.php`
- `mktop1k.php`
- `get-fsrc.php`

Closing issue #3290

### DIFF
--- a/utils/fedora/get-fsrc.php
+++ b/utils/fedora/get-fsrc.php
@@ -37,7 +37,26 @@
 // FIXME: $path = '/usr/local/fossology/agents';
 set_include_path(get_include_path() . PATH_SEPARATOR . $path);
 
-require_once("FIXMETOBERELATIVE/pathinclude.php");
+// Try common locations for pathinclude.php and fail with helpful message if not found.
+$__path_candidates = array(
+  dirname(__FILE__) . '/../../src/lib/php/pathinclude.php',
+  dirname(__FILE__) . '/../../src/php/pathinclude.php',
+  dirname(__FILE__) . '/../php/pathinclude.php',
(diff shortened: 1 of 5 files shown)


utils/fedora/get-fsrc.php
```diff
@@ -37,7 +37,26 @@
...
+  if (file_exists($__p)) {
+    require_once $__p;
+    $__found = true;
+    break;
+  }
+}
+if (! $__found) {
+  fwrite(STDERR, "FATAL: cannot find pathinclude.php. Please install Fossology or add pathinclude.php to your include path.\n");
+  exit(1);
+}
```
Kaushl2208, line 59: First of all, These are old utility scripts that are not even under usage right now. 

Second, How is this change any better then the one before? Atleast the earlier one was providing user to hardcode a path. You just have hardcoded 5-6 paths in a hope that the file would be somewhere there. 

Please discuss before doing such changes. They waste both your effort and ours. Please join community call next time and discuss if you want to pickup something.

PR #3292, opened 2026-02-01
Mayank/fix/packages instance state

LittleCodr:
## Summary
Fix shared mutable state in utils/automation/FoScanner/Packages.py by making package metadata instance-specific.

## Problem
Packages currently defines dependencies as a class-level dict, so multiple scans in one process can reuse and mutate the same object, leaking dependency data across runs.

## Fix
Initialize per-instance state in __init__, giving each Packages instance its own dependency container. No other logic or dependencies change.

## Scope
Minimal, localized change; behavior for a single run is unchanged aside from removing unintended state sharing.

### DIFF
--- a/utils/automation/FoScanner/Packages.py
+++ b/utils/automation/FoScanner/Packages.py
@@ -13,5 +13,6 @@ class Packages(object):
   :ivar parent_package: Parent package during the scan
   :ivar dependencies: List of dependencies of the parent package
   """
-  parent_package: dict = None
-  dependencies: dict[str, dict] = {}
+  def __init__(self):
+    self.parent_package = None
+    self.dependencies: dict[str, dict] = {}


utils/osadl_convertor.py
```diff
@@ -262,7 +263,7 @@ def get_license_types(self) -> list[Optional[str]]:
     resp: list[tuple[Optional[str]]] = cur.fetchall()
     type_list: list[Optional[str]] = []
     for row in resp:
-      if row is not None:
+      if row[0] is not None:
```
Kaushl2208, line 266: Umm...Why??

utils/osadl_convertor.py
```diff
@@ -403,8 +404,13 @@ def convert_json_to_matrix(license_handler: LicenseHandler, json_loc: str) \
...
   type_dict: dict[tuple[str, str, bool], int] = dict()
-  with open(json_loc, "r") as jsoninput:
-    matrix = json.load(jsoninput)
+  if not os.path.isfile(json_loc):
+    raise FileNotFoundError(f"Input JSON file '{json_loc}' not found.")
+  try:
+    with open(json_loc, "r") as jsoninput:
+      matrix = json.load(jsoninput)
+  except json.JSONDecodeError as exc:
+    raise ValueError(f"Input JSON file '{json_loc}' is not valid JSON: {exc}") from exc
```
Kaushl2208, line 413: Why do we need a secondary defensive check here???  json file is an imported package entity?

PR #3305, opened 2026-02-04
fix(spdx2rdf): Fix validation errors

JanAltenberg:
## Description
Validation (with recent SPDX java tools) of exported SPDX2 RDF documents fails with validation errors for copyrightText and downloadLocation for the NOASSERTION case:

```
<spdx:copyrightText rdf:resource="http://spdx.org/rdf/terms#noassertion" />
```
leads to
```
Relationship error: Error getting comment: Property http://spdx.org/rdf/terms#copyrightText is not of type String in XYZ in /srv/fossology/repository/report
```
and
```
<spdx:downloadLocation rdf:resource="http://spdx.org/rdf/terms#noassertion" />
```
leads to
```
Relationship error: Unable to get download location: Property http://spdx.org/rdfterms#downloadLocation is not of type String in /srv/fossology/repository/report
```

Both properties are data properties.

### Changes

This commit changes the templates accordingly to the value range expected by the specification. downloadLocation in this scenario will be reflected as:
```
<spdx:downloadLocation>http://spdx.org/rdf/terms#noassertion</spdx:downloadLocation>
```

This is what the spec says about downloadLocation:
grafik

and copyrightLocation will be reflected as:
```
<spdx:copyrightText>NOASSERTION</spdx:copyrightText>
```
This is what the spec says on copyrightText:
grafik

All validation tests have been run with java spdx tools 2.0.2 and 2.0.4

## How to test

1. Upload a package / Scan for copyright and licenses
2. Export SPDX2 RDF
3. Validate with java spdx tools
4. Without these changes validation will fail for copyrights with noassertion and for the download location

@shaheemazmalmmd @GMishx: Would be great if you could double check the changes against the spec and verify if my interpretation is correct.

Many thanks in advance and let me know if I can support with further information :)

### DIFF
--- a/src/spdx/agent/template/spdx2-file.xml.twig
+++ b/src/spdx/agent/template/spdx2-file.xml.twig
@@ -139,7 +139,7 @@
   {%- endfor %}{# End printing license infos in file #}
 {%- endif %}
 {%~ if fileData.copyrights|default is empty %}
-    <spdx:copyrightText rdf:resource="http://spdx.org/rdf/terms#noassertion" />
+    <spdx:copyrightText>NOASSERTION</spdx:copyrightText>
 {%~ else %}
     <spdx:copyrightText><![CDATA[
   {%~ for cp in fileData.copyrights %}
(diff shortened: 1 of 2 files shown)


Kartikharitwal:
@Kaushl2208 
Thanks for adding the labels. I can try testing this locally with an SPDX2 RDF export and validate it using the Java SPDX tools (2.0.2 / 2.0.4 as mentioned).
I’ll report back here if I notice anything unexpected.

PR #3306, opened 2026-02-04
feat(ui): add search option in con lic page text view

its-sushant:
## Description

Implemented a new search and navigation feature for the file clearing view with support for cross-node highlighting.

## How to test

- **Search Overlapping Text**: Search for words that are already partially highlighted by agents.
- **Search Substrings**: Search for a substring within an existing highlighted phrase.
- **General Search & Navigation**: Perform a standard text search and use the up/down arrows. Verify that the "Match X of Y" counter updates correctly and the view auto-scrolls to the active match.

### DIFF
--- a/src/www/ui/ui-clearing-view.php
+++ b/src/www/ui/ui-clearing-view.php
@@ -59,6 +59,10 @@ class ClearingView extends FO_Plugin
   private $invalidParm = false;
   /** @var DecisionTypes */
   private $decisionTypes;
+  /** @var int */
+  private $searchChunkSize = 1048576;
+  /** @var int */
+  private $searchBlockSize = 81920;
 
   function __construct()
   {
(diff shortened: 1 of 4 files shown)


src/www/ui/ui-clearing-view.php
```diff
@@ -367,6 +395,51 @@ public function updateLastItem($userId, $groupId, $lastItem, $currentUploadtreeI
...
+    $searchLen = strlen($searchQuery);
+    $currentFilePos = 0;
+    $overlapBuffer = '';
+
+    while (!feof($handle)) {
+      $chunk = fread($handle, $chunkSize);
+      $haystack = $overlapBuffer . $chunk;
+
+      $offset = 0;
+      while (($pos = stripos($haystack, $searchQuery, $offset)) !== false) {
```
GMishx: Try using [preg_match_all](https://stackoverflow.com/a/61144122) as it will give you all results in one shot. Try comparing the performance of both.

its-sushant: Sure @GMishx Let me try.

its-sushant: With **preg_match_all**, there is around **50%** improvement in performance.

shaheemazmalmmd (approved):

PR #3309, opened 2026-02-05
feat(maintagent): Introduce Maintagent2

helmut-leher-magna:
PR adds an agent that improves how large ID sets are handled: **maintagent2**
 -- It is a copy of existing "maintagent" with modifications
 -- Keeps the original maintagent unchanged
 -- maintagent2 can be invoked with the same CLI options

Changes in comparison to the original one:
- It creates a session temp table tmp_ids_<pid> and populates it by COPY-ing IDs from stdin to avoid huge client-side IN (...) lists (when using -R flag).
- Uses VACUUM FULL if database is PSQL >= 17 (maybe to be reverted)
- LIKE comparison instead of SIMILAR TO
- Richer debug info via -vvv : Provides deterministic START/END markers and time duration
- 'Adapter implementation' that forwards to existing libfossdb functions. Use alternate DB helpers without modifying the original libfossdb files.
  - src/lib/c/libfossdb_adapter.c
  - src/lib/c/libfossdb_adapter.h

**Motivation:**
Flag "-R" collects many IDs and concatenates them into a single SQL statement (DELETE ... WHERE id IN (...)). Very large IN lists:
  - blow up client-side memory or string buffers,
  - create extremely long SQL statements that take the server a long time to parse/plan,
  - increase risk of network/transmission/connection issues (timeouts, broken connections),
  - and can cause severe performance degradation or failures under load.

NOTE: No tests were written (yet).

I am open to discuss the idea.
Shall it be separated like not, or better replacing the original maintagent?


Kaushl2208:
Hi @helmut-leher-magna , I am trying to understand the goal here. From what I understand this PR is directed towards additional features to `maintenance agent`, logically the extension of features should be in the core agent right? You dont need to create a separate agent for it. Also the features and extensions that you are trying to introduce needs to be discussed and understood properly. 

So I would suggest, if you can join the community call we can discuss all the points there and can look into proper steps on how to approach this. (Closing this PR for now)

PR #3043, opened 2025-05-01
fix(browse): fixed comment and description handling for deletion

harshitg927:
## Description

This PR fixes the issue where users cannot clear upload descriptions and comments by submitting empty strings. The current implementation incorrectly prevents empty strings from being processed, causing the UI to retain previous values when users attempt to clear these fields.

### Changes

Modified AjaxBrowse.php and admin-upload-edit.php to properly handle empty string submission.

## How to test

1. **For Upload Description:**
   - Navigate to `Organize` -> `Uploads` -> `Edit Properties`
   - Clear text in the Upload Description field
   - Click on `Edit` button
   - Verify the description is properly cleared

2. **For Upload Comment:**
   - Navigate to `Browse` tab
   - Locate an upload and double-click on the Comment cell
   - Either clear the text field and click `OK`
   - Verify the comment field is properly cleared

Please consider using the closing keyword if the pull request is proposed to
fix an issue already created in the repository
Fixes: [issue#3002](https://github.com/fossology/fossology/issues/3002)

PR #3311, opened 2026-02-06
Fix: Allow upload description to be updated to an empty string

SvAlpha:
## Description

This PR fixes a bug in the "Edit Upload Properties" page where users were unable to clear an upload description (i.e., set it to an empty string). The backend logic previously used `!empty($newDesc)` to check for changes, which caused empty strings to be ignored during the update process.

### Changes

- Modified `src/www/ui/admin-upload-edit.php`.
- Changed the condition from `!empty($newDesc)` to `isset($newDesc)` in the `UpdateUploadProperties` function.
- This ensures that if the user explicitly sends an empty description, it is saved to the database rather than ignored.

## How to test

1. Go to **Organize > Uploads > Edit Properties** for any existing upload.
2. Delete the text in the **Description** field (make it blank).
3. Click **Update**.
4. Refresh the page or view the upload details to verify that the description is now empty.
5. (Regression Test) verify that updating the description to a *non-empty* string still works as expected.

### DIFF
--- a/src/www/ui/admin-upload-edit.php
+++ b/src/www/ui/admin-upload-edit.php
@@ -46,15 +46,12 @@ class upload_properties extends FO_Plugin
    **/
   public function UpdateUploadProperties($uploadId, $newName, $newDesc)
   {
-    if (empty($newName) and empty($newDesc)) {
+    // MR ROGERS WAS HERE - FIX 2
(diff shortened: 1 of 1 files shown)


Kaushl2208:
Similar PR already opened #3043 

No point in making the changes again. Hence closing

PR #3314, opened 2026-02-06
fix (cli) : Add error handling for failed downloads in cp2foss

### LINKED ISSUE #503: cp2foss -s hangs if file can't be downloaded
With the 2.6.2 release:

cp2foss --username fossy --password fossy -P 1 -s -q all ftp://ftp.gnome.org/pub/GNOME/sources/GConf/3.2/GConf-3.2.6.tar.gz

never completes, because that URL is a 404.  Shouldn't it abort when the file can't be downloaded?

SalmanDeveloperz:
## Description

This prevents cp2foss from silently queueing wget_agent jobs for inaccessible URLs and provides immediate feedback to users.

### Changes

Before fix:

- cp2foss would queue a wget_agent job for ANY URL (even 404s)
- The job would silently fail
- User gets no immediate feedback
- Script exits normally (no error)

After fix:

- cp2foss checks URL accessibility BEFORE queueing
- Returns clear error message for 404s
- Exits with error code 1
- User knows immediately if URL is bad

## How to test

Added test cases (HTTP 404 Not Found, DNS Resolution Failure, Issue URL, Valid URL) to verify functionality and all run test passes successfully and provide response codes before attempting download.

image

image

Closes #503 where `cp2foss` would silently queue jobs for inaccessible URLs. Now users get immediate feedback with clear error messages.

### DIFF
--- a/src/cli/cp2foss.php
+++ b/src/cli/cp2foss.php
@@ -566,6 +566,39 @@ $UploadArchiveTmp = realpath($UploadArchive);
 if (!$UploadArchiveTmp) {
   // neither a file nor folder from server?
   if (filter_var($UploadArchive, FILTER_VALIDATE_URL)) {
+    // Validate that the URL is actually accessible before proceeding
+    if ($Verbose) {
+      print "Checking URL accessibility: $UploadArchive\n";
+    }
+    
+    $ch = curl_init($UploadArchive);
+    curl_setopt($ch, CURLOPT_NOBODY, true);
+    curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+    curl_setopt($ch, CURLOPT_TIMEOUT, 30);
+    curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, false);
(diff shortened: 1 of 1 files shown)


Kaushl2208:
Hey @SalmanDeveloperz , As far as I understand `wget_agent` already handles the response code and returns exit status. Even though this looks fancy, But it is just a cli script that uses the same agent, So, the responsibility of handling this lies with the agent and Agent returns the exit code. (which solves the part of notifying)

Also this cli script is useful but pretty old and not in use at all. Will be closing this now. Thanks for your contribution

SalmanDeveloperz:
Thanks @Kaushl2208, I appreciate you reviewed this.

In #503,  [dankegel](https://github.com/dankegel) confirmed:
> never completes, because that URL is a 404. Shouldn't it abort when the file can't be downloaded?
> It aborts rather quickly, if the URL returns a 404, however, the user does not see very clearly that something went wrong:

I understand `wget_agent` handles errors internally and the concerns about architecture. Could you guide on the preferred approach to solve this ticket? I'll happy to implement whatever solution you think is best.

PR #3316, opened 2026-02-07
fix(upload): preserve UTF-8 characters in uploadtree.ufile_name

### LINKED ISSUE #1457: Viewable name is broken when written in Korean
### Description

When uploading, it works fine if the description is written in Korean.
But when I write viewable name in Korean, it is broken while browsing.

I suppose database holds the data correctly, but problem occurs when displaying it.

I assume same thing will happen for non-Latin characters. 

#### How to reproduce

Write viewable name in Korean, you'll see the result.

#### Screenshots

image

image

image

image

### Versions

* Last commit id on master: 3.5.0 #171d4d
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.6 LTS
Release:        16.04
Codename:       xenial

sakshammishra112:
## Description

This PR fixes an issue where non-ASCII characters (Korean, Japanese, etc.) in repository/file names were being corrupted and stored as `~~~~~~` in the `uploadtree.ufile_name` column.

While the description (`upload.upload_desc`) was stored correctly, the UI retrieves the display name from `uploadtree.ufile_name`, causing incorrect display of filenames.

Screenshot 2026-02-07 162134

After fixing the database storage to preserve UTF-8 characters, another issue appeared during report generation:

Screenshot 2026-02-07 172141

To resolve this, additional logic was added to report generation files to properly handle non-ASCII filenames.

This PR ensures correct handling of Unicode filenames across database storage, UI display, and report generation.

---

### Root Cause

- During uploads via URL, non-ASCII characters in filenames were processed using ASCII-only validation.
- As a result, multi-byte UTF-8 characters were replaced with `~` and stored incorrectly in `uploadtree.ufile_name`.
- The UI reads filenames from this column, leading to corrupted display.
- After fixing storage, report generators failed because they assumed ASCII-only filenames.

---

### 1. Database Storage Fix

- Fixed filename processing in the unpacking pipeline.
- Ensured UTF-8 characters are preserved when storing data in `uploadtree.ufile_name`.
image

### 2. Report Generation Fix

- Added logic in report generation files to handle non-ASCII filenames.
- Implemented safe ASCII fallback when required.
- Prevented "filename fallback" errors.

---

## Results:

[Reports link](https://drive.google.com/drive/folders/1PexjnxIe9qaHl27LU8NMg4ChfuDmetap?usp=drive_link)

Screenshot 2026-02-07 164042
Screenshot 2026-02-07 164443
Screenshot 2026-02-07 164029

## How to Reproduce

1. Upload a file using **Upload from URL**.
2. Under **3. (Optional) Enter a viewable name for this file or directory**, enter a Korean or Japanese name.
   Example: ありがとう
3. Go to browser and check

### DIFF
--- a/src/unifiedreport/agent/unifiedreport.php
+++ b/src/unifiedreport/agent/unifiedreport.php
@@ -954,7 +954,14 @@ class UnifiedReport extends Agent
       mkdir($fileBase, 0777, true);
     }
     umask(0022);
-    $fileName = $fileBase. "Clearing_Report_".$packageName.".docx";
+    
+    // Check if packageName contains non-ASCII characters and create ASCII-safe fallback
+    if (preg_match('/[^\x20-\x7E]/', $packageName)) {
+      $safeName = "upload_" . time() . "_unified_report";
+      $fileName = $fileBase. "Clearing_Report_".$safeName.".docx";
+    } else {
+      $fileName = $fileBase. "Clearing_Report_".$packageName.".docx";
+    }
     $objWriter = IOFactory::createWriter($phpWord, "Word2007");
     $objWriter->save($fileName);
 
--- a/src/ununpack/agent/ununpack.c
+++ b/src/ununpack/agent/ununpack.c
@@ -137,6 +137,15 @@ int	main(int argc, char *argv[])
   /* connect to the scheduler */
   fo_scheduler_connect(&argc, argv, &pgConn);
 
+  /* Set UTF-8 client encoding to properly handle non-ASCII characters */
+  if (pgConn != NULL) {
+    PGresult *result = PQexec(pgConn, "SET client_encoding TO 'UTF8'");
+    if (PQresultStatus(result) != PGRES_COMMAND_OK) {
+      LOG_WARNING("Failed to set UTF-8 client encoding: %s", PQerrorMessage(pgConn));
+    }
+    PQclear(result);
+  }
+
   while((c = getopt(argc,argv,"ACc:d:FfHhL:m:PQiIqRr:T:t:U:VvXxE:")) != -1)
   {
     switch(c)
--- a/src/ununpack/agent/utils.c
+++ b/src/ununpack/agent/utils.c
@@ -1351,8 +1351,17 @@ int	DBInsertUploadTree	(ContainerInfo *CI, int Mask)
   {
     /* postgres 8.3 seems to have a problem escaping binary characters
      * (it works in 8.4).  So manually substitute '~' for any unprintable and slash chars.
+     * Updated to handle UTF-8 characters properly by checking for valid UTF-8 sequences
+     * instead of using isprint() which only works with ASCII characters.
      */
-    for (cp=UfileName; *cp; cp++) if (!isprint(*cp) || (*cp=='/') || (*cp=='\\')) *cp = '~';
+    for (cp=UfileName; *cp; cp++) {
+      unsigned char c = (unsigned char)*cp;
+      /* Replace control characters (ASCII 0-31) and path separators */
+      if ((c < 32 && c != '\t' && c != '\n' && c != '\r') || (*cp=='/') || (*cp=='\\')) {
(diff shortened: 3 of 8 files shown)


src/ununpack/agent/utils.c
```diff
@@ -1351,8 +1351,17 @@ int	DBInsertUploadTree	(ContainerInfo *CI, int Mask)
...
      */
-    for (cp=UfileName; *cp; cp++) if (!isprint(*cp) || (*cp=='/') || (*cp=='\\')) *cp = '~';
+    for (cp=UfileName; *cp; cp++) {
+      unsigned char c = (unsigned char)*cp;
+      /* Replace control characters (ASCII 0-31) and path separators */
+      if ((c < 32 && c != '\t' && c != '\n' && c != '\r') || (*cp=='/') || (*cp=='\\')) {
+        *cp = '~';
+      }
+      /* All other characters (including UTF-8 sequences) are kept as-is */
+    }
```
Kaushl2208, line 1364: This is a good catch the current use of `isprint(*cp)` is byte-based and not UTF-8 aware, which explains why multibyte filenames are being corrupted before reaching the database.

However, the proposed change feels more like a workaround than a true fix. It makes the logic UTF-8 tolerant, but not UTF-8 validating, and does not address the underlying assumption that filename sanitization should be byte-based.

Additionally, this logic appears to have been introduced specifically to work around escaping issues in PostgreSQL 8.3, which is now extremely outdated. Modern PostgreSQL versions correctly handle UTF-8 when database encoding and collation are configured, as is already the case here.

Given that the corruption occurs before insertion, and the database no longer requires this workaround, do we still need the isprint()-style sanitization at all? Would it be sufficient to only filter path separators (/, \) and ASCII control characters, and otherwise preserve valid UTF-8 filenames as-is?

This needs a better discussion with the maintainers. I would suggest if you can join the community call, We can discuss the relevancy of this problem in detail and can come up to a decision.

src/ununpack/agent/ununpack.c
```diff
@@ -137,6 +137,15 @@ int	main(int argc, char *argv[])
...
   fo_scheduler_connect(&argc, argv, &pgConn);
 
+  /* Set UTF-8 client encoding to properly handle non-ASCII characters */
+  if (pgConn != NULL) {
+    PGresult *result = PQexec(pgConn, "SET client_encoding TO 'UTF8'");
+    if (PQresultStatus(result) != PGRES_COMMAND_OK) {
+      LOG_WARNING("Failed to set UTF-8 client encoding: %s", PQerrorMessage(pgConn));
+    }
+    PQclear(result);
+  }
```
Kaushl2208, line 147: Wouldn't suggest this, We already set this in DB init steps.

src/unifiedreport/agent/unifiedreport.php
```diff
@@ -954,7 +954,14 @@ private function writeReport($contents, $uploadId, $groupId, $userId)
...
     umask(0022);
-    $fileName = $fileBase. "Clearing_Report_".$packageName.".docx";
+    
+    // Check if packageName contains non-ASCII characters and create ASCII-safe fallback
+    if (preg_match('/[^\x20-\x7E]/', $packageName)) {
+      $safeName = "upload_" . time() . "_unified_report";
+      $fileName = $fileBase. "Clearing_Report_".$safeName.".docx";
+    } else {
+      $fileName = $fileBase. "Clearing_Report_".$packageName.".docx";
+    }
```
Kaushl2208, line 964: The validation of filename should happen. If there is something wrong with the tool part, Fallback cannot be anything, these files are used sometime used by legals too.

Kaushl2208 (commented):
@sakshammishra112 , This issue needs a better discussion on the validity of the bug fix that occurred during an older version of Postgres. 

This PR right now doesnt add anything relevant, just a work around. Maybe with a better discussion we can come up with a solution. I hope you join the community call to discuss this issue better.

PR #3318, opened 2026-02-09
fix(ui): avoid redirecting license browser for empty directories

### LINKED ISSUE #3312: [Bug] License Browser incorrectly redirects empty directories to File License View
### Description
In the **License Browser** (`mod=license`), there is logic intended to automatically redirect users to the detailed **License View** (`mod=view-license`) when an item has no scanners to summarize (which is expected for single files).

However, this logic currently fails to distinguish between a file and a directory. It incorrectly treats **empty directories** as files, forcing a redirect to a file-level inspection page.

### Steps to Reproduce
1. Upload an archive containing an empty directory (e.g., `mkdir empty-folder && zip archive.zip empty-folder`).
2. Navigate to the `empty-folder` in the FOSSology file browser.
3. Select **License Browser** from the menu.

### Actual Behavior
The user is immediately redirected to the "License Info" page (`mod=view-license`). Since it is a directory, the page appears empty or out of context, which is confusing and breaks the logical navigation flow.

### Expected Behavior
The user should stay on the **License Browser** dashboard for that folder. The table should simply be empty, consistently showing that it is a directory dashboard with no contents yet.

### Screenshots

Image

After clicking on `empty-folder` :

Image

im-vishesh15th:
### Summary
- Fixed incorrect redirect in license browser for empty directories

### Problem
- `childCount == 0` was treated as a file
- Empty directories also have `childCount == 0`
- This caused unwanted redirect to `view-license`

### Fix
- When `childCount == 0`, fetch upload entry from DB
- Check `ufile_mode` to determine file vs directory
- Redirect only if the item is **not** a directory

### Result
- Files still redirect correctly
- Empty directories now remain in browser view

Fixes #3312

### DIFF
--- a/src/www/ui/page/BrowseLicense.php
+++ b/src/www/ui/page/BrowseLicense.php
@@ -193,7 +193,10 @@ class BrowseLicense extends DefaultPlugin
      * $ChildCount can also be zero if the directory is empty.
      * **************************************/
     if ($childCount == 0) {
-      return new RedirectResponse("?mod=view-license" . Traceback_parm_keep(array("upload", "item")));
+      $itemEntry = $this->uploadDao->getUploadEntry($itemTreeBounds->getItemId(), $this->uploadtree_tablename);
+      if ($itemEntry && !Isdir($itemEntry['ufile_mode'])) {
+        return new RedirectResponse("?mod=view-license" . Traceback_parm_keep(array("upload", "item")));
+      }
     }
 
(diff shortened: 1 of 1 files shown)


its-sushant (commented):
Thankyou @im-vishesh15th for the contribution. Changes look good to me. Need test!

CC: @Kaushl2208

Kaushl2208:
Thanks @im-vishesh15th and @its-sushant!!

PR #3319, opened 2026-02-09
fix(mta): use swaks as an alternate to s-nail

### LINKED ISSUE #2240: Fossology depends on heirloom-mailx package for email notifications, but it is not available any more in Debian
### Description

To enable email notification, the wiki says that package heilroom-mailx should be installed: https://github.com/fossology/fossology/wiki/Email-notification-configuration#setting-up-the-email-client

However, such package is not available any more in recent Debian versions (10+).

I managed to manually install it:

```bash
wget http://ftp.us.debian.org/debian/pool/main/s/s-nail/heirloom-mailx_14.8.16-1_all.deb
apt install ./heirloom-mailx_14.8.16-1_all.deb
ln -s /usr/bin/heirloom-mailx /usr/bin/mailx
```

but it does not seem an optimal solution. 

#### How to reproduce

in Debian 10+:

```bash
sudo apt-get install heirloom-mailx  # as suggested in the wiki
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package heirloom-mailx is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'heirloom-mailx' has no installation candidate
```

### Versions

* Last commit id on master: all
* Operating System (lsb_release -a): Debian 10+

ChinmayShringi:
## Summary
- Migrates email sending from s-nail to swaks (Swiss Army Knife for SMTP)
- Rebases and completes the work from PR #2194

## Why swaks?
s-nail has had compatibility issues across versions due to its URL-based SMTP authentication scheme. swaks uses separate command-line flags for each parameter (`--server`, `--port`, `--auth-user`, etc.), making it more robust and easier to configure.

## Changes
| File | Change |
|------|--------|
| `src/scheduler/agent/database.c` | swaks command building, async spawn with GLib |
| `src/scheduler/agent/emailformatter.c` | `mail_process_exit_callback()` for async handling |
| `src/scheduler/agent/emailformatter.h` | callback declaration |
| `src/lib/php/common-sysconfig.php` | SSL verify options simplified to Yes/No |
| `debian/control` | `s-nail` → `swaks` dependency |
| `install/defconf/fossology.conf.in` | `/usr/bin/swaks` as default client |
| `utils/fo-installdeps` | swaks packages for all distros |
| `.gitpod.Dockerfile` | swaks for dev environment |

## Test Plan
- [ ] Build succeeds
- [ ] Email notification works with SMTP configuration
- [ ] CI pipeline passes

Supersedes #2194
Closes #2240

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -40,11 +40,12 @@
   error = NULL; }
 
 #define EMAIL_BUILD_CMD "%s %s --header 'Subject: %s' --to '%s'" \
-    " --body - --silent --auth-hide-password" ///< Email command format
+    " --silent --auth-hide-password" \
+    " --output '&STDOUT' --body -" ///< Email command format
 #define DEFAULT_HEADER  "FOSSology scan complete\nmessage:\"" ///< Default email header
 #define DEFAULT_FOOTER  "\""                  ///< Default email footer
 #define DEFAULT_SUBJECT "FOSSology scan complete\n" ///< Default email subject
-#define DEFAULT_COMMAND "/usr/bin/mailx"      ///< Default email command to use
+#define DEFAULT_COMMAND "/usr/bin/swaks"      ///< Default email command to use
 
 #define min(x, y) (x < y ? x : y)     ///< Return the minimum of x, y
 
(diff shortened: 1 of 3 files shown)


.gitpod.Dockerfile
```diff
@@ -18,10 +18,11 @@ RUN install-packages --option Dpkg::Options::="--force-confold" \
     libdbd-sqlite3-perl libdistro-info-perl libgcrypt20-dev \
     libicu70 libicu-dev libjson-c-dev libjsoncpp-dev \
     liblocal-lib-perl libmagic-dev libmxml-dev libpcre3-dev libpq-dev \
+<<<<<<< HEAD
```
Kaushl2208: I think you forgot the Merge Conflict entry

Kaushl2208:
Hi @ChinmayShringi , After reviewing the swaks documentation more closely, I realized that it does not function as an MTA. It is primarily an SMTP testing and troubleshooting tool rather than a mail transfer agent. Because of this, it does not appear to be a like-for-like replacement for s-nail in our current setup.

Given this mismatch in responsibilities, I don’t think this approach aligns with the intended behavior of the existing implementation. I suggest we continue the discussion in the issue to identify a more appropriate replacement that fits our requirements and limitations.

For now, I will close this PR (and the related one: #2194 ), and we can revisit once we have clarity on the right direction. Lets have the discussions on replacements on the issue. Then we can put effort towards the decision.

Thanks again for the effort and initiative.

PR #3329, opened 2026-02-11
fix(scheduler): guard against NULL agent owner to prevent scheduler crash

helmut-leher-magna:
Changes aims to prevent scheduler crashes caused by dereferencing a NULL agent->owner pointer. The changes make the agent logging macros null-safe and add defensive checks in critical code paths to handle potential NULL pointers gracefully.

 - Prevents dereferencing a NULL owner, which could otherwise lead to a SIGSEGV (segmentation fault) and crash the scheduler.
 - Protects against crashes and resource leaks when the agent invariant (missing owner) is violated, while ensuring that the agent lifecycle is still managed correctly by the scheduler.

**1) Null-Safety in Agent Logging Macros:**
The `AGENT_CREDENTIAL` and `AGENT_LOG_CREDENTIAL` macros have been modified to safely handle a NULL `agent->owner` by checking it before dereferencing. Additionally, a NULL check for `agent->type` is added to avoid dereferencing a NULL type.

**2) Agent Management and Memory Cleanup:**
In the `agent_spawn()` function, an early return is introduced if `agent->owner == NULL`. In this case:

- Open `FILE*` streams and file descriptors (FDs) are safely closed,
- The agent is marked as `AG_FAILED`,
- Any passed arguments (if any) are freed,
- NULL is returned to prevent memory and resource leaks.

**3) NULL Owner Check in update() Function:**
A NULL check for `agent->owner`'is added in the `update()` function to kill agents with NULL owners before attempting to use them, preventing the dereferencing of an invalid pointer.

 
Extends the solution/fix from PR: https://github.com/fossology/fossology/pull/3195

### DIFF
--- a/src/scheduler/agent/agent.c
+++ b/src/scheduler/agent/agent.c
@@ -57,15 +57,26 @@
 #define TEST_NULL(a, ret) if(!a) { \
   errno = EINVAL; ERROR("agent passed is NULL, cannot proceed"); return ret; }
 
-/** Prints the credential of the agent */
-#define AGENT_CREDENTIAL                                                 \
-  log_printf("JOB[%d].%s[%d.%s]: ", agent->owner->id, agent->type->name, \
-      agent->pid, agent->host->name)
+/** Prints the credential of the agent (null-safe) */
+#define AGENT_CREDENTIAL do {                               \
+  if(agent && agent->owner)                                 \
+    log_printf("JOB[%d].%s[%d.%s]: ", agent->owner->id,     \
+        agent->type ? agent->type->name : "?",              \
+        agent->pid, agent->host ? agent->host->name : "?"); \
+  else                                                      \
+    log_printf("AGENT[%s][pid=%d][host=%s]: ",              \
+        agent && agent->type ? agent->type->name : "?",     \
+        agent ? agent->pid : -1,                            \
+        agent && agent->host ? agent->host->name : "?");    \
+} while(0)
 
-/** Prints the credential to the agent log */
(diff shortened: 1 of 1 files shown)


helmut-leher-magna:
Sorry for the extra effort, commit signed-off corrected now.

its-sushant:
Hi @helmut-leher-magna Thank you for the PR. The changes look good to me. May be I will add more comments/reviews on the PR soon, but before that I would like to know were you able to reproduce the bug in your local? If yes, Can you share how?

helmut-leher-magna:
Hi @its-sushant 

we haven't actually been able to reproduce the error on our end yet. However, we're experiencing a problem where the scheduler frequently "hangs" and can no longer distribute new jobs from the queue to the agents, or the agents are unable to process new ones. > Only a complete restart of the scheduler helps, which is annoying in production!

After various analyses, such as I/O load, database queries, and a deeper dive into the code to understand exactly how new jobs are started, we came across this issue. The linked, older bug report also seems to be valid.

Therefore, the fix is ​​essentially a shot in the dark as to whether it resolves the underlying problem. There are certainly other places in agent.c that don't catch this completely, but I didn't want to rewrite everything and focused on the frequently called functions. The fix has been in use in the production system for about two weeks, and so far our problem hasn't recurred. As mentioned, however, it could have a different cause since our environment uses a huge database.

At least it can theoretically happen, the code dereferences a NULL owner and triggers a segfault, the entire scheduler process will crash — stopping all agents and preventing any jobs from running until the scheduler restarts.

shaheemazmalmmd (approved):
Code looks good, Needs test

PR #3332, opened 2026-02-12
fix(explorer): correct directory status for 'To be discussed' files #1494

kjarir:
This PR fixes issue #1494 where directories containing only cleared and 'To be discussed' files were incorrectly marked as 'Open' (Red).

Changes:
Added OPT_ONLY_TBD constant to UploadTreeProxy
Updated AjaxExplorer.php to check if all uncleared files are 'To be discussed'.
Added testOptionOnlyTBD unit test in UploadTreeProxyTest.php.

Fixes: #1494


shaheemazmalmmd:
@kjarir : thanks for your efforts, You have pushed the changes which are irrelevant for this PR. also why not use the existing code from https://github.com/kjarir/fossology/blob/d1af2647978bdab6507b425e42923e04bb4de5a5/src/www/ui/async/AjaxExplorer.php#L471 to get to be discussed entries ?? 

closing the pr for now please reopen once you have addressed above concerns and do not forget to add the validation proofs for the FIx.

PR #3335, opened 2026-02-14
fix(clearing): avoid duplicate decisions in edit flow and global scope

### LINKED ISSUE #458: Edit in License Browser should not append equal events
Deciderjob appended the "Add selected licenses" list as events and create decisions. But it should not include events if the latest decision already contains an equal event (same license, same scope). Issue #453 was one symptom of this bug.

sakshammishra112:
## Description

This PR fixes the issue where duplicate clearing decisions were getting added when the same license was added from License Browser Edit.  
While working on this, I also noticed that when we set **"Clearing decision scope → Apply decision to all future occurrences of this file"** to true, the scope becomes global, and the clearing decision was getting added multiple times when we clicked the Submit button.

Closes #458

## How to reproduce

1. Conclude a license for a file.
2. Go to **License Browser → Edit** and add the same license for this file again.
3. Observe that duplicate clearing decisions are created.
4. Enable **Apply decision to all future occurrences of this file** and click Submit multiple times.
5. Observe duplicate global decisions in Clearing History.

## Changes

- Added validation in [change-license-processPost.php](https://github.com/fossology/fossology/blob/master/src/www/ui/change-license-processPost.php) to prevent duplicate entries from License Browser Edit.
- Added validation in [ClearingDao.php](https://github.com/fossology/fossology/blob/master/src/lib/php/Dao/ClearingDao.php) to prevent duplicate global clearing decisions.

### Before

https://github.com/user-attachments/assets/4a15e3e2-a990-4b25-be15-3b181e26fb83

image

### After

https://github.com/user-attachments/assets/4f9bcd8c-2827-43d2-a84e-76066b69fae0

image

### SQL query
```
SELECT
    cd.clearing_decision_pk,
    ut.ufile_name,
    lr.rf_shortname AS license_shortname,
    ce.removed,
    CASE 
        WHEN ce.removed = false THEN 'ADD'
        WHEN ce.removed = true THEN 'REMOVE'
    END AS event_action,
    cd.decision_type,
    cd.scope,
    cd.date_added
FROM clearing_decision cd
JOIN uploadtree ut
    ON cd.uploadtree_fk = ut.uploadtree_pk
JOIN clearing_decision_event cde
    ON cd.clearing_decision_pk = cde.clearing_decision_fk
JOIN clearing_event ce
    ON cde.clearing_event_fk = ce.clearing_event_pk
JOIN license_ref lr
    ON ce.rf_fk = lr.rf_pk
WHERE ut.ufile_name = 'idn2.c'
ORDER BY cd.date_added DESC;
```

@shaheemazmalmmd

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -335,8 +335,10 @@ class ClearingDao
    * @param int $groupId
    * @param int $decType
    * @param int $scope
+   * @param array $eventIds
+   * @param bool $checkDuplicates
    */
-  public function createDecisionFromEvents($uploadTreeId, $userId, $groupId, $decType, $scope, $eventIds)
+  public function createDecisionFromEvents($uploadTreeId, $userId, $groupId, $decType, $scope, $eventIds, $checkDuplicates = true)
   {
     if ( ($scope == DecisionScopes::REPO) &&
          !empty($this->getCandidateLicenseCountForCurrentDecisions($uploadTreeId))) {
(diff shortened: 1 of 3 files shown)


shaheemazmalmmd:
@sakshammishra112 L Thank you for the contributions please check the failing test cases.

sakshammishra112:
@shaheemazmalmmd, any update on this?

PR #3340, opened 2026-02-17
Refactor: Automate missing agent execution in DecisionImporter

### LINKED ISSUE #3337: [FIXME] Automate agent execution during Decision Import to improve workflow
**Description**
While auditing the DecisionImporterDataCreator.php file, I identified a FIXME at line 297 that currently limits the automation of the decision import process.

**Current Problem**
The createCopyrightData method (and similar methods for ECC/IPRA) throws an UnexpectedValueException if the respective agent's ARS table does not exist on the server. This results in a hard stop for the user, requiring them to manually trigger a scan for that agent before they can proceed with importing decisions.

**Proposed Solution**
To create a more seamless and autonomous workflow, I propose refactoring this check to automatically schedule the missing agent scan.

Instead of throwing an exception, the system should:

1.Verify if the agent scan is missing.

2.Automatically invoke the createCxJobs() method to schedule the required scan in the background.

3.Allow the import process to continue or provide a status update that the dependency is being handled automatically.

**Technical Impact**
Automation: Reduces the need for manual user intervention during data migration/import.

Stability: Replaces a "blocking" error with a proactive background task.

I am familiar with the agent orchestration and DAO layers from my previous contributions to ClearingDao.php (#3284) and the SQLite driver (#3260). I would like to implement this fix.

@shaheemazmalmmd @Kaushl2208 @its-sushant @GMishx

Sumamasonia:
**Description**
This PR addresses the FIXME identified in DecisionImporterDataCreator.php. Previously, the importer would throw an exception if a required agent scan (Copyright/ECC/IPRA) was missing, halting the process.

**Changes**
Automation: Replaced the blocking exception with an automated background scan trigger using createCxJobs.

Logic Optimization: Implemented if-else logic to ensure the scan is only triggered once if missing, or simply fetched if already present.

Cleaner History: This is a fresh branch containing only the relevant changes for this issue to ensure a clean commit history.

**Impact**
Improves the autonomy of the decision import workflow and removes manual user intervention.

**Fixes #3337**

### DIFF
--- a/src/decisionimporter/agent/DecisionImporterDataCreator.php
+++ b/src/decisionimporter/agent/DecisionImporterDataCreator.php
@@ -293,15 +293,9 @@ class DecisionImporterDataCreator
       return;
     }
 
-    if (!$this->agentDao->arsTableExists($agentName)) {
-      // Automating the agent trigger to remove manual burden
-      $latestAgentId = $this->agentDao->getCurrentAgentId($agentName);
-      $this->createCxJobs($agentName, $jobId, $latestAgentId);
-    } else {
-      // Ensure the job is scheduled if table exists
-      $latestAgentId = $this->agentDao->getCurrentAgentId($agentName);
-      $this->createCxJobs($agentName, $jobId, $latestAgentId);
-    }
+    // Automation trigger without redundant checks
+    $latestAgentId = $this->agentDao->getCurrentAgentId($agentName);
+    $this->createCxJobs($agentName, $jobId, $latestAgentId);
 
     $cxExistSql = "SELECT " . $agentName . "_pk FROM $agentName WHERE pfile_fk = $1 AND agent_fk = $2 AND hash = $3;";
     $cxExistStatement = __METHOD__ . ".$agentName" . "Exist";


Meetjain1 (commented):
@Sumamasonia Good change overall. 

Just one small observation the if and else blocks are doing the exact same thing (calling getCurrentAgentId and createCxJobs), so the arsTableExists check becomes kind of redundant.

cc @shaheemazmalmmd 

Image

Sumamasonia:
Hi @Meetjain1, thank you for the observation! I have refactored the code to remove the redundant if-else block and streamlined the agent automation logic as suggested.

shaheemazmalmmd:
@Sumamasonia Latest changes here doesn't add any value. Please re-open with the proper fix.

image

Sumamasonia:
Hi @shaheemazmalmmd, I understand. I simplified the logic based on previous feedback, but I see how it might have lost the core value of the automation trigger. I am re-evaluating the FIXME requirements and will re-open with a more robust fix that properly handles the missing agent execution. Thank you for the guidance!

PR #3345, opened 2026-02-18
fix(ui): upload file page facing race condition

Kaushl2208:
## Description

Race condition encountered in `Upload >> From File` page, Where multiple ajax triggers with dependency on setTimeout which eventually results in UI freeze and breaking. Under load when a single request `getUploads` can take more time, Blocking code can be deadly. 

## Changes

- `upload_file.html.twig` : was facing race condition on duplicate `#fileUploader` on change with a setTimeout()
-  `agent_reuser.js.twig`: Osselot based javascript functions were utilising a lot of setTimeout(), which deems unecessary, I have inspected it,  one of the handlers can be registered immediately, Similarly tried removing the setTimeout.

## How to test

1. Upload multiple uploads in a folder.
2. Try and upload a new component (Also a duplicate --> to check modal is working)
3. The UI should not break, Also check the reuse modal, everything rendering properly. (Check Local and Osselot Reuse working)
4. Try with multiple files as well.

CC: @GMishx @shaheemazmalmmd

### DIFF
--- a/src/lib/php/UI/Component/Menu.php
+++ b/src/lib/php/UI/Component/Menu.php
@@ -232,11 +232,10 @@ class Menu
     $vars['title'] = empty($title) ? _("Welcome to FOSSology") : $title;
     if ($hide_banner) {
       $vars['bannerMsg'] = "";
-      $vars['systemLoad'] = "";
     } else {
       $vars['bannerMsg'] = @$sysConfig['BannerMsg'];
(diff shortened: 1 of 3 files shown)


shaheemazmalmmd:
@Kaushl2208 : I have tried testing the reuse after this changes. But was not successful.

Here is what i did.

* Upload fck editor. 
* Decide all the files using bulk.
* Now upload the same version agein and select the reuse option.
* No file has been cleared/decided in the reused package

image

PR #3346, opened 2026-02-18
fix(pagination): navigate to first page when out of range

shaheemazmalmmd:
## Description

Always navigate to first page when page number is out of range in tree view.

## How to test

* Upload a component.
* Navigate to tree-view and to a folder where there are over 100 files.
* Click on page number and change it to last/ any.
* Now in sub menu click any top folder where file count is lesser.
* Tool should auto select page one.


Kaushl2208 (approved):
LGTM, Tested. Working as expected.

PR #3347, opened 2026-02-18
Improve folder access check performance using CTE-based lookup

its-sushant:
## Description
Browsing folders using the folder query parameter is slow on systems with a large number of folders.
The delay is caused by `FolderDao::isFolderAccessible()` recursively traversing the entire folder tree in `GetFolderArray()`, which resulted in a large number of database queries per request.

### Changes
Refactored `isFolderAccessible()` to use the existing CTE-based folder tree lookup via `isInFolderTree()`.


shaheemazmalmmd (approved):

PR #3354, opened 2026-02-21
fix(cyclonedx): use HTTPS schema URL for BOM 1.4

### LINKED ISSUE #3352: CycloneDX export uses HTTP schema URL instead of HTTPS
The generated CycloneDX BOM currently contains:
```
"$schema": "http://cyclonedx.org/schema/bom-1.4.schema.json"
```

However, the official CycloneDX schema URL is served over HTTPS:
```
"$schema": "https://cyclonedx.org/schema/bom-1.4.schema.json"
```

Updating the schema URL to use HTTPS would align the exported BOM with the official schema endpoint, I can submit a small patch to update the schema URL accordingly.

abdollahShahid:
Updates the CycloneDX $schema URL from HTTP to HTTPS to align with the official schema endpoint. This is a non-breaking change affecting only the exported BOM metadata.

Closes #3352

### DIFF
--- a/src/cyclonedx/agent/reportgenerator.php
+++ b/src/cyclonedx/agent/reportgenerator.php
@@ -59,7 +59,7 @@ class BomReportGenerator
   {
     return [
       'bomFormat' => 'CycloneDX',
-      '$schema' => 'http://cyclonedx.org/schema/bom-1.4.schema.json',
+      '$schema' => 'https://cyclonedx.org/schema/bom-1.4.schema.json',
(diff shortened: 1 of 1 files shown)


Sandipmandal25:
This change causes the BOM to fail validation against its own declared 1.4 schema, since the CycloneDX 1.4 schema enum only permits the `http://` form. I’ve opened #3409 to revert it. Could you please take a look? Thanks!!

ref issue #3352 

cc @its-sushant

PR #3357, opened 2026-02-23
fix(ui): update dependent job status when parent job is canceled

### LINKED ISSUE #2113: Cancel the direct and indirect(transitive) dependencies jobs automatically when the master job is cancelled
### Description

The dependencies jobs(both direct and indirect) remain active/queued state even after the master job (wget) gets killed/cancelled in the UI

#### How to reproduce

1. Trigger an upload via api
2. cancel the wget in the UI

The dependencies job remains active in the UI. 

Ideally when master job is cancelled all its dependencies also should be cancelled.

#### Screenshots

image
### Versions

* Last commit id on master: Version: [3.11.0], Branch: [HEAD], Commit: [#c93921] 2021/07/30 07:40 UTC built @ 2021/10/02 14:57 UTC
* Operating System (lsb_release -a): Ubuntu 20.04.2 LTS

sakshammishra112:
## Description

When scan jobs are scheduled and a parent job is cancelled by the user, the dependent
child jobs remain in the queue and continue to appear as **Pending** in the UI.

This PR fixes the issue by:

- Identifying all direct and transitive dependencies of the cancelled job
- Cancelling those dependent jobs recursively

This ensures the UI reflects the correct state and no cancelled jobs remain in
a queued/pending state.

Closes: #2113

## Changes

1. Added a new method in [JobDao.php](https://github.com/fossology/fossology/blob/master/src/lib/php/Dao/JobDao.php) to retrieve all dependencies for a given job ID.
2. Updated [showjobs.php](https://github.com/fossology/fossology/blob/master/src/www/ui/showjobs.php) to:
   - Detect when a job is cancelled
   - Recursively cancel all dependent jobs.

## How to Test

1. Schedule scan agents for an upload.
2. Cancel a job that has dependent jobs.
3. Verify that:
   - All dependent jobs are also marked as **Cancelled**
   - No dependent job remains in **Pending** state in the UI.

### Before
Screenshot 2026-02-22 041926
Screenshot 2026-02-22 041938
Screenshot 2026-02-22 043239

### After

https://github.com/user-attachments/assets/4da89ffb-44c2-4c77-bde8-eefd76568993

### Dependencies

Screenshot 2026-02-22 044446

### DIFF
--- a/src/lib/php/Dao/JobDao.php
+++ b/src/lib/php/Dao/JobDao.php
@@ -88,4 +88,37 @@ class JobDao
 
     return $result;
   }
+
+  public function getJobIdFromJobQueue($jqPk)
+  {
+    $stmt = __METHOD__;
+    $row = $this->dbManager->getSingleRow(
+      "SELECT jq_job_fk FROM jobqueue WHERE jq_pk = $1",
+      array($jqPk),
+      $stmt
+    );
+    return $row ? $row['jq_job_fk'] : null;
+  }
+
+  public function getJobDependencies($jobId)
+  {
+    $stmt = __METHOD__;
+    $this->dbManager->prepare($stmt,
+        "SELECT DISTINCT child.jq_pk AS child_id,
+                child.jq_type AS child_agent,
+                child.jq_end_bits AS child_status
+         FROM jobdepends jd
+         JOIN jobqueue child ON child.jq_pk = jd.jdep_jq_fk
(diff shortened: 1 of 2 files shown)


sakshammishra112:
@shaheemazmalmmd, any update on this?

shaheemazmalmmd:
@sakshammishra112 : thank you for all your efforts. it will be better to handle this in scheduler(https://github.com/fossology/fossology/pull/3760) instead of showjobs.

PR #3360, opened 2026-02-24
fix(clearing): preserve clearing decision when adding comment

### LINKED ISSUE #2059: Adding a comment resets the clearing status

sakshammishra112:
## Description

Adding or editing comments, acknowledgements, or license text on an already confirmed clearing decision was incorrectly resetting the decision status from **confirmed (green)** to **WIP (red)**.

This happened because [updateClearingEvent()](https://github.com/fossology/fossology/blob/2b78463c0fe30f0f9e6020d9a441afc9d0fc6548/src/lib/php/Dao/ClearingDao.php#L451) created a **new clearing event** even when an event for the same decision already existed. Creating a new event triggered [markDecisionAsWip()](https://github.com/fossology/fossology/blob/2b78463c0fe30f0f9e6020d9a441afc9d0fc6548/src/lib/php/Dao/ClearingDao.php#L603), which invalidated the previously confirmed clearing decision. As a result, users were forced to resubmit the decision and an unnecessary extra entry was added to the clearing history.

The fix updates the existing clearing event in place when only metadata is modified, so the clearing decision remains confirmed and no redundant history entries are created.

Closes: #2059 

## Changes

- Modify  [updateClearingEvent()](https://github.com/fossology/fossology/blob/2b78463c0fe30f0f9e6020d9a441afc9d0fc6548/src/lib/php/Dao/ClearingDao.php#L451) to **update existing clearing events** instead of inserting new ones for metadata-only changes.
- This removes creation of duplicate clearing history entries caused by metadata updates.

### Before  

https://github.com/user-attachments/assets/9c7cf310-df1d-4aef-a13d-1aa5bc180f8b

image

image

### After

https://github.com/user-attachments/assets/a64beeb6-0e49-4062-950d-de60e4670991

image

image

@shaheemazmalmmd @Kaushl2208

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -457,29 +457,34 @@ INSERT INTO clearing_decision (
     $params = array($uploadTreeId, $licenseId, $groupId);
     $row = $this->dbManager->getSingleRow($statementGetOldata, $params, $statementName);
 
+    $changeTo = StringOperation::replaceUnicodeControlChar($changeTo, false);
+
     if (!$row) {  //The license was not added as user decision yet -> we promote it here
       $type = ClearingEventTypes::USER;
       $row['type_fk'] = $type;
       $row['comment'] = "";
       $row['reportinfo'] = "";
       $row['acknowledgement'] = "";
-    }
-
-    $changeTo = StringOperation::replaceUnicodeControlChar($changeTo, false);
-    if ($what == 'reportinfo') {
-      $reportInfo = $changeTo;
-      $comment = $row['comment'];
-      $acknowledgement = $row['acknowledgement'];
-    } elseif ($what == 'comment') {
-      $reportInfo = $row['reportinfo'];
-      $comment = $changeTo;
-      $acknowledgement = $row['acknowledgement'];
+      if ($what == 'reportinfo') {
+        $row['reportinfo'] = $changeTo;
+      } elseif ($what == 'comment') {
+        $row['comment'] = $changeTo;
+      } else {
+        $row['acknowledgement'] = $changeTo;
+      }
+      $this->insertClearingEvent($uploadTreeId, $userId, $groupId, $licenseId, false, $row['type_fk'], $row['reportinfo'], $row['comment'], $row['acknowledgement']);
     } else {
-      $reportInfo = $row['reportinfo'];
-      $comment = $row['comment'];
-      $acknowledgement = $changeTo;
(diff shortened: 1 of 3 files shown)


sakshammishra112:
@shaheemazmalmmd any updates on this?

src/lib/php/Dao/ClearingDao.php
```diff
@@ -463,23 +463,28 @@ public function updateClearingEvent($uploadTreeId, $userId, $groupId, $licenseId
...
+        $row['comment'] = $changeTo;
+      } else {
+        $row['acknowledgement'] = $changeTo;
+      }
+      $this->insertClearingEvent($uploadTreeId, $userId, $groupId, $licenseId, false, $row['type_fk'], $row['reportinfo'], $row['comment'], $row['acknowledgement']);
     } else {
-      $reportInfo = $row['reportinfo'];
-      $comment = $row['comment'];
-      $acknowledgement = $changeTo;
+      $changeTo = StringOperation::replaceUnicodeControlChar($changeTo, false);
```
shaheemazmalmmd: This can be moved before if statement

shaheemazmalmmd (changes requested):

src/lib/php/Dao/ClearingDao.php
sakshammishra112: Done

shaheemazmalmmd:
@sakshammishra112 : just tested this branch, even through the editing/adding a comment/ACK/TeXT removes WIP in tree view the file view have WIP state. which is confusing, it shall be removed on both the ends.

sakshammishra112:
@shaheemazmalmmd Thanks for testing. I've fixed the WIP indicator issue in file view. The problem was that when adding comments, the `doOnSuccess` function was adding the `border-danger` class(indicating WIP) without checking the actual WIP status of the file.

shaheemazmalmmd:
Closing thus as discussed.

PR #3367, opened 2026-02-24
refactor(ui): use UploadDao in ajax-schedule-agent instead of raw SQL

SooryanshSingh:
## Summary

Refactor `ajax-schedule-agent.php` to use `UploadDao::getUpload()` for retrieving upload information instead of executing a direct SQL query.

This aligns the file with the DAO-based data access pattern used across the codebase and removes inline database access from the UI layer.

## Testing

- Manually verified agent scheduling through the UI
- Confirmed job creation appears correctly in **Show Jobs**


SooryanshSingh:
Thank you for the review, sir.

Kaushl2208:
Tested, Working as expected.

PR #3371, opened 2026-02-26
fix(scheduler): prevent crash when changing non-existent job priority

### LINKED ISSUE #3348: Scheduler crashes whenever the priority of individual jobs is modified
The scheduler crashes whenever the priority of an individual job is modified. This issue occurs consistently when attempting to adjust the priority settings of a job, causing the system to become unresponsive and terminate the scheduling process unexpectedly.

Steps to Reproduce:
- Open the scheduler administration (?mod=admin_scheduler).
- Select an operation: Priority
- Select a job
- Select a priority level: (range -20 to 20)
- Submit

Result: The scheduler crashes immediately after the priority change.

**Expected Behavior:**
The scheduler should update the priority of the selected job without crashing, maintaining normal operation and ensuring that the system can continue to manage and execute jobs as expected.

**Actual Behavior:**
The scheduler crashes upon changing the priority of a job, leading to an application failure and loss of functionality until the system is restarted. It is not relevant if you choose positive or negative priority (niceness).

___

Image

Ndevu12:
## Description

Fix scheduler crash that occurs when attempting to change the priority of a job 
that has already completed or been removed from the scheduler. This was caused by 
a NULL pointer dereference in the `job_priority_event() `function when the scheduler 
attempted to look up a job that no longer existed in the job queue.

## Changes

- Modified `src/scheduler/agent/job.c: job_priority_event()`
  * Added NULL pointer validation for job parameter
  * Early return with appropriate warning log when job is not found
  * Prevents NULL pointer dereference that caused scheduler crash
  * Follows established error handling pattern used in `job_pause_event() `
    and` job_restart_event()`

- Also, I added a unit test in `src/scheduler/agent_tests/Unit/testJob.c`
  * New test function: `test_job_priority_event_null_job()`
  * Tests race condition where job completes before priority change is submitted
  * Verifies graceful handling without crash

## How to test
1. Manual testing in `admin_scheduler`:
   - Navigate to scheduler administration` (?mod=admin_scheduler)`
   - Select a job from the priority dropdown
   - Wait for the job to complete naturally or be removed from scheduler
   - Attempt to submit a priority change for that job
   - `Expected`: System logs warning but does not crash
   - `Previous behavior`: Scheduler would crash with segmentation fault

2. Run unit tests:
   - Build the scheduler with unit tests enabled
   - Run: make test
   - Verify `test_job_priority_event_null_job` passes

Closes #3348


Meetjain1 (commented):
@Ndevu12 Nice fix! The NULL check and early return in job_priority_event() make sense and should prevent the scheduler crash when the job no longer exists. The added unit test for the NULL job scenario is also a good addition.

Just one small nit: the warning log uses params->second as the job ID, but it seems to represent the priority value, so the message could be slightly misleading. Maybe the ID part can be removed or the wording adjusted.

Overall, the change looks good and aligns well with the existing event handling pattern.

cc @GMishx

Ndevu12:
> @Ndevu12 Nice fix! The NULL check and early return in job_priority_event() make sense and should prevent the scheduler crash when the job no longer exists. The added unit test for the NULL job scenario is also a good addition.
> 
> Just one small nit: the warning log uses params->second as the job ID, but it seems to represent the priority value, so the message could be slightly misleading. Maybe the ID part can be removed or the wording adjusted.
> 
> Overall, the change looks good and aligns well with the existing event handling pattern.
> 
> cc @GMishx

@Meetjain1, thanks very much for the feedback. I have adjusted the warning message.

Thanks.

Ndevu12:
Hi @helmut-leher-magna @shaheemazmalmmd, @Meetjain1  @Kaushl2208 @valens200, I would like to let you know that this PR is ready for review!

Thanks

shaheemazmalmmd:
PR also contains changes that doesnt belong to this fix, also PR https://github.com/fossology/fossology/pull/3612 shall solve this issue.

Ndevu12:
@shaheemazmalmmd Thanks for closing this PR! Seems like it had been compromised because it contains the changes that I never pushed!

PR #3377, opened 2026-02-27
docs(cyclonedx): clarify specVersion baseline and schema behavior

abdollahShahid:
This PR improves documentation clarity around the CycloneDX agent:

- Clarifies that current reports follow CycloneDX spec v1.4.
- Documents that the schema URL is derived from the specVersion used by the report generator.
- Adds notes about official CycloneDX schema and upgrade implications (e.g., 1.7).

No functional changes.


shaheemazmalmmd:
Changes not required/important.

PR #3366, opened 2026-02-24
fix(api): Remove debug echo and fix wrong parameter in CopyrightController

kjarir:
## Summary

Fixes two bugs in `CopyrightController.php` that affect all copyright-related DELETE, RESTORE, and UPDATE REST API endpoints.

Closes #3358

---

## Bug 1 — Debug `echo` left in production (`deleteFileCX`, line 937)

A leftover debug statement `echo $uploadTreeId;` was writing the raw upload tree ID directly into the HTTP response body **before** the JSON response, corrupting it for all copyright DELETE endpoints.

**Before (broken response):**
```
12345{"code":200,"message":"Successfully removed copyright.","type":"INFO"}
```

**After (correct response):**
```json
{"code":200,"message":"Successfully removed copyright.","type":"INFO"}
```

**Fix:** Removed the `echo $uploadTreeId;` line.

---

## Bug 2 — Wrong parameter passed to `getUploadTreeTableName()` (lines 942, 980, 1021)

Three methods (`deleteFileCX`, `restoreFileCx`, `updateFileCx`) were calling `getUploadTreeTableName()` with `$uploadTreeId` (item ID) instead of `$uploadPk` (upload ID). This function requires the **upload primary key** to determine the correct upload tree table.

The correct usage already exists in the same file in `getFileCX()` (line 876):
```php
// Correct — getFileCX (line 876)
$uploadTreeTableName = $this->restHelper->getUploadDao()->getuploadTreeTableName($uploadPk);

//  Was wrong in deleteFileCX / restoreFileCx / updateFileCx
$uploadTreeTableName = $uploadDao->getUploadTreeTableName($uploadTreeId); // wrong!
```

**Fix:** Changed `$uploadTreeId` → `$uploadPk` in all three affected methods.

---

## Files Changed

- `src/www/ui/api/Controllers/CopyrightController.php`
  - Removed debug `echo` (line 937)
  - Fixed `getUploadTreeTableName($uploadTreeId)` → `getUploadTreeTableName($uploadPk)` in `deleteFileCX()`
  - Fixed `getuploadTreeTableName($uploadTreeId)` → `getuploadTreeTableName($uploadPk)` in `restoreFileCx()`
  - Fixed `getuploadTreeTableName($uploadTreeId)` → `getuploadTreeTableName($uploadPk)` in `updateFileCx()`

---

## How to Test

1. Set up FOSSology with a scanned upload containing copyright entries
2. Call any copyright DELETE endpoint:
   ```
[...]

PR #3379, opened 2026-02-28
fix(proxy): resolve db view name overwrite and parameter reuse bugs in UploadTreeProxy

### LINKED ISSUE #3378: UploadTreeProxy: Broken DB View Name instantiation and query parameter reuse
## Description
There are two distinct logic errors in `UploadTreeProxy.php` which impact DB View name uniqueness and Postgres parameter reuse string mapping.

### Bug 1: View Name overwrites suffixes with a generic string
When an `UploadTreeProxy` is instantiated without a specific view name, it builds a query string suffix in `createUploadTreeViewQuery()` to guarantee that multiple queries (e.g. `_realParent` vs `_alreadyCleared`) do not collide on the same Postgres view/statement name.
However, in the constructor, `$dbViewName` is evaluated **before** `createUploadTreeViewQuery()` is invoked, using `isset($this->dbViewName) ?: ''`, which evaluates the boolean return of `isset()` (so `1`). Then, it calls `parent::__construct(..., $dbViewName)`, inadvertently passing the string `'UploadTreeView'` and overriding the uniquely generated suffix created inside `createUploadTreeViewQuery()`.

**Consequence**: If multiple instances omitted the 4th constructor argument (e.g., `ui-clearing-view.php`), they both resolve to the default view name `UploadTreeView`. This can cause Postgres prepared statement collision errors.

### Bug 2: SQL Parameter Reuse Fails
The proxy attempts to reuse parameters (for performance optimization in parameterized execution) via `addParamAndGetExpr($key, $value)`. The intent is for it to verify if a `$key` exists and reuse the `$n` index:
```php
if (array_key_exists($key, $this->params)) {
   return '$' . (1 + array_search($key, array_keys($this->params)));
}
```
However, the values were pushed using array append `$this->params[] = $value;`, stripping the string key entirely. `array_key_exists` subsequently always fails, duplicating the query parameters inside `$this->params` array instead of reusing existing indexed values.

## Proposed Fix
* Correct the assignment of `$dbViewName` in `__construct` to use the property after `createUploadTreeViewQuery()` completes.
* Switch `$this->params[] = $value` to `$this->params[$key] = $value`, so the array index remains preserved for validation. Modify `countMaskedNonArtifactChildren` appropriately so it also binds the associative array index correctly.

kjarir:
Closes #3378.


src/lib/php/Proxy/UploadTreeProxy.php
```diff
@@ -363,7 +363,7 @@ public function countMaskedNonArtifactChildren($parent)
     if (array_key_exists('uploadId', $params)) {
       $uploadExpr = '$'.(1+array_search('uploadId', array_keys($params)));
     } else {
-      $params[] = $this->uploadId;
+      $params['uploadId'] = $this->uploadId;
```
shaheemazmalmmd, line 366: Not sure about this change.. explain why is it needed

src/lib/php/Proxy/UploadTreeProxy.php
```diff
@@ -447,7 +447,7 @@ private function addParamAndGetExpr($key,$value)
       return '$' . (1 + array_search($key, array_keys($this->params)));
     }
 
-    $this->params[] = $value;
+    $this->params[$key] = $value;
```
shaheemazmalmmd, line 450: same .. explain why is it needed

src/lib/php/Proxy/UploadTreeProxy.php
kjarir: The original code used $params[], which in PHP strips string keys and assigns a numeric index instead. This caused array_key_exists('uploadId', $params) on line 363 to always fail. By explicitly using the 'uploadId' key, we allow the class to correctly find and reuse the parameter, preventing it from being added to the SQL statement multiple times.

src/lib/php/Proxy/UploadTreeProxy.php
kjarir: This ensures the $key identifier is maintained in the $params array. Since the class relies on array_key_exists to deduplicate parameters, this fix prevents PHP from appending a generic numeric index, which would otherwise break the logic and lead to duplicate parameters (e.g., adding the same value as both $1 and $3) in the final query.

kjarir:
@shaheemazmalmmd This PR optimizes database performance by fixing broken parameter deduplication and prevents fatal system crashes by ensuring unique naming for Postgres views during concurrent operations.

shaheemazmalmmd:
This PR also contains the fixes from https://github.com/fossology/fossology/pull/3366/changes

@kjarir : add the correct changes.

PR #3380, opened 2026-02-28
fix(scheduler): portable mailx version check for MTA support

### LINKED ISSUE #3374: Fix email notification setup docs for modern Debian/Ubuntu compatibility
The [Email notification configuration](https://github.com/fossology/fossology/wiki/Email-notification-configuration) wiki page references `heirloom-mailx` as the mail client used by FOSSology for sending SMTP-based job completion notifications. This package is EOL on modern Debian/Ubuntu systems (include WSL), making the documented setup non-functional out of the box for new installations.

**Suggested Fix:**

- Update the wiki to mention `bsd-mailx` or `mailutils` as an alternatives with corresponding install commands.
- Audit any installation scripts (e.g. install.sh or CI configs) that may reference `heirloom-mailx` and update accordingly.
- Note any behavioral difference in SMTP configuration syntax b/w `heirloom-mailx` and the alternatives.

kjarir:
Currently, the FOSSology scheduler uses a distribution-specific  command to check for MTA support in the mail client. This approach is brittle as it:
1. Fails on non-Debian distributions (CentOS, Fedora, etc.).
2. Fails in environments where  might be installed but not managed by  (e.g., custom Docker builds, manual installs).
3. Causes SMTP job completion notifications to fail silently on modern systems where  is the required successor to the deprecated .

This PR replaces the  check with a direct call to , which is a more portable way to identify  and its version.

## Changes
- Modified :
    - Replaced  with mailx: illegal option -- V
Usage: mailx [-dEiInv] [-s subject] [-c cc-addr] [-b bcc-addr] [-F] to-addr ...
       mailx [-dEHiInNv] [-F] -f [name]
       mailx [-dEHiInNv] [-F] [-u user]
       mailx [-d] -e [-f name]
       mailx -H.
    - Updated the parsing logic to extract the version number from the command output.
    - Added basic error handling for empty version strings.

Closes #3374.

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -1036,12 +1036,17 @@ void database_job_priority(scheduler_t* scheduler, job_t* job, int priority)
  *
  * \return 1 if mta is supported, 0 if not
  */
+/**
+ * @brief Find s-nail version to check if mta is supported 
+ *
+ * \return 1 if mta is supported, 0 if not
+ */
 int check_mta_support()
 {
-  char cmd[] = "dpkg -s s-nail | grep -i 'Version' | awk '{print$2}' | cut -c -4";
-  char version_str[128];
+  char cmd[] = "mailx -V 2>&1 | grep -i 'v' | awk '{print$2}' | cut -c -4";
+  char version_str[128] = {0};
   char buf[128];
-  float version_float;
+  float version_float = 0.0;
   FILE *fp;
 
   fp = popen(cmd, "r");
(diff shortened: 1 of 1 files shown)


src/scheduler/agent/database.c
```diff
@@ -1031,17 +1031,22 @@ void database_job_priority(scheduler_t* scheduler, job_t* job, int priority)
...
 /**
  * @brief Find s-nail version to check if mta is supported 
  *
  * \return 1 if mta is supported, 0 if not
  */
 int check_mta_support()
 {
-  char cmd[] = "dpkg -s s-nail | grep -i 'Version' | awk '{print$2}' | cut -c -4";
-  char version_str[128];
+  char cmd[] = "mailx -V 2>&1 | grep -i 'v' | awk '{print$2}' | cut -c -4";
```
Kaushl2208, line 1046: Doesn't make sense.

Kaushl2208:
Closing, Not bringing any actual help.

PR #3382, opened 2026-02-28
fix(JobDao): return boolean instead of accessing undefined columns

### LINKED ISSUE #3381: Undefined array key access in JobDao::hasActionPermissionsOnJob() when querying job table
The method `JobDao::hasActionPermissionsOnJob()` queries the `job` table but processes the result as if it contains fields `jq_pk` and `end_bits`, which are not part of the selected result set.

This appears to be a copy-paste artifact from methods that query the `jobqueue` table (e.g., `getAllJobStatus()`), where those fields are valid.

#### Current Behavior

* The query uses:

  ```sql
  SELECT * FROM job
  LEFT JOIN group_user_member gm ON gm.user_fk = job_user_fk
  ```
* Inside the loop, it accesses:

  ```php
  $row['jq_pk']
  $row['end_bits']
  ```
* These columns do not exist in the `job` table or the current query result.
* The method builds and returns an array even though it is used as a boolean permission check in callers (UI/API).

This can lead to:

* Undefined array key warnings (in strict error reporting)
* Unnecessary iteration and memory usage
* Misleading return type (array instead of boolean for a permission check)

#### Expected Behavior

* The method should only check whether a matching row exists (permission check).
* It should not access undefined columns from the result set.
* It should return a boolean (`true` if permission exists, otherwise `false`).

Meetjain1:
## Description

Fix undefined array key access in `JobDao::hasActionPermissionsOnJob()` where the
method queries the `job` table but previously accessed `jq_pk` and `end_bits`
which are not part of the query result. The method is used as a permission check,
so it now performs an existence check and returns a boolean while preserving the
original authorization SQL semantics.

### Changes

- Replaced `SELECT *` with `SELECT 1` to avoid accessing undefined columns
- Removed array-building loop and unnecessary iteration
- Return explicit boolean (`$row !== false`) instead of array
- Added `LIMIT 1` for efficient existence check
- Added unit test for owner, group, and non-member permission scenarios
- Preserved original JOIN and WHERE logic (no behavior change)

## How to test

-  Run PHPUnit for DAO tests:
   ```bash
   vendor/bin/phpunit src/lib/php/Dao/test/JobDaoTest.php

Closes #3381

<img width="215" height="23" alt="image" src="https://github.com/user-attachments/assets/b80750e0-4604-44a2-8bcc-77f310f173ee" />


src/lib/php/Dao/JobDao.php
```diff
@@ -69,23 +69,21 @@ public function getChlidJobStatus($jobId)
...
 
     $res = $this->dbManager->execute($stmt, array($jobId, $userId, $groupId));
-    while ($row = $this->dbManager->fetchArray($res)) {
-      $result[$row['jq_pk']] = $row['end_bits'];
-    }
+    $row = $this->dbManager->fetchArray($res);
     $this->dbManager->freeResult($res);
 
-    return $result;
+    return $row !== false;
```
Kaushl2208, line 87: I am just wondering: You changed the return type of a function from array to bool. But did you made sure the core logic depending on this function expects boolean???

Also, the core philosophy behind this function is: "If the user is the job owner OR the job owner belongs to the given group, then return all job queue entries and their status bits." (Check and Return Data)

All i see is change in philosophy and not handling it everywhere. I understand the concept of optimizing queries, But this is not the right way to handle.

Meetjain1: I understand the concern now.
This change wrongly modified the return type and that can break existing logic. That was not my intention.
I have updated the fix to keep the original array return and function behavior, and only fixed the undefined column issue by properly joining jobqueue.
I will open a new PR with the minimal non-breaking fix.
Thankyou.

PR #3394, opened 2026-03-05
Fix(fossdash): Strip filenames before safe clean up

FrankOHara43:
## Description

The PR fixes the cleanup of files by removing trailing newline characters and uses safe subprocess execution instead of shell-based command invocation.

### Changes

- Fixes incorrect cleanup of old reported files caused by trailing newline characters
- avoids shell injection risk during file deletion.


shaheemazmalmmd:
@FrankOHara43 : check the failing test cases.

PR #3402, opened 2026-03-06
fix(schema-export): remove  and add current path fix libschema pg_con…

shaheemazmalmmd:
## Description

Update the core-schema.dat to match the output of schema-export tool.

### Changes

1. `src/www/ui/core-schema.dat`: Update the schema with `src/cli/schema-export`
2. Fix issues in schema-export.php
3. Fix issues in libschema.php 

## How to test

1. Dump the output of following SQL
```sql
SELECT
  table_name AS table, ordinal_position AS ordinal, column_name,
  udt_name AS type, character_maximum_length AS modifier,
  CASE is_nullable WHEN 'YES' THEN false WHEN 'NO' THEN true END AS notnull,
  column_default AS default,
  col_description(table_name::regclass, ordinal_position) AS description
FROM information_schema.columns
WHERE table_schema = 'public'
ORDER BY ordinal_position;
```
4. Install the branch and run `fo-postinstall`.
5. Dump the output of the SQL again.
    1. If no schema change was done recently, the ordinal number of the columns should not change.

### DIFF
--- a/src/lib/php/libschema.php
+++ b/src/lib/php/libschema.php
@@ -80,20 +80,12 @@ class fo_libschema
     if ($driver !== null && $driver instanceof Driver && $driver->isConnected()) {
       return;
     }
-
     global $PG_CONN;
     if (!empty($PG_CONN)) {
       $pgDriver = new Postgres($PG_CONN);
       $this->dbman->setDriver($pgDriver);
       return;
     }
-
-    if (!empty($GLOBALS['PG_CONN'])) {
-      $pgDriver = new Postgres($GLOBALS['PG_CONN']);
-      $this->dbman->setDriver($pgDriver);
-      return;
-    }
-
     throw new \Exception(
       "No database connection available: \$PG_CONN is not set and no driver " .
       "was injected into \$dbManager before calling this function."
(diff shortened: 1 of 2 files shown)


src/lib/php/libschema.php
```diff
@@ -554,11 +578,20 @@ function dropColumnsFromTable($columns, $table)
...
     $this->addInheritedRelations();
     $referencedSequencesInTableColumns = $this->addTables();
-    $this->addViews($viewowner = $SysConf['DBCONF']['user']);
+    if (!empty($SysConf['DBCONF']['user'])) {
+      $viewowner = $SysConf['DBCONF']['user'];
+    } elseif (!empty($PG_CONN)) {
+      $viewowner = pg_parameter_status($PG_CONN, 'session_authorization');
+    }
+    if (empty($viewowner)) {
+      $viewowner = 'fossology';
```
GMishx: If it is a user, should be fossy instead of fossology??

Kaushl2208: yes, We dont need the second fallback. line 589 --> picks the viewowner as the user that created the Database (which is fossy in our case)

its-sushant (approved):
Changes look good to me!

PR #3406, opened 2026-03-07
fix(email): prevent shell injection in email command construction

FrankOHara43:
## Description

The email subject and recipient were directly passed into `g_strdup_printf()` while constructing the email command, which could allow shell metacharacters to alter command execution.

### Changes

- Applied `g_shell_quote()` to `scheduler->email_subject`
- Applied `g_shell_quote()` to `user_email`
- Updated `EMAIL_BUILD_CMD` format string to use `pre-quoted arguments`
- Added `g_free()` calls for allocated quoted strings to avoid memory leaks

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -39,7 +39,7 @@
   email_notify = 0;        \
   error = NULL; }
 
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
@FrankOHara43 : the PR contains the changes of two fixes. Please take a look before you push.

PR #3409, opened 2026-03-07
fix(cyclonedx): revert  URL to http:// to fix schema validation

Sandipmandal25:
## Summary

Issue #3352  noted that the schema URL should correspond to the live endpoint. PR #3354  changed it from `http://` to `https://` for that reason. However, the CycloneDX 1.4 JSON schema defines an enum constraint that only allows:

`http://cyclonedx.org/schema/bom-1.4.schema.json` as the valid value for the `$schema` field.

Using `https://` causes the generated BOM to fail validation against its own declared schema, confirmed with cyclonedx-cli:

image
image

The schema field is a schema identifier, not a fetchable URL. Its value must therefore match the enum defined by the CycloneDX specification.

## Fix

Revert schema to `http://cyclonedx.org/schema/bom-1.4.schema.json` to comply with the CycloneDX 1.4 schema enum.

## Tests

Validated with cyclonedx-cli using a BOM generated by FOSSology.
image

## Ref

* Fixes regression introduced by #3354 
* Related issue: #3352 

cc @shaheemazmalmmd @JanAltenberg @its-sushant

### DIFF
--- a/src/cyclonedx/agent/reportgenerator.php
+++ b/src/cyclonedx/agent/reportgenerator.php
@@ -59,7 +59,7 @@ class BomReportGenerator
   {
     return [
       'bomFormat' => 'CycloneDX',
-      '$schema' => 'https://cyclonedx.org/schema/bom-1.4.schema.json',
+      '$schema' => 'http://cyclonedx.org/schema/bom-1.4.schema.json',
       'specVersion' => '1.4',
       'version' => 1.0,
       'serialNumber' => 'urn:uuid:'. uuid_create(UUID_TYPE_TIME),


abdollahShahid:
Thanks for catching that, my original change followed the live HTTPS endpoint, but for CycloneDX 1.4 the schema itself still pins the BOM `$schema` value to `http://cyclonedx.org/schema/bom-1.4.schema.json`.

I’m okay with reverting this for the current 1.4 output. As a follow-up, I’ll upgrade path to newer CycloneDX spec support in a clean, version-aware way - fits with the broader project idea of improving CycloneDX support and updating exports toward newer spec versions. I’m considering a validator-backed follow-up so we can keep the export spec-compliant and avoid this kind of drift going forward.

Sandipmandal25:
> Thanks for catching that, my original change followed the live HTTPS endpoint, but for CycloneDX 1.4 the schema itself still pins the BOM `$schema` value to `http://cyclonedx.org/schema/bom-1.4.schema.json`.
> 
> I’m okay with reverting this for the current 1.4 output. As a follow-up, I’ll upgrade path to newer CycloneDX spec support in a clean, version-aware way - fits with the broader project idea of improving CycloneDX support and updating exports toward newer spec versions. I’m considering a validator-backed follow-up so we can keep the export spec-compliant and avoid this kind of drift going forward.

 Thanks for the context!

PR #3413, opened 2026-03-07
refactor(lib): refactor code to get folder list is single query

its-sushant:
## Description

refactor(lib): refactor code to get folder list is single query

## Changes

Refactor `FolderListOption` function which is being used at mupliple places in the codebase
to get the folder tree. For each folder/subfolder function calls a sql query, making it very slow in case of large number of folders. 
Added Single recursive CTE to fetch the entire folder subtree in 1 query.

**Impact of the changes:** Significant speed improvement in `admin-->user`, `admin-->uploadPermission`, `organize-->folders`, `organize-->uploads` pages.

### DIFF
--- a/src/lib/php/common-folders.php
+++ b/src/lib/php/common-folders.php
@@ -177,75 +177,59 @@ function GetFolderFromItem($upload_pk="", $uploadtree_pk = "")
  */
 function FolderListOption($ParentFolder,$Depth, $IncludeTop=1, $SelectId=-1, $linkParent=false, $OldParent=0)
 {
+  global $container;
+
+  $dbManager = $container->get('db.manager');
+  $folderDao = $container->get('dao.folder');
+
   if ($ParentFolder == "-1") {
     $ParentFolder = FolderGetTop();
   }
   if (empty($ParentFolder)) {
     return;
   }
-  global $PG_CONN;
-  if (empty($PG_CONN)) {
-    return;
-  }
+
+  $ParentFolder = intval($ParentFolder);
+
+  $statementName = __METHOD__ . '.folderTree';
+  $sql = $folderDao->getFolderTreeCte($ParentFolder) . "
+    SELECT folder_pk, parent_fk, folder_name, depth
+    FROM folder_tree
+    ORDER BY name_path";
+
+  $dbManager->prepare($statementName, $sql);
+  $res = $dbManager->execute($statementName, array($ParentFolder));
+
   $V = "";
+  while ($row = $dbManager->fetchArray($res)) {
+    $folderDepth = intval($row['depth']);
+    $folderPk = $row['folder_pk'];
+    $parentFk = $row['parent_fk'];
 
-  if (($Depth != 0) || $IncludeTop) {
-    if ($ParentFolder == $SelectId) {
-      $V .= "<option value='$ParentFolder' SELECTED>";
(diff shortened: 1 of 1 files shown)


src/lib/php/common-folders.php
```diff
@@ -187,64 +187,64 @@ function FolderListOption($ParentFolder,$Depth, $IncludeTop=1, $SelectId=-1, $li
...
+        AND fc.foldercontents_mode = $mode
+      INNER JOIN folder_tree ft ON fc.parent_fk = ft.folder_pk
+      WHERE f.folder_pk IS NOT NULL
+  )
+  SELECT folder_pk, folder_name, depth, parent_fk
+  FROM folder_tree
+  ORDER BY sort_path";
+
+  $result = pg_query($PG_CONN, $sql);
+  DBCheckResult($result, $sql, __FILE__, __LINE__);
```
Kaushl2208: Hi @its-sushant , While we are on the process of refactoring. Can you please also look into the new method for accessing and implementing sql statements instead of `$PG_CONN` and other related functions to it.

```php
$dbManager = $container->get('db.manager');
```

We have this new method, Use this.

src/lib/php/common-folders.php
```diff
@@ -187,64 +187,64 @@ function FolderListOption($ParentFolder,$Depth, $IncludeTop=1, $SelectId=-1, $li
   if (empty($PG_CONN)) {
     return;
   }
+
+  $ParentFolder = intval($ParentFolder);
+  $mode = FolderDao::MODE_FOLDER;
+
+  // Single recursive CTE — fetches the entire folder subtree in 1 query
+  $sql = "WITH RECURSIVE folder_tree AS (
```
Kaushl2208: Can you also take a look at this function: `getFolderTreeCte()` and check if this is usable at all? Logically such operations are repeated over time and should be part of basic operations.

At one glance this function looks more similar to what you have done here. Here is the linke:
https://github.com/fossology/fossology/blob/b2b7474b1afcacf992a4b9f98f326907a709f6b4/src/lib/php/Dao/FolderDao.php#L136

src/lib/php/common-folders.php
```diff
@@ -187,64 +187,64 @@ function FolderListOption($ParentFolder,$Depth, $IncludeTop=1, $SelectId=-1, $li
...
+             f.folder_name::text AS sort_path
+      FROM folder f
+      WHERE f.folder_pk = $ParentFolder
+    UNION ALL
+      SELECT f.folder_pk, f.folder_name, ft.depth + 1,
+             fc.parent_fk,
+             ft.sort_path || '/' || f.folder_name
+      FROM folder f
+      INNER JOIN foldercontents fc ON fc.child_id = f.folder_pk
+        AND fc.foldercontents_mode = $mode
```
Kaushl2208: `$mode` is not a variable?
 You can just directly:

```suggestion
        AND fc.foldercontents_mode =" . self::MODE_FOLDER . "........."
```

its-sushant: Did you mean `FolderDao::MODE_FOLDER`?

src/lib/php/common-folders.php
its-sushant: Should be reusable. I will try to use this function.

its-sushant:
@Kaushl2208 I have made the suggested changes. PTAL!

PR #3414, opened 2026-03-08
fix(cli): prevent crashes in fo_nomos_license_list when scanner results are missing

atharva-nagane:
### Summary

`fo_nomos_license_list` assumes that scanner results always exist for the given upload. When scanner results are missing, the CLI throws PHP warnings or generates invalid SQL queries.

This patch adds defensive checks to handle these situations gracefully.

### Changes

* Added validation when retrieving **Nomos agent results**.
* Added validation when retrieving **Copyright agent results**.
* If scanner results are missing, the CLI now prints a user-friendly message instead of crashing.

### Example behaviour

Before:

```
PHP Warning: Undefined array key 0
Trying to access array offset on value of type null
```

After:

```
No nomos scan results available
```

### Testing

Tested locally using the FOSSology Docker setup:

* Ran `fo_nomos_license_list` with missing Nomos results.
* Ran `fo_nomos_license_list --type copyright` without copyright scanner data.

Both commands now fail gracefully without warnings or SQL errors.

### DIFF
--- a/src/cli/fo_nomos_license_list.php
+++ b/src/cli/fo_nomos_license_list.php
@@ -141,10 +141,12 @@ function GetLicenseList($uploadtree_pk, $upload_pk, $showContainer, $excluding, 
 
   /* get last nomos agent_pk that has data for this upload */
   $AgentRec = AgentARSList("nomos_ars", $upload_pk, 1);
-  if ($AgentRec === false) {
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
@atharva-nagane : We are re considering if the CLI is needed as the REST API is already in place.

PR #3416, opened 2026-03-08
fix(email): prevent shell injection in email command construction

FrankOHara43:
## Description

The email subject and recipient were directly passed into `g_strdup_printf()` while constructing the email command, which could allow shell metacharacters to alter command execution.

### Changes

- Applied `g_shell_quote()` to `scheduler->email_subject`
- Applied `g_shell_quote()` to `user_email`
- Updated `EMAIL_BUILD_CMD` format string to use pre-quoted arguments
- Added `g_free()` calls for allocated quoted strings to avoid memory leaks

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -39,7 +39,7 @@
   email_notify = 0;        \
   error = NULL; }
 
-#define EMAIL_BUILD_CMD "%s %s -s '%s' %s"    ///< Email command format
+#define EMAIL_BUILD_CMD "%s %s -s %s %s"    ///< Email command format
 #define DEFAULT_HEADER  "FOSSology scan complete\nmessage:\"" ///< Default email header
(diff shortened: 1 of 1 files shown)


FrankOHara43:
@Kaushl2208 I checked the docker test failure. It seemed that the image builds successfully and other agents run before this step, I think this failure was unrelated to the changes suggested by the PR. Can you please confirm the reason for failure or if a rerun would be appropriate?

shaheemazmalmmd (approved):
Code looks good. Needs test

shaheemazmalmmd:
@FrankOHara43 : please squash your commits to one

FrankOHara43:
@shaheemazmalmmd squashed the commits to one as suggested in the above comment.

shaheemazmalmmd:
@FrankOHara43 : The email subject will be modified by admins only who should be responsible this, So this changes are not valuable.

PR #3417, opened 2026-03-08
fix(nomos): prevent command buffer overflow in mySystem

FrankOHara43:
## Description

An overflow could occur when formatting shell commands with long file paths, potentially leading to memory corruption and arbitrary code execution.

### Changes

- Increased the size of the static command buffer (`cmdBuf`) to safely accommodate long file paths and command strings.
- Replaced unsafe `vsprintf` with `vsnprintf` for secure formatting, adding bounds checking to prevent buffer overflow.
- Added error handling to log and return an error if the command is truncated due to buffer limits.

### DIFF
--- a/src/nomos/agent/util.c
+++ b/src/nomos/agent/util.c
@@ -32,7 +32,7 @@ static char grepzone[10485760]; /* 10M for now, adjust if needed */
 static va_list ap;
 static char utilbuf[myBUFSIZ];
 static struct mm_cache mmap_data[MM_CACHESIZE];
-static char cmdBuf[512];
+static char cmdBuf[PATH_MAX * 2 + 256];
 
 
 #ifdef MEMORY_TRACING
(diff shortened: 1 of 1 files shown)


FrankOHara43:

shaheemazmalmmd:
@FrankOHara43 : please squash the commits to one.

FrankOHara43:
@shaheemazmalmmd squased the commits to one as suggested in the above comment.

src/nomos/agent/util.c
```diff
@@ -32,7 +32,7 @@ static char grepzone[10485760]; /* 10M for now, adjust if needed */
 static va_list ap;
 static char utilbuf[myBUFSIZ];
 static struct mm_cache mmap_data[MM_CACHESIZE];
-static char cmdBuf[512];
+static char cmdBuf[PATH_MAX * 2 + 256];
```
shaheemazmalmmd, line 35: Can you explain that is the math behind calculating this change ??

FrankOHara43:
@shaheemazmalmmd  `mySystem("mv '%s' %s", oldpath, newpath);` passes two full file paths into the command string, this can lead to `oldpath` upto  `PATH_MAX` and `newpath` upto `PATH_MAX` and will also have  fixed overhead. So, `512` can easily overflow if either of them reached even half of `PATH_MAX`.  The formula directly derives from the buffer's actual usage pattern and preventing truncation with the extra `256` bytes  as a conservative margin for command formatting and null termination. On Linux `PATH_MAX` is of `4096` adding upto the total `8448` bytes.

shaheemazmalmmd:
@FrankOHara43 : can you please squash the commits ro one ?

FrankOHara43:
@shaheemazmalmmd squashed commits into one as suggested in the above comment.

Issue #64, opened 2015-01-08
wget shouldn't be using NOKILL special option


yaobinshi:
---

Author Name: **Alex Norton** 
Original Redmine Issue: 1557, http://www.fossology.org/issues/1557
Original Date: 2012/03/15
Original Assignee: Mary Laser

---

Currently the delagent uses the NOKILL special flag so that the scheduler will not kill it when it is fetching large uploads. If an agent does not call the fo_scheduler_heart() function periodically, the scheduler will kill the agent because it believes that it is no longer going to complete its task. When performing very large uploads, the wget agent will no call fo_scheduler_heart() frequently enough and the scheduler was killing it.

As a patch for this problem, the NOKILL option was added to the wget agent. Wget should be changed to instead correctly call fo_scheduler_heart() since it is still working.

yaobinshi:
---

Original Redmine Comment
Author Name: **Bob Gobeille** 
Original Date: 2012/10/08 18:39:14 +0000

---

Change unranked issues to rank 99 so they show up in a ranking report at the bottom.  Otherwise, they show up on top.

mcjaeger:
it is using heartbeat, but should be during download:
https://github.com/fossology/fossology/blob/fa15f2615bf2373d566869812cd674671ce2ad51/src/wget_agent/agent/main.c#L186

mcjaeger:
but to keep track on this,  NO_Kill is still there:

https://github.com/fossology/fossology/blob/fa15f2615bf2373d566869812cd674671ce2ad51/src/wget_agent/wget_agent.conf

shaheemazmalmmd:
I have observed that the Wget is still using NO_KILL option. https://github.com/fossology/fossology/blob/master/src/wget_agent/wget_agent.conf

PR #3419, opened 2026-03-08
fix(wget): send periodic heartbeats during long-running wget operations

sakshammishra112:
## Description

The wget agent previously relied on the `NOKILL` special tag to prevent the scheduler from terminating it during long downloads. This PR removes the `NOKILL` workaround and introduces periodic heartbeats so the scheduler can detect that the agent is still active during long-running operations. 

closes: #64 
## Changes

1. Added the following functions:

   - `heartbeat_monitor()` – Background thread that sends scheduler heartbeats every 30 seconds.
   - `start_heartbeat_monitoring()` – Safely starts the monitoring thread before long-running operations.
   - `stop_heartbeat_monitoring()` – Stops the monitoring thread after the operation completes.

2. Added heartbeat monitoring around long-running operations such as:
   - wget downloads
   - tar/mv operations
   - tar operations in **GetVersionControl**

3. Updated `CMakeLists.txt` to link the agent with the `pthread` library.

4. Removed the `NOKILL` special option from `wget_agent.conf`.

## How To Test
1. Tested the download of CentOS iso file ~14 GB
``` bash
https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/iso/CentOS-Stream-9-latest-x86_64-dvd1.iso
```
Screenshot 2026-03-08 221124

The wget agent runs correctly even for long-running downloads.

2. Tested the download 500mb.bin ~500mb

```bash
https://mirror.nforce.com/pub/speedtests/500mb.bin
```

CC: @shaheemazmalmmd

### DIFF
--- a/src/wget_agent/agent/wget_agent.c
+++ b/src/wget_agent/agent/wget_agent.c
@@ -16,6 +16,9 @@
 #define ASPRINTF_MEM_ERROR_LOG LOG_FATAL("Not enough memory for asprintf before line %d", __LINE__)
 
 #include "wget_agent.h"
+#include <pthread.h>
+#include <signal.h>
+#include <time.h>
 
 char SQL[STRMAX];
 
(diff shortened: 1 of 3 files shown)


src/wget_agent/wget_agent.conf
```diff
@@ -21,5 +21,4 @@ max = -1
 ; Directives:
 ;     NOKILL: do not kill the agent if it isn't sending status updates
 ;     LOCAL:  only run the agent on the same computer as the scheduler
-special[] = NOKILL
```
Kaushl2208, line 24: Was there a problem with NOKILL option??
I dont think we need to remove it.

src/wget_agent/agent/wget_agent.c
```diff
@@ -425,7 +485,11 @@ int GetURL(char *TempFile, char *URL, char *TempFileDir)
      'http://a.org/file' -l 1 -R index.html*  2>&1"
    */
   LOG_VERBOSE0("CMD: %s", cmd);
+  
+  /* Start heartbeat monitoring for the long-running wget operation */
+  start_heartbeat_monitoring();
```
Kaushl2208, line 490: I am very skeptical of this implementation. Because I dont see a reason of using heartbeat process for `system(cmd)`, It already returns the required flags in case something doesnt work or process has failed etc. And we have handled that as well in further steps.

If there was a problem in downloading or getting the files, It still doesnt solve it, instead it will share the heartbeat nonetheless. 

Please try and discuss such architectural changes before spending any energy onto it.

sakshammishra112: I did discuss this with @shaheemazmalmmd before implementing this.

src/wget_agent/wget_agent.conf
sakshammishra112: True, I thought the same thing that their was no problem with NOKILL. But thought that this kind of behaviour was need for wget/

PR #3422, opened 2026-03-09
fix(bucketbrowser): handle missing bucket analysis dataset

### LINKED ISSUE #3418: Bucket Browser fails when no bucket analysis dataset exists for an upload
## Description

When opening the **Bucket Browser** for an upload where **Bucket Analysis was not selected during upload**, the page does not render correctly.

The Bucket Browser expects a valid bucket analysis dataset (`ars_pk`). If the bucket agent has not been executed yet (for example, when **Bucket Analysis is not checked in the upload analysis options**), the dataset does not exist and the page attempts to query `bucket_ars` using an undefined value.

This can result in an SQL error or an empty page.

As a consequence, users cannot access the Bucket Browser view for uploads that have not yet been analyzed by the bucket agent.

## Steps to Reproduce

1. Upload a file or archive in Fossology
2. In the **analysis options**, ensure that **Bucket Analysis is NOT selected**
3. Complete the upload
4. Navigate to:
`Browser->Bucket Browser`
5. Open the uploaded file or directory in the Bucket Browser

---

## Actual Behavior

The Bucket Browser fails to render correctly because no bucket dataset (`ars_pk`) exists for the upload.

Possible outcomes include:

- SQL errors
- blank page
- missing bucket results

## Expected Behavior

The Bucket Browser should handle the case where no bucket dataset exists for an upload and provide a clear user experience instead of failing.

---

## Screenshots

**Upload analysis options (Bucket Analysis not selected)**

Image

**Bucket Browser error / blank page**

Image
Image

---

im-vishesh15th:
### Summary
Handle missing bucket analysis dataset in Bucket Browser

### Problem
Bucket Browser assumes that a bucket analysis dataset (`ars_pk`) exists for every upload.

If **Bucket Analysis** is not selected during upload, the bucket agent is never executed and no dataset is created in the `bucket_ars` table. When the Bucket Browser is opened in this state, the page attempts to query `bucket_ars` using a missing dataset, which can result in SQL errors or an empty page. As a result, the Bucket Browser cannot be used for uploads that have not yet been analyzed by the bucket agent.

### Fix
The Bucket Browser now handles the absence of a bucket dataset gracefully.

The implementation adds a detection step before performing bucket queries:

1. If `ars_pk` is not provided in the request, the system attempts to fetch the latest bucket dataset for the upload from `bucket_ars`.
2. If no dataset exists, the browser does not continue with bucket queries.
3. Instead, a message is displayed informing the user that no bucket analysis is available.
4. A **"Run Bucket Agent"** button is shown, allowing the user to trigger bucket analysis directly from the Bucket Browser.
5. Once the agent completes, refreshing the page displays the bucket results normally.

This ensures the page does not fail and provides a clear path for users to generate the required analysis.

### Result
- Bucket Browser no longer fails when bucket analysis was not executed
- Users receive a clear message explaining the situation
- Bucket analysis can be triggered directly from the Bucket Browser
- Existing behavior for uploads that already contain bucket datasets remains unchanged

Fixes #3418

### DIFF
--- a/src/buckets/ui/ui-buckets.php
+++ b/src/buckets/ui/ui-buckets.php
@@ -186,15 +186,77 @@ class ui_buckets extends FO_Plugin
     $upload_pk = $row["upload_fk"];
     pg_free_result($result);
 
-    /* Get the ars_pk of the scan to display, also the select list  */
-    $ars_pk = GetArrayVal("ars", $_GET);
-    $BucketSelect = SelectBucketDataset($upload_pk, $ars_pk, "selectbdata",
-                                        "onchange=\"addArsGo('newds','selectbdata');\"");
-    if ($ars_pk == 0)
+    $ars_pk = GetParm("ars", PARM_INTEGER);
+
+/* If ars not provided → try latest bucket dataset */
+if (!$ars_pk)
+{
+  $sql = "SELECT ars_pk
+          FROM bucket_ars
+          WHERE upload_fk = $upload_pk
+          ORDER BY ars_pk DESC
+          LIMIT 1";
+
+  $result = pg_query($PG_CONN, $sql);
+  DBCheckResult($result, $sql, __FILE__, __LINE__);
+
+  if ($row = pg_fetch_assoc($result))
+  {
+    $ars_pk = $row['ars_pk'];
+  }
+
+  pg_free_result($result);
+}
+
+/* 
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
https://github.com/fossology/fossology/issues/3418#issuecomment-4074315068

PR #3423, opened 2026-03-10
fix(ununpack): prevent gcrypt handle leak in checksum error paths

FrankOHara43:
## Description

Previously, three error paths returned `NULL` without `calling gcry_md_close(checksumhandler),` which caused the allocated gcrypt handle to leak.

### Changes

-Added `gcry_md_close(checksumhandler)` before `return(NULL)` when `gcry_md_enable()` fails for MD5.
-Added `gcry_md_close(checksumhandler)` before `return(NULL)` when `gcry_md_enable()` fails for SHA1.
-Added `gcry_md_close(checksumhandler)` before `return(NULL)` when `calloc()` fails during checksum structure allocation.


FrankOHara43:

PR #3424, opened 2026-03-10
fix(unpack): use MAP_FAILED for mmap failure check in CopyFile

FrankOHara43:
## Description

A failed `mmap()` call was never detected due to the incorrect `NULL` comparison. This allowed execution to continue with an invalid pointer, which could later trigger a segmentation fault during the `write()` loop around.

### Changes

-Replaced `NULL` check with `MAP_FAILED` in `CopyFile()`.


FrankOHara43:

shaheemazmalmmd:
@FrankOHara43 : i see that there are two commits in this branch can you squash the commits to one ??

PR #3425, opened 2026-03-10
Fix NULL pointer dereference in SMTP configuration handling

### LINKED ISSUE #3384: Unsafe Hash Table Lookup Leading to Incorrect Logical Evaluation
## Location
`src/scheduler/agent/database.c:1157-1158`

## Description

The code performs direct comparisons using values returned from `g_hash_table_lookup()` without verifying whether the keys exist in the hash table:

```c
if ((g_strcmp0((char *)g_hash_table_lookup(smtpvariables, "SMTPPort"), "25") != 0) 
    || g_strcmp0((char *)g_hash_table_lookup(smtpvariables, "SMTPStartTls"), "1") == 0)
```

If either key:
- "SMTPPort"
- "SMTPStartTls"

does not exist in the hash table,
 then g_hash_table_lookup(...)  →  NULL
And you are effectively doing:
```
g_strcmp0(NULL, "25")
g_strcmp0(NULL, "1")
```
Even though g_strcmp0() is designed to handle NULL safely, the problem is:

### Incorrect Logical Evaluation
Example:
```
g_strcmp0(NULL, "25") != 0
```
This evaluates to TRUE.
This means:
- If "SMTPPort" is missing, the condition behaves as if the port is not 25.
- The system may enable or alter SMTP/TLS behavior incorrectly.

### Silent Misconfiguration Risk

If both keys are missing:

(TRUE) || (FALSE)
The condition evaluates to TRUE, potentially:

Enabling TLS unexpectedly
Forcing non-default SMTP behavior
Causing incorrect mail configuration handling
This can lead to runtime issues that are difficult to trace.

Aditya30ag:
## Description

This PR resolves potential NULL pointer dereference crashes in `src/scheduler/agent/database.c` when retrieving SMTP configuration from the hash table. 

This wouldcrash when:

SMTPHostName key missing from smtpvariables hash table
SMTPPort key missing from smtpvariables hash table

Crash Reason: g_hash_table_lookup() returns NULL, and g_string_append_printf() with %s format cannot handle NULL pointers - causes segmentation fault.

### Changes

* **Memory Safety:** Implemented NULL checks for `g_hash_table_lookup` calls related to `SMTPPort`, `SMTPStartTls`, and `SMTPHostName`.
* **Logic Improvements:** Refactored nested lookups into local variables to improve readability and ensure `g_strcmp0` is only called on non-NULL pointers.
* **Graceful Degradation:** Used ternary operators to provide empty string defaults in `g_string_append_printf` to prevent crashes while maintaining command string structure.

## How to test

1.  **Environment Setup:** Ensure Fossology is built and the scheduler agent is running.
2.  **Triggering the Bug:** Manually remove or comment out `SMTPHostName` or `SMTPPort` from your configuration database/hash table.
3.  **Execution:** Run a task that triggers an email notification or an SMTP-related agent action.
4.  **Expected Result:** The agent should now handle the missing keys gracefully (by skipping the logic or using empty strings) instead of triggering a Segmentation Fault (SIGSEGV).

---
Fixes #3384

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -1154,8 +1154,10 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
       g_string_append_printf(client_cmd, " -S mta=\"");
     }
     /* use smtps only if port is not 25 or SMTPStartTls is provided */
-    if((g_strcmp0((char *)g_hash_table_lookup(smtpvariables, "SMTPPort"), "25") !=  0) 
-        || g_strcmp0((char *)g_hash_table_lookup(smtpvariables, "SMTPStartTls"), "1") == 0)
+    char *smtpPort = (char *)g_hash_table_lookup(smtpvariables, "SMTPPort");
+    char *smtpStartTls = (char *)g_hash_table_lookup(smtpvariables, "SMTPStartTls");
+    if((smtpPort && g_strcmp0(smtpPort, "25") != 0) 
+        || (smtpStartTls && g_strcmp0(smtpStartTls, "1") == 0))
     {
       g_string_append_printf(client_cmd, "smtps://");
     }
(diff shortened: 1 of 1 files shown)


Aditya30ag:
Hello @Kaushl2208 ,
Could you please review this PR when you have time? I would really appreciate your feedback because

Before:
missing SMTPPort → TRUE

After:
missing SMTPPort → FALSE

the fix changes logic so this changes are valid or not?

src/scheduler/agent/database.c
```diff
@@ -1154,8 +1154,10 @@ char* get_email_command(scheduler_t* scheduler, char* user_email)
       g_string_append_printf(client_cmd, " -S mta=\"");
     }
     /* use smtps only if port is not 25 or SMTPStartTls is provided */
-    if((g_strcmp0((char *)g_hash_table_lookup(smtpvariables, "SMTPPort"), "25") !=  0)
```
Kaushl2208, line 1157: `g_strcmp0()` already handles the NULL gracefully (returns 0)`

What is the point of this??

Aditya30ag: I misunderstood the behavior of g_strcmp0().
Since it already handles NULL safely, the additional checks around it were unnecessary.

My original intention was to prevent potential crashes from g_string_append_printf() when NULL values are passed with %s. However, I see now that the changes I made also altered the existing logic like handling of missing SMTPPort, which was not appropriate for a fix.

thanks for reviewing

PR #3430, opened 2026-03-11
fix(scheduler): safely handle NULL folder_pk during email replacement

### LINKED ISSUE #3429: Unsafe integer conversion in agent configuration lookup
### Description
In several agent components (specifically within `src/scheduler/agent/`), configuration values are retrieved from a `GHashTable` and passed directly to `atoi()` or `atol()` without a NULL check.

### Technical Root Cause
The current implementation:
`int limit = atoi(g_hash_table_lookup(table, "key"));`

If "key" is missing from the database/hash table, `g_hash_table_lookup` returns `NULL`. In C, passing `NULL` to `atoi()` results in **Undefined Behavior**, which can lead to unpredictable agent crashes or incorrect analysis limits.

### Suggested Fix
Implement a safe retrieval pattern that checks for `NULL` and provides a sensible default value to ensure system stability even if a configuration entry is missing.

I would like to submit a PR to fix these instances in the scheduler agents.

Yamini04-oss:
## Description

This PR addresses the unsafe integer conversion identified in issue #3429.

### Changes

1. Added validation to ensure the result of PQget() is not NULL before passing it to atoi().
2.  Implemented a safe fallback to 0 for folder_pk if the database returns a null value.
 
3. Updated the $UPLOADFOLDERNAME replacement logic in src/scheduler/agent/database.c to handle both the initial folder lookup and the recursive parent-folder traversal loop safely.

- ## How to test
1.Compilation Check: Run make and sudo make install in the repository root to ensure the syntax in src/scheduler/agent/database.c is correct and the scheduler builds without errors.

2.Manual Code Audit: Verified that folderPkStr is checked for NULL before every atoi() call in the $UPLOADFOLDERNAME block.

3.Logic Verification: Verified that the while loop still correctly terminates and uses the fallback value 0 for the parent_folder_name query if a primary key is missing, preventing a segmentation fault.

Fixes #3429

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -213,7 +213,8 @@ static gboolean email_replace(const GMatchInfo* match, GString* ret,
     {
       rows = g_ptr_array_new();
       GString *foldername = g_string_new(PQgetvalue(db_result, 0, 0));
-      guint folder_pk = atoi(PQget(db_result, 0, "folder_pk"));
+      char *folderPkStr = PQget(db_result, 0, "folder_pk");
+      guint folder_pk = (folderPkStr != NULL) ? (guint)atoi(folderPkStr) : 0;
       g_ptr_array_add(rows, foldername);
       SafePQclear(db_result);
       g_free(sql);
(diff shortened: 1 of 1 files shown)


Yamzxx:
Hi @shaheemazmalmmd  

Sir, could you please let me know the status of this PR

Aditya30ag:
Hey @Yamini04-oss ,
Nice improvement adding the NULL checks before atoi().
Would it also make sense to replace atoi() with something like strtol() for safer parsing and better error detection?

Yamzxx:
Hello @Aditya30ag! You're totally right strtol() is much safer here since atoi() just fails silently with a 0. I'll refactor the folder_pk logic to use strtol() and add those error checks to make it more robust. I'll push the update shortly. Thanks for the catch

abdollahShahid:
@Yamini04-oss CI is failing due to the **DCO check**. Please sign off your commit (`git commit --amend -s`) and push again.

Yamzxx:
Yeahh! i have did that @abdollahShahid

Yamzxx:
Screenshot (513)
@shaheemazmalmmd Attached is the screenshot confirming the test results for the recent validation. As you can see, the outputs align with the expected requirements.

shaheemazmalmmd:
@Yamini04-oss : the testcases still fails. can you please check

Aditya30ag:
@Yamini04-oss the DCO check is failing 
please do sign-off for commit

PR #3431, opened 2026-03-12
fix(adj2nest): validate PGresult before processing upload rows

FrankOHara43:
## Description

Improvement in error handling by validating `PostgreSQL` query result before processing upload rows.

### Changes

- Added `PQresultStatus(pgResult) != PGRES_TUPLES_OK `validation after `PQexec()`.
- Cleared `pgResult` using `PQclear()` when the query result is invalid.
- Added an early `return` to avoid calling `PQntuples()` on an invalid result.

### DIFF
--- a/src/adj2nest/agent/adj2nest.c
+++ b/src/adj2nest/agent/adj2nest.c
@@ -70,6 +70,7 @@
 #include <ctype.h>
 #include <signal.h>
 #include <libgen.h>
+#include <libpq-fe.h>
 
 #include "libfossology.h"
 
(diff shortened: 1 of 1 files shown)


FrankOHara43:

src/adj2nest/agent/adj2nest.c
```diff
@@ -346,6 +347,11 @@ void    ListUploads     ()
   pgResult = PQexec(pgConn, SQL);
   fo_checkPQresult(pgConn, pgResult, SQL, __FILE__, __LINE__);
 
+  if (PQresultStatus(pgResult) != PGRES_TUPLES_OK) {
```
shaheemazmalmmd, line 350: Function fo_checkPQresult already does this

PR #3432, opened 2026-03-12
fix(api): remove stray debug echo from deleteFileCX and correct limit error messages

### LINKED ISSUE #3433: bug(api): stray debug echo in deleteFileCX() corrupts DELETE response body
**Describe the bug**

There's a leftover `echo $uploadTreeId;` inside `CopyrightController::deleteFileCX()` that outputs the raw item ID as plain text right before the JSON body is written. Any client trying to parse the response gets something like `2{"code":200,...}` which is invalid JSON.

**Affected endpoints**

Every DELETE endpoint that routes through `deleteFileCX()`:
- copyrights, user-copyrights, scancode-copyrights
- emails, scancode-emails
- urls, scancode-urls
- authors, scancode-authors
- ecc, keywords, ipra

**To reproduce**

Call any of the DELETE endpoints above — the response body will have the item ID prepended as plain text before the JSON payload.

**Expected behavior**

Clean JSON response with no extra output.

**Additional context**

Also noticed the `limit` validation error message in `CopyrightController`, `LicenseController` and `UploadController` says `"> 1"` but the actual check is `< 1`, meaning `limit=1` is valid. The message should say `>= 1`. `UploadTreeController` already has it right.

Fix tracked in #3432.

mandar1045:
Found two small bugs while reading through the API controllers.

**Debug echo left in production**
`deleteFileCX()` has `echo $uploadTreeId;` sitting right before the JSON response. This leaks the item ID as raw text into the response body, which breaks JSON parsing on the client. Affects all the DELETE endpoints for copyrights, emails, urls, authors, ecc, keywords, ipra — scanner variants too.

**Wrong error message for `limit`**
The check is `if ($limit < 1)` so `limit=1` is fine, but the error says `"> 1"` which makes it look like 1 is rejected. Changed it to `>= 1`. Same issue in three controllers — `LicenseController` and `UploadController` had the same copy-paste. `UploadTreeController` already said it right.

Closes #3433

Signed-off-by: Mandar Joshi <mandarjoshi1045@gmail.com>

### DIFF
--- a/src/www/ui/api/Controllers/CopyrightController.php
+++ b/src/www/ui/api/Controllers/CopyrightController.php
@@ -850,7 +850,7 @@ class CopyrightController extends RestController
       $limit = filter_var($limit, FILTER_VALIDATE_INT);
       if ($limit < 1) {
         throw new HttpBadRequestException(
-          "limit should be positive integer > 1");
+          "limit should be a positive integer >= 1");
       }
     } else {
       $limit = self::COPYRIGHT_FETCH_LIMIT;
@@ -933,8 +933,6 @@ class CopyrightController extends RestController
     $copyrightHash = $args['hash'];
     $userId = $this->restHelper->getUserId();
     $cpTable = $this->copyrightHist->getTableName($dataType);
-
-    echo $uploadTreeId;
 
     $this->uploadAccessible($uploadPk);
     $this->isItemExists($uploadPk, $uploadTreeId);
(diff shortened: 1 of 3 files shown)


mandar1045:
Hi @GMishx @Kaushl2208 @shaheemazmalmmd — could you approve the workflows so the CI checks can run? Thanks!

abdollahShahid:
@mandar1045 CI is failing due to the **DCO check**. Please sign off your commit (`git commit --amend -s`) and push again.

mandar1045:
Could a maintainer please approve the workflow runs? The DCO issue has been fixed.

shaheemazmalmmd:
PR https://github.com/fossology/fossology/pull/3366 also removes the echo statement, rest of the code doesnt add any value

PR #3436, opened 2026-03-12
fix(cyclonedx): store main license info using license key instead of overwriting licensesInDocument

### LINKED ISSUE #3435: CycloneDX export overwrites licensesInDocument when processing main licenses
While reviewing the CycloneDX exporter code, I noticed that `$licensesInDocument` is defined as an array of `SpdxLicenseInfo`, but inside the loop that processes main licenses it gets overwritten with a single `SpdxLicenseInfo` object. In `src/cyclonedx/agent/cyclonedx.php`, the code assigns `$this->licensesInDocument = (new SpdxLicenseInfo())` instead of adding the object to the array using the computed `$licId`. Since `$licensesInDocument` is later used with array indexing, this could lead to inconsistent behavior when multiple main licenses are present. The screenshots below show the array declaration and the assignment inside the loop.

#### Declaration of `licensesInDocument` as an array
Image

#### `licensesInDocument` overwritten with a `SpdxLicenseInfo` object inside the main license loop
Image

abdollahShahid:
Fixes #3435

`licensesInDocument` is declared as an array of `SpdxLicenseInfo`, but in the main license loop the array was overwritten with a single object. This change stores the license info using the computed `$licId` as key, preserving the array structure and avoiding potential issues when multiple main licenses are processed.

Screenshot 2026-03-13 012904

CLI reproduction showing that replacing the licensesInDocument array with a SpdxLicenseInfo object causes array_key_exists() to fail on the next iteration.

### DIFF
--- a/src/cyclonedx/agent/cyclonedx.php
+++ b/src/cyclonedx/agent/cyclonedx.php
@@ -215,13 +215,6 @@ class CycloneDXAgent extends Agent
       $reportedLicenseId = $this->licenseMap->getProjectedId($licId);
       $mainLicObj = $this->licenseDao->getLicenseById($reportedLicenseId, $this->groupId);
       $licId = $mainLicObj->getId() . "-" . md5($mainLicObj->getText());
-      if (!array_key_exists($licId, $this->licensesInDocument)) {
-        $this->licensesInDocument = (new SpdxLicenseInfo())
-          ->setLicenseObj($mainLicObj)
-          ->setCustomText(false)
-          ->setTextPrinted(true)
-          ->setListedLicense(true);
-      }
       $licensedata['id'] = $mainLicObj->getSpdxId();
       $licensedata['url'] = $mainLicObj->getUrl();
       $mainLicenses[] = $this->reportGenerator->createLicense($licensedata);


src/cyclonedx/agent/cyclonedx.php
```diff
@@ -216,7 +216,7 @@ protected function renderPackage($uploadId)
       $mainLicObj = $this->licenseDao->getLicenseById($reportedLicenseId, $this->groupId);
       $licId = $mainLicObj->getId() . "-" . md5($mainLicObj->getText());
       if (!array_key_exists($licId, $this->licensesInDocument)) {
-        $this->licensesInDocument = (new SpdxLicenseInfo())
+        $this->licensesInDocument[$licId] = (new SpdxLicenseInfo())
           ->setLicenseObj($mainLicObj)
           ->setCustomText(false)
           ->setTextPrinted(true)
```
its-sushant: Good catch, but this `if` block is irrelevant here since `$this->licensesInDocument` is never used after this assignment. We can remove the entire if block.

abdollahShahid: > Good catch, but this `if` block is irrelevant here since `$this->licensesInDocument` is never used after this assignment. We can remove the entire if block.

sure

PR #3438, opened 2026-03-13
fix(api): null safety checks, wrong status code, typo and misleading error messages

### LINKED ISSUE #3439: API controllers: null crashes, wrong status code, typo and bad error messages
While going through the REST API controllers, I found a handful of small but real issues that were either causing silent failures, wrong HTTP responses, or just misleading error messages for the caller.

**UserController — null crash in V2 mode**
`getUser`, `deleteUser`, and `updateUser` all call `getUserByName()` and immediately read `['user_pk']` off the result. If the username doesn't exist, the function returns null and PHP throws a fatal error. The caller just gets a 500 instead of a proper 404.

**GroupController — 200 instead of 201 on group creation**
Creating a group returns 200 OK. Should be 201 Created since a new resource is being made.

**GroupController — typo `apiVerison`**
`deleteGroup` uses `$apiVerison` (misspelled) for the version check, so the check was silently broken.

**ReportController — no null check before reading DB rows**
`downloadReport` reads from the DB and then directly accesses `$row['jq_type']` and `$row['job_upload_fk']` without checking if the query returned anything. Bad job IDs would cause an undefined index warning instead of a clean 404.

**JobController — inaccurate error message for pagination params**
The error said "cannot be smaller than 1 and has to be numeric" but limit is allowed to be 0 (check is `< 0`) and the values are already cast to int at that point.

---

Fixed in: #3438

mandar1045:
## Summary

Picked up a few issues in the API controllers that I noticed while reading through the code. Nothing major, but they were causing either silent failures or confusing behavior.

Fixes #3439

---

**Null crash in UserController (V2)**

In V2 mode, `getUser`, `deleteUser`, and `updateUser` were calling `getUserByName()` and directly accessing `['user_pk']` on the result without checking if the user actually exists. If the username isn't found, `getUserByName` returns null and PHP blows up. Added a null check and throw a 404 before it gets there.

**Wrong status code in GroupController**

Group creation was returning 200. Should be 201 since we're creating a new resource, not just doing some action on an existing one.

**Typo — `$apiVerison` in GroupController**

`deleteGroup` had `$apiVerison` instead of `$apiVersion`. PHP just treated it as an undefined variable so the version check was silently broken in that method.

**Missing null checks in ReportController**

`downloadReport` was reading `$row['jq_type']` and `$row['job_upload_fk']` straight from DB queries without checking if anything was actually returned. Added `empty($row)` guards before both accesses so it throws a proper 404 instead of an undefined index warning.

**Confusing error message in JobController**

The error for bad pagination params said "cannot be smaller than 1 and has to be numeric" but limit is allowed to be 0 (the check is `< 0`), and the values are already integers at that point so the numeric part doesn't make sense. Reworded it to match what the code actually validates.

### DIFF
--- a/src/www/ui/api/Controllers/GroupController.php
+++ b/src/www/ui/api/Controllers/GroupController.php
@@ -70,8 +70,9 @@ class GroupController extends RestController
    */
   public function createGroup($request, $response, $args)
   {
+    $apiVersion = ApiVersion::getVersion($request);
     $groupName = '';
-    if (ApiVersion::getVersion($request) == ApiVersion::V2) {
+    if ($apiVersion == ApiVersion::V2) {
       $queryParams = $request->getQueryParams();
       $groupName = $queryParams['name'] ?? '';
     } else {
@@ -83,7 +84,8 @@ class GroupController extends RestController
     $userDao = $this->restHelper->getUserDao();
     $groupId = $userDao->addGroup($groupName);
     $userDao->addGroupMembership($groupId, $this->restHelper->getUserId());
-    $returnVal = new Info(200, "Group $groupName added.", InfoType::INFO);
+    $statusCode = $apiVersion == ApiVersion::V2 ? 201 : 200;
+    $returnVal = new Info($statusCode, "Group $groupName added.", InfoType::INFO);
     return $response->withJson($returnVal->getArray(), $returnVal->getCode());
   }
 
@@ -98,7 +100,7 @@ class GroupController extends RestController
    */
   public function deleteGroup($request, $response, $args)
   {
-    $apiVerison = ApiVersion::getVersion($request);
+    $apiVersion = ApiVersion::getVersion($request);
     if (empty($args['pathParam'])) {
       throw new HttpBadRequestException("ERROR - No group name or id provided");
     }
@@ -109,7 +111,7 @@ class GroupController extends RestController
     $groupMap = $userDao->getDeletableAdminGroupMap($userId,
       $_SESSION[Auth::USER_LEVEL]);
     $groupId = null;
-    if ($apiVerison == ApiVersion::V2) {
+    if ($apiVersion == ApiVersion::V2) {
       $groupName = $args['pathParam'];
       $groupId = intval($userDao->getGroupIdByName($groupName));
     } else {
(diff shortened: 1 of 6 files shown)


src/www/ui/api/Controllers/GroupController.php
```diff
@@ -83,7 +83,7 @@ public function createGroup($request, $response, $args)
     $userDao = $this->restHelper->getUserDao();
     $groupId = $userDao->addGroup($groupName);
     $userDao->addGroupMembership($groupId, $this->restHelper->getUserId());
-    $returnVal = new Info(200, "Group $groupName added.", InfoType::INFO);
+    $returnVal = new Info(201, "Group $groupName added.", InfoType::INFO);
```
Kaushl2208: Pre-built clients on FOSSology's api will start breaking due to this change in response . Leave the old response for V1, Do the bug fix for V2.
(Also add the changes in the relevant api document file)

src/www/ui/api/Controllers/GroupController.php
```diff
@@ -98,7 +98,7 @@ public function createGroup($request, $response, $args)
    */
   public function deleteGroup($request, $response, $args)
   {
-    $apiVerison = ApiVersion::getVersion($request);
+    $apiVersion = ApiVersion::getVersion($request);
```
Kaushl2208, line 103: Good catch!

src/www/ui/api/Controllers/GroupController.php
mandar1045: Ok, thanks , I'll resolve it asap and update the PR.

mandar1045:
Updated createGroup to keep V1 response code 200 and use 201 for V2; updated openapiv2.yaml accordingly.

Tests:
- php -d session.save_path=/tmp/php_sessions vendor/bin/phpunit -c www/ui_tests/api/tests.xml --filter GroupControllerTest
- php -d session.save_path=/tmp/php_sessions vendor/bin/phpunit -c www/ui_tests/api/tests.xml

mandar1045:
Could a maintainer please approve the workflow runs?

mandar1045:
Hi! @shaheemazmalmmd @Kaushl2208  I've addressed the earlier feedback and rebased this PR — tests are passing and it should be ready for review now.

Please let me know if you'd prefer splitting this into smaller PRs. Thanks!

PR #3443, opened 2026-03-14
fix(test): replace deprecated assertRegExp in test_common_menu.php

mandar1045:
## What this fixes

`assertRegExp` was deprecated in PHPUnit 9 and removed in PHPUnit 10.
`test_common_menu.php` was still using the old method, causing test warnings
(and failures on newer PHPUnit). Replaced with `assertMatchesRegularExpression`.

## Files changed

- `src/lib/php/tests/test_common_menu.php` — 2 occurrences replaced

## What was NOT changed

`LicenseStdCommentDaoTest.php` and `cliTest.php` also call `assertRegExp` but
they already wrap it in a PHPUnit version check (`>= 9`), so they are
intentionally backwards-compatible and left as-is.

---

### DIFF
--- a/src/lib/php/tests/test_common_menu.php
+++ b/src/lib/php/tests/test_common_menu.php
@@ -12,6 +12,7 @@
 namespace Fossology\Tests;
 
 use PHPUnit\Framework\TestCase;
+use PHPUnit\Runner\Version as PHPUnitVersion;
 
 require_once dirname(__FILE__) . '/../common-menu.php';
 require_once dirname(__FILE__) . '/../common-parm.php';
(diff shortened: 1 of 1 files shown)


its-sushant:
@mandar1045  There is one extra commit which I believe is not in the scope of this PR.

mandar1045:
@its-sushant Thanks for catching that! You're right  -- the API fix commit (`fix(api): remove debug echo and fix misleading limit error messages`) was not in scope of this PR. I've removed it and will submit it as a separate PR. This PR now only contains the single in-scope commit for replacing the deprecated `assertRegExp` method.

mandar1045:
@its-sushant 

can you please check my other pr #3438 and #3432  as i opened them 2 days ago

mandar1045:
Could a maintainer please approve the workflow runs? The DCO issue and PHPUnit 8.x compatibility issue have been fixed.

mandar1045:
thank you , any other suggestions ?

PR #3444, opened 2026-03-14
fix(api): enforce admin-only access on user management endpoints

Sandipmandal25:
## Summary

The following methods in `UserController` were missing an admin permission check:

- `getUsers`
- `addUser`
- `deleteUser`
- `updateUser`

Although documented as Admin-only in the OpenAPI spec and enforced in frontend plugins (`Auth::PERM_ADMIN`), the backend did not validate admin privileges.

This allowed non-admin API users to:

- List all users
- Delete any user
- Modify user details

Added `throwNotAdminException()` as the first statement in all four methods, consistent with the pattern used in other admin controllers.

## Tests

- Obtain a non-admin write token
- GET /api/v1/users with non-admin token returns 403 Forbidden (was 200)
- DELETE /api/v1/users/{id} with non-admin token returns 403 Forbidden (was 202)
- GET /api/v1/users with admin token returns 200 OK
- POST /api/v1/users with admin token returns 201 Created
## Screenshots

**Before Using Non Admin Token**

image
image

**After Using Non Admin Token**

image
image

### DIFF
--- a/src/www/ui/api/Controllers/UserController.php
+++ b/src/www/ui/api/Controllers/UserController.php
@@ -46,6 +46,7 @@ class UserController extends RestController
    */
   public function getUsers($request, $response, $args)
   {
+    $this->throwNotAdminException();
     $apiVersion = ApiVersion::getVersion($request);
     $id = null;
     if (isset($args['pathParam'])) {
@@ -77,6 +78,7 @@ class UserController extends RestController
    */
   public function addUser($request, $response, $args)
   {
+    $this->throwNotAdminException();
     $apiVersion = ApiVersion::getVersion($request);
     $userDetails = $this->getParsedBody($request);
     $userHelper = new UserHelper();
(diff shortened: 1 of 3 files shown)


Sandipmandal25:
Hi @shaheemazmalmmd as discussed in the community meeting, you mentioned that this is not the intended behavior and suggested opening a PR for it. Following that, I have opened this PR. Thank you!

src/www/ui/api/Controllers/UserController.php
```diff
@@ -183,6 +186,7 @@ public function getCurrentUser($request, $response, $args)
    */
   public function updateUser($request, $response, $args)
   {
+    $this->throwNotAdminException();
```
its-sushant: @Sandipmandal25 Apart from the admin, the current user should be able to edit their own account. This check will prevent all users with non-admin access rights from editing their accounts.

Sandipmandal25: Thanks for the feedback! @its-sushant  I have updated `updateUser` to skip the admin check when a user is editing their own account. but `UserHelper::modifyUserDetails()` also has its own independent admin check that also need to be updated for self editing to work. Should I include that change too? or any other approach you would prefer? Thanks!

shaheemazmalmmd: Please proceed with changes in UserHelper::modifyUserDetails() as well.

Sandipmandal25: hi @shaheemazmalmmd  I have updated `UserHelper::modifyUserDetails()` to allow self editing as requested.

its-sushant (approved):
LGTM! Need test.

its-sushant:
Tested working as expected!

PR #3446, opened 2026-03-14
Fix: Prevent memory bloat in Scancode agent by using safe map lookupsFix memory bloat in scancode agent map lookup

### LINKED ISSUE #3441: Prevent implicit memory bloat in fileIdsMapReverse lookups
## **Description**

While reviewing the agent's map handling, I identified a potential memory efficiency issue in how `fileIdsMapReverse` lookups are performed. 

Currently, the code utilizes `operator[]` for data retrieval. In C++, `std::map::operator[]` is not a read-only operation; if the key (a `fileName`) is missing, the map implicitly inserts a default-constructed entry.

### Technical Impact
* **Memory Bloat:** Repeated lookups of missing or invalid filenames cause the map to grow silently with "ghost" entries.
* **Side Effects:** This can lead to incorrect logic in any part of the system that relies on map size or iteration.

### Proposed Refactor
I suggest replacing the direct access with a `find()` iterator check. This ensures the map remains immutable during lookups.

**Current Unsafe Logic:**
```cpp
auto id = fileIdsMapReverse[fileName]; // Causes implicit insertion if missing

Yamini04-oss:
## Description

This resolves a memory efficiency issue in the Scancode agent.

Previously, the agent used `operator[] `for map lookups. In C++, this causes "implicit insertion" meaning every time a filename wasn't found, the map would grow by creating a new, empty entry. For large uploads with thousands of files, this resulted in significant memory bloat.

### Changes

Safe Lookups: Updated` scancode_utils.cc` to use .find() and .end() for map lookups instead of the index operator.

Memory Optimization: By checking if a key exists before accessing it, we prevent the map from growing unnecessarily during the scanning process.

Logic Preservation: Maintained the existing behavior where fileId defaults to 0 if a match isn't found.

## How to test
1.To verify this fix, I focused on ensuring that the map no longer grows when a file is missing (the root cause of the bloat):

2.Logic Verification: I replaced the implicit insertion map[key] with a find() iterator. This ensures that if scancodeValue["file"] returns a path not present in our database, the fileIdsMapReverse remains the same size instead of allocating a new entry.

3.Behavioral Consistency: I verified that fileId still defaults to 0 when a match is not found, preserving the original error-handling flow of the agent.

CI/CD Build: Relying on the GitHub Actions pipeline to confirm that the change compiles correctly within the full FOSSology environment (bypassing local dependency issues with jsoncpp and librpm).

#Closes #3441


its-sushant:
@Yamini04-oss You just deleted a file.

Yamzxx:
So sorry Sir, @its-sushant, I made a mistake with my local commit.

Yamzxx:
@its-sushant Sir,that was truly unintentional! I had a local environment sync issue that wiped the file content during the commit without me realizing it

its-sushant:
> So sorry Sir, @its-sushant, I made a mistake with my local commit.

Hey. No need to be sorry, It's your first contribution. In future make sure to verify the changes before creating the PR. Chill!

PR #3265, opened 2026-01-21
feat(decider): auto decide OJO LicenseRef when Nomos finds no license

Meetjain1:
## Description

When using SPDX `LicenseRef-*` licenses, Nomos may report `NO_LICENSE_FOUND` while Ojo detects a valid `LicenseRef`.
In this case the Decider currently refuses to conclude a license.
This PR adds a safe fallback so Decider can automatically conclude the license based on Ojo findings **only when there is no contradiction** and the `LicenseRef` exists in Fossology’s main license list.

### Changes

* Added a new decider fallback rule to auto-decide Ojo `LicenseRef-*` results when Nomos reports no license found.
* Implemented contradiction checks to ensure we do not auto-decide if Nomos reports a different license.
* Added unit tests covering:

  * Ojo-only case
  * Nomos `NO_LICENSE_FOUND` case
  * contradiction case
  * valid/invalid `LicenseRef` in main license list

## How to test

Run unit tests:

```bash
composer install
./vendor/bin/phpunit src/decider/agent_tests/Unit/DeciderAgentTest.php --testdox
```

### Added Unit Tests

* `testAreOjoLicenseRefsWithoutNomosContradiction_ojoOnly()` -> Verifies the OJO-only LicenseRef case
* `testAreOjoLicenseRefsWithoutNomosContradiction_withContradiction()` -> Verifies contradiction handling
* `testIsValidLicenseRefInMainList_valid()` -> Verifies a valid LicenseRef exists in the main license list
* `testIsValidLicenseRefInMainList_invalid()` -> Verifies an invalid LicenseRef is rejected

Expected output includes:

* `OK (4 tests, 4 assertions)`

image

Fixes #2147

PR #3449, opened 2026-03-14
decider: ignore nomos no-license fallback for OJO/RESO contradiction #3265

Shivangisriva:
#3265 

Problem
When a file contains an SPDX LicenseRef-* license, OJO can correctly detect the LicenseRef-* while Nomos may report a fallback result like No_license_found (or equivalent “license not found” value).

In this situation, the Decider ojoNoContradiction logic may treat Nomos fallback as contradiction and skip auto-conclusion, even when the LicenseRef-* is already known in the main license list.

Root cause
The contradiction check compared OJO/RESO findings with Nomos findings without explicitly ignoring Nomos “no-license-found” fallback values.

Fix
Updated Decider logic to:
Ignore Nomos fallback findings (No_license_found, License not found, Void) during OJO/RESO contradiction checks.
Keep contradiction behavior unchanged for real Nomos license findings.
Require at least one actual OJO/RESO finding before auto-concluding (to avoid fallback-only false positives).
Files changed
DeciderAgent.php
DeciderAgentTest.php

Tests added
Added unit tests for:
OJO + Nomos fallback => should agree.
Nomos fallback-only group => should not be treated as contradiction.
OJO + real Nomos conflicting license => should still disagree.

### DIFF
--- a/src/decider/agent/DeciderAgent.php
+++ b/src/decider/agent/DeciderAgent.php
@@ -323,9 +323,14 @@ class DeciderAgent extends Agent
   private function autodecideIfOjoMatchesNoContradiction(ItemTreeBounds $itemTreeBounds, $matches)
   {
     $licenseMatchExists = count($matches) > 0;
+    $haveOjoFinding = false;
     foreach ($matches as $licenseMatches) {
+      $haveOjoFinding = $haveOjoFinding ||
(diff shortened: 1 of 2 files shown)


its-sushant:
Duplicate and Irrelevant.

Issue #3461, opened 2026-03-16
Optimize Hash Table Lookups in `nomos_utils.c`


Aditya30ag:
## Issue

The current hash table implementation used in `nomos_utils.c` for license cache lookups is not optimized for performance. The implementation relies on a simple hashing mechanism and performs expensive string comparisons during collision resolution.

This becomes inefficient when the cache size grows or when the lookup function is executed frequently, which is common in the Nomos license detection pipeline.

### Problems Identified

1. **Weak Hash Function**
   - The existing hash function only uses the first `sizeof(long)` bytes of the key.
   - This causes poor distribution of hash values.
   - Increased collisions lead to slower lookups.

2. **Expensive String Comparisons**
   - `strcmp()` is called directly during collision resolution.
   - This results in unnecessary character-by-character comparisons.

3. **Function Call Overhead**
   - Frequently used functions like `lrcache_hash()` and `lrcache_lookup()` are not inlined.
   - This adds overhead during heavy lookup operations.

4. **Cache Inefficiency During Collision Handling**
   - Collisions cause repeated memory access patterns that degrade CPU cache performance.

Kaushl2208:
I think we should step back a bit before jumping to conclusions or introducing a new hash strategy.

First, the current cache lookup operates on `rf_shortname`, which represents OSS license shortnames. These are typically small, well-defined, and fairly distinct strings (e.g., MIT, GPL-2.0, Apache-2.0). This is not a high-entropy or adversarial dataset where hash distribution becomes a critical bottleneck. The nature of the input matters a lot when evaluating hashing strategies.

Second, I’d like to understand how the claim of “slower lookups” was established.

- Do we have any benchmarks comparing the current implementation vs alternatives?
- Any profiling data showing lrcache_hash() or strcmp() as hotspots in the Nomos pipeline?

Without data, it’s hard to justify that the current approach is actually a problem in practice.

Third, I’m not sure how FNV came into the picture here. It’s a solid general-purpose hash, but it’s not the only option and definitely not the default answer for every hash table concern. Given our input characteristics, introducing a more complex hash needs justification.

If the concern is genuinely about distribution, there are simpler and well-understood alternatives like:

```C
hashval = 5381;
while (*p)
  hashval = ((hashval << 5) + hashval) + *p++; 
```

This would already be a step up from the current approach without adding unnecessary complexity.

[...]

PR #3463, opened 2026-03-16
perf(nomos): optimize hash table lookups in nomos_utils.c

Aditya30ag:
## Description

Improve hash table lookup performance in `nomos_utils.c`.

The existing hash function used only a few bytes of the key, which could lead to poor hash distribution and increased collisions. This change introduces the FNV-1a hash algorithm and optimizes lookup operations to reduce unnecessary string comparisons.

## Changes

- Replace the existing hash function with the FNV-1a algorithm for better distribution.
- Store the length of `rf_shortname` in the cache node to avoid repeated `strlen()` calls.
- Add a quick length check before performing `strcmp()` during lookups.
- Mark critical lookup functions for inlining.
- Use bit masking for hash indexing when the table size is a power of two.

## How to test

1. Build Fossology normally.
2. Run the Nomos agent on a sample repository or test dataset.
3. Verify that license detection results remain unchanged.
4. Confirm that the application runs without errors and cache lookups function correctly.

Closing issue: #3461

### DIFF
--- a/src/nomos/agent/nomos_utils.c
+++ b/src/nomos/agent/nomos_utils.c
@@ -104,7 +104,7 @@ FUNCTION long add2license_ref(char *licenseName)
 }
 
 /**
- \brief calculate the hash of an rf_shortname
+ \brief calculate the hash of an rf_shortname using FNV-1a hash algorithm
  rf_shortname is the key
 
  @param pcroot Root pointer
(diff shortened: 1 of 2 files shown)


Yamzxx:
Nice improvement! Using FNV-1a for hashing and adding the length check before `strcmp()` seems like a good way to reduce collisions and unnecessary comparisons.

Just curious, did you notice any performance difference when running Nomos on larger repositories after this change?
@Aditya30ag

Aditya30ag:
The improvements are mainly based on reducing hash collisions and avoiding unnecessary string comparisons, so they should scale better with larger datasets. I’m planning to run more comprehensive performance tests to quantify the impact.

Kaushl2208:
@Aditya30ag sir, `Performance tests and quantifying impact` is something that is considered as the evidence of an issue. 
There are gaps in each and every library/function but they are designed to fulfill a purpose, Yes they can totally be extended and its a good thing to upgrade but there has to be a very specific reason to do so.

You have opened an issue on specific performance functions, Then please show how did you come up to that ?? I am hoping it was not an AI that suggested to jump 5 level of hoops and integrate a SWORD to solve a small fix. 

I will wait for the metrics from you on the issue. Meanwhile, I am closing this PR.

PR #3465, opened 2026-03-16
fix(buckets): prevent NULL dereference when parsing scheduler input

DARSHAN-R-DARSHAN:
The buckets agent parses scheduler input using strtok_r and atoi.

If malformed input is received (missing value after bppk= or upk=),
strtok_r may return NULL, which is then passed to atoi(). This results
in undefined behavior, which may cause a crash.

This patch adds validation before converting tokens to integers to
prevent NULL dereference and improve robustness.

### DIFF
--- a/src/buckets/agent/buckets.c
+++ b/src/buckets/agent/buckets.c
@@ -236,12 +236,19 @@ int main(int argc, char **argv)
       {
         if (strcmp(token, "bppk") == 0)
         {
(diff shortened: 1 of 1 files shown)


Kaushl2208:
No value addition, Closing this now.
Also please open an issue in future and wait for a maintainer to actually take a look into the issue, saves both of our time.

PR #3466, opened 2026-03-16
Fix(ui): rename "Copy" action to "Alias" in content_move (#613)

### LINKED ISSUE #613: Fix(content_move): the action **Copy** should be renamed to something like **Alias**

Shivangisriva:
### Description

Renamed the "Copy" action to "Alias" in the content_move UI.

The current "Copy" functionality actually creates an alias to the same folder instead of creating a new copy. Renaming it to "Alias" better reflects the behavior and avoids confusion for users.

### Changes
- Updated UI label from "Copy" to "Alias" in the content_move page.

### Issue
Fixes #613

image


shaheemazmalmmd:
Code deletes decision importer agent file and adds a new file package.json which is irrelevant here.

Shivangisriva:
Thanks for the feedback.

I understand that the changes I made in the previous PR were not relevant to the scope of the issue. I’ve now prepared a cleaned-up version that only includes the intended changes.

Could you please take a look at the new PR and let me know if it looks correct? Also, if possible, guidance on which files to focus on for future contributions would be very helpful.

Thanks!

PR #3468, opened 2026-03-17
fix(showjobs): optimize N+1 queries and fix pagination

its-sushant:
## Description

fix(showjobs): optimize N+1 queries and fix pagination

### Changes

- Batch permission checks for uploads in `myjobs()`
- Fix Pagination in `uploads2Jobs()`
- Reduce 4 weeks to 2 weeks for showing recent jobs.
- Get `getJobInfo()` in batch.

### DIFF
--- a/src/lib/php/Dao/ShowJobsDao.php
+++ b/src/lib/php/Dao/ShowJobsDao.php
@@ -23,13 +23,23 @@ class ShowJobsDao
   /** @var maxUploadsPerPage */
   private $maxJobsPerPage = 10; /* max number of jobs to display on a page */
   /** @var nhours */
-  private $nhours = 672;  /* 672=24*28 (4 weeks) What is considered a recent number of hours for "My Recent Jobs" */
+  private $nhours = 336;  /* 336=24*14 (2 weeks) What is considered a recent number of hours for "My Recent Jobs" */
 
   function __construct(DbManager $dbManager, UploadDao $uploadDao)
   {
     $this->dbManager = $dbManager;
     $this->uploadDao = $uploadDao;
     $this->logger = new Logger(self::class);
+  }
+
+  /**
+   * @brief Convert an array of integers to a Postgres array literal string.
+   * @param array $ints Array of integer values
+   * @return string Postgres array literal, e.g. "{1,2,3}"
+   */
+  private function toPgIntArray(array $ints)
+  {
+    return '{' . implode(',', array_map('intval', $ints)) . '}';
   }
 
   /**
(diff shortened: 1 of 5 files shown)


its-sushant:
CC: @Kaushl2208 @shaheemazmalmmd

src/lib/php/Dao/ShowJobsDao.php
```diff
@@ -49,29 +49,36 @@ function uploads2Jobs($upload_pks, $page = 0)
       return $upload_pk !== null && $this->uploadDao->isAccessible($upload_pk, Auth::getGroupId());
     });
 
-    // get count of upload pks, return empty array if count equals 0
-    $jobCount = count($upload_pks);
-    if ($jobCount == 0) {
-      return $jobArray;
+    if (empty($upload_pks)) {
+      return array();
```
Kaushl2208: Should use already created `$jobArray`??

src/lib/php/Dao/ShowJobsDao.php
```diff
@@ -49,29 +49,36 @@ function uploads2Jobs($upload_pks, $page = 0)
...
-    for (; $offset < $lastOffset; $offset++) {
-      $upload_pk = $upload_pks[$offset];
+    /* Count total jobs for pagination */
+    $countRow = $this->dbManager->getSingleRow(
+      "SELECT count(*) AS cnt FROM job WHERE job_upload_fk = ANY($1)",
+      array($uploadList),
+      __METHOD__ . ".countJobs"
+    );
+    $totalJobCount = intval($countRow['cnt']);
+    $totalPages = floor($totalJobCount / $this->maxJobsPerPage);
```
Kaushl2208, line 76: shouldn't it be `ceil`, 

Lets say if we have `totalJobCount = 95` and `maxJobsPerPage = 10`
Then, `95 / 10 = 9.5 → floor → 9`

But we will need 10 pages to show 95 values right?? Cause 9 will end up showing 90 jobs??

src/lib/php/Dao/ShowJobsDao.php
```diff
@@ -114,11 +121,21 @@ public function myJobs($allusers, $page = 0)
...
     $result = $this->dbManager->execute($statementName);
+
+    $rows = array();
+    $uploadIds = array();
     while ($row = $this->dbManager->fetchArray($result)) {
-      if (! empty($row['job_upload_fk'])) {
-        $uploadIsAccessible = $this->uploadDao->isAccessible(
-          $row['job_upload_fk'], Auth::getGroupId());
-        if (! $uploadIsAccessible) {
+      $rows[] = $row;
```
Kaushl2208, line 138: Do we need to store values duplicated in an array?? If nor, Lets release this? 
This doubles memory usage (DB buffer + PHP array)

src/lib/php/Dao/ShowJobsDao.php
```diff
@@ -163,74 +180,215 @@ public function myJobs($allusers, $page = 0)
   **/
   public function getJobInfo($job_pks)
   {
-    /* Output data array */
+    if (empty($job_pks)) {
+      return array();
+    }
+
+    $jobPkArray = '{' . implode(',', array_map('intval', $job_pks)) . '}';
```
Kaushl2208: We are implicitly building this string in multiple functions. If it makes sense can we create a helper function that can do that??

Or do we actually need to typecast these as string?? Can we :

`$job_pks = array_map('intval', $job_pks);` and in sql  use `ANY($1::int[])` 

Leaving this upto you.

src/lib/php/Dao/ShowJobsDao.php
```diff
@@ -163,74 +180,215 @@ public function myJobs($allusers, $page = 0)
...
+
+    $jobData = $this->fetchJobRecords($jobPkArray);
+    $uploadPks = $this->getUploadPksFromJobs($jobData);
+
+    $uploadRows = $this->fetchUploadRecords($uploadPks);
+    $uploadtreeRows = $this->fetchUploadtreeRoots($uploadRows);
+
+    $this->attachUploadData($jobData, $uploadRows, $uploadtreeRows);
+    $this->attachDeletedUploadData($jobData);
+    $this->attachJobQueueData($jobData, $jobPkArray);
```
Kaushl2208, line 204: Nice I like the modular approach to this. 
Can you also check if we can merge some of these queries??? 
Because now, We are handling data twice, Once in DB and Once with PHP operations.

src/lib/php/Dao/ShowJobsDao.php
its-sushant: Yes

src/lib/php/Dao/ShowJobsDao.php
its-sushant: Since pagination is based on 0 base index, floor should work!

src/lib/php/Dao/ShowJobsDao.php
its-sushant: Yup, Could be released earlier.

src/lib/php/Dao/ShowJobsDao.php
its-sushant: Created a helper function for this.

src/lib/php/Dao/ShowJobsDao.php
its-sushant: Could be merged partially!

its-sushant:
@Kaushl2208 Thanks for the review. Made the suggested changes. PTAL!

src/lib/php/Dao/ShowJobsDao.php
```diff
@@ -163,75 +189,177 @@ public function myJobs($allusers, $page = 0)
...
+   * @brief Fetch job records for the given job_pk's.
+   * @param string $jobPkArray Array of job_pk's in Postgres array format (e.g. "{1,2,3}")
+   * @return array of job records indexed by job_pk
+   */
+  private function fetchJobRecords($jobPkArray)
+  {
+    $jobData = array();
+    $stmt = __METHOD__;
+    $this->dbManager->prepare($stmt,
+      "SELECT * FROM job WHERE job_pk = ANY($1::int[])");
```
Kaushl2208, line 226: Hey sushant, Either lets use the helper function or we try and caste in query.

I will prefer the latter.

its-sushant: Since Postgres expects array literals in the format '{1,2,3}' and PHP don't automatically convert arrays, this helper function would be necessary here.

PR #3478, opened 2026-03-18
fix(nomos): preserve getline base pointer during line trimming

FrankOHara43:
## Description

Fixed undefined behavior in `nomos` line trimming. Currently the heap buffer returned by `getline()`  is modified directly during leading whitespace removal. This shifts the original pointer and can result in invalid memory deallocation when `free()` is called later.

### Changes
- A separate working pointer is initialized to preserve the original `getline()` allocation.
- Whitespace trimming updated to operate on the working pointer instead of owned buffer pointer.
- Added bounds checking before trailing whitespace trimming.
- Used `(unsigned char)` casts in `isspace()` calls for safe character handling.
- Passed the trimmed working pointer to `processFile()` while keeping the original pointer for `free()`.

### DIFF
--- a/src/nomos/agent/nomos.c
+++ b/src/nomos/agent/nomos.c
@@ -200,22 +200,28 @@ void list_dir (const char * dir_name, int process_count, int *distribute_count, 
 void read_file_grab_license(int file_number, FILE **pFile)
 {
   char *line = NULL;
+  char *lineToProcess = NULL;
   size_t len = 0;
   int lenth_tmp = 0;
   ssize_t read = 0;
 
   /*read line by line, then start to scan licenses */
   while ((read = getline(&line, &len, pFile[file_number])) != -1) {
-    if (line && line[0]) // line is not empty
+    lineToProcess = line;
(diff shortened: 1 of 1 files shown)


abdollahShahid:
@FrankOHara43 Hey,  CI is failing because one PHPUnit test is breaking (LicenseDaoTest). I think the change might be slightly affecting behavior, not just fixing the pointer handling. It’d be good to double-check that the output/parsing stays exactly the same and only the memory handling is improved. Maybe try running the tests locally and compare before/after.

FrankOHara43:
@abdollahShahid I've applied the changes suggested above, please take a look.

Kaushl2208:
Hey @FrankOHara43, Did you face an issue regarding this? Do you have logs or anything for `invalid memory deallocation`??

Cause looking at the changes, I am not seeing anything valuable. Could you please open an issue in future if you face any issue or bugs or want to suggest a feature? Let a maintainer take a look, replicate??? 

If you have faced an issue linked to this, and you have something, Open an issue we will discuss there. Meanwhile I am closing this PR.

PR #3484, opened 2026-03-18
fix(upload): prevent orphan DB records when file move fails

winniem30:
### Problem
During multi-file uploads, a database record is created before the file is successfully moved to disk. If the file move operation fails, the database entry remains without a corresponding file, leading to orphaned upload records.

### Root Cause
The function JobAddUpload() is called before verifying successful file storage.

### Fix
Added cleanup logic to delete the database record when file movement fails.

### Impact
- Prevents database inconsistency
- Ensures upload records always correspond to actual files

### Testing
- Simulated file upload failure
- Verified that no orphan records remain in the database

Closes #<3483>

### DIFF
--- a/src/www/ui/page/UploadFilePage.php
+++ b/src/www/ui/page/UploadFilePage.php
@@ -28,10 +28,10 @@ class UploadFilePage extends UploadPageBase
   public function __construct()
   {
     parent::__construct(self::NAME, array(
-        self::TITLE => _("Upload a New File"),
-        self::MENU_LIST => "Upload::From File",
-        self::DEPENDENCIES => array("agent_unpack", "showjobs"),
-        self::PERMISSION => Auth::PERM_WRITE
+      self::TITLE => _("Upload a New File"),
+      self::MENU_LIST => "Upload::From File",
+      self::DEPENDENCIES => array("agent_unpack", "showjobs"),
+      self::PERMISSION => Auth::PERM_WRITE
     ));
   }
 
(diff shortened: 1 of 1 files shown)


sakshammishra112:
Hey @winniem30 I think there are commits which are not part of this PR. can u remove them

winniem30:
> Hey @winniem30 I think there are commits which are not part of this PR. can u remove them

Thanks for pointing that out! I've cleaned up the branch and removed unrelated commits.

winniem30:
Sorry for the messy commit history! I was resolving rebase conflicts while cleaning up unrelated commits, which led to multiple force pushes. The branch is now clean with a single focused commit. Apologies for the noise!

would love your feedback on the fix and let me know if the approach looks good or needs any changes. @shaheemazmalmmd Thanks!

src/www/ui/page/UploadFilePage.php
```diff
@@ -173,16 +190,21 @@ protected function handleUpload(Request $request)
           $uploadedFile['file']->getFilename() . '-uploaded'
         )->getPathname();
       } catch (FileException $e) {
+        if (!empty($uploadId)) {
+          $this->dbManager->getDbDriver()->exec(
+            "DELETE FROM upload WHERE upload_pk = " . intval($uploadId)
+          );
+        }
```
Kaushl2208, line 197: @winniem30 , I am guessing you never setup fossology and used it??

First of all, The interesting PR title suggests that you have completed the removal of orphan records. But did you go through the Function that creates DB entries?? I am guessing this was suggested through AI. 

Please go through the workflow once. You will understand why a failed task doesnt mean you can just clean one record from a table. You still have to show the failed jobs too, The details of the failed job as well. 

If you ask the maintainers first before putting up such changes, you will know why they are not mandatory.


Anyways, 

AI suggested, un helpful change. Can be closed

Kaushl2208 (commented):
Closing.

src/www/ui/page/UploadFilePage.php
winniem30: Thank you for the feedback @Kaushl2208. I did set up FOSSology locally, but I acknowledge I didn't trace through the full workflow deeply enough. I'll study the complete upload flow more carefully before raising further changes. Appreciate the guidance. 
image

Issue #1271, opened 2019-01-16
Multiple uploads can not be reused


GMishx:
### Description

While choosing multiple uploads for reuse, only one goes to the DB.

#### How to reproduce

1. Upload two or more packages.
    - Performing clearing does not matter.
1. Upload a new package.
    - While uploading, select more than one package for reuse.
1. Only one package is selected for reuse.
    - Check entries from `SELECT * FROM upload_reuse;`

### Versions

* Last commit id on master: 0995f55454fa0ec31a77ef1bdb19e02e29e622b5
* Operating System (lsb_release -a):
```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```

### Logs

Nothing shows up in any logs.

NicolasToussaint:
I confirm the problem still exists.

I also have a problem selecting multiple  uploads for reuse with the REST API.  
It looks like the Reuser JSOPN schema [only accepts an integer](https://github.com/fossology/fossology/blob/d1a900b9d0d8ab9609d306bba809608c29637cfb/src/www/ui/api/documentation/openapi.yaml#L889).

I guess an array would be necessary here?  
```
  "reuse": {
    "reuse_upload": [108,110],
    "reuse_group": "3",
    "reuse_main": true,
    "reuse_enhanced": true
  }
```

nikhilsingh7777:
Hello @shaheemazmalmmd  Is this problem still need to fix , Do i look into this ?

PR #3488, opened 2026-03-19
fix(reuser): support multiple upload reuse selections

sakshammishra112:
## Description

When selecting multiple uploads in the UI, only one upload was processed for reuse. This occurred because the UI was sending only a single value to the backend, as `uploadToReuseSelectorName` was not defined as an array.

Closes: #1271 

### Changes

- `agent_reuser.html.twig`: Changed `uploadToReuseSelectorName` to an array to support multiple selections.
- `agent-reuser.php`: Added logic to handle multiple uploads.

### Context
- In `zlib-1.3.1.tar.gz` i have cleared licenses for `zlib-1.3.1.tar.gz/zlib-1.3.1.tar/zlib-1.3.1/contrib/ada`
- In `zlib-1.2.12.tar.gz`  i have cleared licenses for `zlib-1.3.1.tar.gz/zlib-1.3.1.tar/zlib-1.3.1/contrib/dotzlib/DotZib`
### Before

https://github.com/user-attachments/assets/1ba4e189-bd19-4159-987c-e44cbfddbc4d

Only reused decison from one upload file. ie. `zlib-1.3.1.tar.gz`

Screenshot 2026-03-21 131516

### After

https://github.com/user-attachments/assets/c9c014a4-0841-4000-8acf-9ec14975ac0b

Used decisons form both uploads i.e.  `zlib-1.3.1.tar.gz` and  `zlib-1.2.13.tar.gz`

Screenshot 2026-03-21 132120

## How to test
- Select multiple uploads for reuse and verify that multiple items are processed in jobs.
- Execute the following query:
 ```sql
 SELECT * FROM upload_reuse WHERE upload_fk = <upload_id>;
```
### Uploads used
```bash
wget https://zlib.net/fossils/zlib-1.2.13.tar.gz
wget https://zlib.net/fossils/zlib-1.2.8.tar.gz
wget https://zlib.net/fossils/zlib-1.3.1.tar.gz
```

### DIFF
--- a/src/reuser/ui/agent-reuser.php
+++ b/src/reuser/ui/agent-reuser.php
@@ -105,13 +105,6 @@ class ReuserAgentPlugin extends AgentPlugin
         return $this->scheduleOsselotImportDirect($jobId, $uploadId, $errorMsg, $request);
     }
 
-    $reuseUploadPair = explode(',', $request->get(self::UPLOAD_TO_REUSE_SELECTOR_NAME), 2);
-    if (count($reuseUploadPair) == 2) {
-      list($reuseUploadId, $reuseGroupId) = $reuseUploadPair;
-    } else {
-      $errorMsg .= 'no reuse upload id given';
-      return -1;
-    }
     $groupId = $request->get('groupId', Auth::getGroupId());
     $getReuseValue = $request->get(self::REUSE_MODE) ?: array();
     $reuserDependencies = array("agent_adj2nest");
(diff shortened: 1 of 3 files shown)


src/reuser/ui/agent-reuser.php
```diff
@@ -134,15 +127,44 @@ public function scheduleAgent($jobId, $uploadId, &$errorMsg, $request)
...
 
+    $reuseSelections = $request->get(self::UPLOAD_TO_REUSE_SELECTOR_NAME);
+    if (!is_array($reuseSelections)) {
+      $reuseSelections = [$reuseSelections];
+    }
+
+    $createdLinks = 0;
+    foreach ($reuseSelections as $reuseSelection) {
+      if (empty($reuseSelection) || !is_string($reuseSelection)) {
+        continue;
```
Kaushl2208: Tho, I understand this case is more of a defensive check and very less chances of happening. But dont you think this should be logged or atleast gives some response, Instead of just quietly pressing the issue and doing continue.

It will be easier to catch the future issues if we just dont quietly suppress it.

src/reuser/ui/agent-reuser.php
```diff
@@ -134,15 +127,44 @@ public function scheduleAgent($jobId, $uploadId, &$errorMsg, $request)
...
+        intval($reuseUploadId),
+        intval($groupId),
+        intval($reuseGroupId),
+        $reuseMode
+      );
+      $createdLinks++;
+    }
+
+    if ($createdLinks === 0) {
+      $errorMsg .= 'no reuse upload id given';
```
Kaushl2208:
```suggestion
      $errorMsg .= 'No valid reuse upload selections found';
```

Current one is pretty vague.

sakshammishra112: Done

src/reuser/ui/agent-reuser.php
sakshammishra112: Done. Also added log for `count($reuseUploadPair) !== 2`.

src/reuser/ui/agent-reuser.php
```diff
@@ -134,15 +127,46 @@ public function scheduleAgent($jobId, $uploadId, &$errorMsg, $request)
...
+    $reuseSelections = $request->get(self::UPLOAD_TO_REUSE_SELECTOR_NAME);
+    if (!is_array($reuseSelections)) {
+      $reuseSelections = [$reuseSelections];
+    }
+
+    $createdLinks = 0;
+    foreach ($reuseSelections as $reuseSelection) {
+      if (empty($reuseSelection) || !is_string($reuseSelection)) {
+        error_log("Reuser: Invalid reuse selection found - empty or non-string value");
+        continue;
```
Kaushl2208: I still think, if this is not a expected behaviour we should throw, Exception at this stage instead of just logging and continuing. Let the reuser fail.

src/reuser/ui/agent-reuser.php
```diff
@@ -134,15 +127,46 @@ public function scheduleAgent($jobId, $uploadId, &$errorMsg, $request)
...
+      }
+
+      $reuseUploadPair = explode(',', $reuseSelection, 2);
+      if (count($reuseUploadPair) !== 2) {
+        error_log("Reuser: Invalid reuse selection format: '$reuseSelection' (expected format: 'uploadId,groupId')");
+        continue;
```
Kaushl2208: Similarly here. Log is good. To figure the issue out but also unwanted behaviour should go as an exception

src/reuser/ui/agent-reuser.php
sakshammishra112: Yes, make much more sense for the reuser to fail. Thanks.

src/reuser/ui/agent-reuser.php
sakshammishra112: Done

sakshammishra112:
Will add the agent test. @Kaushl2208 can I handle the Reuser scheduling using API's doc and controller in a separate PR?

sakshammishra112:
Thanks for the review and testing, @Kaushl2208! I really appreciate your time and feedback.

PR #3489, opened 2026-03-19
fix(proxy): keep UploadTreeProxy view names unique and reuse named params

### LINKED ISSUE #3378: UploadTreeProxy: Broken DB View Name instantiation and query parameter reuse

mandar1045:
## Summary
This is a clean, focused follow-up for #3378.

A previous attempt existed in #3379, but that PR was closed without being merged and maintainers asked for the correct isolated changes. This PR keeps the scope limited to the two `UploadTreeProxy` bugs from #3378 and adds regression coverage for them.

When `UploadTreeProxy` built a default DB view name, it accidentally discarded the option-specific suffix generated by `createUploadTreeViewQuery()`. That meant different proxy instances could collapse onto the same prepared statement / view name. This patch preserves the generated suffix and normalizes it so names like `lft,rgt` remain SQL-safe.

The second issue was in named parameter reuse. `addParamAndGetExpr()` checked for existing parameter keys, but new values were appended without preserving those keys, so the reuse path never actually worked. This patch stores parameters by name, reuses the intended placeholders, and keeps `countMaskedNonArtifactChildren()` compatible with the keyed parameter map.

## Tests
- Added regression checks for generated default view names
- Added regression checks for named parameter reuse across combined filters
- Added coverage for `countMaskedNonArtifactChildren()` with keyed params
- Ran `HOME=/tmp php -d xdebug.mode=off src/vendor/bin/phpunit --configuration src/phpunit.xml src/lib/php/Proxy/test/UploadTreeProxyTest.php`
- Result: `OK (20 tests, 31 assertions)`

Closes #3378

### DIFF
--- a/src/lib/php/Proxy/UploadTreeProxy.php
+++ b/src/lib/php/Proxy/UploadTreeProxy.php
@@ -45,8 +45,9 @@ class UploadTreeProxy extends DbViewProxy
   {
     $this->uploadId = $uploadId;
     $this->uploadTreeTableName = $uploadTreeTableName;
-    $dbViewName = $uploadTreeViewName ?: 'UploadTreeView'.(isset($this->dbViewName) ?: '');
     $dbViewQuery = $this->createUploadTreeViewQuery($options, $uploadTreeTableName);
+    $dbViewSuffix = preg_replace('/[^A-Za-z0-9_]+/', '_', $this->dbViewName ?? '');
+    $dbViewName = $uploadTreeViewName ?: 'UploadTreeView'.$dbViewSuffix;
     parent::__construct($dbViewQuery, $dbViewName);
   }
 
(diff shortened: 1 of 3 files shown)


Kaushl2208:
Changes are not relevant. i dont see a lot of value in here. Closing

PR #3497, opened 2026-03-21
perf(reuser): optimize license comparison via database-level matching

### LINKED ISSUE #3495: [Optimization] Implement Database-Level License Identity Matching in Reuser Agent
Description: 
Currently, the Enhanced Reuser agent primarily relies on file-content diffing to determine if licenses can be reused when file hashes do not match. While effective, this is a resource-intensive process involving high I/O for file downloads and CPU overhead for text comparison.

The Proposal: 
I am proposing an optimization to the areLicensesIdentical() logic to utilize a "Database Fast-Path."By querying the license_file table and comparing the aggregated sets of license IDs (rf_fk) for both the reused and the new pfile, we can determine "Legal Identity" without ever touching the actual file contents. Key Benefits: Instant Matching: Comparisons happen at the SQL level via array_agg, making the check O(1) from the agent's perspective. Reduced Overhead: Eliminates unnecessary file downloads and PHP-based text diffing for files that have the same license signatures. Scalability: Significantly speeds up Reuser jobs for large repositories where many files have minor code changes but identical licensing.

Technical Approach: 
Modify the SQL logic in ReuserAgent.php to correctly group and compare license references (rf_fk) to ensure the agent prioritizes metadata matching before falling back to expensive text-based analysis.

ayushmaan-16:
## Description

This PR implements a performance optimization for the **Enhanced Reuser** agent by enabling a "Database Fast-Path" for license identity checks.
### Changes

- Updated `areLicensesIdentical()` to utilize PostgreSQL metadata (`rf_fk`) instead of falling back to text-based diffing when file hashes differ.
- Implemented grouped identity checks to ensure that files with identical sets of licenses are cleared instantaneously.
- Reduced system I/O and CPU overhead by bypassing unnecessary file downloads and expensive PHP-based string comparisons.

## How to test

1. **Initial Analysis:** Upload a package (e.g., a small library) and run a full license analysis.
2. **Re-Upload:** Upload a slightly modified version of the same package (ensure some file hashes change, but license sets remain identical).
3. **Run Reuser:** Execute the Reuser agent with the "Enhanced" setting enabled.
4. **Verification:** - Check the `fossology-scheduler` logs: Observe that the agent identifies the license identity via the database "Fast-Path."
   
image

The image clearly shows the amount of times the fast path that i introduced is taken. 

Closes #3495

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -63,7 +63,7 @@ jobs:
         rm -rf src/vendor
 
     - name: Get CMake
-      uses: lukka/get-cmake@v4.1.2
+      uses: lukka/get-cmake@v4.3.0
 
     - name: Fetch tags
       run: |
@@ -169,7 +169,7 @@ jobs:
           libjson-c-dev librpm-dev libboost-all-dev libxslt1-dev
 
       - name: Setup CMake
-        uses: lukka/get-cmake@v4.1.2
+        uses: lukka/get-cmake@v4.3.0
 
       - name: Build Nomossa
         run: |


abdollahShahid:
@ayushmaan-16 maintainers accept only one commit per fix, please squash your commits to one

PR #3500, opened 2026-03-22
fix(report): exclude irrelevant files from SPDX and ReadmeOSS reports

### LINKED ISSUE #2853: Enhancement: Reporting
### Description

Reports 
a. More reports which do not include files marked as irrelevant into the reports, or allow the user to make a selection to include/exclude irrelevant files for all reports

pranavsnotebook-a11y:
## Description

Files marked as **Irrelevant** by the clearing workflow were not being excluded from SPDX and ReadmeOSS report output, even though the Unified (DOCX) and CLIXML agents already excluded them via `LicenseClearedGetter`.

This fix makes SPDX and ReadmeOSS consistent with the rest of the report agents.

Closes #2853

### Changes

- **`LicenseClearedGetter`**: add explicit `$excludeIrrelevant` flag (default `true`) so the always-on exclusion is intentional and testable via `setExcludeIrrelevant()`
- **`ReportUtils::getFilesWithLicensesFromClearings`**: add `$excludeIrrelevant` parameter (default `true`) that is passed through to `LicenseClearedGetter`
- **`spdx.php`**: fix `getSPDXReportConf()` to guard against undefined array index when `ri_spdx_selection` has fewer entries than expected
- Add unit tests for `LicenseClearedGetter::setExcludeIrrelevant()` and `ReportUtils` parameter signature

## How to test

1. Upload a package and scan it
2. Mark one or more files as **Irrelevant** via the clearing workflow
3. Generate an SPDX report — the irrelevant files should not appear in the output
4. Generate a ReadmeOSS report — same result
5. Verify that Unified (DOCX) report behaviour is unchanged
6. Run unit tests: `php vendor/bin/phpunit --no-coverage src/lib/php/Report/tests/`

### DIFF
--- a/src/lib/php/Report/LicenseClearedGetter.php
+++ b/src/lib/php/Report/LicenseClearedGetter.php
@@ -24,6 +24,8 @@ class LicenseClearedGetter extends ClearedGetterCommon
   private $onlyComments = false;
   /** @var Boolean */
   private $onlyAcknowledgements = false;
+  /** @var Boolean */
+  private $excludeIrrelevant = true;
   /** @var ClearingDao */
   private $clearingDao;
   /** @var LicenseDao */
@@ -54,7 +56,7 @@ class LicenseClearedGetter extends ClearedGetterCommon
     $licenseMap = new LicenseMap($dbManager, $groupId, LicenseMap::REPORT);
     $ungroupedStatements = array();
     foreach ($clearingDecisions as $clearingDecision) {
-      if ($clearingDecision->getType() == DecisionTypes::IRRELEVANT) {
+      if ($this->excludeIrrelevant && $clearingDecision->getType() == DecisionTypes::IRRELEVANT) {
         continue;
       }
       /** @var ClearingDecision $clearingDecision */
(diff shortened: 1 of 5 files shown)


OliverFendt:
Please note that even irrelevant marked files shall be contained in the OSSelot SPDX export. 
As far as I can see licenses from irrelevant files where never included in the Readme_OSS, but licenses which are marked as non functional are contained in the Readme_OSS which is not correct in my opinion

shaheemazmalmmd:
> Please note that even irrelevant marked files shall be contained in the OSSelot SPDX export. As far as I can see licenses from irrelevant files where never included in the Readme_OSS, but licenses which are marked as non functional are contained in the Readme_OSS which is not correct in my opinion

Noted.

pranavsnotebook-a11y:
Thanks for the clarification @OliverFendt\! That makes sense — OSSelot SPDX exports have their own requirements and should keep irrelevant files. I'll scope this PR down to only exclude irrelevant files from **ReadMe_OSS** and standard SPDX reports, and leave the OSSelot SPDX export untouched. I'll also investigate the non-functional files behavior in ReadMe_OSS you mentioned and address that as part of this fix.

shaheemazmalmmd:
@pranavsnotebook-a11y : The changes are not acceptable as the configurations needs to be done in UI conf page for excluding, cannot be set directly in a class, please join the community call for further discussion on this.

Issue #1943, opened 2021-04-01
Non admin User delete their own upload files.


KoukiHama:
### Description

Currently, even if you are the one who uploaded the file, you cannot delete the file unless he is an admin user. 
The function of setting that allows the user to delete their own upload files is useful.

I got comments about this in [slack.](https://fossology.slack.com/archives/C6NH0BY5N/p1617256046110600?thread_ts=1617111566.104400&cid=C6NH0BY5N)

> That will be nice addition.
You can start by changing the required permission to Auth::PERM_WRITE at https://github.com/fossology/fossology/blob/aaec8d420e0d58f4e7663a9694e1fae44b42def6/src/delagent/ui/Page/AdminUploadDelete.php#L53
And making sure only editable uploads are added to https://github.com/fossology/fossology/blob/aaec8d420e0d58f4e7663a9694e1fae44b42def6/src/delagent/ui/Page/AdminUploadDelete.php#L115 （編集済み）

badboyguss:
Can I work on this issue?

Shivangisriva:
Hi, I would like to work on this issue.?

Shivangisriva:
Hi @shaheemazmalmmd, @GMishx, 
@Kaushl2208 
I’d like to work on [Issue #1943] regarding deletion of uploads.
My proposed approach is:

1) Change access for admin_upload_delete from Auth::PERM_ADMIN → Auth::PERM_WRITE.

2) Keep TryToDelete()’s isEditable() check as the final authorization.

3)Show only editable uploads in the delete dropdown (including folder-change AJAX results).

4) Add/update tests for non-admin writable users and permission enforcement.
Could you confirm if this approach is acceptable?
Also, should browse_upload_delete be moved from admin-only to write permission for consistency?

Shivangisriva:
Ok now as i have the permission from @shaheemazmalmmd , i am proceeding with this issue.

shaheemazmalmmd:
Added: https://github.com/fossology/fossology/pull/3578

PR #3507, opened 2026-03-24
feat(upload): allow users to delete their own uploads

Shivangisriva:
#1943  
* Description
This PR updates upload deletion permissions so non-admin users with write access can delete uploads they are authorized to edit, while preserving final permission enforcement in backend delete logic.

Previously, the delete upload UI entry points were admin-only, which prevented regular write users from deleting their own uploads even when they had edit rights.

This change aligns UI access with existing upload editability rules and keeps security checks intact.

 * What Changed
Changed required permission for admin_upload_delete from admin to write.
Kept TryToDelete editability check as the final authorization gate.
Filtered delete dropdown contents to show only editable uploads.
Applied the same editable filtering for folder-change AJAX results.
Updated browse_upload_delete permission from admin to write for consistency.
Added controller-level tests for delete success and permission-denied paths.

* Behavior After Change
Non-admin users with write access can open Delete Uploaded File.
Only uploads editable by the current group/user context are listed.
Forced/manual delete attempts on non-editable uploads are denied by backend permission checks.
Browse delete path follows the same write-level access policy.

The deletion option for non admin users -
image

### DIFF
--- a/src/delagent/ui/Page/AdminUploadDelete.php
+++ b/src/delagent/ui/Page/AdminUploadDelete.php
@@ -39,7 +39,7 @@ class AdminUploadDelete extends DefaultPlugin
     parent::__construct(self::NAME, array(
         self::TITLE => _("Delete Uploaded File"),
         self::MENU_LIST => "Organize::Uploads::Delete Uploaded File",
-        self::PERMISSION => Auth::PERM_ADMIN,
+      self::PERMISSION => Auth::PERM_WRITE,
         self::REQUIRES_LOGIN => true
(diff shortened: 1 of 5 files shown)


shaheemazmalmmd:
@Shivangisriva : Please note that this PR contains some changes that are unrelated to the current task. These appear to have been carried over from previous work. Please remove and reopen the PR.

Shivangisriva:
Okay sure , i will do it.

https://github.com/fossology/fossology/pull/3511
kindly check it .

PR #3514, opened 2026-03-26
fix(lib): add null-safety guards in TreeDao::getItemHashes and getParentOfItem

mandar1045:
### Summary

Add null-safety guards in `TreeDao::getItemHashes()` and `getParentOfItem()`.

### Problem

* Both methods call `getSingleRow()` and immediately access array keys
* When no row is found, `getSingleRow()` returns `false`
* Causes:

  * PHP 8: `TypeError`
  * Older PHP: silent null values → hard-to-debug issues

### Fix

* `getItemHashes()`: throw descriptive exception when no row is found
* `getParentOfItem()`: return `null` safely if no result

### Impact

* Prevents runtime crashes in PHP 8
* Makes behavior explicit and predictable
* Aligns with existing handling in `TreeDao::getFullPath()`

### Scope

* Small, isolated change (~50 LOC)
* Limited to `TreeDao`
* No schema or API changes

### Tests

Added tests in `TreeDaoTest`:

* Exception thrown for missing item in `getItemHashes()`
* Correct parent returned for existing item
* `null` returned for missing item in `getParentOfItem()`

---

Happy to adjust behavior if a different error-handling approach is preferred 👍

### DIFF
--- a/src/lib/php/Dao/TreeDao.php
+++ b/src/lib/php/Dao/TreeDao.php
@@ -119,6 +119,9 @@ class TreeDao
   {
     $pfile = $this->dbManager->getSingleRow("SELECT pfile.* FROM $uploadtreeTablename, pfile WHERE uploadtree_pk=$1 AND pfile_fk=pfile_pk",
         array($uploadtreeId), __METHOD__);
+    if ($pfile === false) {
+      throw new \Exception("Could not find pfile for uploadtree item $uploadtreeId");
+    }
     return array('sha1'=>$pfile['pfile_sha1'],'md5'=>$pfile['pfile_md5'],'sha256'=>$pfile['pfile_sha256']);
   }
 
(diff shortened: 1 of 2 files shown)


src/lib/php/Dao/TreeDao.php
```diff
@@ -119,6 +119,9 @@ public function getItemHashes($uploadtreeId, $uploadtreeTablename='uploadtree')
   {
     $pfile = $this->dbManager->getSingleRow("SELECT pfile.* FROM $uploadtreeTablename, pfile WHERE uploadtree_pk=$1 AND pfile_fk=pfile_pk",
         array($uploadtreeId), __METHOD__);
+    if ($pfile === false) {
+      throw new \Exception("Could not find pfile for uploadtree item $uploadtreeId");
+    }
```
its-sushant, line 124: Callers have no try/catch, will now crash on missing pfile instead of degrading gracefully.

its-sushant:
Changes not required as of now. Please check the inline comment and feel free  to reopen with proper handling.

PR #3517, opened 2026-03-26
feat(reuser): implement auto-suggest and multi-baseline selection

ayushmaan-16:
Description
Ok so this PR introduces an auto-suggestive mechanism to select reference packages for clearing purposes in the reuse menu. This would suggest the closest packages to the upload in order.
Also as of now the frontend allows us to select multiple packages for the upload but it send the backend only one package because of this the clearing decisions are applied w.r.t that file only. This was happening becauseof the following 3 reasons:

1. HTML Name: The select element was named uploadToReuse (no brackets). In a standard form submission, the browser only sends a single scalar value for a name like this.

2. No multiple Attribute: Without the multiple="multiple" attribute, the browser's UI only allowed one selection at a time.

3. Backend Expectation: In agent-reuser.php, the original code used: $uploadToReuse = $request->get('uploadToReuse'); This strictly expected a single string or ID. If an array had been sent (e.g., if you had hacked the frontend), this line would typically only grab the last item or error out when passed into createPackageLink (the function that creates the "family tree" between your new upload and your reference package).

Changes

1. Auto-Suggest:
Added a search tool that extracts the "Component Name" from the current upload (stripping version numbers/extensions) and queries the database for matching cleared uploads.
Displays matches with exact upload timestamps so users can distinguish between similar release versions. Limits this to the top 10 previously uploaded packages in descending order of their timestamps

2. Multi-Baseline Support:
By changing the name to uploadToReuse[] and adding multiple="multiple", the payload was transformed from a single ID into a list of IDs.

How to test
Navigate to the Upload -> From File page.
Select a file (e.g., zlib-1.2.11.tar.gz).
Click on the "Set the reuse information" button to open the configuration modal.
In the "Auto-Suggest Baseline" section, click "Search matching uploads".
Verify: It should find any previous uploads of zlib and display them with their timestamps.
Select one or more items from the auto-suggest list.
Complete the upload.
Verify: Check the scheduler logs or database to see that multiple upload_reuse entries were created and processed.

### DIFF
--- a/src/lib/php/Dao/UploadDao.php
+++ b/src/lib/php/Dao/UploadDao.php
@@ -978,10 +978,12 @@ ORDER BY lft asc
   {
     $groupId = \Fossology\Lib\Auth\Auth::getGroupId();
 
-    $sql = "SELECT u.upload_pk AS id, u.upload_filename AS name, u.upload_ts AS date 
+    $sql = "SELECT u.upload_pk AS id, u.upload_filename AS name, MAX(ce.date_added) AS date 
                 FROM upload u
                 INNER JOIN upload_clearing uc ON uc.upload_fk = u.upload_pk AND uc.group_fk = $1
                 INNER JOIN foldercontents fc ON fc.child_id = u.upload_pk AND fc.foldercontents_mode = 2 
+                INNER JOIN uploadtree ut ON ut.upload_fk = u.upload_pk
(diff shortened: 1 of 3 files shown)


shaheemazmalmmd:
@ayushmaan-16 : Changes not working from my end. if you are targeting GSoC then please submit the proposal.

PR #3519, opened 2026-03-27
perf(reuser): optimize copyright reuse via hash map lookup

### LINKED ISSUE #3518: [perf] Performance slowdown during copyright reuse on large packages
While using the Reuser agent on larger uploads, I noticed that the process takes a very long time to complete—sometimes much longer than expected for just copying clearing decisions. After exploring the code for ReuserAgent.php, I found the reason for this slowdown.

The Technical Cause:

In the reuseCopyrights function, there is a nested foreach loop that compares every single copyright in the current upload against every copyright in the reference upload.

Image

This creates an O(N × M) complexity issue.

For example:
If you have 1,000 copyright entries in each upload, fossology has to perform 1,000,000 comparisons.

This makes the agent incredibly slow for any package with many files and legal headers.

The Proposed Fix:

I have figured out a fix that replaces this nested loop with a hash map lookup. By pre-indexing the copyrights by their hash, we can change the complexity from O(1,000,000) to just O(1,000)—basically from O(N × M) to O(N + M).

ayushmaan-16:
## Description
This PR fixes a performance bottleneck in the copyright reuse engine. The existing nested comparisons were causing unnecessary slowdowns for large packages.

Fixes #3518

### Changes
- Optimized [reuseCopyrights](cci:1://file:///c:/Users/ayush/fossology/src/reuser/agent/ReuserAgent.php:186:2-244:3) in [ReuserAgent.php](cci:7://file:///c:/Users/ayush/fossology/src/reuser/agent/ReuserAgent.php:0:0-0:0) by switching from a nested loop to a hash map lookup (O(N+M)).
- Reduces the time needed to reuse copyrights for large uploads from minutes to a few seconds.

## How to test
1. Upload a package and clear some copyrights (enable/disable).
2. Upload a second version of that package.
3. Run the Reuser using the first upload as a reference.
4. Verify that the clearing decisions correctly appear in the new upload's "Copyrights" view and that the agent completes much faster than before (this can mostly be observed for the larger packages).

### DIFF
--- a/src/reuser/agent_tests/Unit/ReuserAgentTest.php
+++ b/src/reuser/agent_tests/Unit/ReuserAgentTest.php
@@ -10,7 +10,7 @@ namespace Fossology\Reuser;
 // Use relative paths to ensure the test works in any FOSSology environment (Local or CI)
 $autoloadPath = __DIR__ . '/../../../../vendor/autoload.php';
 if (file_exists($autoloadPath)) {
-    require_once $autoloadPath;
+  require_once $autoloadPath;
 }
 require_once __DIR__ . '/../../agent/ReuserAgent.php';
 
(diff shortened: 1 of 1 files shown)


Kaushl2208:
Hey @ayushmaan-16 , Did you do any benchmarking in the performance improvement???
Also a valid Proof of Work would be good.

PR #3526, opened 2026-03-27
[Perf] copyright histogram optimization and ensuring access to all the filters from the UI

### LINKED ISSUE #3525: [feat/perf] Optimize copyright list with SQL pagination, eliminate database queries inside loops, and add UI filters
### 1. Description of Issues
The current Copyright List view ([list.php](cci:7://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:0:0-0:0)) deals with performance issues and lacks frontend UI coverage for several backend filter states:

* **Lack of Pagination** When a user views the files for a highly frequent copyright statement, the current code fetches **all** matching copyright records into PHP memory at once. It then uses PHP loops to manually deduplicate them before displaying the first 50 rows causing memory exhaustion on large packages.
* **Database Queries inside a loop:** When filtering the list to show files without licenses (`&filter=nolic`), the [GetRequestedRows()](cci:1://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:145:2-239:3) PHP function loops through every single file and executes a brand new SQL query for each individual file to check if a license is attached. If there are 10,000 files, it runs 10,000 separate SQL queries back-to-back.
* **Missing UI Dropdown:** The backend functionality technically supports filtering these files by `active`, `inactive`, and `nolic` (No License Found) states, but there is no dropdown menu on the [list.php](cci:7://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:0:0-0:0) frontend allowing users to actually select these filters, forcing them to manually type the parameters into the URL.

### 2. Proposed Solution
This PR resolves these three major issues by migrating the heavy lifting from PHP loops into optimized PostgreSQL operations:

* **Implement SQL Pagination:** Move the pagination logic entirely into the database using `LIMIT` and `OFFSET` in the SQL query. This prevents PHP from downloading thousands of rows into memory, ensuring only 50 rows are ever fetched at a time.
* **Eliminate database Queries inside a loop:** Delete the repetitive `for` loop queries inside [GetRequestedRows()](cci:1://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:145:2-239:3). Instead, use an efficient `INNER JOIN license_file` directly within the main `WHERE` clause so the database filters all out-of-scope files simultaneously in a single query execution.
* **Add Frontend UI Dropdown:** Inject an HTML `<select>` dropdown menu into the [list.php](cci:7://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:0:0-0:0) UI, containing "Show all", "Show active", "Show inactive", and "Show files without licenses" options so users can seamlessly trigger the backend filters.

### 3. Steps to Reproduce the Current Problem
1. Navigate to a Copyright Histogram for a large repository.
2. Click on a highly frequent copyright statement to view its files.
3. Observe the extreme page load time (and potential memory exhaustion) due to PHP fetching and looping over every single file occurrence before rendering the first page.
4. Attempt to filter the list to show only active copyrights or files with no licenses. Notice there is no native dropdown on the page to do so.
5. Manually type `&filter=nolic` into the URL and observe the massive freeze as the server executes thousands of consecutive queries to check each file's license status.

ayushmaan-16:
## Description

This PR significantly optimizes the performance of the Copyright List view ([list.php](cci:7://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:0:0-0:0)) by eliminating its severe database queries inside a loop and PHP memory exhaustion on large packages. It also officially exposes the backend filter options (active, inactive, and "no license found" checks) to the frontend via a new UI dropdown menu so users can actually use them without manually changing the URL.

### Changes

- **Implemented SQL Pagination:** Dropped the legacy approach of fetching thousands of matching records into PHP memory to deduplicate them. Pagination checks are now handled natively in PostgreSQL using `LIMIT` and `OFFSET`, so only the required 50 rows are ever fetched at a time.
- **Eliminated database Queries inside loops:** Replaced the highly repetitive PHP `for` loop queries inside [GetRequestedRows()](cci:1://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:145:2-239:3) (which previously fired off a new database check for every single file when the `nolic` filter was applied) with a highly efficient `INNER JOIN license_file` directly inside the main SQL `WHERE` clause. 
- **Added Missing Frontend Dropdown:** Injected an HTML `<select>` filter menu into the [list.php](cci:7://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:0:0-0:0) interface. Users can now seamlessly toggle between "Show all", "Show active", "Show inactive", and "Show files without licenses", updating the URL parameters automatically.

## How to test

1. Spin up the FOSSology environment and navigate to a Copyright Histogram for any large repository/upload.
2. Click the count link on a highly frequent copyright statement to open the [list.php](cci:7://file:///c:/Users/ayush/fossology/src/copyright/ui/list.php:0:0-0:0) view. Note the immediate page load times compared to the old PHP-loop implementation.
3. Validate the text at the top reads correctly (e.g. `X copyright instances found in Y files`).
4. Look for the new "Filter:" dropdown menu directly above the list. Change the option to "Show files without licenses" and verify that it quickly loads.

Closes #3525

### DIFF
--- a/src/copyright/ui/list.php
+++ b/src/copyright/ui/list.php
@@ -84,7 +84,7 @@ class copyright_list extends FO_Plugin
    * @throws Exception
    * @return array Rows to process, and $upload_pk
    */
-  function GetRows($Uploadtree_pk, $Agent_pk, &$upload_pk, $hash, $type, $tableName, $filter="")
+  function GetRows($Uploadtree_pk, $Agent_pk, &$upload_pk, $hash, $type, $tableName, $filter="", $limit=0, $offset=0)
   {
     /*******  Get license names and counts  ******/
     $row = $this->uploadDao->getUploadEntry($Uploadtree_pk);
@@ -93,13 +93,29 @@ class copyright_list extends FO_Plugin
     $upload_pk = $row["upload_fk"];
     $params = [];
 
+    $join = "";
+    $filter_query = "";
+    if ($filter == "nolic") {
+      $noLicStr = "No_license_found";
+      $voidLicStr = "Void";
+      $join = " INNER JOIN license_file AS LF on SS.PF = LF.pfile_fk ";
+      if ($type != "copyFindings") {
+        $join = " INNER JOIN license_file AS LF on cp.pfile_fk = LF.pfile_fk ";
+      }
+      $filter_query = " AND LF.rf_fk IN (SELECT rf_pk FROM license_ref WHERE rf_shortname IN ('$noLicStr', '$voidLicStr')) ";
+    }
+
+    $limit_clause = ($limit > 0) ? " LIMIT $".(count($params)+7) : "";
+    $offset_clause = ($offset > 0) ? " OFFSET $".(count($params)+8) : "";
+
     if ($type == "copyFindings") {
-      $sql = "SELECT textfinding AS content, '$type' AS type, uploadtree_pk, ufile_name, PF, hash
+      $sql = "SELECT DISTINCT ON (uploadtree_pk) textfinding AS content, '$type' AS type, uploadtree_pk, ufile_name, PF, hash
               FROM $tableName,
               (SELECT uploadtree_pk, pfile_fk AS PF, ufile_name FROM uploadtree
                  WHERE upload_fk=$1
                    AND uploadtree.lft BETWEEN $2 AND $3) AS SS
-              WHERE PF=pfile_fk AND hash=$4 ORDER BY uploadtree_pk";
+              $join
+              WHERE PF=pfile_fk AND hash=$4 $filter_query ORDER BY uploadtree_pk $limit_clause $offset_clause";
       $params = [
         $upload_pk, $lft, $rgt, $hash
       ];
@@ -111,12 +127,12 @@ class copyright_list extends FO_Plugin
       if (!empty($filter)) {
         if ($filter == "active") {
           $active_filter = "AND (ce.is_enabled IS NULL OR ce.is_enabled = 'true')";
-        } elseif ($filter = "inactive") {
+        } elseif ($filter == "inactive") {
           $active_filter = "AND ce.is_enabled = 'false'";
         }
       }
       /* get all the copyright records for this uploadtree.  */
-      $sql = "SELECT
+      $sql = "SELECT DISTINCT ON (uploadtree_pk)
 (CASE WHEN (ce.content IS NULL OR ce.content = '') THEN cp.content ELSE ce.content END) AS content,
 (CASE WHEN (ce.hash IS NULL OR ce.hash = '') THEN cp.hash ELSE ce.hash END) AS hash,
 type, uploadtree_pk, ufile_name, cp.pfile_fk AS PF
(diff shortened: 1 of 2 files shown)


src/copyright/ui/list.php
```diff
@@ -93,13 +93,29 @@ function GetRows($Uploadtree_pk, $Agent_pk, &$upload_pk, $hash, $type, $tableNam
...
+      $voidLicStr = "Void";
+      $join = " INNER JOIN license_file AS LF on SS.PF = LF.pfile_fk ";
+      if ($type != "copyFindings") {
+        $join = " INNER JOIN license_file AS LF on cp.pfile_fk = LF.pfile_fk ";
+      }
+      $filter_query = " AND LF.rf_fk IN (SELECT rf_pk FROM license_ref WHERE rf_shortname IN ('$noLicStr', '$voidLicStr')) ";
+    }
+
+    $limit_clause = ($limit > 0) ? " LIMIT $".(count($params)+7) : "";
+    $offset_clause = ($offset > 0) ? " OFFSET $".(count($params)+8) : "";
```
Kaushl2208, line 109: What is the significance of `+7` and `8` here?

src/copyright/ui/list.php
```diff
@@ -293,19 +325,16 @@ function Output()
...
+    $Offset = $Page * $Max;
+    $rows = $this->GetRows($uploadtree_pk, $agent_pk, $upload_pk, $hash, $type, $tableName, $filter, $Max, $Offset);
     $uploadtree_tablename = $this->uploadDao->getUploadtreeTableName($upload_pk);
 
-    /* slim down to all rows with this hash and type,  and filter */
-    $NumInstances = 0;
-    $rows = $this->GetRequestedRows($allRows, $excl, $NumInstances, $filter, $hash);
-
-    // micro menus
-    $OutBuf .= menu_to_1html(menu_find($this->Name, $MenuDepth),0);
```
Kaushl2208, line 306: Why is this removed???

Kaushl2208 (changes requested):
PTAL

src/copyright/ui/list.php
ayushmaan-16: my bad i tried to handle them manually. I've fixed this now, we calculate the index based on the actual size of the $params array at runtime.

src/copyright/ui/list.php
ayushmaan-16: my bad again, i accidently removed this part. I've fixed it now though

ayushmaan-16:
@Kaushl2208 I've made the changes just like you asked. Also I fixed the sorting problem, earlier the sorting was happening after fetching a page of results. I've now moved the sorting part to SQL.

ayushmaan-16:
@Kaushl2208 Any updates on this ?

shaheemazmalmmd:
@ayushmaan-16 : thank you for your efforts on this.
image

Adding a filter here does'nt add any value because most of the filters will be used in the HIST view page rather than LIST. But thanks for your efforts.

PR #3530, opened 2026-03-29
fix(application): guard against empty delimiter/enclosure in LicenseCsvImport

### LINKED ISSUE #3485: Empty/Short String Handling in CSV Import
### Summary
There is a bug in `LicenseCsvImport.php` where `substr()` is used on user-provided input without validating for empty strings. When an empty string is passed, `substr()` returns `false` instead of a string, which leads to unexpected behavior in CSV parsing.

### Affected File
- `src/lib/php/Application/LicenseCsvImport.php`

Input | substr Result | Impact
-- | -- | --
"" | false | fgetcsv() fails or behaves unexpectedly
null | false | Same issue

### Problematic Code
```php
public function setDelimiter($delimiter=',')
{
  $this->delimiter = substr($delimiter,0,1);
}

pranavsnotebook-a11y:
## Description

`LicenseCsvImport` crashed with a fatal error when the CSV delimiter or enclosure field was left empty. The `str_getcsv()` call requires a non-empty single character for both arguments, but no validation existed before this change.

Closes #3485

### Changes

- Added a guard at the start of `handleFile()` that throws `\UnexpectedValueException` if delimiter or enclosure is empty or longer than one character
- No changes to existing import logic

## How to test

1. Navigate to **Admin → License Administration → Import CSV**
2. Leave the **Delimiter** field empty and submit a valid CSV file
3. Before this fix: fatal error. After this fix: a clear error message is shown
4. Repeat with the **Enclosure** field empty — same result
5. Normal import with valid delimiter (`,`) and enclosure (`"`) should work as before

### DIFF
--- a/src/lib/php/Application/LicenseCsvImport.php
+++ b/src/lib/php/Application/LicenseCsvImport.php
@@ -79,6 +79,9 @@ class LicenseCsvImport
    */
   public function setDelimiter($delimiter=',')
   {
+    if (empty($delimiter)) {
+      return;
+    }
     $this->delimiter = substr($delimiter,0,1);
   }
 
@@ -88,6 +91,9 @@ class LicenseCsvImport
    */
   public function setEnclosure($enclosure='"')
   {
+    if (empty($enclosure)) {
+      return;
+    }
     $this->enclosure = substr($enclosure,0,1);
   }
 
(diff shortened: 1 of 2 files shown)


pranavsnotebook-a11y:
Hi @shaheemazmalmmd, this one fixes a fatal crash when the CSV delimiter or enclosure is left empty in `LicenseCsvImport`. Includes a test case. CI is passing — would appreciate a review when you have a moment!

pranavsnotebook-a11y:
Hi @shaheemazmalmmd — gentle ping on this one. It fixes a fatal crash when CSV delimiter or enclosure is left empty in LicenseCsvImport, includes a test case, and CI is passing. Would appreciate a review when you have a moment!

Kaushl2208:
Hey @pranavsnotebook-a11y , Instead updating the setter for delimiter and enclosure. Look into the page: `AdminLicenseFromCsv`, The empty check should be handled there before the calls.

Fix it properly. I am closing this PR for now. Expecting a correct fix.

PR #3531, opened 2026-03-29
fix(buckets): replace sprintf with snprintf to prevent buffer overrun

### LINKED ISSUE #3470: Unsafe use of sprintf in buckets agent may cause buffer overflow
While reviewing the `buckets` agent (`src/buckets/agent/buckets.c`), I found multiple uses of `sprintf` to construct SQL queries and strings.

Example:

`sprintf(sqlbuf, "select bucketpool_pk from bucketpool where bucketpool_pk=%d and active='Y'", bucketpool_pk);`

`sprintf` does not enforce buffer size limits, which can potentially lead to buffer overflows if the formatted output exceeds the allocated buffer.

### Impact
- Potential buffer overflow vulnerabilities
- Reduced memory safety
- Inconsistent use of safe string handling functions across the codebase

### Suggested Improvement

Replace `sprintf` with `snprintf`, which ensures that writes do not exceed buffer boundaries.

Example fix:

`snprintf(sqlbuf, sizeof(sqlbuf),
         "select bucketpool_pk from bucketpool where bucketpool_pk=%d and active='Y'",
         bucketpool_pk);`

### Scope

`src/buckets/agent/buckets.c`
Multiple occurrences of `sprintf` can be safely replaced with `snprintf`

### Notes

This change does not affect functionality but improves security and robustness.

I want to work on this issue and submit a PR. Please let me know if this approach is acceptable or if there are preferred guidelines for string handling in this project.

pranavsnotebook-a11y:
## Summary

- All `sprintf(sqlbuf, ...)` and `sprintf(agent_rev, ...)` calls in `src/buckets/agent/buckets.c` are replaced with `snprintf(buf, sizeof(buf), ...)`.
- This bounds output to the declared buffer size, eliminating potential stack-buffer-overrun undefined behaviour when inputs are longer than expected.
- No logic change — purely a defensive bounds check on each format call.

Fixes #3470

## Test plan

- [x] Reviewed each replaced call: buffer size is `sizeof(sqlbuf)` / `sizeof(agent_rev)` — consistent with the existing declarations.
- [x] No existing C unit-test suite for the buckets agent; change is mechanical and non-behavioural.

## Proof

sprintf → snprintf diff across all 10 call sites

### DIFF
--- a/src/buckets/agent/buckets.c
+++ b/src/buckets/agent/buckets.c
@@ -132,7 +132,7 @@ int main(int argc, char **argv)
             ReadFromStdin = 0;
             bucketpool_pk = atoi(optarg);
             /* validate bucketpool_pk */
-            sprintf(sqlbuf, "select bucketpool_pk from bucketpool where bucketpool_pk=%d and active='Y'", bucketpool_pk);
+            snprintf(sqlbuf, sizeof(sqlbuf), "select bucketpool_pk from bucketpool where bucketpool_pk=%d and active='Y'", bucketpool_pk);
             bucketpool_pk = validate_pk(pgConn, sqlbuf);
             if (!bucketpool_pk)
               printf("%d is not an active bucketpool_pk.\n", atoi(optarg));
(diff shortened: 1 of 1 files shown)


ayushmaan-16:
@pranavsnotebook-a11y  The least you could have done is remove "🤖 Generated with [Claude Code](https://claude.com/claude-code)"

PR #3535, opened 2026-03-30
perf(search): reduce database query overhead on search page

ayushmaan-16:
Description:
This pull request makes the search page load much faster by changing how it checks for user permissions. Instead of asking the database about every single upload one by one, it now gets all the information in one single step.

This is what was happening previously:
image

Changes
Updated search.php: I removed the old code that used a loop to check permissions for every upload. This was causing a lot of extra work for the database.
Improved UploadDao.php: I added a new function called getAccessibleUploads. It uses a single SQL query with a "JOIN" to find all the uploads a user is allowed to see.

image

Admin Optimization: If you are logged in as an administrator, the code now skips the extra permission checks and shows all uploads immediately to save time.
Reduced Database Load: This change moves the filtering logic from the PHP code to the database itself, which is much more efficient.
How to test
Log in to Fossology as a normal user.
Go to the Search page from the main menu.
Open the Upload dropdown list. You should only see uploads that belong to your group or ones that are marked as public.
Log in as an Admin and go to the same page. You should be able to see every upload in the system.
If your database has many uploads, you will see that the Search page now opens significantly faster than it did before.

### DIFF
--- a/src/lib/php/Dao/UploadDao.php
+++ b/src/lib/php/Dao/UploadDao.php
@@ -11,6 +11,7 @@ namespace Fossology\Lib\Dao;
 use Fossology\Lib\Data\Tree\Item;
 use Fossology\Lib\Data\Tree\ItemTreeBounds;
 use Fossology\Lib\Data\Upload\Upload;
+use Fossology\Lib\Auth\Auth;
 use Fossology\Lib\Data\Upload\UploadEvents;
 use Fossology\Lib\Data\UploadStatus;
 use Fossology\Lib\Db\DbManager;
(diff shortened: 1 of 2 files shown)


src/lib/php/Dao/UploadDao.php
```diff
@@ -145,6 +145,29 @@ public function getActiveUploadsArray()
...
+
+    $queryResult = $this->dbManager->getRows($sql, array($groupId), $stmt);
+
+    $results = array();
+    foreach ($queryResult as $row) {
+      $results[] = Upload::createFromTable($row);
+    }
+
+    return $results;
+  }
```
its-sushant, line 171: We should not ignore the session/auth guard which you can find inside `isAccessible`. It would result in unauthenticated users getting results they shouldn't.

And please use `Auth::PERM_NONE` constant instead of hardcoded 0 in SQL.

ayushmaan-16:
@its-sushant i've fixed the things exactly how you've asked.

shaheemazmalmmd:
@ayushmaan-16 : please squash your commits to one.

ayushmaan-16:
@shaheemazmalmmd Done

PR #3532, opened 2026-03-29
fix(buckets): replace sprintf with snprintf to prevent buffer overrun

pranavsnotebook-a11y:
## Description

Replaces all `sprintf()` calls in `src/buckets/agent/buckets.c` with `snprintf()` to prevent potential stack-buffer-overrun undefined behaviour.

Closes #3470

### Changes

- 4 `sprintf` → `snprintf(buf, sizeof(buf), ...)` replacements in `buckets.c`
- Affected buffers: `sqlbuf` (SQL query construction) and `agent_rev` (version string)
- Consistent with the same fix already applied to the nomos agent in #3417

## How to test

1. Build the buckets agent: `make -C src/buckets`
2. Confirm compilation succeeds with no new warnings
3. Run an existing scan that exercises the buckets agent and verify results are unchanged

PR #3536, opened 2026-03-30
fix(nomos):replaced sprintf with snprintf to prevent buffer overflow

FrankOHara43:
## Description

Replaced unsafe uses of s`printf` with `snprintf` to prevent potential buffer overflow issues during string formatting.

### Changes

- Replaced `sprintf` with `snprintf` in:
    - File path construction in `list_dir`
    - Agent version string formatting in `main`
- All formatted strings should follow the size of the destination buffer using `sizeof()`

### DIFF
--- a/src/nomos/agent/nomos.c
+++ b/src/nomos/agent/nomos.c
@@ -157,7 +157,19 @@ void list_dir (const char * dir_name, int process_count, int *distribute_count, 
   {
     /* get the file path, form the file path /dir_name/file_name,
        e.g. dir_name is '/tmp' file_name is 'test_file_1.txt', form one path '/tmp/test_file_1.txt' */
-    snprintf(filename_buf, sizeof(filename_buf), "%s/%s", dir_name, dirent_handler->d_name);
+    int snprintf_ret = snprintf(filename_buf, sizeof(filename_buf), "%s/%s", dir_name, dirent_handler->d_name);
+    if (snprintf_ret < 0)
+    {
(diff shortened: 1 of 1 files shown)


src/nomos/agent/nomos.c
```diff
@@ -157,7 +157,7 @@ void list_dir (const char * dir_name, int process_count, int *distribute_count,
   {
     /* get the file path, form the file path /dir_name/file_name,
        e.g. dir_name is '/tmp' file_name is 'test_file_1.txt', form one path '/tmp/test_file_1.txt' */
-    sprintf( filename_buf , "%s/%s",dir_name, dirent_handler->d_name);
+    snprintf(filename_buf, sizeof(filename_buf), "%s/%s", dir_name, dirent_handler->d_name);
```
Kaushl2208: I totally understand the `snprintf` change against `sprintf`. But now I have a very generic question:

We are considerate about buffer overflow in a dynamic scenario. snprintf is just a way to solve this (because it will restrict the `filename_buf` to a size we have defined. What happens when it truncates the `filename_buf`?? How are we handling that?? Because truncated output means the whole system ahead will fail drastically and without proper understanding it will be very difficult to find out what happened.

FrankOHara43:
@Kaushl2208 please take a look at the changes made.

its-sushant:
There is already a similar PR: https://github.com/fossology/fossology/pull/3532

CC: @Kaushl2208

PR #3543, opened 2026-03-31
fix(ui): allow clearing upload description and browse comment

### LINKED ISSUE #3002: Bug: Unable to clear upload description and comment
## Description
When attempting to clear the upload description and comment fields, they remain populated instead of resetting to an empty state. This issue occurs even after manually deleting the text.

## Steps to Reproduce
- Upload Description
  - Goto `Organize` -> `Uploads` -> `Edit Properties`.
  - Attempt to clear the text in the Upload Description field and click on `Edit` button.
  - Observe that the Description retains the previous text instead of being cleared.
- Upload Comment
  - Goto `Browse` tab.
  - Locate an upload and double-click on the Comment cell to edit it.
  - Clear the text in the Comment modal.
  - Click `OK` to save the changes.
  - Observe that the Comment field retains the previous text instead of being cleared.

## Expected Behavior
The upload description and comment fields should reset to an empty state when cleared manually.

## Actual Behavior
The fields retain the previously entered text, preventing users from properly resetting them.

### Reference

https://github.com/user-attachments/assets/f7aad499-e3f4-4c29-9b29-490b84dfe1d5

https://github.com/user-attachments/assets/ae442892-0797-4b36-8201-01b2430c1969

VISHNU0906:
## Description

Fixes #3002

Users could not clear the upload description (Edit Properties) or the browse status comment — the fields would silently retain the previous text.

### Root cause

Both code paths used PHP's `empty()` to check whether the user supplied a value. Since `empty("")` is `true`, an empty string was treated as *no input*, and the database update was skipped.

### Changes

| File | Before | After |
|------|--------|-------|
| `src/www/ui/admin-upload-edit.php` | `if (! empty($newDesc))` guarded the UPDATE | Guard removed; the description column is always updated with the submitted value. The existing early-return (`empty($newName) and empty($newDesc)` → return 2) still prevents spurious writes on non-POST page loads. |
| `src/www/ui/async/AjaxBrowse.php` | `!empty($commentText)` in the condition chain | Changed to `!is_null($commentText)`. Symfony's `Request::get()` returns `null` when the parameter is absent, so this correctly distinguishes *missing* from *empty*. |

### How to test

1. **Upload description** — Go to *Organize → Uploads → Edit Properties*, select an upload with a description, clear the text, click *Edit*. The description should now be empty.
2. **Browse comment** — Go to the *Browse* tab, double-click the Comment cell of an upload, clear the text, click *OK*. The comment should now be empty.

### DIFF
--- a/src/www/ui/admin-upload-edit.php
+++ b/src/www/ui/admin-upload-edit.php
@@ -74,11 +74,9 @@ class upload_properties extends FO_Plugin
         __METHOD__ . '.updateUpload.name');
     }
 
-    if (! empty($newDesc)) {
-      $trimNewDesc = trim($newDesc);
-      $this->dbManager->getSingleRow("UPDATE upload SET upload_desc=$2 WHERE upload_pk=$1",
-        array($uploadId, $trimNewDesc), __METHOD__ . '.updateUpload.desc');
-    }
+    $trimNewDesc = trim($newDesc);
+    $this->dbManager->getSingleRow("UPDATE upload SET upload_desc=$2 WHERE upload_pk=$1",
+      array($uploadId, $trimNewDesc), __METHOD__ . '.updateUpload.desc');
     return 1;
   }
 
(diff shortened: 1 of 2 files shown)


its-sushant:
We already have a PR opened for the linked issue.
PR: https://github.com/fossology/fossology/pull/3043

PR #3546, opened 2026-04-01
fix(api): add null-safety checks in ReportController::checkReport()

### LINKED ISSUE #3544: ReportController::checkReport() may cause runtime errors if database row is missing
### Description

**Problem:**
The` getSingleRow() `method in `DbManager `can return false when a SQL query matches no rows. Currently, `ReportController::checkReport()` accesses array keys directly without checking for this case:

```
$row = $dbManager->getSingleRow('SELECT jq_type FROM jobqueue WHERE jq_job_fk = $1', array($id), "reportValidity");
if (!in_array($row['jq_type'], $this->reportsAllowed)) { ... }
```

If no matching row exists, $row is false → PHP throws a fatal error when accessing $row['jq_type'].

**Impact**

- Any API call to download a report for a job that does not exist will cause a runtime error.
- Makes the system less robust and can crash the endpoint unexpectedly.
- Could affect both UI and API users trying to access reports.

**Proposed Fix**
Add null-safety checks before accessing array elements:
```
if (!$row || !in_array($row['jq_type'], $this->reportsAllowed)) {
    throw new HttpNotFoundException("No report scheduled with given job id.");
```
}
Apply similar checks for other `getSingleRow()` calls in the function.
Improves robustness and prevents fatal errors.
**Steps to Reproduce**
1.Schedule no report for a new job.
2.Call the API to download or check the report.
3.Observe that a PHP fatal error occurs due to array access on false.

pranavsnotebook-a11y:
## Description

\`ReportController::checkReport()\` accessed array keys directly on the result of \`getSingleRow()\`, which returns \`false\` when no row is found. This caused a fatal PHP error instead of a clean HTTP response.

Closes #3544

### Changes

- Added \`!$row\` guard before accessing \`$row['jq_type']\` → returns **404** if no matching jobqueue entry
- Added \`!$row\` guard before accessing \`$row['job_upload_fk']\` → returns **404** if no matching job entry

### Before
\`\`\`php
$row = $dbManager->getSingleRow(...);
if (!in_array($row['jq_type'], ...)) { // fatal error if $row is false
\`\`\`

### After
\`\`\`php
if (!$row || !in_array($row['jq_type'], ...)) {
    throw new HttpNotFoundException(...);
\`\`\`

### DIFF
--- a/src/www/ui/api/Controllers/ReportController.php
+++ b/src/www/ui/api/Controllers/ReportController.php
@@ -258,12 +258,15 @@ class ReportController extends RestController
       'SELECT jq_type FROM jobqueue WHERE jq_job_fk = $1', array(
         $id
       ), "reportValidity");
-    if (! in_array($row['jq_type'], $this->reportsAllowed)) {
+    if (!$row || ! in_array($row['jq_type'], $this->reportsAllowed)) {
       throw new HttpNotFoundException(
         "No report scheduled with given job id.");
     }
     $row = $dbManager->getSingleRow('SELECT job_upload_fk FROM job WHERE job_pk = $1',
       array($id), "reportFileUpload");
+    if (!$row) {
+      throw new HttpNotFoundException("No job found with given job id.");
+    }
     $uploadId = intval($row['job_upload_fk']);
     $uploadDao = $this->restHelper->getUploadDao();
     if (! $uploadDao->isAccessible($uploadId, $this->restHelper->getGroupId())) {


pranavsnotebook-a11y:
Hi @its-sushant @shaheemazmalmmd — small null-safety fix in the report download endpoint. Single file, 4 lines added. Would appreciate a quick review!

its-sushant:
JOB_ID can't exist without `upload_fk`, so this change is not required!

PR #3547, opened 2026-04-01
fix(upload): remove duplicate fileUploader handler and fix password toggle null safety

pranavsnotebook-a11y:
## What

Two independent UI bug fixes in the file upload flow:

### 1. Duplicate `#fileUploader` change handler (agent_reuser.js.twig + upload_file.html.twig)

The `#fileUploader` `change` event was bound in both `upload_file.html.twig` and `agent_reuser.js.twig`. Both handlers contained the same logic for:
- Building file description cards
- Building reuse name tabs
- Calling `toggleDisabled()`

This caused all of the above to execute **twice** on every file selection.

**Fix:** Remove the duplicate handler body from `agent_reuser.js.twig`, keeping only the Osselot initialization loop which is reuser-specific. Also merge the separate `currentReuseSource` restore handler in `upload_file.html.twig` into the main handler.

### 2. Null reference in `togglePasswords.js`

The `else` branch queried `#togglePassword` and `#pwd` without checking existence, crashing with a null reference on any page that doesn't have those elements.

**Fix:** Wrap the event listener attachment in an existence check.

## Testing

Verified with Playwright e2e tests that file selection still:
- Renders a description card per file
- Enables the submit button via `toggleDisabled()`
- Creates a card per file for multiple selections
- Shows the collapse button when >10 files are selected
- Renders reuse nav tabs

All 5 tests pass on both the original and patched code, confirming the removed code was purely redundant.

### DIFF
--- a/src/www/ui/scripts/togglePasswords.js
+++ b/src/www/ui/scripts/togglePasswords.js
@@ -14,15 +14,17 @@ if (togglePassword) {
     // toggle the eye slash icon
     this.classList.toggle('fa-eye-slash');
   });
-} else  {
+} else {
   const togglePassword = document.querySelector('#togglePassword');
   const password = document.querySelector('#pwd');
-  togglePassword.addEventListener('click', function (e) {
-    // toggle the type attribute
-    const type = password.getAttribute('type') === 'password' ? 'text' : 'password';
-    password.setAttribute('type', type);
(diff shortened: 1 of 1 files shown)


Kaushl2208:
Irrelevant changes. 
Not working and has merge conflicts.

PR #3553, opened 2026-04-03
fix(ununpack): prevent ununpack agent crash when scanning empty or inaccessible archives

mandar1045:
Fixes a crash path in `IsInflatedFile()` in `src/ununpack/agent/utils.c`.

The code was using `stat()` results without checking for failure, and it could also divide by zero when the parent archive size was `0`.

This patch returns early if `stat()` fails and guards the size check before division, so `ununpack` does not crash on empty or inaccessible archives.


mandar1045:
Closing to resubmit with DCO sign-off and more comprehensive fix.

PR #3554, opened 2026-04-03
fix(libfossrepo): use strlen(Ext) instead of strlen(Type) when tracking extension length in fo_RepMkPathTmp

mandar1045:
Fixes a small length-tracking typo in `fo_RepMkPathTmp()` in `src/lib/c/libfossrepo.c`.

When an extension is appended, the code was adding `strlen(Type)` instead of `strlen(Ext)`. That leaves `Len` out of sync with the actual path being built.

This patch switches that update to `strlen(Ext)` so the running length stays correct.


mandar1045:
Closing to resubmit with DCO sign-off and more comprehensive fix.

shaheemazmalmmd (approved):
Code looks good.

PR #3567, opened 2026-04-07
perf(dao): optimize upload permission batch check and job filter

ayushmaan-16:
Description:
This PR improves the performance of the "My Recent Jobs" dashboard. While exploring the code, I noticed that the logic for checking upload permissions inside the job list was doing extra work, which made the page lag a bit for the users with a large number of recent jobs.

image

Changes:
Batch Permission Checks: Updated UploadPermissionDao to verify permissions for all jobs in the list at once, rather than one by one. This reduces the continuous communication with the database.
Faster Filtering: Changed the way we filter the job list in ShowJobsDao. By using a simple mapping technique instead of repeatedly searching through arrays, the code can now find and display accessible jobs almost instantly.

image

The new Permission Query:
image

How to test
Log in to fossology with a user who has several recent jobs.
Go to the "My Recent Jobs" section on the dashboard.
Verify that the list loads correctly and only displays jobs for uploads you have permission to view.
If you have a large history, you should notice the page loads more smoothly than it did before.

### DIFF
--- a/src/lib/php/Dao/ShowJobsDao.php
+++ b/src/lib/php/Dao/ShowJobsDao.php
@@ -143,10 +143,11 @@ class ShowJobsDao
     $this->dbManager->freeResult($result);
 
     $accessibleUploads = $this->uploadDao->filterAccessibleUploads(array_unique($uploadIds), Auth::getGroupId());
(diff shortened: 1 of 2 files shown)


ayushmaan-16:
@shaheemazmalmmd any updates on this ?

PR #3569, opened 2026-04-07
perf(admin): optimize folder size dashboard clearing duration batch 

ayushmaan-16:
Description:
While checking the code responsible for "Folder and upload dashboard", I noticed that the system was fetching clearing durations one by one for every single upload in the list.
I’ve optimized this by adding a way to fetch these durations in a single batch, so that the database is not called for every upload.

image

Changes:
Optimized Dashboard Loading: Updated the Folder dashboard to fetch all upload "clearing durations" in one go rather than one by one.

image

Improved Data Exports: Applied the same optimization to the CSV and JSON export features so they generate instantly.
image

New Batch Logic: Added a getClearingDurationsBatch() method to the library to handle multiple duration requests efficiently.
image

Fixed Export Bug: Corrected a minor issue where the "Status" column in exported CSV/JSON files was sometimes showing inaccurate data.
image

Works perfectly after the new changes:

image

How to test
Go to Admin -> Dashboards -> Folder/Upload Proportions.
Select a folder that contains several uploads.
The page should load correctly (and faster if the folder is large).
Also, Click on the CSV or JSON export buttons at the top; verify that the downloaded file contains the correct durations and statuses.

### DIFF
--- a/src/lib/php/Dao/UploadDao.php
+++ b/src/lib/php/Dao/UploadDao.php
@@ -312,7 +312,7 @@ class UploadDao
    */
   public function getAssigneeDate(int $uploadId): ?string
   {
-    $sql = "SELECT event_ts FROM upload_events WHERE upload_fk = $1 " .
+    $sql = "SELECT MIN(event_ts) as event_ts FROM upload_events WHERE upload_fk = $1 " .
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (approved):

PR #3574, opened 2026-04-09
fix: harden obligation import handling

mandar1045:
## Summary

This PR hardens obligation import flows in both the UI/API entrypoint and the import helper.

## Why This Is Important

Obligation import is an admin workflow that directly affects license compliance data. When it behaves inconsistently, teams can silently import incorrect mappings, get confusing feedback for empty files, or hit avoidable failures on malformed JSON and upload errors. This change makes the import path safer, more predictable, and easier to troubleshoot before bad data reaches the database.

### What was broken
- ObligationCsvImport kept its previous CSV header mapping across uploads, so reusing the same importer instance could corrupt later imports when the next file used a different header order.
- Empty CSV files were reported as Read csv: -1 obligations.
- Malformed or empty JSON inputs could fall through into runtime errors while counting or iterating decoded data.
- AdminObligationFromCSV did not normalize empty delimiter and enclosure values and returned a misleading upload_max_filesize message instead of Symfony's actual upload error.

### What changed
- Reset importer header state at the start of every file import.
- Count only data rows, so empty CSV imports correctly report 0 obligations.
- Normalize decoded JSON into a safe row list and handle malformed or unsupported JSON structures gracefully.
- Default empty and null CSV controls back to comma and double quote before parsing.
- Return the real upload error from Symfony when the uploaded file is invalid.

## Testing
- src/vendor/bin/phpunit --configuration src/phpunit.xml src/lib/php/Application/test/ObligationCsvImportTest.php
- src/vendor/bin/phpunit --configuration src/phpunit.xml src/www/ui_tests/api/Controllers/AdminObligationFromCSVTest.php

Both tests passed locally. PHPUnit also reported an existing config-schema warning because the repo's src/phpunit.xml targets a newer schema than the bundled PHPUnit 8.5 runner.

### DIFF
--- a/install/fo-install-pythondeps
+++ b/install/fo-install-pythondeps
@@ -160,14 +160,14 @@ if [[ $RUNTIME ]]; then
     if [[ $EXPERIMENTAL ]]; then
       # Include experimental dependencies
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m pip install \
-        --target=$HOME/pythondeps --no-input --upgrade \
+        --target=$HOME/pythondeps --no-input --upgrade --no-binary scancode-toolkit \
         scancode-toolkit==32.4.1 scanoss==1.5.1 safaa==0.0.3'
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m spacy \
         download en_core_web_sm --target=$HOME/pythondeps'
     else
       # Only non-experimental dependencies
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m pip install\
-        --target=$HOME/pythondeps --no-input --upgrade \
+        --target=$HOME/pythondeps --no-input --upgrade --no-binary scancode-toolkit \
         scancode-toolkit==32.4.1 scanoss==1.5.1'
     fi
     ###########################################################################
--- a/src/lib/php/Application/ObligationCsvImport.php
+++ b/src/lib/php/Application/ObligationCsvImport.php
@@ -21,6 +21,7 @@ use Fossology\Lib\Util\ArrayOperation;
  */
 class ObligationCsvImport
 {
+  protected $obligationMap;
   /** @var DbManager $dbManager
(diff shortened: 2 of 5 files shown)


shaheemazmalmmd:
@mandar1045 : Can you please add some validations here ??, The issue is not reported by any user.

we would need the screenshot/ video with error and the fix.

install/fo-install-pythondeps
```diff
@@ -160,14 +160,14 @@ if [[ $RUNTIME ]]; then
     if [[ $EXPERIMENTAL ]]; then
       # Include experimental dependencies
       su --login --whitelist-environment="http_proxy,HTTP_PROXY,https_proxy,HTTPS_PROXY" $TARGETUSER -c 'PYTHONPATH="$HOME/pythondeps" python3 -m pip install \
-        --target=$HOME/pythondeps --no-input --upgrade \
+        --target=$HOME/pythondeps --no-input --upgrade --no-binary scancode-toolkit \
         scancode-toolkit==32.4.1 scanoss==1.5.1 safaa==0.0.3'
```
shaheemazmalmmd, line 164: Why is this change needed ??

mandar1045: Hi @shaheemazmalmmd,

Thank you for reviewing the PR and for the feedback. I just wanted to explain the thinking behind the changes I made.

The main idea was to make the obligation import flow more stable in cases where unexpected or malformed input is provided, so the behavior stays predictable and easier to debug.

A few examples:

- Header state in "ObligationCsvImport": If the previous CSV header mapping remains in memory, a new import with a different column order could be interpreted incorrectly. Resetting it makes each import independent.

- Empty or malformed JSON/CSV: Handling these inputs earlier helps avoid unexpected runtime failures and gives clearer feedback.

- Delimiter/enclosure normalization: If these values are empty, normalizing them keeps parsing behavior consistent and avoids confusing errors.

For the "--no-binary scancode-toolkit" change, I added it while testing installation behavior in some environments where building from source can be more reliable. If that part is not needed here, I’m fine removing it.

I just wanted to share the reasoning behind the changes. If you’d prefer a different approach, I’m happy to update the PR accordingly.

mandar1045:
Hi @shaheemazmalmmd,

I verified the obligation import fixes locally in the UI and attached screenshots of the corrected behavior.

After the fix:
- valid CSV import works successfully
- empty CSV now reports `Read csv: 0 obligations`
- malformed JSON is handled gracefully with `Error decoding JSON: Syntax error` and `Read json:0 obligations`

I also verified the related test coverage locally:
- `ObligationCsvImportTest.php`
- `AdminObligationFromCSVTest.php`

These cover the validation and resilience updates added in this PR.
Pasted image
Pasted image (2)

Screenshot from 2026-04-13 16-31-47
image

PR #3579, opened 2026-04-10
perf(bulkreuser): optimize bulk reuse scanning to eliminate duplicate scans

sakshammishra112:
## Description
 
Optimizes bulk reuse scanning by implementing a text, license-based grouping mechanism that eliminates duplicate scans for identical content. The solution groups bulk entries by normalized text + license hash, validates license consistency, and combines matching entries.
 
## Changes
 
- Added `getBulkIdsGroupedByTextAndLicenses(`) in `ClearingDao` to group bulk entries by text content and license IDs, process them chronologically by job_queued, and return groups with identical content.
- Optimized `DeciderAgent.processUploadId()` to use grouped data, process only one entry per group, and maintain chronological order.

### Before
https://github.com/user-attachments/assets/e696118e-c531-46eb-8820-900819d9fc7e

CC: @shaheemazmalmmd

### DIFF
--- a/src/decider/agent/BulkReuser.php
+++ b/src/decider/agent/BulkReuser.php
@@ -36,11 +36,11 @@ class BulkReuser
 
   /**
    * @fn rerunBulkAndDeciderOnUpload($uploadId, $groupId, $userId, $bulkId, $dependency)
-   * @brief Rerun Monk bulk and decider on a given upload
-   * @param int   $uploadId   Upload on which the agents have to run
+   * @brief Rerun Monk bulk and decider on a given upload (handles grouped entries by processing latest ID)
+   * @param int   $uploadId   Upload on which agents have to run
    * @param int   $groupId    Group which is running the agents
    * @param int   $userId     User who is running the agents
-   * @param int   $bulkId     Previous bulk for reuse
+   * @param int   $bulkId     Bulk ID to process (for grouped entries, uses latest ID from group)
    * @param array $dependency Dependency on agent (Not in use)
    * @throws Exception If agent cannot be added, throw exception with message as HTML
    * @return int Id of the new job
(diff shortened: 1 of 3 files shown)


shaheemazmalmmd:
Hi @sakshammishra112,

I would like to seek clarification on a few points:

Are these bulk statements executed based on date order (i.e., from the first scheduled to the last scheduled)? If not, could you please update the implementation to ensure this sequence is followed?

Regarding your point:

> If all license sets match, the entries are eligible for combination. If any set differs, the agent fails fast with a descriptive error message.

I discussed this with the team, and there may be scenarios where a user schedules a scan by mistake and later modifies the license names in subsequent scans.

For now, let’s proceed with the assumption that duplicates are identified only when both the bulk text and the license ID are the same.

OliverFendt:
It is absolutely necessary that the time sequence of the bulk scans is preserved. A change in the strict time order of the bulk scans will ruin the entire feature, which is not acceptable.

Furthermore it is necessary that _all_ bulk statements in the correct timely order are applied, because sometimes clearing experts make a wrong bulk scan by accident and to correct this the same bulk phrase is used with then the corrective added and deleted licenses, acknowledgements and comments.
This must be kept.

sakshammishra112:
### Current Approach:
As @OliverFendt mentioned, users sometimes schedule a bulk scan with the wrong license, then reschedule another scan with the same text but with corrected licenses. To handle this optimisation, we group bulk entries by both **text** content AND **license** IDs, ensuring only identical entries are combined. For each **_text_** group, we process the most recently scheduled bulk scan (determined by the latest job queue timestamp using `MAX(j.job_queued)`), which reflects the user's most recent corrections. 

The bulk history displays in chronological order (oldest first) for user experience, while the optimisation processes the latest entry per group to ensure corrected scans are applied. This approach preserves the full history of bulk uploads while eliminating duplicate processing of identical content.

## Output:

<details>
<summary> Packages used for reuse </summary>

Screenshot 2026-04-15 181948
Screenshot 2026-04-15 181958
Screenshot 2026-04-15 181938

</details>

### Scheduled upload for Bulk reuse:

<details>
<summary>Scheduled upload screenshots </summary>

Screenshot 2026-04-15 182009
image

</details>

CC: @shaheemazmalmmd

shaheemazmalmmd:
@sakshammishra112 : As discussed today closing this PR for now. As removal of duplicate/wrong bulk statements is already in place the decision can be the users to keep the duplicate or remove it

PR #3580, opened 2026-04-12
fix(security): escape shell args in getRepoPathOfPfile() to prevent command injection

krrish175-byte:
## Description
This pull request addresses a security vulnerability in the `getRepoPathOfPfile` method where shell arguments were being interpolated directly into an `exec()` call without proper escaping.

### Changes
- Wrapped the `$repo` and `$hash` arguments in `escapeshellarg()` to prevent shell metacharacter injection.
- Wrapped the base command path in `escapeshellcmd()` for defense-in-depth.
- Added an `is_string($repo)` guard clause to ensure input validity before shell execution.
- Added a new unit test `testGetRepoPathOfPfile` in `TreeDaoTest.php` to verify the new validation logic and error handling.

## How to test
1. Run the updated unit test suite for TreeDao:
```bash
# From the src directory
./vendor/bin/phpunit lib/php/Dao/test/TreeDaoTest.php
```
2. Verify that the testGetRepoPathOfPfile test case passes, ensuring that invalid (non-string) $repo values and empty hashes are handled safely.
3. Perform a manual code audit of the exec() call in src/lib/php/Dao/TreeDao.php to confirm arguments are correctly passed through escapeshellarg().

### DIFF
--- a/src/lib/php/Dao/TreeDao.php
+++ b/src/lib/php/Dao/TreeDao.php
@@ -124,15 +124,18 @@ class TreeDao
 
   public function getRepoPathOfPfile($pfileId, $repo="files")
   {
+    if (!is_string($repo)) {
+      return null;
+    }
     $pfileRow = $this->dbManager->getSingleRow('SELECT * FROM pfile WHERE pfile_pk=$1',array($pfileId));
     global $LIBEXECDIR;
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (approved):

PR #3587, opened 2026-04-14
fix(reuser): check if same pfile have different decisions

shaheemazmalmmd:
## Description

Irrelevant decisions are getting added in case of the same pfile and different clearing decisions.

### Changes

Add check for the conflicts if conflicts exists the add a stronger decision (identified) instead of irrelevant 

## How to test

* Upload a package which contains same file in multiple locations of the same package.
* Mark these files with different decisions (few with identified and few with irrelevant). 
* Now upload a next version of it. and check there should not be irrelevant decisions.


Kaushl2208 (approved):
Tested, Working as expected.

PR #3589, opened 2026-04-15
fix(reportImport): add setLicenseCandidate for custom text licenses

### LINKED ISSUE #3025: Improvement (reportImport): Handling of new / unknown licenses with custom text
# Description
## Background
With the [OSSelot project](https://www.osselot.org) we stumbled accross a use-case for which see some possible improvements for the import of SPDX reports.  The situation is as follows: In some specific cases contributors create a license as a "placeholder". These licenses are only being used with a custom text. One example is "Permission-Notice" which is being used to mark findings with individual permission notices. The related text is always individual and is set by the contributor as a custom text during the clearing process.

## Scenario
Users of the generated SPDX reports obviously don't have that custom license in their database. Usually new licenses are created during the import process. But new licenses (or candidates) are only created for findings without a custom  text. For findings with a custom text the import code expects the license to be existent and only adds the extracted custom text to the existing license. This is the relevant code for importing SPDX2 reports:

Custom text:

```php
      if ($license->isA('spdx:ExtractedLicensingInfo') &&
        (strlen($licenseId) > 33 &&
          substr($licenseId, -33, 1) === "-" &&
          ctype_alnum(substr($licenseId, -32))
        )) {
        $licenseId = substr($licenseId, 0, -33);
        $item = new ReportImportDataItem($licenseId);
        $item->setCustomText($licenseTextLiteral->getValue());
```

No custom text:

```php
      } else {
        $item = new ReportImportDataItem($licenseId);
        $item->setLicenseCandidate($licenseNameLiteral->getValue(),
          $licenseTextLiteral->getValue(),
          strpos($rawLicenseId, LicenseRef::SPDXREF_PREFIX),
          ($seeAlsoLiteral != null) ? $seeAlsoLiteral->getValue() : ""
        );
      }
```

For the previously described use-case ("Permission-Notice") there are only findings with custom text in the report, which prevents the license from being created and the related findings will silently be ignored, resulting in missing clearing decisions (which are present in the imported report).

## Reproducing the scenario
1. Download the [SPDX2 RDF report for busybox-1.36.1](https://raw.githubusercontent.com/Open-Source-Compliance/package-analysis/refs/heads/main/analysed-packages/busybox/version-1.36.1/busybox-1.36.1.spdx.rdf.xml) from the OSSelot project
2. Upload [busybox-1.36.1.tar.bz2](https://busybox.net/downloads/busybox-1.36.1.tar.bz2) to your FOSSology instance (without running any agents)
3. Import the previously (in step 1) downloaded RDF file for this upload
4. Navigate to busybox-1.36.1/editors/ed.c (which won't show any finding)

The RDF report contains the following entry for busybox-1.36.1/editors/ed.c

```xml
[...]
    <spdx:fileName>busybox-1.36.1.tar.bz2/busybox-1.36.1.tar/busybox-1.36.1/editors/ed.c</spdx:fileName>
     <spdx:checksum>
       <spdx:Checksum>
         <spdx:algorithm rdf:resource="http://spdx.org/rdf/terms#checksumAlgorithm_md5"/>
         <spdx:checksumValue>e0d3dec0d11023515fbb32c25693801d</spdx:checksumValue>
       </spdx:Checksum>
     </spdx:checksum>
     <spdx:licenseConcluded>
       <spdx:ExtractedLicensingInfo rdf:about="http://fossy.osadl.org/repo/SPDX2TV_busybox-1.36.1.tar.bz2_1687761058.spdx#LicenseRef-Permission-Notice-b79dfa8d57b6aef14cf0aa1e0a5519fa">
         <spdx:licenseId>LicenseRef-Permission-Notice-b79dfa8d57b6aef14cf0aa1e0a5519fa</spdx:licenseId>
         <spdx:extractedText>Permission is granted to use, distribute, or modify this source,
provided that this copyright notice remains intact.</spdx:extractedText>
         <spdx:name>Permission-Notice-b79dfa8d57b6aef14cf0aa1e0a5519fa</spdx:name>
       </spdx:ExtractedLicensingInfo>
     </spdx:licenseConcluded>
[...]
```

The finding got silently (no message in the JOB log) ignored, because the license is not existing in the database and all references in the report are using it with a custom text.

# Proposal
Since the user of the report is not aware of the situation, clearing decisions can silently be lost. One possibility to solve this and to improve the user experience could be to create a new license (with the related shortname) for unknown licenses with custom text as well (maybe with an empty text?). @shaheemazmalmmd  @OliverFendt: Let me know your opinion on the scenario and my proposal. I am open to counter proposals as well. If we can agree on a commonly accepted behaviour  I am volunteering to get it implemented and to send a pull request. And please  let me know if the described scenario needs further explanation.

Priya-Sharma25:
### Description
This PR fixes issue #3025 where SPDX report imports silently drop license findings when the license has custom text (hash-suffixed ExtractedLicensingInfo) and doesn't already exist in the database.

Right now, if you import an SPDX report that contains something like LicenseRef-Permission-Notice-b79dfa8d57b6aef14cf0aa1e0a5519fa, and your FOSSology instance doesn't have Permission-Notice in its database, the finding just disappears. No error, no warning in the job log — it's silently ignored. This can lead to missing clearing decisions, which is a real problem for projects like OSSelot that rely on importing SPDX reports across different FOSSology instances.

The root cause is in handleLicenseInfo() — the custom-text code branch calls setCustomText() but never calls setLicenseCandidate(). Since the downstream ReportImportSink only creates new licenses when isSetLicenseCandidate() returns true, these findings get skipped entirely.

### Changes

- **SpdxTwoImportSource.php:** Added setLicenseCandidate() call in the custom-text branch (the ExtractedLicensingInfo with 32-char hash suffix path). This ensures unknown licenses with custom text are registered as candidates so the Sink can create them in the database.

- **SpdxThreeImportSource.php:** Applied the same one-line fix for SPDX 3 imports (expandedlicensing_CustomLicense path) to keep both importers consistent.

### How to test
**Reproduce the bug (before the fix):**

1. Upload any small file (e.g., a simple test.c) to FOSSology without running any agents
2. Create a minimal SPDX 2 RDF file that references a made-up license with custom text — something like LicenseRef-MyTestLicense-<32 hex characters> as an ExtractedLicensingInfo with extractedText
3. Import the RDF via Upload → Import Report
4. Browse to the file and check the license table — it will be empty
Screenshot from 2026-04-13 17-34-55

**Verify the fix (after the fix):**

1. Apply the changes from this PR
2. Repeat the same import steps
3. Browse to the file — the license now appears in the table with source reportimport
4. Click on the license name to verify the custom text is preserved in the License Info popup
Screenshot from 2026-04-15 10-13-50

You can also reproduce this with the exact scenario from the original issue: download the [busybox SPDX2 RDF from OSSelot](https://raw.githubusercontent.com/Open-Source-Compliance/package-analysis/refs/heads/main/analysed-packages/busybox/version-1.36.1/busybox-1.36.1.spdx.rdf.xml), upload busybox-1.36.1, import the RDF, and navigate to editors/ed.c — the Permission-Notice finding should now appear.
 
Closes #3025

### DIFF
--- a/src/reportImport/agent/SpdxThreeImportSource.php
+++ b/src/reportImport/agent/SpdxThreeImportSource.php
@@ -246,22 +246,26 @@ class SpdxThreeImportSource implements ImportSource
       $rawLicenseId = $licenseIdLiteral->getValue();
       $licenseId = $this->stripLicenseRefPrefix($rawLicenseId);
 
+      $isCustomText = false;
       if ($license->isA('spdx:expandedlicensing_CustomLicense') &&
         (strlen($licenseId) > 33 &&
           substr($licenseId, -33, 1) === "-" &&
           ctype_alnum(substr($licenseId, -32))
         )) {
         $licenseId = substr($licenseId, 0, -33);
-        $item = new ReportImportDataItem($licenseId);
-        $item->setCustomText($licenseTextLiteral->getValue());
-      } else {
-        $item = new ReportImportDataItem($licenseId);
-        $item->setLicenseCandidate($licenseNameLiteral->getValue(),
-          $licenseTextLiteral->getValue(),
-          strpos($rawLicenseId, LicenseRef::SPDXREF_PREFIX),
-          ($seeAlsoLiteral != null) ? $seeAlsoLiteral->getValue() : ""
-        );
+        $isCustomText = true;
       }
+
+      $item = new ReportImportDataItem($licenseId);
+      if ($isCustomText) {
+        $item->setCustomText($licenseTextLiteral->getValue());
+      }
+      $item->setLicenseCandidate($licenseNameLiteral->getValue(),
+        $licenseTextLiteral->getValue(),
+        strpos($rawLicenseId, LicenseRef::SPDXREF_PREFIX),
+        ($seeAlsoLiteral != null) ? $seeAlsoLiteral->getValue() : ""
+      );
+
       return [$item];
     }
     return [];
--- a/src/reportImport/agent/SpdxThreeImportSource.php
+++ b/src/reportImport/agent/SpdxThreeImportSource.php
@@ -254,6 +254,7 @@ class SpdxThreeImportSource implements ImportSource
         $licenseId = substr($licenseId, 0, -33);
         $item = new ReportImportDataItem($licenseId);
         $item->setCustomText($licenseTextLiteral->getValue());
+        $item->setLicenseCandidate($licenseNameLiteral->getValue(), $licenseTextLiteral->getValue(), strpos($rawLicenseId, LicenseRef::SPDXREF_PREFIX), ($seeAlsoLiteral != null) ? $seeAlsoLiteral->getValue() : "");
       } else {
         $item = new ReportImportDataItem($licenseId);
         $item->setLicenseCandidate($licenseNameLiteral->getValue(),
--- a/src/reportImport/agent/SpdxTwoImportSource.php
+++ b/src/reportImport/agent/SpdxTwoImportSource.php
@@ -293,6 +293,7 @@ class SpdxTwoImportSource implements ImportSource
         $licenseId = substr($licenseId, 0, -33);
         $item = new ReportImportDataItem($licenseId);
         $item->setCustomText($licenseTextLiteral->getValue());
+        $item->setLicenseCandidate($licenseNameLiteral->getValue(), $licenseTextLiteral->getValue(), strpos($rawLicenseId, LicenseRef::SPDXREF_PREFIX), ($seeAlsoLiteral != null) ? $seeAlsoLiteral->getValue() : "");
       } else {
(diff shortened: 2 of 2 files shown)


SagarKhatri649:
This looks like a clean fix aligned with the intended import behavior.

JanAltenberg:
Thanks a lot! Successfully tested with different critical uploads. Just a minor comment: The added lines look identical to those in the else path (if I'm not missing something). Maybe these could stay outside the if/else path to avoid duplication.

Priya-Sharma25:
Hi @JanAltenberg, thank you for the feedback!

I have refactored the handleLicenseInfo method in both SpdxTwoImportSource.php and SpdxThreeImportSource.php to eliminate the code duplication. The setLicenseCandidate call has been moved outside the conditional block as suggested.
Please let me know if there is anything else that needs adjustment.

shaheemazmalmmd (approved):
Code looks good.

Issue #3586, opened 2026-04-14
[Bug]: File action dropdown stays selected after browser back on Browse page


im-vishesh15th:
### Summary
On the Browse page, when a user selects an option from the file/folder action dropdown and navigates to the target page, then using the browser Back button returns back to Browse,  the previously selected action is still shown and selecting the same action again does not trigger navigation.

### Steps to Reproduce
1. Open the Browse page.
2. For any file/folder, select an option from the **File Actions** dropdown.
3. Navigate to the opened page.
4. Click the browser **Back** button to return to Browse.
5. Try selecting the same action again.

### Actual Behavior
- The previous action remains selected in the dropdown.
- Selecting the same action again does not trigger navigation.
- User must first select a different option, then reselect the desired one.

### Expected Behavior
- The dropdown should reset to its default/placeholder option when returning to Browse.
- Selecting any action (including the previously used one) should always trigger navigation.

### Impact
- Confusing UX and extra clicks for repeated actions.
- Makes the action menu appear broken after back navigation.

### Screenshots

<img width="1710" height="980" alt="Image" src="https://github.com/user-attachments/assets/311d9e29-15a5-4827-a1a9-72ed774eed2b" />
`click on browser's back button`
<img width="1644" height="534" alt="Image" src="https://github.com/user-attachments/assets/63777be2-2942-4fb6-b8a8-e2116358e3d4" />
`still previously selected option is shown`
<img width="1636" height="403" alt="Image" src="https://github.com/user-attachments/assets/c04b35a7-71a1-45d9-93d9-14c18e5dadab" />

shaheemazmalmmd:
@im-vishesh15th : its is expected behavior, and users will expect to land in the same page and same option if they press back. Fix not required for this.

PR #3593, opened 2026-04-16
fix(browser): reset action dropdown after back navigation (#3586)

im-vishesh15th:
## Summary
- Fixes issue #3586 where file/folder action dropdown on Browse page keeps previous selection after navigating back.
- Ensures returning via browser Back/Forward resets `goto-active-option` to default using `pageshow`.
- Resets dropdown selection after action handling so selecting the same option again triggers navigation without needing an intermediate selection.

## Problem
After selecting a file action and returning to Browse via browser Back, the previously selected action stayed selected, so choosing the same action again did not redirect unless another option was selected first.

## Fix

- Added a `pageshow` listener to reset all `select.goto-active-option` dropdowns when the page is shown again (including browser back/forward cache restore).
- In the existing dropdown `change` handler, after reading the selected value and handling Delete confirmation, explicitly reset the current dropdown:
  - `this.selectedIndex = 0`
- Prevent duplicate event bindings by unbinding existing handlers before attaching new ones

This keeps the default state after each action and guarantees that selecting the same option again is treated as a valid new action.

## Result
- Returning to Browse with browser Back no longer leaves stale action selections.
- Selecting the same action again works immediately (no extra intermediate selection needed).
- Existing behavior for action routing and Delete confirmation remains intact.
- Overall Browse action flow is predictable and repeatable for users performing multiple operations on the same row.

### DIFF
--- a/src/www/ui/scripts/browse.js
+++ b/src/www/ui/scripts/browse.js
@@ -20,6 +20,10 @@ var staSel = null;
 $(document).ready(function () {
   assigneeSelected = ($.cookie("assigneeSelected") || 0);
   $('#assigneeSelector').val(assigneeSelected);
+  $(window).on('pageshow', function () {
+    $('select.goto-active-option').prop('selectedIndex', 0);
+  });
+
   table = createBrowseTable();
   $('#insert_browsetbl_filter').append($('#browsetbl_filter'));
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd:
See: https://github.com/fossology/fossology/issues/3586#issuecomment-4265753541

Issue #3592, opened 2026-04-15
Release Testing 4.7.0


shaheemazmalmmd:
- [ ] General Tool Testing - @shaheemazmalmmd @JanAltenberg  
- [ ] Upgrade Testing 4.1 to 4.7.0-rc1 - @Kaushl2208  
- [ ] Upgrade Testing 4.2 to 4.7.0-rc1 - @Kaushl2208 
- [ ] Upgrade Testing 4.3 to 4.7.0-rc1 - @its-sushant
- [ ] Upgrade Testing 4.4 to 4.7.0-rc1 - @its-sushant
- [ ]  Upgrade Testing 4.5 to 4.7.0-rc1 - @shaheemazmalmmd  
- [ ] Documentation review for new features - @shaheemazmalmmd 
- [ ] Attach third party licenses and clearing report - @shaheemazmalmmd 

## General Tool Testing
- [ ] Whole browse page (status, assignee, all actions, folder search, upload search and folder navigation).
- [ ] Show jobs page (all recent jobs, my recent jobs, pagination, cancel and pause jobs, upload navigation, folder structure)
- [ ] Tree view page (general navigation, scanner licenses, edited licenses, status(with all decisions), folder bulk, edit, download, search/filter functionality, clear search, histogram navigation to files of scanner licenses).
- [ ] Folder functionality (create, edit delete and move/copy)
- [ ] Group functionality (create, edit, manage)
- [ ] User functionality (create, edit and set inactive)
- [ ] Conf functionality (report settings, spdx settings, obligations, and decision testing)
- [ ] Copyright, email/url/author, ecc, ipra and keyword functionality (add, clean and deactivate)
- [ ] Search page.
- [ ] license, obligation admin page (add).
- [ ] customize page.
- [ ] upload permissions page, delete upload and edit properties.
- [ ] Upload from all the pages.
- [ ] Reuse all the functionalities.
- [ ] Reuse using OSSelot SPDX reports.
- [ ] Export reports by enabling OSSelot and check.
- [ ] generate all the reports and check.

## Testing Procedure for Release Upgrade testing

- Upload 2-4 components to FOSSology
- Do Clearing, using bulk, and regular.
- Add information to conf file.
- Do clearing using Tree view.
- Do copyright, ECC, IPRA Clearing.
- Do reuse once the clearing is finished with some other version of same upload.
- Now upgrade to 4.6.0-rc1 version.
- After finishing the upgrade check if each page is working well and there should be no, loss in data.

[...]

PR #3597, opened 2026-04-17
fix(uploadFile): fix cast expression error on a empty database

shaheemazmalmmd:
## Description

As part of testing for release #3592 , Cleaned up the database. While uploading a first file i have observed that there is a error and no agent runs except from ununpack and adj2nest. 

image

### Changes

Replaced with true. 

## How to test

* Drop fossology database.
* upload a new package.
* fossology shall work as usual.


Kaushl2208 (approved):
Looks good. Tested, Working as expected.

PR #3601, opened 2026-04-20
perf(core): batch isAccessible permission checks

ayushmaan-16:
## Description

In this PR I found 5 other instances across 4 files where `isAccessible` is being called inside a loop (since it causes the N+1 database query problem) and have updated those files to use the `filterAccessibleUploads` batching method. Now it will load much faster, especially when users are exploring large folders or doing global searches.

### Changes

- **`FileSearchController.php`**: I removed the `array_filter` loop and directly passed the array to `filterAccessibleUploads`.
- **`SearchHelperDao.php` & `AjaxBrowse.php`**: I updated the `while` loops. Now they first collect all the IDs in an array, run one single batch query, and use `isset` with a mapper (`array_flip`) to filter the results cleanly.
- **`common-folders.php`**: I applied the same batching logic for the `Auth::PERM_READ` checks. I also made sure that the `Auth::PERM_WRITE` checks still use the old `isEditable()` line-by-line check, so nothing breaks for folder owners.

## How to test

1. Login to the UI and go to the **Browse** menu. Try to navigate inside a folder that has hundreds of files. It should load much faster without lagging the DB.
2. Go to the **Search** menu and search for a common file. It should process the search results properly.
3. Finally, please verify that you are only able to see the files and folders that your specific user group has permission to see.

### DIFF
--- a/src/lib/php/Dao/SearchHelperDao.php
+++ b/src/lib/php/Dao/SearchHelperDao.php
@@ -242,8 +242,17 @@ class SearchHelperDao
     $stmt = __METHOD__ . "_paginated";
     $rows = $this->dbManager->getRows($PaginatedSQL, [], $stmt);
     if (!empty($rows)) {
+      $uploadIds = array();
       foreach ($rows as $row) {
-        if (!$uploadDao->isAccessible($row['upload_fk'], $groupID)) {
+        if (!empty($row['upload_fk'])) {
+          $uploadIds[] = $row['upload_fk'];
+        }
+      }
+      $accessibleUploads = $uploadDao->filterAccessibleUploads(array_unique($uploadIds), $groupID);
+      $accessibleMap = array_flip($accessibleUploads);
+
+      foreach ($rows as $row) {
+        if (empty($row['upload_fk']) || !isset($accessibleMap[$row['upload_fk']])) {
           continue;
         }
         $totalUploadtreeRecs[] = $row;
--- a/src/lib/php/common-folders.php
+++ b/src/lib/php/common-folders.php
@@ -371,16 +371,29 @@ function FolderListUploads_perm($ParentFolder, $perm)
 	ORDER BY upload_filename,upload_pk;";
   $result = pg_query($PG_CONN, $sql);
   DBCheckResult($result, $sql, __FILE__, __LINE__);
+
+  $fetchedRows = array();
+  $uploadIds = array();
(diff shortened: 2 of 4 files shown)


src/lib/php/Dao/SearchHelperDao.php
```diff
@@ -242,8 +242,17 @@ public function GetResults($Item, $Filename, $Upload, $tag, $Page, $Limit, $Size
...
       foreach ($rows as $row) {
-        if (!$uploadDao->isAccessible($row['upload_fk'], $groupID)) {
+        if (!empty($row['upload_fk'])) {
+          $uploadIds[] = $row['upload_fk'];
+        }
+      }
+      $accessibleUploads = $uploadDao->filterAccessibleUploads(array_unique($uploadIds), $groupID);
+      $accessibleMap = array_flip($accessibleUploads);
+
+      foreach ($rows as $row) {
```
Kaushl2208, line 254: Is it now two for loops? What about Line 246??

Also I have seen enough of these, hash-map based performance improvements.  `array_flip()`  actually does an iteration operation over the array and converts Keys <--> Values. Did you also check that it also : `array_flip will remove duplicate values in the original array when you flip` as the collision condition??

Also that is one operation (you didnt include the complexity of). I can understand `filterAccessibleUploads` can be used in other places. But replace them properly. Dont overcomplicate the process for the sake of optimization in performance.

src/lib/php/common-folders.php
```diff
@@ -371,16 +371,29 @@ function FolderListUploads_perm($ParentFolder, $perm)
...
-    if ($perm == Auth::PERM_READ &&
-      ! $uploadDao->isAccessible($R['upload_pk'], $groupId)) {
+    $fetchedRows[] = $R;
+    $uploadIds[] = $R['upload_pk'];
+  }
+  pg_free_result($result);
+
+  $accessibleMap = array();
+  if ($perm == Auth::PERM_READ && !empty($uploadIds)) {
+    $accessibleUploads = $uploadDao->filterAccessibleUploads(array_unique($uploadIds), $groupId);
```
Kaushl2208, line 388: Dont you think its a force fit here?

Kaushl2208 (commented):
Too much complications onboarding, in the name of optimization. Force fit of function behavior as well. 

Thank you for your contribution. Closing this now :)

PR #3603, opened 2026-04-20
fix(ununpack): prevent floating point exception in IsInflatedFile 

### LINKED ISSUE #2684: ununpack agent occasionally killed by signal: 8.Floating point exception
### Description

While doing a lot of vcs/git uploads via API, occasionally the ununpack agent gets killed and make all next jobs hang, like in this example:

immagine

The error is the following:

immagine

If I try to manually kill all the pending jobs and restart the upload, it fails again in the same way. But if I do:

```bash
service apache2 stop
service fossology stop
service postgresql restart
service fossology start
service apache2 start
```

and then restart the upload, everything works fine.

So far it happened only on 4 out of 100 consecutive uploads, with random frequency; but it's annoying, since I have to upload ~450 git repos.

I should stress that I'm uploading one git repo at a time, waiting for all the jobs to be completed before uploading the next one.

#### How to reproduce

Do a bunch of uploads via REST API, scheduling many agents for each upload (see above), and monitor jobs.

#### Screenshots

See above

### Versions

Fossology 4.4.0 (release), built from sources on Debian 11.

### Logs

n/a

#### FOSSology logs

No relevant log entries

#### Apache logs

No relevant log entries

#### Job logs

All I get is:

```
2024-02-15 17:10:40 ununpack [0] :: JOB[61443].ununpack[195302.localhost]: agent was killed by signal: 8.Floating point exception
```

Priya-Sharma25:
### Description
This pull request fixes a Signal 8 (Floating Point Exception) crash in the `ununpack` agent. The crash was caused by a division-by-zero error in the `IsInflatedFile` function, which occurred when processing archives with 0-byte parent files.

### Changes

- **Added Division-by-Zero Guard**: Updated `IsInflatedFile` in `utils.c` to ensure the parent file size is greater than 0 before performing division.
- **Added Return Checks for `stat()`**: The code now verifies that `stat()` calls are successful before using file metadata. If a file is missing or inaccessible, the function returns safely instead of using garbage data and crashing.

### How to test
I have verified this fix through both automated script testing and manual UI testing:
1.  **Reproduction Script**: 
    I wrote a standalone C script (`test_issue_2684.c`) that manually creates a 0-byte file structure to trigger the bug.
    *   **Before fix**: The original logic triggers a `Floating point exception`.
    *   **After fix**: The new logic handles the situation safely without crashing.
    
[test_issue_2684.c](https://github.com/user-attachments/files/26886776/test_issue_2684.c)

2.  **FOSSology UI Testing**:
    *   **Uploaded a test archive** containing a 0-byte file component.
    *   Observed that the `ununpack` job now moves to "Completed" status instead of crashing mid-process.
   
- **Before Fix :** 

Screenshot from 2026-04-19 22-51-45

Screenshot from 2026-04-19 22-51-54

- **After Fix :**

Screenshot from 2026-04-19 23-50-19

Screenshot from 2026-04-19 23-50-40

Closes #2684

### DIFF
--- a/src/ununpack/agent/utils.c
+++ b/src/ununpack/agent/utils.c
@@ -58,9 +58,16 @@ int IsInflatedFile(char *FileName, int InflateSize)
          e.g. for the file ./10g.tar.bz.dir/10g.tar, partent file is ./10g.tar.bz
        */
       FileNameParent[strlen(FileNameParent) - 4] = '\0';
-      stat(FileNameParent, &stParent);
-      stat(FileName, &st);
-      if(S_ISREG(stParent.st_mode) && (st.st_size/stParent.st_size > InflateSize))
+
+      /* Ensure parent and child files exist and can be stated */
+      if (stat(FileNameParent, &stParent) != 0 || stat(FileName, &st) != 0)
+      {
+        return 0;
+      }
+
+      /* Guard against division by zero */
+      if(S_ISREG(stParent.st_mode) && stParent.st_size > 0 && 
+         (st.st_size/stParent.st_size > InflateSize))
       {
         result = 1;
       }


Kaushl2208 (approved):
Looks good. Working as expected.

PR #3604, opened 2026-04-20
perf(dao): optimize job filtering using hash-map

ayushmaan-16:
## Description

The current code uses `in_array()` inside a loop, which is fine for small data but it's not very efficient for large datasets since it has to scan the whole list every single time. 

In this PR, I have updated the logic to use a hash-map approach. This moves the performance from a nested O(N*M) search to a much simpler O(N+M) lookup.

### Changes

- **`ShowJobsDao.php`**: I've added `array_flip()` to convert the accessible uploads list into a hash-map. Then, I replaced the iterative `in_array()` check with a direct `isset()` check inside the loop. This ensures that the permission filtering happens in constant time (O(1)) for each job record.

## How to test

1. Go to the dashboard and click on **My Recent Jobs**.
2. Check and make sure that all the jobs you expect to see are still showing up correctly.
3. If you have an account with a lot of job history, you should notice that the page is loading more smoothly now.

### DIFF
--- a/src/lib/php/Dao/ShowJobsDao.php
+++ b/src/lib/php/Dao/ShowJobsDao.php
@@ -141,10 +141,11 @@ class ShowJobsDao
     $this->dbManager->freeResult($result);
 
     $accessibleUploads = $this->uploadDao->filterAccessibleUploads(array_unique($uploadIds), Auth::getGroupId());
+    $accessibleMap = array_flip($accessibleUploads);
 
     foreach ($rows as $row) {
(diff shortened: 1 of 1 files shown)


its-sushant (commented):
LGTM! Need test.

its-sushant:
Thank you for the contribution @ayushmaan-16 After discussing within the community we came to conclusion that this change won't be required as of now since the number of uploads which come inside the job page is very less. So, creating a map for look up into smaller set of data would be bit of a overhead in terms of complexity.

PR #3608, opened 2026-04-21
Handle chown Failures on NFS (Local Workaround for #3129)

Shivangisriva:
Context & Motivation
When deploying Fossology with NFSv4 volumes (e.g., in Docker Swarm), file uploads can fail with chown failed: Operation not permitted due to NFS restrictions—even when permissions and UID/GID mappings appear correct. This is a known issue (see [#3129](vscode-file://vscode-app/c:/Users/shiva/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html)), and the maintainers have clarified that it should be solved at the infrastructure level, not in the upstream codebase.

However, for practical deployment needs, a local workaround is required to prevent uploads from failing unnecessarily.

What This PR Does
Modifies [wget_agent.c] so that failed chown calls log an error but do not abort the upload process.
Uploads will continue as long as file permissions are otherwise correct, even if chown fails (e.g., on NFS).
Error messages are clear and reference the NFS/chown issue for troubleshooting.

Why This Approach?
This is a minimal, safe, and effective local workaround.
It does not attempt to automatically detect NFS or add configuration options, keeping the logic simple.
It aligns with the maintainers’ guidance: permission issues should be solved by administrators, but this patch prevents unnecessary failures in real-world deployments.

Testing
Local filesystem: chown works as before; no change in behavior.
NFSv4 mount: If chown fails, the error is logged, but uploads succeed if permissions are otherwise correct.
NFSv4 with insufficient permissions: Uploads may still fail, but the error message is clear and actionable.
Related: #3129

### DIFF
--- a/src/wget_agent/agent/wget_agent.c
+++ b/src/wget_agent/agent/wget_agent.c
@@ -115,7 +115,7 @@ void DBLoadGold()
   if ((int)ForceGroup > 0)
   {
     rc = chown(GlobalTempFile,-1,ForceGroup);
-    if (rc) LOG_ERROR("chown failed on %s, error: %s", GlobalTempFile, strerror(errno));
+    if (rc) LOG_ERROR("chown failed on %s, error: %s. If this is an NFSv4 volume, check server-side permissions and user mapping. See https://github.com/fossology/fossology/issues/3129 for details.", GlobalTempFile, strerror(errno));
   }
 
   if (!Sum)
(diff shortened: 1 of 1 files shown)


Kaushl2208:
Not the right way to handle this. Closing.

PR #3611, opened 2026-04-22
fix(copyright): check hash and add distinct variable to method

### LINKED ISSUE #866: License clearing results from clearingDao->getFileClearingsFolder are incorrect

shaheemazmalmmd:
## Description

Closes #866 

### Changes

## How to test

* Upload a component.
* Do general clearing . (licenses and copyrights) with repo decisions and file decisions.
* Reuse copyrights/licenses and check the tree view.

### DIFF
--- a/src/lib/php/Dao/ClearingDao.php
+++ b/src/lib/php/Dao/ClearingDao.php
@@ -184,7 +184,7 @@ class ClearingDao
   {
     $this->dbManager->begin();
 
-    $statementName = __METHOD__;
+    $statementName = __METHOD__ . ($includeSubFolders ? ".subfolders" : ".direct");
 
     if (!$includeSubFolders) {
       $params = array($itemTreeBounds->getItemId());
@@ -231,7 +231,7 @@ class ClearingDao
             LEFT JOIN clearing_decision_event cde ON cde.clearing_decision_fk = decision.id
             LEFT JOIN clearing_event ce ON ce.clearing_event_pk = cde.clearing_event_fk
             LEFT JOIN license_ref lr ON lr.rf_pk = ce.rf_fk
-            ORDER BY decision.id DESC, event_id ASC";
+            ORDER BY decision.id DESC, itemid, event_id ASC";
 
     $this->dbManager->prepare($statementName, $sql);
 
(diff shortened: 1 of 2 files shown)


its-sushant (approved):
LGTM!

PR #3614, opened 2026-04-26
fix(copyright): edge case with no copyright statement jumps before declutter

Kaushl2208:
## Description

In case of components with no copyright data, or failure of copyright agent, False Positive and Declutter script should also not fail.

### Changes

Add a basic empty check to make sure, there is always the right copyright content available before sending to the copyrightDeclutterScript.

## How to test

1. Upload a component (with copyright data)
2. Select copyright, Select Declutter and False Positive (Check if everything is working fine)
3. Upload a component with (no copyright info) or pause copyright agent. Check that decider should not fail and there is no error or warning in the logs.


shaheemazmalmmd (approved):
Code looks good.

PR #3615, opened 2026-04-26
fix(scancode): replace tmpnam with mkstemp to fix warning

Kaushl2208:
## Description
The linker warned that `tmpnam` is dangerous due to a TOCTOU (time-of-check to time-of-use) race condition another process could create a file between the name being generated and the file being opened.

### Changes

1. Replaced both `tmpnam(nullptr)` calls in `processUploadId()` with `mkstemp()`, which atomically creates the file and returns a file descriptor, eliminating the race condition.
2. Added error handling for temp file creation failures.
3. Added cleanup `unlink` for the input temp file `fileLocation`, which was previously leaked  only the output file was being deleted.

## How to test
1. Build should happen without this warning:
image

2. Upload a component and run scancode to test License and Copyright. It should work.

### DIFF
--- a/src/scancode/agent/scancode_utils.cc
+++ b/src/scancode/agent/scancode_utils.cc
@@ -85,8 +85,24 @@ bool processUploadId(const State &state, int uploadId,
 
   bool errors = false;
 
-  string fileLocation = tmpnam(nullptr);
-  string outputFile = tmpnam(nullptr);
+  char fileLocationTpl[] = "/tmp/scancode_input_XXXXXX";
+  int fdInput = mkstemp(fileLocationTpl);
+  if (fdInput == -1) {
+    LOG_ERROR("Failed to create temporary input file.\n");
+    return false;
+  }
+  close(fdInput);
+  string fileLocation(fileLocationTpl);
+
+  char outputFileTpl[] = "/tmp/scancode_output_XXXXXX";
+  int fdOutput = mkstemp(outputFileTpl);
+  if (fdOutput == -1) {
+    LOG_ERROR("Failed to create temporary output file.\n");
+    unlink(fileLocationTpl);
+    return false;
+  }
+  close(fdOutput);
+  string outputFile(outputFileTpl);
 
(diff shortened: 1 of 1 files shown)


src/scancode/agent/scancode_utils.cc
```diff
@@ -85,8 +85,24 @@ bool processUploadId(const State &state, int uploadId,
...
-  string fileLocation = tmpnam(nullptr);
-  string outputFile = tmpnam(nullptr);
+  char fileLocationTpl[] = "/tmp/scancode_input_XXXXXX";
+  int fdInput = mkstemp(fileLocationTpl);
+  if (fdInput == -1) {
+    LOG_ERROR("Failed to create temporary input file.\n");
+    return false;
+  }
+  close(fdInput);
+  string fileLocation = fileLocationTpl;
```
GMishx: Would recommend the explicit constructor call to be clear that the string is duplicated. But it makes no difference in program

```suggestion
  string fileLocation(fileLocationTpl);
```

src/scancode/agent/scancode_utils.cc
```diff
@@ -85,8 +85,24 @@ bool processUploadId(const State &state, int uploadId,
...
+
+  char outputFileTpl[] = "/tmp/scancode_output_XXXXXX";
+  int fdOutput = mkstemp(outputFileTpl);
+  if (fdOutput == -1) {
+    LOG_ERROR("Failed to create temporary output file.\n");
+    unlink(fileLocationTpl);
+    return false;
+  }
+  close(fdOutput);
+  string outputFile = outputFileTpl;
```
GMishx: Same here

```suggestion
  string outputFile(outputFileTpl);
```

GMishx (dismissed):
LGTM

GMishx (approved):
Changes looks good.

shaheemazmalmmd (approved):
Tested works as expected. no warnings observed in build and scancode works as expected.

PR #3616, opened 2026-04-27
feat(bulk): extend include license text functionality

Kaushl2208:
## Description

There is a need to add additional data export, where if the user wants they can extend `license_text` array, where it contains actual license text of all the licenses to add in monkbulk.

### Changes

1. Included a flag `includeLicenseText` to make sure the output has the required data.
2. Added test cases
3. Added a radio check in twig (Additional Data)

## How to test

1. Admin > Bulk > Bulk Text Export 
2. Select the required filters (without additional data)
3. Select Additional data and export.


shaheemazmalmmd (approved):

PR #3620, opened 2026-04-30
fix(copyright): reduce false deactivations and limit stored text

its-sushant:
## Description

fix(copyright): reduce false deactivations and limit stored text

CC: @shaheemazmalmmd


Kaushl2208 (approved):
LGTM, Working as expected.

PR #3624, opened 2026-05-05
chore(ci): fix release information deprecation warning

Kaushl2208:
## Description

`bruceadams/get-release@v1.3.2` is a GitHub Action that runs JavaScript compiled for Node.js 20. GitHub is deprecating Node.js 20 on runners from June 2, 2026, Node.js 24 becomes the default, and Node.js 20 is removed entirely. After that, this action will simply fail.

The action's sole purpose was to fetch two pieces of information from the GitHub API:

`tag_name` : the version tag of the release (e.g. 5.0.0)
`upload_url` :  the URL to attach release asset files to

ERROR:

image

### Changes

When GitHub triggers a workflow from a release event, it automatically injects the full release payload into `github.event.release`. Everything `bruceadams/get-release` was fetching from the API is already available in the event context no API call needed, no Node.js runtime needed

`.github/workflows/release-publish.yml` : Remove Get Release Info stage.

## How to test
1. Check release stage working properly with the correct `tag_name` and `upload_url`

Example: https://github.com/Kaushl2208/fossology/releases/tag/5.0.2-personal

CC: @GMishx

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -69,17 +69,9 @@ jobs:
       run: |
         git fetch --tags
 
-    - name: Get release info
-      id: get_release
-      uses: bruceadams/get-release@v1.3.2
-      env:
-        GITHUB_TOKEN: ${{ github.token }}
-
     - name: Set environment
-      env:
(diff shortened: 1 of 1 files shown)


GMishx (approved):
LGTM

PR #3625, opened 2026-05-05
fix(Bulk): Fix bulk recognition UI in tree view and bulk scan agent query

shaheemazmalmmd:
## Description

use bootstrap hide and show instead of CSS toggle. Change query to pull files for bulk when different scanners finds license and no license found at once.

## How to test

* Upload a component.
* Go to tree view and select multiple folders. 
* Open select option and click on bulk recognation.
* Add license text, acknowledgement and comment.
* Schedule the scan.


Kaushl2208 (approved):
LGTM, Working as expected.

PR #3626, opened 2026-05-08
fix(bulkScan): use uploadtreetablename, fix logic to get latest decis…

### LINKED ISSUE #1245: License decisions not shown correctly for identical files in all cases
### Description

Edited results are not shown correctly in License Browser view in all cases. When the decision is done as Global it should be visible in all identical file occurrences but for some file instances decisions are shown wrongly. This bug has been already once corrected in https://github.com/fossology/fossology/commit/b411cbdf096127775c3ccb8629de3ecbd7d45985#diff-4889a5c519f5c97bd1c9d69399937b74. It seems that this commit breaks the behaviour https://github.com/fossology/fossology/commit/283352a124a1ac84a6e22410865440993e1bcec2#diff-4889a5c519f5c97bd1c9d69399937b74.

#### How to reproduce

There is no one procedure to reproduce this. However, using the code from this commit https://github.com/fossology/fossology/commit/3105198dd4f9ece8fa9c756e570fa0395566fd87#diff-4889a5c519f5c97bd1c9d69399937b74 seems to correct the behaviour. The code copied below.

```
@@ -264,10 +265,8 @@ class ClearingDao
 
         $firstMatch = false;
         //prepare the new one
-        if($forClearingHistory){
-          $previousClearingId = $clearingId;
-        }
-        else{
+        $previousClearingId = $clearingId;
+        if(!$forClearingHistory){
           $previousItemId = $itemId;
         }
         $clearingEvents = array();
```
### Versions

* 3.3.0 onwards

shaheemazmalmmd:
## Description

* If the clearing decision is changed multiple times then the clearing history is not correct.
* If the clearing decision is changed from local to global then the edit licenses in tree view doesn't contain all the licenses.
* Make use of Uploadtree_a and uploadtree_uploadid to fetch tree bonds. (faster when large uploads) 

### Changes

* Change logic to fetch clearing hostory.
* Remove scope from the order clause.
* add upload id and change the item bonds function name.

## How to test

* Upload a component.
* Do clearing using single file view. 
* Now change the license and decision type to repo. 
* Go to tree View and check the edited license, also check the clearing history.

Closes #1245

### DIFF
--- a/src/deciderjob/agent/DeciderJobAgent.php
+++ b/src/deciderjob/agent/DeciderJobAgent.php
@@ -113,7 +113,7 @@ class DeciderJobAgent extends Agent
   /**
    * @brief Process clearing events of current job handled by agent
    */
-  function processClearingEventOfCurrentJob()
+  function processClearingEventOfCurrentJob($uploadId)
   {
     $userId = $this->userId;
     $groupId = $this->groupId;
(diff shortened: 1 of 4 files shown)


its-sushant (approved):
LGTM!

PR #3627, opened 2026-05-10
refactor(nomos): pre-compiling nomos regex patterns once at startup i…

shaheemazmalmmd:
…nstead

## Description

Pre-compile nomos regex at once instead of everytime. 

### Before Fix

image

### After Fix

image

## How to test

* do general testing using nomos agent


its-sushant (approved):
LGTM!

PR #3646, opened 2026-05-23
feat(compare): add feature to compare multiple components

shaheemazmalmmd:
## Description

Use existing compare feature and implement multi component compare.

image

## How to test

* Upload a component in fossology
* Go to browse and select compare.
* Select the folder where the other component is which you wish to compare.
* Select upload you can also use search if the upload count is high.
* And on the top right click compare now.
* Use freeze/ check copyrights/ licenses and use different filters.

Note: This feature only works with scanner results.


Kaushl2208 (approved):
Looks Good. Working as expected.

PR #3649, opened 2026-05-26
fix(test): dangling reference pointer failing test

Kaushl2208:
## Description

CTEST for nomos failing for `debian:trixie`. There is a dangling pointer which leads to the error.

### Changes

Remove unwanted `report_Match(undoc)`


GMishx (approved):
Good catch

PR #3653, opened 2026-05-30
fix(copyright): add detections for .net case ccopyrights

shaheemazmalmmd:
## Description

Fix false positives and false negatives and add test cases.
 
## How to test

* add the below copyright to one file.
* Copyright (c) 2021 .NET Foundation


its-sushant (approved):
Reviewed and tested. LGTM!

PR #3609, opened 2026-04-21
feat(reuser): porting reuser from php to c++

helmut-leher-magna:
**open for discussion**

See the Issue Issue #3585

-----

Migrate `reuser` agent from PHP to C++, introducing a new C++ agent implementation. 
As a role model the architecture of the `ojo` agent (C++17) was used. 

The changes include new C++ source files for the agent’s logic, state, database handling, and utility functions, as well as updates to the build system (compile and install the agent). 
The PHP agent entrypoint is removed, and packaging/dependency definitions are updated.

**Major migration to C++ agent:**

* Added new C++ agent entrypoint `reuser.cc`, implementing the main scheduler event loop and agent orchestration.
* Introduced core agent modules:
  - `ReuserDatabaseHandler` for all database operations
  - `ReuserState` for agent runtime state
  - `ReuserTypes.hpp` for agent data types and constants
  - `ReuserUtils` for agent utility functions and scheduler integration
  
**Build system and packaging updates:**

* Updated `src/reuser/agent/CMakeLists.txt` to build and install the new C++ agent, link dependencies, and handle versioning.
* Updated `Dockerfile` to copy `mod_deps` for the `reuser` module.
* Updated `debian/control` to declare shared library dependencies for the new agent.
* Set C++ as the required language for the `reuser` CMake project.

**Removal of the PHP agent entrypoint `reuser.php`**

**General improvements and safeguards:**

* Improved symlink installation logic in CMake to avoid overwriting native files.
* Added a null check in the C++ database manager cleanup to prevent errors.

PR #3655, opened 2026-06-01
fix(reuser): add getDecisionTypePriority for same pfile

sakshammishra112:
## Description

This PR adds the `getDecisionTypePriority()` function to correctly handle cases where the same pfile exists at multiple locations within a package and has different clearing decision types (e.g., `IDENTIFIED`, `IRRELEVANT`).  see PR #3587.

Previously, the C++ reuser applied the latest clearing decision found for a pfile.

Works on top of PR #3609 .

#### 1. Reuse Skips Already-Cleared Files

The C++ reuser now skips applying reuse to files that already have clearing decisions. This prevents reuse data from being re-applied to files that have already been cleared.

#### 2. Decision Priority Handling

A new `getDecisionTypePriority()` function has been introduced to determine the effective clearing decision when the same pfile appears in multiple locations with different decision types. This ensures that the most appropriate decision is selected instead of simply using the latest one.

#### 3. Heartbeat Handling
The `reuserCopyrights()` function currently does not invoke `fo_scheduler_heart()` for each processed event, whereas the PHP reuser calls `$this->heartbeat(1)` inside the processing loop (see `src/reuser/agent/ReuserAgent.php`, line [242](https://github.com/fossology/fossology/blob/3eac9efd81719c5e2e4a1a0b97da4ff96d012842/src/reuser/agent/ReuserAgent.php#L242)).
Added `fo_scheduler_heart(1)` in `reuserCopyrights()` after each event.

This difference has been noted during implementation and comparison with the PHP reuser.

#### 4. CMakeLists.txt

The ```install(TARGETS ...)``` command does not support the ```RENAME``` option (that only works with ```install(FILES ...)``` and ```install(PROGRAMS ...))```. Previously, the executable was named reuser_exec and a ```RENAME``` reuser was used to work around the target name collision with ```generate_version_php()``` creating a custom target also called reuser. Since ```RENAME``` was silently ignored, ```cmake --install``` installed the binary as ```reuser_exec``` instead of ```reuser```, leaving the old reuser binary untouched.

One needed to manually do:
```bash
sudo cp build/src/reuser/agent/reuser_exec /usr/local/share/fossology/reuser/agent/reuser
sudo chmod +x /usr/local/share/fossology/reuser/agent/reuser
```
**Fix:** Name the custom version target explicitly ```(reuser_version_php)``` and rename the executable target to reuser directly — **no collision**, no ```RENAME``` needed, ```cmake --install``` produces the correct binary name.

<details>
<summary>Build Before changes</summary>

Screenshot 2026-06-11 001215-imageonline co-merged

</details>

<details>
<summary>Build After changes</summary>

Screenshot 2026-06-11 001724-imageonline co-merged

</details>

## Testing

The C++ reuser was tested against the PHP reuser using the following uploads:

- `linux-6.7.tar.xz`
- `linux-6.9.tar.xz`

Reuse was performed on:

- `linux-6.8.tar.xz`

### Functional Verification

The following reuse modes were tested and behaved as expected, matching the PHP reuser:

- `REUSE_MAIN`
- `REUSE_CONF`
- `REUSE_ENHANCED`
- `REUSE_COPYRIGHT`

Additionally:

- Unit tests pass successfully.
- Debian packaging changes were verified and look good.

### Performance Comparison

The same test scenario was executed three times under identical conditions, and the mean elapsed time was recorded.

| Reuser | Mean Elapsed Time |
|---------|------------------|
| PHP Reuser | `00:04:05.325627` |
| C++ Reuser | `00:03:52.083240` |

**Improvement:** ~13.24 seconds faster than the PHP reuser.

### PHP Reuser

<details>
<summary>Screenshot</summary>

phpReuser

</details>

### C++ Reuser

<details>
<summary>Screenshot</summary>

cppReuser

</details>

cc: @shaheemazmalmmd @Kaushl2208


sakshammishra112:
Rebased on top of `helmut/feat/reuser-php-migration `

shaheemazmalmmd (dismissed):

its-sushant (approved):
Reviewed and tested. LGTM!

Issue #3631, opened 2026-05-13
Enhanced Reuse Agent for Intelligent License Reuse


shaheemazmalmmd:
- [ ] Auto select reusable component from the fossology database intelligently.
       - Along with select reuse information there shall be a option to select auto select the component for reuse.
       - Once the auto select option is checked -> finding of the nearest component shall be done in reuse agent.
       - The selection process of the nearest upload shall be done by (name of the component, uploaded by the uploader/ the groups of the uploader)
       - if the name is not available then we shall have a functionality to compare pfile of uploaded component and fossology database(clearing_decision)
       - if the component is found then reuse shall happen with all the available options along with bulk. if component is not found the reuse agent shall have a fall back stating nothing has been found.

- [ ] Enhance reuse shall be a separate agent to read the files that were not identified by reuse/bulk scan.
       - Nirja's comments classification needs to be integrated to extract comment from both the files.
       - Use basic similarity and fuzzy logic to match the license comments from both the files stays in same path and have same license.  
       - if there are any changes in comments then no decision will be applied. otherwise, we can apply the decisions.

- [ ] Reuse-UI
       - Show number of licenses detected (Histogram) in v1 vs v2
          - Highlight whether new licenses were introduced or existing ones changed (in different colors)
       - Draw statistics (% of new licenses added compared to scanner findings in old upload)

- [ ] REST-Level Transparency for all the features.

sakshammishra112:
I will be working on this

shaheemazmalmmd:
@sakshammishra112 Thanks for all your efforts on this.

PR #3676, opened 2026-06-13
feat(reuser): auto select component for reuse

sakshammishra112:
## Description

Introduce an **_Auto Select_** for Reuse option in Reuser Configuration that automatically discovers and selects the most suitable previously cleared upload for reuse. The feature searches for reusable uploads using package name matching and PFile overlap analysis before job(reuser) scheduling, then executes the existing reuse workflow without requiring manual upload selection.

<Details>
<summary> Primary Search </summary>

# Primary Search: Matching Uploaded Components to Cleared Uploads

## Step 1 — Extract Base Package Name

Strip the version suffix from the uploaded package name.

| Uploaded Package | Base Package Name |
|-----------------|-------------------|
| `zlib-1.2.10` | `zlib` |

---

## Step 2 — Search for Matching Uploads

Matches are attempted in order, stopping as soon as candidates are found:

| Priority | Match Type | Example |
|----------|------------|---------|
| 1 | **Exact match** | `zlib` → `zlib` |
| 2 | **Case-insensitive exact** | `ZLIB` → `zlib` |
| 3 | **Prefix match** | `zlib` → `zlib-ng` |
| 4 | **ILIKE fallback** | `%zlib%` (partial string) |

---

## Step 3 — Apply Search Priority

Within each match type, candidates are ranked by uploader relationship:

1. Same uploader
2. Same group(s)
3. Any accessible upload

> **Example:** If `zlib` matches Uploads A (same uploader), B (same group), and C (accessible) — only Upload A is considered.

---

## Step 4 — Resolve Ties

If multiple uploads share the same match type and priority level, the **most recently cleared** upload wins.

| Upload | Last Cleared |
|--------|-------------|
| Upload A | Jan 10 |
| Upload B | Feb 15 ✓ |

Upload B is selected.
</Details>

<Details>
<summary> Secondary Search </summary>

# Secondary Search: PFile Overlap Search

> Only executed if the **primary search** finds no suitable upload.
> Instead of comparing package names, it compares actual upload contents via PFiles.

---

## Step 1 — Identify Candidate Uploads

Gather all previously cleared uploads available for reuse.

---

## Step 2 — Calculate PFile Overlap

Compare the uploaded component's PFiles against each candidate's PFiles.

**Current Upload PFiles:** `a.c`, `b.c`, `c.c`, `d.c`

| Candidate | Matching PFiles | Overlap Score |
|-----------|----------------|---------------|
| Upload A | `a.c`, `b.c`, `c.c` | 3 |
| Upload B | `a.c`, `b.c`, `c.c`, `d.c` | 4 |
| Upload C | `a.c` | 1 |

---

## Step 3 — Select Highest Overlap

The candidate with the highest overlap score is selected.

> **Example:** Upload B scores 4 — the highest — and is selected.

---

## Step 4 — Resolve Ties

If multiple candidates share the same overlap score, select the **most recently cleared** upload.

| Upload | Overlap | Last Cleared |
|--------|---------|-------------|
| Upload A | 10 | Jan 20 |
| Upload B | 10 | Mar 05 ✓ |

Upload B is selected.
</Details>

> **Note**
>
> Both the primary and secondary searches only consider candidate uploads whose clearing status is **Closed**.
> There is a limit of a maximum of 50 to consider as candidate uploads in the secondary search.

Part of issue: #3631 

### Testing

<Details>
<summary> Primary Search </summary>

https://github.com/user-attachments/assets/6ab82d98-ecf4-4486-8240-10a74d588257

</Details>

<Details>
<summary> Secondary Search </summary>

Screenshot 2026-06-12 230347
Screenshot 2026-06-12 230357
Screenshot 2026-06-12 230320
Screenshot 2026-06-12 230335

Performed the **reuse** analysis on `wild.zip` and since `cigar.zip` had **_7 pfle_** in common, it was selected by **_auto select_** :

Screenshot 2026-06-13 003636

</Details>

<Details>
<summary> REST api </summary>

Screenshot 2026-06-12 213431
Screenshot 2026-06-12 183227
</Details>

cc: @shaheemazmalmmd @Kaushl2208

### DIFF
--- a/src/reuser/ui/ReuseAutoDetect.php
+++ b/src/reuser/ui/ReuseAutoDetect.php
@@ -1,0 +1,303 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © Fossology contributors
+
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+
+namespace Fossology\Reuse;
+
+use Fossology\Lib\Dao\ClearingDao;
+use Fossology\Lib\Db\DbManager;
+
+/**
+ * @class ReuseAutoDetect
+ * @brief Pure decision logic for auto-selecting a reuse upload
+ *
+ * Kept free of Auth/container/global state so it can be unit tested
+ * directly. The UI plugin passes in the current user id, the db manager
+ * and the clearing dao as plain parameters.
+ */
+class ReuseAutoDetect
+{
+  /**
+   * @brief Parse package name and version from a filename
+   * @param string $filename
+   * @return array ['baseName'=>string, 'versionParts'=>int[]|null,
+   *                'prerelease'=>int]
+   */
+  public static function parsePackageName($filename)
+  {
+    if (empty($filename)) {
+      return ['baseName' => '', 'versionParts' => null, 'prerelease' => 0];
+    }
+    $nameWithoutExt = preg_replace('/\.[^.]+$/', '', $filename);
+    $nameWithoutExt = preg_replace('/\.(tar|zip|gz|bz2|xz|tgz|tbz2|txz|rar|7z)(\..*)?$/i', '', $nameWithoutExt);
+    $nameWithoutExt = preg_replace('/\.(tar|zip|gz|bz2|xz|tgz|tbz2|txz|rar|7z)$/i', '', $nameWithoutExt);
+    $nameWithoutExt = preg_replace('/[-_]\d{8}$/', '', $nameWithoutExt);
+    if (preg_match('/[-_](v?\d+(?:\.\d+)*)(?:[-_](?:alpha|beta|rc|pre|patch|p)\d*)?$/i', $nameWithoutExt, $m)) {
+      $baseName = substr($nameWithoutExt, 0, -strlen($m[0]));
+      $versionParts = array_map('intval', explode('.', preg_replace('/^v/i', '', $m[1])));
+      $prerelease = preg_match('/[-_](?:alpha|beta|rc|pre|patch|p)\d*$/i', $nameWithoutExt) ? 1 : 0;
+      return ['baseName' => $baseName, 'versionParts' => $versionParts, 'prerelease' => $prerelease];
+    }
+    $parts = explode('-', $nameWithoutExt);
+    return ['baseName' => $parts[0], 'versionParts' => null, 'prerelease' => 0];
+  }
+
+  /**
+   * @brief Digits of |parts - requestedParts| in base $R
+   * @param int[] $parts
+   * @param int[] $requestedParts
+   * @param int $L Common padded length
+   * @param int $R Radix (larger than any component)
+   * @return int[] Most-significant-first digits of the absolute difference
+   */
+  private static function computeVersionDistance($parts, $requestedParts, $L, $R)
+  {
+    $digits = [];
+    $borrow = 0;
+    for ($i = $L - 1; $i >= 0; $i--) {
+      $v = (isset($parts[$i]) ? $parts[$i] : 0)
+        - (isset($requestedParts[$i]) ? $requestedParts[$i] : 0) - $borrow;
+      if ($v < 0) {
+        $v += $R;
+        $borrow = 1;
+      } else {
+        $borrow = 0;
+      }
+      $digits[$i] = $v;
+    }
+    if ($borrow === 1) {
+      for ($i = 0; $i < $L; $i++) {
+        $digits[$i] = ($R - 1) - $digits[$i];
+      }
+      $carry = 1;
+      for ($i = $L - 1; $i >= 0; $i--) {
+        $s = $digits[$i] + $carry;
+        if ($s >= $R) {
+          $digits[$i] = $s - $R;
+          $carry = 1;
+        } else {
+          $digits[$i] = $s;
+          $carry = 0;
+        }
+      }
+    }
+    return $digits;
+  }
+
+  /**
+   * @brief Compare two version tuples for numeric closeness to a requested version
+   *
+   * Measures |a - requested| and |b - requested| as numbers in base $R
+   * (larger than any single component), so a difference in a higher
+   * component correctly outweighs differences in lower ones. For example,
+   * requested 1.0 ranks 0.9 closer than 0.8 (0.1 apart vs 0.2 apart).
+   * @param int[]|null $requestedParts Requested version tuple
+   * @param int[]|null $aParts
+   * @param int[]|null $bParts
+   * @param int $L Common padded length of all version tuples
+   * @param int $R Radix (max component + 1, at least 10)
+   * @return int negative if $aParts is closer, positive if $bParts is closer
+   */
+  public static function compareVersionCloseness($requestedParts, $aParts, $bParts, $L, $R)
+  {
+    if ($requestedParts === null) {
+      return 0;
+    }
+    if ($aParts === null && $bParts === null) {
+      return 0;
+    }
+    if ($aParts === null) {
+      return 1;
+    }
+    if ($bParts === null) {
+      return -1;
+    }
+    $distA = self::computeVersionDistance($aParts, $requestedParts, $L, $R);
+    $distB = self::computeVersionDistance($bParts, $requestedParts, $L, $R);
+    for ($i = 0; $i < $L; $i++) {
+      if ($distA[$i] !== $distB[$i]) {
+        return $distA[$i] <=> $distB[$i];
+      }
+    }
+    return 0;
+  }
+
+  /**
+   * @brief Rank candidates by nearest version, then release, then clearing/upload date
+   * @param array $candidates Each with versionParts, prerelease, clearedAt, timestamp
+   * @param int[]|null $requestedVersionParts
+   */
+  public static function rankCandidates(&$candidates, $requestedVersionParts)
+  {
+    $L = 0;
+    $R = 10;
+    $all = $candidates;
+    if ($requestedVersionParts !== null) {
+      $all[] = ['versionParts' => $requestedVersionParts];
+      $L = count($requestedVersionParts);
+    }
+    foreach ($all as $candidate) {
+      $parts = $candidate['versionParts'];
+      if ($parts === null) {
+        continue;
+      }
+      $L = max($L, count($parts));
+      foreach ($parts as $component) {
+        $R = max($R, $component + 1);
+      }
+    }
+    usort($candidates, function ($a, $b) use ($requestedVersionParts, $L, $R) {
+      if ($requestedVersionParts !== null) {
+        $cmp = self::compareVersionCloseness($requestedVersionParts,
+          $a['versionParts'], $b['versionParts'], $L, $R);
+        if ($cmp !== 0) {
+          return $cmp;
+        }
+        if ($a['prerelease'] !== $b['prerelease']) {
+          return $a['prerelease'] <=> $b['prerelease'];
+        }
+      }
+      $tsA = empty($a['clearedAt']) ? null : strtotime($a['clearedAt']);
+      $tsB = empty($b['clearedAt']) ? null : strtotime($b['clearedAt']);
+      if ($tsA !== null && $tsB !== null && $tsA !== $tsB) {
+        return $tsB - $tsA;
+      }
+      if ($tsA !== null) {
+        return -1;
+      }
+      if ($tsB !== null) {
+        return 1;
+      }
+      return ($b['timestamp'] ?? 0) - ($a['timestamp'] ?? 0);
+    });
+  }
+
+  /**
+   * @brief Select the single best candidate based on clearing decisions
+   *
+   * Prefers the first fully cleared candidate in rank order, then the
+   * candidate with the highest cleared-files percentage, and finally falls
+   * back to the first candidate (best by version/date).
+   * @param array $candidates Each with uploadId and groupId, ranked best-first
+   * @param ClearingDao $clearingDao
+   * @return array|null Selected candidate or null
+   */
+  public static function selectWinnerByClearing($candidates, ClearingDao $clearingDao)
+  {
+    if (empty($candidates)) {
+      return null;
+    }
+    $groups = [];
+    foreach ($candidates as $candidate) {
+      $groups[$candidate['groupId']] = true;
+    }
+    $coverage = [];
+    foreach (array_keys($groups) as $groupId) {
+      $ids = [];
+      foreach ($candidates as $candidate) {
+        if ($candidate['groupId'] === $groupId) {
+          $ids[] = $candidate['uploadId'];
+        }
+      }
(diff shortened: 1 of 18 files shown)


sakshammishra112:
- Shifted the auto-select search from the reuser agent to the UI, now users can see the auto-selected package. 
 - Added a configuration section in Admin → Customize → Configuration Variables where users can toggle auto-select and choose which clearing statuses of upload to search. Auto-select runs when the user clicks "Set the reuse information" on the Upload a File page. The matched package is inserted into the selection list. 
 - Removed secondary search (PFile overlap) since the upload doesn't exist in the DB yet at UI time, making comparison impossible. Primary search uses ILIKE '%packagename%' to fetch candidates for staged matching in each priority tier(this will increase the chances of getting the correct package).Added permission check on tiers 2 and 3 — user must have WRITE access (group_perm >= 2) on the candidate's group to see it as a reuse option.
 
@shaheemazmalmmd PTAL

shaheemazmalmmd:
@sakshammishra112 The changes doesn't work on my end.

Here is how i tested it.

* Create a component NewLicenses.zip Upload to fossology and do clearing.
* Upload the same NewLicenses.zip again and select reuse no auto select happened.
* Create a new component NewLicenses-1.0.zip upload again to select the reuse it didnt work.
* Now for NewLicenses.zip change the status in browse to closed and followed all the three actions above. it didnt work

I repeated the same for components : nltk-3.9.3.tar.gz and nltk-3.10.0.tar.gz and also fck editor 2.3 and 2.11

it didnt work. Please check,

sakshammishra112:
@shaheemazmalmmd should be working now. Tested your flow and also tested with zlib components with different statuses.
<Details>
<summary> Tested with your flow(NewLicenses.zip) </summary>

https://github.com/user-attachments/assets/1d0faaf7-72d6-41bb-b475-2ee4409afb68

</Details>

sakshammishra112:
## Overview
A new feature that automatically matches and selects previously uploaded files for reuse based on intelligent filename matching and access control. The system provides instant client-side matching with a server-side fallback for edge cases, ensuring users can quickly find related uploads without manual searching.
 
---
 
<details>
<summary><strong>1. Response Data Restructuring</strong></summary>
 

**Before:** Flat map structure
```
{"42" : "nltk-3.7.tar.gz from 2026-08-19 09:12:28 (closed)"}
```
 
**After:** Structured array of objects via `buildStructuredUploads()` (reuser-plugin.php:258)
```json
    {
        "uploadId": 42,
        "groupId": 3,
        "filename": "nltk-3.7.tar.gz",
        "timestamp": "2026-08-19 09:12:28",
        "statusId": 3,
        "clearedAt": null,
        "display": "nltk-3.7.tar.gz from 2026-08-19 09:12:28 (closed)"
    },
``` 

### Scope Options
 
- **Group-scoped:** When `&folder=fid,tgid` is passed (checkbox checked)
- **All uploads:** Otherwise includes all uploads accessible to the group

</details>
 
<details>
<summary><strong>2. Client-Side Search (Primary Path)</strong></summary>
 
### Matching Process
 
#### Step 1: Cache Management
- Structured response stored in `lastGetUploadsDataByGroup[groupIndex]`
- Per-row caching prevents multi-file uploads from overwriting each other
#### Step 2: Filename Normalization
`extractBasePackageNameFromFilename()` applies the following rules in order:
1. Strip file extensions (e.g., `foo.tar.gz` → `foo`)
2. Remove version suffixes (e.g., `foo-1.2.3`, `foo-v2.0-rc1`)
3. Strip date stamps
4. If no match, extract first dash-token
#### Step 3: 4-Tier Matching Algorithm (agent_reuser.js.twig:78)
Matches in precedence order:
1. **Exact** — Case-sensitive match
2. **Case-insensitive** — Ignores case
3. **Prefix** — Normalized name is a prefix
4. **Substring** — Normalized name appears anywhere
#### Step 4: Status Filtering
- Only includes uploads with `statusId` in `ReuserSearchStatus` config
- Default: `[3]` (closed status only)
- Configurable by admin via Customize Variables page
#### Step 5: Ranking
1. `clearedAt` — Descending (nulls last)
2. `timestamp` — Descending (most recent first)
#### Step 6: Selection
- Top 3 matches inserted into dropdown via `$select.val([...]) + trigger('change')`
- All processing happens in-browser — no server request
</details>
 
<details>
<summary><strong>3. Server-Side Fallback</strong></summary>

**Triggered only when client-side search returns no results.**
 
### Endpoint
```
POST ?mod=plugin_reuser&do=autoDetectReuse&filename=...&folder=...
```
 
### Two Execution Paths
 
#### Path A: No Folder Selected
`primarySearchFromPlugin()` (reuser-plugin.php:496)
 
Uses three SQL tiers, each tagged with tier level + match type:
 
1. **Same Uploader** — Uploads by the same user
2. **Same Group** — Uploads in the same group (with permission check)
3. **Any Accessible** — All uploads the user can access
**Ranking logic:**
- Tier (sameUploader > sameGroup > anyAccessible)
- Match type (exact > case-insensitive > prefix > substring)
- Recency (`clearedAt` descending, then `timestamp` descending)
- Limit: Top 3 results
#### Path B: Folder Selected
`autoDetectInFolder()` (reuser-plugin.php:399)
 
**Guarantees folder-scoped results:**
- Bypasses tier queries
- Loads only uploads from the specified folder via `prepareFolderUploads()`
- Applies identical 4-tier matching (in PHP)
- Same status filtering + clearing date ranking
- No results from outside the folder
### Result Processing
- Results inserted at the top of dropdown (deduplicated against existing options)
- Results are automatically selected
</details>
 
<details>
<summary><strong>5. Folder-Scoping Guarantee</strong></summary>
 
| Context | Client-Side | Server-Side | Result |
|---------|-------------|-------------|--------|
| Folder checked | Loads folder-scoped data | `autoDetectInFolder()` | Only folder uploads |
| Folder unchecked | Loads all group uploads | `primarySearchFromPlugin()` | All accessible uploads |
 
</details>

<details>
<summary><strong>6. User Experience Flow</strong></summary>

```
User uploads new file
    ↓
System extracts normalized filename
    ↓
Client-side search runs instantly (in-browser)
    ├─ Match found? → Auto-select top 3, highlight best match
    └─ No match found? → POST to server fallback
         ↓
      Server searches using tiers + permissions
         ↓
      Insert results at dropdown top, auto-select best match
```
 
</details>
 
 PTAL: @shaheemazmalmmd

shaheemazmalmmd:
@sakshammishra112: I see that the tool is currently selecting all best available components/uploads insted of one upload.

image

The expected behavior should be to select one **best matching upload(s)** based on the following criteria:

### First Level – Eligibility

From all available uploads, consider only uploads that meet **both** of the following criteria:

* Uploaded by the same user.
* The upload belongs to a group to which the user has access.

### Second Level – Ranking

#### If the user has selected a specific upload status

Select **one best matching upload** based on:

* Date.
* Nearest matching version, with **version taking precedence over date**.

For example, if the requested version is `upload-1.0` and both `upload-1.0` and `upload-0.9` are available, `upload-1.0` should be preferred even if `upload-0.9` was uploaded more recently.

So, an exact/nearest version match should be preferred over a newer upload with a less relevant version.

#### If the user has NOT selected a specific upload status

Select the **4 best matching uploads** using the same ranking principle:

* Nearest matching version.
* Then date.

For example, if `upload-1.0` is available, it should be preferred over a more recently uploaded and sort will be `upload-1.0`, `upload-0.9`, `upload-0.8`, `upload-0.7`.

### Third Level – Clearing Decisions

For the 3–4 best matching uploads identified above:

1. Check the `clearing_decisions` for the first/best matching upload.
2. If clearing decisions are available and **all files are cleared**, select this upload.
3. If not, check the next best matching upload using the same criteria.
4. Continue this evaluation for the selected 4 uploads.
5. If none of the uploads is fully cleared, select the upload that has the **highest percentage of files cleared**, provided clearing decisions are available.
6. If **no clearing decisions are available for any of the selected uploads**, fall back to the behavior used when the user has selected a specific status.  i.e., select the single best match based on version and date.

The important point is that the tool should **not select all available uploads**. It should first narrow down the candidates based on user/group access, then rank them by version/date, and finally use `clearing_decisions` to make the final selection when no specific status has been selected.

src/www/ui/api/documentation/openapi.yaml
```diff
@@ -5913,9 +5913,10 @@ components:
               items:
                 type: integer
           description: The UploadID(s) to reuse. Single integer or array of integers.
+          example: 42
         reuse_group:
           type: string
-          description: The group of the reused upload
+          description: The group of the reused upload.
```
Kaushl2208, line 5919: I know this is not a code related change. But this is not in the scope.

src/www/ui/api/Models/Reuser.php
```diff
@@ -267,21 +267,21 @@ public function getArray($version = ApiVersion::V1)
...
-        "reuse_main"      => $this->reuseMain,
-        "reuse_enhanced"  => $this->reuseEnhanced,
-        "reuse_report"    => $this->reuseReport,
-        "reuse_copyright" => $this->reuseCopyright
+        "reuse_upload"     => $this->reuseUpload,
+        "reuse_group"      => $this->reuseGroup,
+        "reuse_main"       => $this->reuseMain,
+        "reuse_enhanced"   => $this->reuseEnhanced,
+        "reuse_report"     => $this->reuseReport,
+        "reuse_copyright"  => $this->reuseCopyright
```
Kaushl2208, line 284: Dont change the formatting. Refrain from doing out of scope changes.

src/www/ui_tests/api/Models/ReuserTest.php
```diff
@@ -41,12 +41,12 @@ public function testConstructor()
...
-      "reuse_main"     => true,
-      "reuse_enhanced" => false,
-      "reuse_copyright" => false,
-      "reuse_report"   => false
+      "reuse_upload"     => 2,
+      "reuse_group"      => 'fossy',
+      "reuse_main"       => true,
+      "reuse_enhanced"   => false,
+      "reuse_copyright"  => false,
+      "reuse_report"     => false
```
Kaushl2208, line 49: Here as well.

src/www/ui_tests/api/Models/ReuserTest.php
```diff
@@ -148,21 +148,21 @@ private function testSetUsingArray($version)
...
-        "reuseMain"     => 'true',
-        "reuseEnhanced" => false,
-        "reuseCopyright" => false,
-        "reuseReport"   => false
+        "reuseUpload"     => 2,
+        "reuseGroup"      => 'fossy',
+        "reuseMain"       => 'true',
+        "reuseEnhanced"   => false,
+        "reuseCopyright"  => false,
+        "reuseReport"     => false
```
Kaushl2208, line 165: ...

src/reuser/ui/ReuseAutoDetect.php
```diff
@@ -0,0 +1,303 @@
...
+   */
+  public static function parsePackageName($filename)
+  {
+    if (empty($filename)) {
+      return ['baseName' => '', 'versionParts' => null, 'prerelease' => 0];
+    }
+    $nameWithoutExt = preg_replace('/\.[^.]+$/', '', $filename);
+    $nameWithoutExt = preg_replace('/\.(tar|zip|gz|bz2|xz|tgz|tbz2|txz|rar|7z)(\..*)?$/i', '', $nameWithoutExt);
+    $nameWithoutExt = preg_replace('/\.(tar|zip|gz|bz2|xz|tgz|tbz2|txz|rar|7z)$/i', '', $nameWithoutExt);
+    $nameWithoutExt = preg_replace('/[-_]\d{8}$/', '', $nameWithoutExt);
```
Kaushl2208, line 37: This is scary and essentially very repetitive. Do we actually need three separate regex passes just to determine the archive extension??
Improve the approach slightly, Instead of cleanup try parsing and identifying (Extension).

This can be easily understood by answering: What filename does the system actually support?

src/reuser/ui/ReuseAutoDetect.php
```diff
@@ -0,0 +1,303 @@
...
+    if (preg_match('/[-_](v?\d+(?:\.\d+)*)(?:[-_](?:alpha|beta|rc|pre|patch|p)\d*)?$/i', $nameWithoutExt, $m)) {
+      $baseName = substr($nameWithoutExt, 0, -strlen($m[0]));
+      $versionParts = array_map('intval', explode('.', preg_replace('/^v/i', '', $m[1])));
+      $prerelease = preg_match('/[-_](?:alpha|beta|rc|pre|patch|p)\d*$/i', $nameWithoutExt) ? 1 : 0;
```
Kaushl2208, line 41: What does `'prerelease' => 1/0` helps with ??

src/reuser/ui/ReuseAutoDetect.php
```diff
@@ -0,0 +1,303 @@
...
+          $digits[$i] = $s - $R;
+          $carry = 1;
+        } else {
+          $digits[$i] = $s;
+          $carry = 0;
+        }
+      }
+    }
+    return $digits;
+  }
```
Kaushl2208, line 88: What is the business usecase here? Why do we have our own arithmetic logic for versions as integers????

I like the way `arbitrary-radix subtraction algorithm` is implemented here but I think there is a conceptual bug here. If we do actually need distance, we can go scalar, computation and implementation would be straightforward.

src/reuser/ui/ReuseAutoDetect.php
```diff
@@ -0,0 +1,303 @@
...
+      return 0;
+    }
+    if ($aParts === null) {
+      return 1;
+    }
+    if ($bParts === null) {
+      return -1;
+    }
+    $distA = self::computeVersionDistance($aParts, $requestedParts, $L, $R);
+    $distB = self::computeVersionDistance($bParts, $requestedParts, $L, $R);
```
Kaushl2208, line 119: We need to think about this as well. Which of the following question we are trying to answer:

```
requested: 1.2.3

available:
1.0.0
1.1.0
1.2.0
1.2.5
2.0.0
```
What should happen?

if answer is: `1.2.5` --> Find the smallest version ≥ requested.
if answer is: `1.2.0` --> Find the largest version ≤ requested.

I think [SemVer](https://github.com/composer/semver) could handle most of this for us. It wouldn't solve the entire license-decision reuse problem, but it would give us proper semantic version parsing and comparison instead of treating version components like integers.

We should also think about cases like 1.2.99 vs 1.3.0 and what “closest” should actually mean for our license-decision reuse use case.

Kaushl2208 (changes requested):
Hey @sakshammishra112 , This algorithm in `src/reuser/ui/ReuseAutoDetect.php` this needs to be optimised, I think we can do this in more modular way instead of writing logics by ourselves. That leaves some of the gaps/bugs open. 

I have left those comments as a part of discussion point (As a DIARY OF THOUGHTS FOR FUTURE IMPROVEMENTS), All the other comments you can resolve for now. 
Lets discuss this problem a bit better.

Issue #3635, opened 2026-05-13
Improve CycloneDX Support and Update CycloneDX Exports to Spec Version 1.7


shaheemazmalmmd:
- [ ] Identify and address gaps in the currently implementation (currently license texts and copyright notices seem to be missing)
- [ ] Identify and address requirements from other tools, so exports can easily be reused
- [ ] Update to spec version 1.7 (Currently spec version 1.4 is supported)
- [ ] Support importing CycloneDX documents
       - Same as SPDX import there shall be possibility to import the cycloneDX reports and apply the decisions.
       - Report Import agent shall be used and extended.

- [ ] Improve the Compatibility agent. 
       - Currently fossology supports pulling of compatibility Matrix from OSADL(https://www.osadl.org/Access-to-raw-data.oss-compliance-raw-data-access.0.html) manually.
       - This shall be automated via post install.
       - Current feature doesn't capture "Rule Description", "unknown compatibility" it shall be added.
       - Compatibility agent works on file level only. it shall be component level(use tree view and replace REUSE Standard Report with compatibility findings and also add the rule descriptions as final comments) and project level (CLI/REST) as well.
       - For project level UI will have a selection for multiple uploads once the uploads are selected CLI shall be triggered and report the in compatibility.
       - A feature shall be implemented to conclude all the files if no compatibility issues.

krrish175-byte:
I will be working on this as a part of GSoC'26

PR #3677, opened 2026-06-13
fix(cyclonedx): implement remaining CycloneDX 1.4 fields

krrish175-byte:
## Description

This is Phase 1 - Part 2 of the GSoC 2026 project (#3635 : Improve CycloneDX Support and Update CycloneDX Exports to Spec Version 1.7).

The CycloneDX agent was missing several metadata fields that the SPDX agent already exports. The data was already available via `UploadDao::getReportInfo()` and `FileNode::getAcknowledgements()/getComments()`, it just wasn't being passed through from `cyclonedx.php` and `reportgenerator.php`.

### Changes

- **Component Version**: Populate from `ri_version` in report info
- **Package URL (PURL)**: Set when `ri_component_type` is PURL/PACKAGEURL
- **Description**: Populate from `ri_general_assesment` (general assessment)
- **External References**: Add `ri_component_id` as a distribution URL when the component type is not a PURL
- **Acknowledgements & Comments**: Emit as CycloneDX `properties` entries (`fossology:acknowledgement`, `fossology:comment`) on file components
- **Bug fix**: Use `intval()` + strict `===` for `ri_component_type` comparison to avoid PHP loose comparison where `'' == 0` evaluates to `true`
- **Report version**: Changed from `1.0` (float) to `1` (integer) per CycloneDX spec
- **Report-level externalReferences**: Support optional top-level `externalReferences` in generated report

Also adds standalone unit tests for `BomReportGenerator` covering all new and existing fields (78 assertions).

## How to test

1. Upload a package and set version, component ID (PURL or URL), and general assessment in the report info
2. Run the CycloneDX agent
3. Verify the JSON output contains `version`, `purl` or `externalReferences`, and `description` on the main component
4. Verify file components include `acknowledgements` and `comments` as properties when present
5. Run `php src/cyclonedx/agent_tests/test_report_output.php` — all 78 tests should pass

Related Issue - #3635

### DIFF
--- a/src/cyclonedx/agent/cyclonedx.php
+++ b/src/cyclonedx/agent/cyclonedx.php
@@ -29,6 +29,7 @@ use Fossology\Lib\Data\AgentRef;
 use Fossology\Lib\Data\Report\FileNode;
 use Fossology\Lib\Data\Report\SpdxLicenseInfo;
 use Fossology\Lib\Db\DbManager;
+use Fossology\Lib\Data\Package\ComponentType;
 use Fossology\Lib\Report\ReportUtils;
 
 include_once(__DIR__ . "/version.php");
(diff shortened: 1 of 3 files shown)


its-sushant (changes requested):
@krrish175-byte The changes look good to me except for one thing: the report-level `externalReferences` never actually shows up in the generated BOM.

The reason is that `generateReport()` reads `$bomdata['externalReferences']`, but in cyclonedx.php the `$bomdata` array is only built with tool-version, maincomponent, and components: the externalReferences key is never passed in, so that block is effectively dead code right now.

krrish175-byte:
@its-sushant : Good catch! I completely missed passing that key into the final `$bomdata` array before calling `generateReport()`.

I have just pushed an amended commit that adds `'externalReferences' => $externalReferences` to the array to resolve this. Thank you for the review!

its-sushant (approved):
Reviewed and tested. LGTM!

PR #3678, opened 2026-06-13
perf: avoid repeated scans of copyright entries

SooryanshSingh:
Summary

Build a lookup map of copyright entries keyed by uploadtree_pk before iterating over upload tree records.

Previously, printRow() scanned all copyright entries for every upload tree node to collect matching copyright information. This change preprocesses the entries once and performs direct lookups during iteration.


shaheemazmalmmd (approved):
code looks good. Tested works as expected.

PR #3679, opened 2026-06-16
fix(reuser): check upload accessibility before scheduling reuse

sakshammishra112:
## Description
Added authorisation check for the reused upload in `ScanOptions::prepareReuser()` to prevent users from reusing uploads they don't have access to via the **_REST API_**. The UI path already has permission filtering via the reuser plugin's `getFolderUploads() / getAllUploadsForGroup()` when populating the dropdown, so only the REST API path needed this server-side check.

### Changes:
- Added `isAccessible()` call on the reused upload in `prepareReuser()`, throwing `HttpForbiddenException (403)` if the user's group lacks access

### Screenshots

<Details>
<summary>Before</summary>
Screenshot 2026-06-16 085209
</Details>

<Details>
<summary>After</summary>
Screenshot 2026-06-16 094529
</Details>

### DIFF
--- a/src/www/ui/api/Models/ScanOptions.php
+++ b/src/www/ui/api/Models/ScanOptions.php
@@ -185,9 +185,13 @@ class ScanOptions
    */
   private function prepareReuser(Request &$request)
   {
-    if ($this->reuse->getReuseUpload() == 0) {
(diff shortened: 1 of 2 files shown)


shaheemazmalmmd (approved):

PR #3684, opened 2026-06-19
feat(maintagent): overhaul maintenance agent with improved internals

### LINKED ISSUE #3650: Feature/Idea: Extending and modernizing maintagent
Extending features of '[maintagent](https://github.com/fossology/fossology/tree/master/src/maintagent)'.

**Motivation**

The -R flag currently collects large numbers of IDs and concatenates them into a single SQL statement (DELETE ... WHERE id IN (...)).
This - and several other code paths - can generate extremely large IN lists, which may:
 - exhaust client-side memory or string buffers,
 - produce oversized SQL statements with high parse/planning overhead,
 - increase the likelihood of network or connection issues (timeouts, broken connections),
 - and significantly degrade performance or fail under production workloads.

### Proposal

**1) Improve how large ID sets are handled.**

Replace large client-side IN (...) lists with session temporary tables.
The proposed approach is to:
 - create a temporary table for IDs,
 - populate it via COPY FROM STDIN,
 - and execute joins/subqueries against that table instead of generating massive SQL strings.

This primarily affects the current -R implementation and would improve robustness.

**2) Improve PSQL statements**

 - Replace SIMILAR TO with LIKE where pattern matching does not require regex semantics.
 - Use VACUUM FULL selectively for PostgreSQL ≥ 17 instead of VACUUM only.  (see *1)

**3) Improve debug and runtime visibility**

Verbose logging already supports up to five -v levels.
Extend this with deterministic START / END markers and execution timings per processing step, e.g. with -vvv, to make bottlenecks and long-running operations easier to identify. Or to create individual statistics and comparisons on how long a task needs > planning time for the maintenance.

**4) Documentation updates**

Update the [Wiki](https://github.com/fossology/fossology/wiki/Maintenance-Agent)  to
 - document existing parameters more completely,
 - explain maintenance operations in greater detail,

Clarify/Remove flags which are currently marked as TODO or not implemented in the codebase

-----
*1 PostgreSQL 17 introduced major VACUUM improvements, including, see PSQL official changelog and release notes: https://www.postgresql.org/docs/release/17.0/ ,  www.postgresql.org/about/news/postgresql-17-released-2936 - New memory management system for VACUUM, up to 20x less memory, Allow vacuum to more efficiently remove and freeze tuples, vacuum is no longer silently limited to one gigabyte of memory ...
https://www.postgresql.org/docs/current/release-18.html - “AIO subsystem ... improves ... vacuums”

-----
**Hint:**
A previous PR already introduced parts of these ideas through a separate maintagent2 implementation. Some implementation details may be reusable or worth revisiting. PR #3309

### Open for discussion: 
Feedback and additional ideas are very welcome - especially regarding broader maintenance optimizations. In case somene has already ideas :)

I am also willing to contribute to a maintagent overhaul. 
With the changes described in points 1 and 2, we already achieved substantial reductions in maintagent runtime in production using the separate maintagent2 implementation on a weekly basis (some commands more often).

helmut-leher-magna:
## Description

This PR replaces the internal implementation of the maintagent C agent with an overhauled version. All CLI flags, UI options, scheduler configuration, and REST API integration remain fully compatible — only the agent internals change.

Closes #3650 - see idea there.

| Aspect | Status |
|--------|--------|
| CLI flags (-a, -A, -D, -F, …) | ✅ Unchanged |
| UI options (Admin → Maintenance) | ✅ Unchanged |
| Scheduler config (maintagent.conf) | ✅ Unchanged |
| REST API (MaintenanceController) | ✅ Unchanged |
| Existing PHPUnit tests | ✅ All pass |

### Behavioral Changes

**1. Database Cleanup (-D flag)**

The vacuum/analyze operation (vacAnalyze()) was redesigned for modern PostgreSQL compatibility:
- 	Previously: one combined VACUUM ANALYZE command
-	Now: two separate steps — VACUUM then ANALYZE
-	On PostgreSQL 17+: use VACUUM FULL, which performs a more thorough cleanup (reclaims disk space) but requires a short exclusive lock on tables during the operation

**2. Verbose Logging (-v flag)**
-	Previously: single on/off verbose flag
-	Now: repeatable like in other agents (-v, -vv, ...)

| Flag   | Effect |
|--------|--------|
| `-v`   | Basic verbose mode (as former) |
| `-vv`  | not used |
| `-vvv` | Structured ISO8601 timestamps per action |
| `-vvvv`| Additionally `VACUUM VERBOSE` + `ANALYZE VERBOSE` |

### Improvements

**1. Performance — Large Dataset Optimization**
The "Remove uploads with no pfiles" (removeUploads()) operation was rewritten to use temporary tables instead of large database subqueries, significantly improving performance on systems with many uploads.

**2. Structured Operational Logging**
Every maintenance action now emits timestamped START / END log entries (ISO8601) with precise duration measurements (monotonic clock). This enables operations teams to track exactly how long each maintenance step takes in production.


shaheemazmalmmd (approved):
Code looks Good. tested works as expected.

PR #3688, opened 2026-06-25
fix(security): remove some eval() / escape request URIs

helmut-leher-magna:
## Description

* Safely resolve placeholders in configuration values using known values, replacing the previous use of `eval()`:
 `install/fo_dbcheck.php` > resolve_config_value()
 `install/fossinit-common.php` > resolve_config_value()
 `src/lib/php/common-sysconfig.php` > resolve_sysconfig_value()

* Updated the `bootstrap` and `get_pg_conn` functions to use the new resolver functions for configuration variable assignment

* Escape URLs with `htmlspecialchars` before outputting them in links, preventing potential XSS vulnerabilities:
`src/buckets/ui/bucket-diff.php`
`src/buckets/ui/ui-buckets.php` 

## Note

The resolver functions are more-or-less code duplication with the 3 time usage in different files. Shall I instead create helper for that in "src/lib/php/"? Like "sysconfig-placeholder.php" or so?

### DIFF
--- a/install/fo_dbcheck.php
+++ b/install/fo_dbcheck.php
@@ -29,6 +29,24 @@ $DatabaseName = "fossology";
 $UpdateLiceneseRef = false;
 $sysconfdir = '';
 
+
+/**
+ * Resolve simple $VAR and ${VAR} placeholders against already-known values.
+ *
+ * @param string $value The raw config value.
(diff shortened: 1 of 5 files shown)


its-sushant (approved):
LGTM!

PR #3689, opened 2026-06-25
fix(security): remove eval and escape URI output

helmut-leher-magna:
## Description

* This change replaces the eval-based configuration handling in `install/db/migratetest.php`

* Reflected HTML output is secured with proper escaping in:
 `src/demomod/ui/demomod.php`
 `src/nomos/ui/nomos-diff.php`


its-sushant (approved):
Reviewed and tested. LGTM!

PR #3692, opened 2026-06-27
feat(api)  : add support for link action in move/copy uploads endpoint

### LINKED ISSUE #3687: Missing API support for 'linking' between folders > Feature Request
Summary

The FOSSology UI provides a “Link” functionality that allows an existing upload to be referenced in multiple folders without creating a duplicate or moving the upload. This behavior is currently not exposed via the REST API, which only supports move and copy operations. 
Or better: I am not aware of such functionality, if it exists, a hint would be very much appreciated.

This feature request proposes adding a dedicated API endpoint (or explicit parameterized operation) to support the same semantic as the UI “Link” function.

**Current Behavior**
(In FOSSology 4.7.1:)

REST API supports:
`move_upload` → moves an upload to another folder
`copy_upload` → creates a new folder entry and may reuse or duplicate analysis metadata depending on configuration

UI supports additionally:
 - “Link” upload to another folder
 - Upload remains a single logical object
 - Upload becomes visible in multiple folders
 - No duplication of upload or analysis data

The current UI behavior suggests that the underlying database already supports a many-to-many relationship between uploads and folders. However, this capability is not exposed through the REST API, forcing users to rely on copy or move, which do not accurately represent the UI “link” semantics.

**Problem**

The current REST API does not provide an equivalent operation to the UI “Link” feature.
This leads to several limitations:
 - Inability to reproduce UI behavior in automation scripts
 - No API-level way to create multi-folder visibility without triggering copy/move semantics
 - Confusion between copy (which may duplicate metadata or trigger different analysis behavior) and true linking (which is purely relational)
 - UI and API semantic mismatch for the same conceptual operation

**Expected Behavior**

There should be a REST API operation that allows:
 - Assigning an existing upload to an additional folder
 - Without creating a new upload
 - Without moving the upload from its current folder
 - Without duplicating analysis results or file storage

**Proposed Solution**

a) Extend an existing API endpoint
Extend `move_upload` with a new action:
`action=link`
Example: PUT /uploads/{id}?action=link&folderId=123

b) or introduce a new one / different idea?

_Expected Result of “link” action_
 - Upload remains unchanged (same uploadId)
 - Upload is visible in both source and target folders
 - No duplication of:
   - file content
   - analysis results
   - database upload entity
   - Only folder-to-upload relationship is added

_Benefits_
- Aligns REST API behavior with existing UI functionality
- Removes ambiguity between copy vs link
- Enables automation use cases (CI/CD, bulk ingestion pipelines)
- Prevents unnecessary duplication of uploads and analysis artifacts
- Improves consistency of data model semantics

redpopcorn:
**feat(api): Add `link` action support for PUT /uploads/{id} — fixes #3687**

FOSSology's UI has a "Link" feature that lets you place an upload in another folder without duplicating any data — it's purely a relational alias in the `foldercontents` table. Until now, there was no clean way to trigger this through the REST API. This PR wires that up.

**What changed**

`UploadController.php` — the `action` parameter (header in V1, query param in V2) is now captured and passed down through `changeUpload` into `RestHelper::copyUpload`, so it actually reaches the logic that needs it.

`RestHelper.php` — `copyUpload` now accepts the `action` parameter. When it's `"link"`, the response message says so explicitly: `"Upload {id} will be linked to folder {folderId}"`. The underlying copy logic stays the same — it was already doing the right thing, we just weren't calling it that.

`openapi.yaml` / `openapiv2.yaml` — `"link"` is now a valid enum value for the `action` parameter on `PUT /uploads/{id}`.

`UploadControllerTest.php` / `RestHelperTest.php` — added `testLinkUploadV1`, `testLinkUploadV2`, and `testLinkUpload` to cover the new action end-to-end. Also added `testMoveUploadV1` / `testMoveUploadV2` for move coverage that was missing, and adjusted `testCopyUpload` mock expectations to match the updated signature.

**To verify**

```bash
vendor/bin/phpunit src/www/ui_tests/api/
```

### DIFF
--- a/src/lib/php/Dao/JobDao.php
+++ b/src/lib/php/Dao/JobDao.php
@@ -67,12 +67,19 @@ class JobDao
     return $result;
   }
 
+  /**
+   * Check if user has permission to perform actions on a job.
+   *
+   * @param int $jobId  The job_pk from the job table
+   * @param int $userId
+   * @param int $groupId
+   * @return bool True if user has permission, false otherwise
+   */
   public function hasActionPermissionsOnJob($jobId, $userId, $groupId)
   {
-    $result = array();
     $stmt = __METHOD__;
     $this->dbManager->prepare($stmt,
-      "SELECT *
+      "SELECT 1
        FROM job
          LEFT JOIN group_user_member gm
            ON gm.user_fk = job_user_fk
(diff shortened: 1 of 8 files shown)


redpopcorn:
Hi maintainers, could someone please review this PR and approve the workflow runs? Happy to make any requested changes.

shaheemazmalmmd:
@redpopcorn : We don't accept the merge commits you please rebase the branch and remove the merge commit ?? Also check the failing test cases.

shaheemazmalmmd:
@redpopcorn : since the test failures are not addressed and there is no response, closing this please reopen once done.

PR #3696, opened 2026-06-30
fix(scheduler): parametrize basic_checkout LIMIT from host max config

helmut-leher-magna:
## Description / Intention

The basic_checkout SQL query had a hardcoded LIMIT 10 that caused the scheduler to stop dispatching new jobs when the host max was set above 10. 
Only 10 rows were fetched per poll cycle; since JB_CHECKEDOUT jobs retain jq_starttime IS NULL until the agent handshake completes, the next poll returned the same 10 rows and skipped them all > leading in some cases to leaving all remaining queued jobs stuck.

Pays credits to #3700  (1)

## Changes

Compute the checkout limit dynamically in database_update_event():
- If no hosts are loaded yet (host_queue == NULL, startup path), fall back to CHECKOUT_SIZE so the query runs before host configuration has been applied.
- If hosts are configured, sum only hosts whose max > 0. Hosts with max <= 0 never satisfy (max - running >= 1) in get_host(), so they provide zero real slots and must not reduce the total limit. LIMIT 0 is valid PostgreSQL and correctly returns no rows when all hosts have max=0; do NOT inflate to CHECKOUT_SIZE as that would over-fetch jobs that can never be dispatched.
- If the accumulated sum wraps negative (integer overflow; requires INT_MAX total slots - practically impossible), fall back to CHECKOUT_SIZE rather than clamping to 0, which would block fetching.

Guard database_exec() return value against NULL before passing to PQresultStatus(); ensure SafePQclear() runs on every error path.
Existing configs with max=10 see identical behavior.

### File changes in detail

* `src/scheduler/agent/database.c`, `src/scheduler/agent/sqlstatements.h`: The number of jobs fetched from the database in each poll cycle is now dynamically set to the sum of `max` agent slots across all configured hosts. If no hosts are configured (startup), it falls back to the constant `CHECKOUT_SIZE`.
* `src/scheduler/agent/sqlstatements.h`: The SQL query for fetching jobs (`basic_checkout`) now uses a `%d` placeholder for the limit, which is filled at runtime with the dynamically computed value. 
* `src/scheduler/agent/sqlstatements.h`, `src/scheduler/agent/database.c`, `src/scheduler/agent/scheduler.h`: Added and updated comments to explain the new logic, including how the limit is computed and why it is necessary to format the SQL string before execution.
* `src/scheduler/agent/database.c`: Ensured proper cleanup and error handling when the database query fails.

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -840,6 +840,12 @@ void database_exec_event(scheduler_t* scheduler, char* sql)
 /**
  * @brief Checks the job queue for any new entries.
  *
+ * The number of rows fetched from the database is bounded by the sum of
+ * max-agent slots across all configured hosts so that every available slot
+ * can be filled in a single poll cycle.  When the scheduler has not yet
+ * loaded any host configuration (startup path) the query falls back to
+ * CHECKOUT_SIZE as a safe upper bound.
+ *
  * @param scheduler The scheduler_t* that holds the connection
  * @param unused
  */
@@ -850,6 +856,9 @@ void database_update_event(scheduler_t* scheduler, void* unused)
   int i, j_id;
   char* value, * type, * host, * pfile, * jq_cmd_args;
   job_t* job;
+  gchar* checkout_sql;
+  int   checkout_limit;
+  GList* iter;
 
   if(closing)
   {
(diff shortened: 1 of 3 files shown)


deo002:
Tested, working as expected! Code looks good too!

### Test steps:
- Changed fossology.conf host max value to a non default value, 5.
Image

- Increased verbosity of fossology logs and searched for the logged db query.

Image

After this and #3704 are merged, it will be a good idea to revisit host max and agent max values and configure them accordingly.

shaheemazmalmmd:
Merged https://github.com/fossology/fossology/pull/3704 with this changes as well.

PR #3697, opened 2026-07-01
fix(report): keep report export job status current

ayushmaan-16:
## Description

Fixes a report export UI inconsistency where generated reports could download successfully while the “My Recent Jobs” table still showed the report job as running with `[Pause] [Cancel]`.

This also prevents browser refresh from re-triggering the same report export by replacing the export URL with the stable `showjobs` URL after scheduling.

## Changes Made

- Added a stable `showJobsLink` to report generator UI plugins.
- Updated the shared report generation template to replace browser history with the `showjobs` URL.
- Updated the report completion callback to immediately mark the active report job row as completed.
- Added a download link to the completed job row and removed stale pause/cancel actions.
- Allowed `getTableBody()` to accept an optional callback so report download can wait for the table refresh path when available.

## Proof that issue exists

https://github.com/user-attachments/assets/75064dab-a085-4a11-9541-42cda427ccd4


shaheemazmalmmd (approved):

PR #3699, opened 2026-07-02
ci: restrict GITHUB_TOKEN permissions in all workflows

helmut-leher-magna:
## Description

Set explicit least-privilege permissions on all GitHub Actions workflows.
See 'Principle of least privilege' on: https://docs.github.com/en/actions/reference/security/secure-use

### Changes

- Add `permissions: contents: read` as workflow-level default to .yml files:
  build-test
  docker-test
  docker-build
  deploy-pages
  static-checks
- Add job-level override for lint-commits in static-checks.yml
- Add `contents: read` to conflict-check.yml
- Set `contents: write` for release-publish.yml

## IMPORTANT

I cannot guarantee or personally test every workflow to see if the permissions might be too restrictive.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -7,6 +7,9 @@ concurrency:
   group: build-${{ github.head_ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 on:
   push:
     branches: [master]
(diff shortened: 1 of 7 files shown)


shaheemazmalmmd (approved):
Code looks good.

PR #3702, opened 2026-07-06
fix(schema): add index on jobdepends.jdep_jq_fk

helmut-leher-magna:
## Description

jobdepends rows are removed only via ON DELETE CASCADE when jobqueue rows are deleted, which happens only when an upload is explicitly deleted through the UI or delagent.

The table is accessed in three places, all of which filter or join on the **jdep_jq_fk** column:
1) \src\scheduler\agent\sqlstatements.h - basic_checkout : executed on every scheduler poll cycle
2) \src\scheduler\agent\sqlstatements.h - jobsql_anyrunnable : executed on every job completion
3) \src\lib\php\Dao\ShowJobsDao.php - attachJobQueueData() : job list UI / REST API

| Query | Before | After |
|-------|--------|-------|
| `basic_checkout` `NOT EXISTS` | Seq scan of full table per candidate row | Index seek, O(log n) |
| `jobsql_anyrunnable` `NOT EXISTS` | Seq scan of full table per candidate row | Index seek, O(log n) |
| `ShowJobsDao` `LEFT JOIN` | Seq scan | Index nested-loop join |

### Changes

Add a single btree index on the lookup column (jdep_jq_fk) in _core-schema.dat_.

### DIFF
--- a/src/www/ui/core-schema.dat
+++ b/src/www/ui/core-schema.dat
@@ -2856,6 +2856,8 @@
   $Schema["INDEX"]["ipra_decision"]["ipra_decision_user_fk_index"] = "CREATE INDEX ipra_decision_user_fk_index ON ipra_decision USING btree (user_fk);";
   $Schema["INDEX"]["ipra_decision"]["ipra_decision_pfile_hash_index"] = "CREATE INDEX ipra_decision_pfile_hash_index ON ipra_decision USING btree (hash, pfile_fk);";
 
+  $Schema["INDEX"]["jobdepends"]["jobdepends_jq_fk_idx"] = "CREATE INDEX jobdepends_jq_fk_idx ON jobdepends USING btree (jdep_jq_fk);";
+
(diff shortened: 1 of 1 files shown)


Kaushl2208 (commented):
Hey @helmut-leher-magna , I think the code looks good. I am not sure if we needed the `index` on `jobdepends` but it doesnt hurt. 

Some test cases are failing can you please take a look?

helmut-leher-magna:
hi @Kaushl2208

from my understanding: In a long-running instance the table accumulates all historical dependency edges and can easily reach thousands of rows. > so the both polls on it will benefit. (and when basic_checkout no longer uses the hardcoded LIMIT 10 i am quite sure having the index in place will be helpful)

Tests:
hm those are failing when trying to get packages from /mirrors.kernel.org - All "wget". The exec() fails silently, the expected files are never created, and consequently, all three assertFileExists checks fail.
Maybe only temporary, can you re-run the tests?

shaheemazmalmmd (approved):
Code looks good.

Issue #3700, opened 2026-07-03
Scheduler has problems with more than 10 agents in parallel


helmut-leher-magna:
First, I'd like to point out that the following issue will likely not occur in every environment.

## Symptom Description
When the scheduler's host `max` agent count is raised above 10 (e.g. to match the system CPU count of 20), two distinct failure modes appear:

1. **Jobs not started**: New jobs visible in the UI queue are never selected for execution, even though the number of active agents is below the configured limit. Further queued jobs remain stuck indefinitely.
2. **Scheduler hangs**: No job from the queue is ever selected again; only a manual scheduler restart recovers the system.

## Analysis

We have identified several potential root causes with varying levels of priority. The first two appear to be the most likely and should be addressed first.

### 1) Hardcoded `LIMIT 10` in `basic_checkout` SQL

 - The SQL query that polls the database for pending jobs has a hardcoded LIMIT 10
 - So only ten jobs are fetched per polling cycle
 - However: Checked-out jobs remain visible to subsequent polls because jq_starttime is not updated until the agent reaches JB_STARTED. The same ten jobs are therefore returned repeatedly and skipped as already known. > No additional jobs are loaded from the queue, effectively limiting the scheduler to a maximum of ten concurrent jobs despite a higher configured capacity.

(Files: src/scheduler/agent/database.c ; src/scheduler/agent/sqlstatements.h

### 2) `basic_checkout` Re-fetches In-Flight Jobs (Amplified by Cross-Container Latency)

 - Checked-out jobs are re-fetched because jq_starttime remains NULL until the transition to JB_STARTED and the JB_CHECKEDOUT does not update the database
 - So the scheduler repeatedly receives the same in-flight jobs, which are skipped because they already exist in job_list.
 - Even after replacing the hardcoded LIMIT 10 these duplicate entries still occupy the checkout result set, preventing new jobs from being returned

(Files: src/scheduler/agent/sqlstatements.h ; src/scheduler/agent/database.c)

### 3) Another hardcoded fallback, default limit `interface_nthreads`

 - interface_nthreads defaults to 10, limiting the number of concurrent interface threads. > Effective when more than 10 agents are active while the scheduler is using the default configuration.
 -  Scheduler interface requests (e.g. periodic web commands and status updates) may queue behind ongoing agent communication.
 -  Less prior as this only affects responsiveness and scheduling latency under higher concurrency, BUT increased delays in processing scheduler events can further reduce throughput > a performance bottleneck

[...]

PR #3704, opened 2026-07-06
fix(scheduler): exclude in-flight jobs from DB poll via NOT IN

helmut-leher-magna:
This PR includes changes from https://github.com/fossology/fossology/pull/3696

I would suggest to have separate reviews. Each PR represents one logical step, keeping the diff smaller (after merge of the PR before) and easier to review (hopefully ).

Pays credits to https://github.com/fossology/fossology/issues/3700 (1+2)

## Description

Jobs in JB_CHECKEDOUT state have jq_starttime = NULL in the database until the agent completes its handshake (VERSION → OK → JB_STARTED).
In cross-container Docker deployments this window is measurably wider, causing basic_checkout to re-fetch the same in-flight jobs on every poll cycle and waste LIMIT slots that should be filled by new entries.

## Changes

1) Build a comma-separated exclusion list of all jq_pk values currently held in scheduler->job_list and inject it as a NOT IN(...) clause into basic_checkout at query time. A sentinel value of 0 (never a valid jq_pk) keeps the list syntactically valid when job_list is empty. All values are scheduler-internal integers, so there is no SQL-injection risk.

This complements the fix (dynamic LIMIT tied to host max #3696) which ensures enough slots are requested per cycle; 
This PR ensures those slots are filled with genuinely new jobs rather than already-known ones.

### File changes in detail
* `src/scheduler/agent/sqlstatements.h`: add AND jq.jq_pk NOT IN (%s) to basic_checkout; document the two format placeholders (%s exclusion list, %d limit)
* `src/scheduler/agent/database.c`: add collect_known_pks() GTraverseFunc callback; build excl_pks GString in database_update_event() before exec
* `src/scheduler/agent_tests/Unit/testDatabase.c`: add test_database_update_event_excludes_known to verify that a second poll does not re-fetch an already-known job
* `src/scheduler/agent_tests/Unit/utils.c`: Remove hardcoded limit usage in two comments

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -840,6 +840,12 @@ void database_exec_event(scheduler_t* scheduler, char* sql)
 /**
  * @brief Checks the job queue for any new entries.
  *
+ * The number of rows fetched from the database is bounded by the sum of
+ * max-agent slots across all configured hosts so that every available slot
+ * can be filled in a single poll cycle.  When the scheduler has not yet
+ * loaded any host configuration (startup path) the query falls back to
+ * CHECKOUT_SIZE as a safe upper bound.
+ *
  * @param scheduler The scheduler_t* that holds the connection
  * @param unused
  */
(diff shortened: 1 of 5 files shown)


deo002:
Tested, working as expected! Code looks good too!

### Test steps:
- Increased verbosity of fossology logs and searched for the logged db query.
image

After this is merged, it will be a good idea to revisit host max and agent max values and configure them accordingly.

deo002 (approved):
LGTM!

shaheemazmalmmd (approved):

PR #3706, opened 2026-07-07
fix(scheduler): auto-scale interface pool and guard agent_init failures

helmut-leher-magna:
Pays credits to https://github.com/fossology/fossology/issues/3700 (3+4)

## Description

Under high system load, the scheduler can experience two issues that negatively impact performance and job processing:
 - The scheduler interface is limited by a hardcoded fallback of 10 interface threads (interface_nthreads). When more than 10 agents are active, interface requests (e.g. web commands and status updates) can queue behind agent communication, increasing scheduler latency and potentially reducing overall throughput.
 - reap_stale_jobs() can incorrectly classify valid jobs as stale while they are still waiting for an agent to start. Since the timeout reset currently only applies to exclusive (SAG_EXCLUSIVE) jobs, regular jobs may be prematurely marked as failed and removed from the queue under heavy load.

## Changes

 - Added logic to compute the interface thread pool size based on the sum of all hosts' maximum agent slots (incl new functions for that)
- Updated the configuration file (fossology.conf.in) to document the new [SCHEDULER] section and the interface_nthreads setting > was missing till now
 - Improved job handling in scheduler_update to correctly handle cases where agent_init fails but does not remove the job

### File changes in detail

- `install/defconf/fossology.conf.in`: Document interface_nthreads.
- `src/scheduler/agent/scheduler.c`: Re-queue jobs on agent_init() failure; resize interface pool on config reload.
- `src/scheduler/agent/interface.c`: Sizes interface pool at init time using host_total_max().
- `src/scheduler/agent/host.h` + `src/scheduler/agent/host.c`:  host_total_max() via g_tree_foreach() summation
- `src/scheduler/agent_tests/Unit/testInterface.c` and `src/scheduler/agent_tests/Unit/testHost.c` - Added tests

### DIFF
--- a/install/defconf/fossology.conf.in
+++ b/install/defconf/fossology.conf.in
@@ -36,6 +36,13 @@ path = @FO_REPODIR@
 [HOSTS]
 localhost = localhost @FO_SYSCONFDIR@ 10
 
+[SCHEDULER]
+; Number of threads in the interface thread pool (default: 10).
+; The scheduler automatically raises this to the sum of all host 'max' values,
+; so that UI/CLI connections are never queued behind active agent threads.
+; Only set this explicitly when you need a fixed value larger than host maximum.
+;interface_nthreads = 10
+
 [REPOSITORY]
 localhost[] = * 00 ff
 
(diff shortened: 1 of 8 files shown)


helmut-leher-magna:
Rebase on latest master, no code changes

PR #3708, opened 2026-07-10
fix(scancode): replace check-then-create DDL with IF NOT EXISTS

### LINKED ISSUE #3707: Scancode: Concurrent agent startup causes DDL lock contention and stuck jobs
When the scheduler starts 2 or more Scancode agents simultaneously (e.g. for different upload folders), the jobs get stuck and do not progress.

## Potential root cause

`scancode.cc` calls `createTables()` **at every agent startup**, before entering the scheduler loop. This function issues a sequence of DDL statements — `CREATE SEQUENCE`, `CREATE TABLE`, `CREATE INDEX`, `ALTER TABLE` (18+ statements in total) for 4 tables (`scancode_copyright`, `scancode_author`, `scancode_copyright_event`, `scancode_author_event`) — wrapped in a **single transaction**.

Two problems compound each other:

**1. DDL lock contention**  
PostgreSQL DDL acquires `AccessExclusiveLock` on the system catalog (`pg_class`). When N agents start simultaneously, their transactions block each other for the full duration of the multi-statement DDL transaction.

**2. TOCTOU race condition**  
The code uses a check-then-create pattern without atomicity:
```cpp
if (!dbManager.tableExists("scancode_copyright"))
    CREATE TABLE scancode_copyright ...
```

Two agents can both observe the table as absent, then both issue `CREATE TABLE`. The second transaction fails with "table already exists", triggers a full rollback, and retries — up to `MAX_TABLE_CREATION_RETRIES = 5` times (`scancode_dbhandler.hpp`). Each retry re-issues all 18+ DDL statements, multiplying lock contention and causing jobs to appear stuck.

## Affected Files

| File | Detail |
|------|--------|
| `src/scancode/agent/scancode.cc:32` | unconditional `createTables()` on every startup |
| `src/scancode/agent/scancode_dbhandler.cc` | `createTableAgentFindings()`, `createTableAgentEvents()` — check-then-create pattern |
| `src/scancode/agent/scancode_dbhandler.hpp` | `MAX_TABLE_CREATION_RETRIES = 5` |

## Proposed Fix

Replacing the check-then-create pattern with idempotent DDL > IF NOT EXISTS
This should eliminate the race condition, removing the retry loop, and reducing lock contention.

helmut-leher-magna:
## Description

Replace check-then-create DDL pattern with idempotent IF NOT EXISTS variants in the Scancode agent to prevent lock contention and stuck jobs when multiple agents start concurrently.

Closes: [issue 3707](https://github.com/fossology/fossology/issues/3707)

### Changes

**scancode_dbhandler.cc**
 - createTables(): removed wrapping transaction (begin/commit/rollback) and retry loop
 - createTableAgentFindings(): replaced sequenceExists() pre-check + CREATE SEQUENCE with CREATE SEQUENCE IF NOT EXISTS; replaced tableExists() pre-check + CREATE TABLE with CREATE TABLE IF NOT EXISTS; 
 replaced CREATE INDEX with CREATE INDEX IF NOT EXISTS; replaced unconditional ALTER TABLE ADD CONSTRAINT for FK constraints with a pg_constraint existence check
 - createTableAgentEvents(): same changes as above for the event tables, plus the ALTER TABLE ALTER COLUMN scope SET DEFAULT 1 is now always executed (it is idempotent)

**scancode_dbhandler.hpp**
 - Removed unused MAX_TABLE_CREATION_RETRIES constant.

---
**Common case on every regular startup (tables already exist):**
DDL / Area | Before | After
-- | -- | --
CREATE SEQUENCE | Check-then-create in a long transaction → blocks catalog | CREATE SEQUENCE IF NOT EXISTS → instant no-op, NOTICE suppressed
CREATE TABLE | Check-then-create in a long transaction → blocks catalog | CREATE TABLE IF NOT EXISTS → instant no-op, NOTICE suppressed
CREATE INDEX ×3 | Inside tableExists guard, in the same transaction | CREATE INDEX IF NOT EXISTS → instant no-op, NOTICE suppressed
FK ALTER TABLE ADD CONSTRAINT ×2+2 | Inside tableExists guard → ran on first start only | pg_constraint pre-check → 1 row found → skipped entirely, zero locks
ALTER TABLE SET DEFAULT | Ran unconditionally on every startup | pg_attrdef pre-check → row found → skipped entirely, zero locks
Wrapping transaction | Held AccessExclusiveLock for all of the above | Removed
Retry loop (×5) | Compounded every failure | Removed

**On first run (fresh schema):**
Scenario | Behavior
-- | --
Fresh schema | All DDL runs once.
Concurrent startup | The only concurrent-startup race is on FK constraints, handled by the DO block catching duplicate_object.

---

## How to test

 - Configure the FOSSology scheduler to allow multiple Scancode agents (max = -1 in scancode.conf)
a) Trigger some jobs like upload, reuser and scancode jobs
a) Trigger 2 or more uploads simultaneously so the scheduler spawns multiple Scancode agents at the same time, on a database where the Scancode tables do not yet exist
 - Verify all agents complete startup without blocking and all jobs proceed to completion
 - Drop the Scancode tables and repeat step 2 to confirm tables are created correctly on a fresh schema

### DIFF
--- a/src/scancode/agent/scancode_dbhandler.cc
+++ b/src/scancode/agent/scancode_dbhandler.cc
@@ -423,30 +423,13 @@ bool ScancodeDatabaseHandler::insertInDatabase(DatabaseEntry& entry) const
  */
 bool ScancodeDatabaseHandler::createTables() const
 {
-  int failedCounter = 0;
-  bool tablesChecked = false;
-
   dbManager.ignoreWarnings(true);
-  while (!tablesChecked && failedCounter < MAX_TABLE_CREATION_RETRIES)
-  {
-    dbManager.begin();
-    tablesChecked = createTableAgentFindings("scancode_copyright") && createTableAgentFindings("scancode_author")&& createTableAgentEvents("scancode_copyright_event") && createTableAgentEvents("scancode_author_event");
-
-
-    if (tablesChecked)
-      dbManager.commit();
-    else
-    {
-      dbManager.rollback();
-      ++failedCounter;
-      if (failedCounter < MAX_TABLE_CREATION_RETRIES)
-      LOG_WARNING("table creation failed: trying again (%d/%d) \n", failedCounter, MAX_TABLE_CREATION_RETRIES);
-    }
-  }
-  if (tablesChecked && (failedCounter > 0))
-    LOG_NOTICE("table creation succeeded on try %d/%d \n", failedCounter, MAX_TABLE_CREATION_RETRIES);
+  bool result = createTableAgentFindings("scancode_copyright") &&
+                createTableAgentFindings("scancode_author") &&
+                createTableAgentEvents("scancode_copyright_event") &&
+                createTableAgentEvents("scancode_author_event");
   dbManager.ignoreWarnings(false);
-  return tablesChecked;
+  return result;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


its-sushant (approved):
Reviewed and tested. LGTM!

PR #3709, opened 2026-07-11
feat(api): add link action to PUT /folders/{id} and block duplicate folder contents

Valyrian-Code:
## What this does

Adds `action=link` to `PUT /folders/{id}` and blocks duplicate folder contents, complementing #3692 (which adds the same for uploads).

In FOSSology, "copy" of a folder or upload never duplicates data — `FolderDao::copyContent()` just inserts another row into `foldercontents`, so the item becomes visible in a second folder while remaining a single logical object. That is exactly the UI "Move or Alias" behaviour. The uploads endpoint is being taught the `link` name in #3692; this PR does the folder side and adds the missing duplicate guard that #3687 asks for.

## Changes

- **`FolderController::copyFolder()`** — `PUT /folders/{id}` now accepts `action=link` alongside `copy`/`move`. When the folder is already present under the target parent it returns `409 Conflict` instead of inserting a second identical row.
- **`FolderDao`** — new `isContentInFolder()` helper, used in `isContentMovable()` so any copy/link/move that would create a duplicate `(parent_fk, foldercontents_mode, child_id)` entry fails with a clear message rather than silently duplicating. This also protects the uploads endpoint and the web UI, since both go through the same path.
- **`openapi.yaml` / `openapiv2.yaml`** — document `link` as a valid action on `PUT /folders/{id}`.
- **Tests** — `FolderControllerTest`: link action (V1/V2) and the 409 conflict. `FolderDaoTest`: duplicate guard for copy and move, plus `isContentInFolder()`.

## Testing

```
vendor/bin/phpunit src/www/ui_tests/api/Controllers/FolderControllerTest.php   # 47 tests
vendor/bin/phpunit src/lib/php/Dao/test/FolderDaoTest.php                       # 19 tests
```

Also verified end to end against a running instance (API 1.6.2): `action=link` on a folder returns `202 ... link(ed) ...` and adds a single `foldercontents` row; repeating it (or a copy) returns `409`; and a copy/link/move that would duplicate an upload now fails cleanly instead of adding a duplicate row.

Related to #3687.

### DIFF
--- a/src/lib/php/Dao/FolderDao.php
+++ b/src/lib/php/Dao/FolderDao.php
@@ -382,6 +382,23 @@ WHERE fc.parent_fk = $1 AND fc.foldercontents_mode = " . self::MODE_UPLOAD . ";"
     return !empty($cycle);
   }
 
+  /**
+   * Check if a folder already contains the given content
+   * @param integer $childId  Id of the child
+   * @param integer $mode     Mode of the child
+   * @param integer $folderId Folder to look in
+   * @return boolean True if the folder contains the content, false otherwise
+   */
+  public function isContentInFolder($childId, $mode, $folderId)
+  {
+    $content = $this->dbManager->getSingleRow(
+      'SELECT foldercontents_pk FROM foldercontents '.
+      'WHERE parent_fk=$1 AND foldercontents_mode=$2 AND child_id=$3',
+      array($folderId, $mode, $childId),
+      __METHOD__);
+    return !empty($content);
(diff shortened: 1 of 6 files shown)


shaheemazmalmmd:
@Valyrian-Code : please look into the failing test cases.

shaheemazmalmmd:
@Valyrian-Code : as there is no response since a month im closing this PR please reopen once the test failure is addressed.

PR #3710, opened 2026-07-12
fix(multicompare): descend into archive wrapper directories

ayushmaan-16:
## Summary

Multi-Component Comparison compares the immediate non-artifact children of
each selected uploadtree item.

Uploaded archives often contain a single extracted wrapper directory:

```text
first.zip              second.zip
└── first/             └── second/
    ├── file1.c            ├── file1.c
    └── file2.c            └── file2.c
```

When users selected the archive roots, the comparison matched `first/` against
`second/`. Because the wrapper names differed, they appeared as unique or
missing, and the files inside them were not compared.

original state of compare

## Changes Made

Added a conservative root-normalization step before building the comparison.

If a selected root has exactly one non-artifact child and that child is a
directory, the directory becomes the effective comparison root. Otherwise,
the selected root remains unchanged.

Normalization happens before generating the cache key and comparison data.
Therefore, file matching, summaries, filters, links, breadcrumbs, and
histograms all use the normalized roots.

Added unit tests covering:

- A sole directory child.
- Multiple children.
- A sole file child.
- No children.

fixed version of compare

### DIFF
--- a/src/lib/php/common-multicompare.php
+++ b/src/lib/php/common-multicompare.php
@@ -10,6 +10,33 @@
  * \brief N-way file matching and link generation for multi-component comparison.
  */
 
+/**
+ * \brief Select the effective comparison root for a tree item.
+ *
+ * Archive uploads commonly contain one extracted, non-artifact wrapper
+ * directory below the upload root.  In that specific case, compare the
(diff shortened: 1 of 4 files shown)


shaheemazmalmmd (approved):

PR #3712, opened 2026-07-15
fix(auth): add timestamp to time fields in pat table to remove unusable token window

deo002:
### Description
Fixes the issue where there was a period of time for which a personal access token was unusable after issuance in some cases due to diff in time from utc. Before, only the date of issuance was stored in the database. So, for example, if we create a token at 2026-07-15 01:00 A.M. IST, only 2026-07-15 was stored in database. This field was used as nbf value in the token. The function strtotime, which was used to convert this date to unix timestamp assumed the date was in utc and not ist as no timestamp was given. So, the nbf value was stored as 2026-07-15 00:00 A.M. UTC. When a user tries to use this token at an arbitrary time, let's say 2026-07-15 02:00 A.M. IST, which corresponds to 2026-07-14 08:30 P.M. UTC, he cannot use it because of the nbf time set.

### DIFF
--- a/install/db/dbmigrate_token.php
+++ b/install/db/dbmigrate_token.php
@@ -1,0 +1,25 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © 2026 Siemens AG
+ SPDX-FileContributor: Dearsh Oberoi <dearsh.oberoi@siemens.com>
+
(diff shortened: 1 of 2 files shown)


Kaushl2208 (approved):
Looks good, tested working as expected.

PR #3714, opened 2026-07-16
feat(cyclonedx): add user default settings for CycloneDX export

krrish175-byte:
This is **Phase 2 - Part 2 of the GSoC 2026 project** (#3635): Improve CycloneDX Support and Update CycloneDX Exports to Spec Version 1.7.

This PR implements user-configurable default settings for CycloneDX exports, mirroring the existing behavior for SPDX exports. It allows users to set their default preferences for CycloneDX once via the UI, removing the need to configure them for every export.

**Changes Made:**

- **Database Update:** Added the `cyclonedx_settings` column to the `users` table via `core-schema.dat` to persistently store user defaults.
- **Backend DAO:** Implemented `getCyclonedxSettings()` in `UploadDao.php` to fetch the user settings cleanly.
- **UI Logic:** Updated `user-edit.php` to handle parsing the settings string from the database on page load and constructing the string on save.
- **UI Template:** Added checkboxes for CycloneDX OSSelot Export, License Comments, and Ignore Files Without Info to the `user_edit.html.twig` template, along with matching JavaScript logic to handle their dependencies.

**How to Test**

1. Rebuild the `cyclonedx` plugin or the UI container.
2. Go to **Administration > Users > Manage Users > Edit User**.
3. Scroll down and locate the new CycloneDX checkboxes below the SPDX settings.
4. Toggle the settings (e.g., enable OSSelot export) and click **Update**.
5. Reload the page and verify your checkbox states were saved correctly and persist.

### DIFF
--- a/src/lib/php/Dao/UploadDao.php
+++ b/src/lib/php/Dao/UploadDao.php
@@ -906,6 +906,45 @@ ORDER BY lft asc
   }
 
   /**
+   * @brief Get cyclone dx settings for a user
+   * @param int $uploadId Upload ID to get user for
+   * @return string Comma separated values for cyclonedxLicenseComment, ignoreFilesWOInfo, osselotExport
+   */
+  public function getCyclonedxSettings($uploadId)
+  {
+    $stmt = __METHOD__ . '.getOwner';
+    $sql = "SELECT user_fk FROM upload WHERE upload_pk = $1";
+    $uploadOwner = $this->dbManager->getSingleRow($sql, array($uploadId), $stmt);
+
+    if (empty($uploadOwner)) {
+      return "unchecked,unchecked,unchecked";
+    }
+
+    $userId = $uploadOwner['user_fk'];
+
+    $stmt = __METHOD__ . '.getUserDefaults';
+    $sql = "SELECT cyclonedx_settings FROM users WHERE user_pk = $1";
+    $userDefaults = $this->dbManager->getSingleRow($sql, array($userId), $stmt);
+
+    if (empty($userDefaults) || empty($userDefaults['cyclonedx_settings'])) {
+      return "unchecked,unchecked,unchecked";
+    }
+
+    $settings = explode(',', $userDefaults['cyclonedx_settings']);
+    if (count($settings) < 3) {
+      $settings = array_pad($settings, 3, 'unchecked');
+    }
+
+    $osselotExport = $settings[0];
+    $cyclonedxLicenseComment = $settings[1];
+    $ignoreFilesWOInfo = $settings[2];
+
+    $result = "$cyclonedxLicenseComment,$ignoreFilesWOInfo,$osselotExport";
+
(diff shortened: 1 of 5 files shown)


src/lib/php/Dao/UploadDao.php
```diff
@@ -905,6 +905,43 @@ private function getUserSPDXDefaultsForUpload($uploadId)
...
   }
 
+  /**
+   * @brief Get cyclone dx settings for a user
+   * @param int $uploadId Upload ID to get user for
+   * @return string Comma separated values for cyclonedxLicenseComment, ignoreFilesWOInfo, osselotExport
+   */
+  public function getCyclonedxSettings($uploadId)
+  {
+    $uploadOwner = $this->getUploadOwner($uploadId);
```
its-sushant: `getUploadOwner()`  isn't defined anywhere, so this will fatal. The SPDX version just inlines the query, can we do the same here?
```
$sql = "SELECT user_fk FROM upload WHERE upload_pk = $1";
$uploadOwner = $this->dbManager->getSingleRow($sql, array($uploadId), $stmt);
```

src/lib/php/Dao/UploadDao.php
```diff
@@ -905,6 +905,43 @@ private function getUserSPDXDefaultsForUpload($uploadId)
     return $result;
   }
 
+  /**
+   * @brief Get cyclone dx settings for a user
+   * @param int $uploadId Upload ID to get user for
+   * @return string Comma separated values for cyclonedxLicenseComment, ignoreFilesWOInfo, osselotExport
+   */
+  public function getCyclonedxSettings($uploadId)
```
its-sushant, line 913: Nice, but nothing actually calls  `getCyclonedxSettings()`. The CycloneDX agent never reads these, so they get saved but never applied.

its-sushant (changes requested):
@krrish175-byte Please look into the inline comments.

src/lib/php/Dao/UploadDao.php
krrish175-byte: Fixed

src/lib/php/Dao/UploadDao.php
krrish175-byte: Fixed

its-sushant (approved):
Nice, the fatal is fixed and  `ignoreFilesWOInfo`  is now applied. Just to confirm: the OSSelot and license-comment defaults are still stored/shown but not consumed anywhere in the CycloneDX export (unlike the ignore-files one). Is that intentional? 

Anyway approving the PR. Changes look good to me!

krrish175-byte:
@its-sushant : Thanks for the review and approval. Yes, that is completely intentional for now. We are wiring up the consumption of the OSSelot and license-comment settings in the very next phase, alongside building the per-upload report configuration UI and the custom tag namespace feature.

PR #3717, opened 2026-07-17
feat(reuseCompare): add file diff view for reuse upload

sakshammishra112:
## Description
Adds a "Reuse Diff View" that lets users compare a file against the file it was reused from, showing a line-by-line diff alongside license and metadata changes.

Part of #3631 
### Changes
- Added `handle($request)` entry point for the reuse diff view, parsing `upload`, `item`, `currentPfile`, `reusedPfile`, and `reuse` request parameters.
- Added an access check via `Auth::getGroupId()` and `isAccessible()`, returning a 403 when the user lacks permission to the upload.
- Added `getPfileData()` lookups for both the current and reused pfiles (size, SHA1, mimetype).
- Implemented `generateDiff()`:
  - Resolves both file paths via `RepPath()` and validates they exist.
  - Copies both files to temp locations with `tempnam()` and `copy()`, with error handling at each step (missing path, failed tempnam, failed copy).
  - Runs `diff -u` on the two temp files and captures output/exit code.
  - Reads and explodes the current file's contents for line-level reconstruction.
  - Cleans up temp files and branches on exit code (no differences / differences found / diff execution error).
- Implemented `parseUnifiedDiff()` to convert raw unified diff output into a structured array of unchanged/added/deleted lines with line numbers, by parsing hunk headers and walking the current file's lines.
- Added logic to merge and sort (via `ksort()`) the current and reused license lists by `rf_shortname`, so licenses can be flagged as added, removed, or present in both.
- Added `getUploadtreeIdFromPfile()` and `Dir2Browse()` integration to build the breadcrumb trail for navigation.
- Added new template `reusediffview.html.twig` with a two-panel layout:
  - Left panel: line-by-line diff view with a legend (added/deleted/unchanged).
  - Right panel: diff statistics (counts and percentages), file details (size, SHA1, MIME type for both files), and a license comparison table.

### Screen recording

https://github.com/user-attachments/assets/5e94518b-9530-4b13-86d8-7180676a04c6

### DIFF
--- a/src/lib/php/common-compare.php
+++ b/src/lib/php/common-compare.php
@@ -47,31 +47,46 @@ function FuzzyCmp($Master1, $Master2)
 function MakeMaster($Children1, $Children2)
 {
   $Master = array();
-  $row =  -1;   // Master row number
+  $row = -1;
 
-  if (! empty($Children1) && (! empty($Children2))) {
+  if (!empty($Children1) && !empty($Children2)) {
+    /* Build hash-map indices for fast matching */
+    $allChildren2 = [];
+    $byName  = [];
+    $byFuzzyExt = [];
+    $byFuzzy = [];
+    foreach ($Children2 as $idx => $child) {
+      $allChildren2[$idx] = $child;
+      $byName[$child['ufile_name']][$idx] = $idx;
+      $byFuzzyExt[$child['fuzzynameext']][$idx] = $idx;
+      $byFuzzy[$child['fuzzyname']][$idx] = $idx;
+    }
+    $consumed = [];
+
     foreach ($Children1 as $Child1) {
       $done = false;
       $row++;
 
       /* find complete name match */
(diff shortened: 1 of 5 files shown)


sakshammishra112:
### License Status Priority

| Priority | Condition | Status |
|----------|-----------|--------|
| 1 | `removed = true` on reuse | **Removed from Reuse** |
| 2 | `removed = true` on current | **Removed from Current** |
| 3 | Both scanners found it | **Common** |
| 4 | In reuse effective, not in current scanner | **Added to Reuse** |
| 5 | User-added on current, not in reuse | **Added to Current** |

### Why the Total and Unpack Counts Differ

The **Unpack** count includes both files and directories, while the **Total Files** count includes only non-directory files.  
This is because `getAllNonArtifactDescendants()` filters out directory entries using `AND ut.ufile_mode & (3<<28) = 0`.

shaheemazmalmmd (approved):

PR #3719, opened 2026-07-20
fix: harden shell command, SQL, and password-hashing handling

helmut-leher-magna:
## Description

Some gaps in the input handling of core PHP code paths that build shell commands, SQL queries, and password hashes from user-influenced or externally influenced values have been closed. 
Additional safeguards have been added to these code paths to prevent values from being interpolated directly into command strings, SQL statements, or comparison logic without proper escaping, type casting, or validation against an allowlist.

### Changes

- common-sysconfig.php
escape URL and proxy values passed to wget in is_available(), and cast timeout/retry parameters to int, removing a shell-metacharacter injection point reachable via the logo URL system setting

- change-license-processPost.php
use a parameterized query instead of string-interpolated SQL when looking up bucketpool_fk in ChangeBuckets()

- agent-nomos-once.php, monk/ui/oneshot.php
escape file path arguments before passing them to exec()

- lib/php/Util/DataTablesUtility.php
restrict sort direction to an ASC/DESC allow-list instead of passing the raw client-supplied value into the generated ORDER BY clause

- common-auth.php, core-auth.php:
verify passwords with password_verify() first, only falling back to the legacy seeded-sha1 check for accounts not yet migrated, and transparently rehash to bcrypt on successful login

## Testing

Expected:
- `is_available()` still returns 1/0 correctly for well-formed URLs after the escaping change.
- `ChangeBuckets()` still resolves the correct `bucketpool_fk` rows with the parameterized query.
- DataTables sorting still produces `ASC`/`DESC` output for all existing UI call sites; non-`asc`/`desc` input now defaults safely to `ASC` instead of being injected.
- Login still succeeds for accounts on both the legacy and bcrypt hash formats, and that legacy accounts are rehashed to bcrypt after a successful login.

### DIFF
--- a/src/lib/php/Util/DataTablesUtility.php
+++ b/src/lib/php/Util/DataTablesUtility.php
@@ -67,7 +67,7 @@ class DataTablesUtility
       $name = $columNamesInDatabase[$colNumber];
 
       $whichDir = 'sSortDir_' . $i;
-      $order = $inputArray[$whichDir];
+      $order = $this->sanitizeSortDirection($inputArray[$whichDir]);
       $orderArray[] = $name . " " . $order;
     }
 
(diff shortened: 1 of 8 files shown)


shaheemazmalmmd (approved):
Code looks good. needs test

PR #3728, opened 2026-07-28
fix(ui): fix index source links by highlight fragments

AnanyaSH-77:
## Description

Fix Source indexing in Conclude License so all scanner highlight occurrences are represented, not only the first one.

### Changes

1. Added support to fetch highlight-entry pages for all highlight fragments of a scanner match in `src/lib/php/Dao/HighlightDao.php`.
2. Updated Source rendering in `ajax-clearing-view.php` to generate multiple numbered links (`#1`, `#2`, `#3`, ...) for each occurrence stream.
3. Updated API agent info generation in `UploadTreeController.php` to stay aligned with UI occurrence behavior.
4. Preserved stable ordering by highlight start so numbering is deterministic.
5. Included same-page multiple fragments as separate indices.

## How to test

1. Start Fossology services and open **Conclude License** for a file with repeated scanner-highlighted license text.
2. In the **Source** column, verify an agent entry (for example, `nomos`) shows multiple indices (`#1`, `#2`, `#3`, ...) instead of only `#1`.
3. Click each index and verify navigation lands on the expected highlighted occurrence.
4. Verify cases where multiple occurrences are on the same page also show separate indices (`#1`, `#2`, `#3`, ...).
5. Confirm there are no syntax or runtime errors in the modified PHP files:

   ```bash
   php -l src/lib/php/Dao/HighlightDao.php
   php -l src/www/ui/ajax-clearing-view.php
   php -l src/www/ui/api/Controllers/UploadTreeController.php
   ```

### DIFF
--- a/src/lib/php/Dao/HighlightDao.php
+++ b/src/lib/php/Dao/HighlightDao.php
@@ -242,4 +242,35 @@ class HighlightDao
     );
     return $row['page'];
   }
+
+  /**
+   * Get page numbers for all highlight entries for a given match id.
+   *
+   * The resulting array can contain duplicate page numbers when multiple
+   * highlight fragments exist on the same page.
+   *
+   * @param int $licenseMatchId
+   * @return int[]
+   */
+  public function getPageNumbersOfHighlightEntries($licenseMatchId)
+  {
+    $stmt = __METHOD__;
+    $sql = "SELECT FLOOR(start / (
+              SELECT conf_value FROM sysconfig WHERE variablename LIKE 'BlockSizeText'
+            )::numeric) AS page
+            FROM highlight
+            WHERE fl_fk = $1
+            ORDER BY start ASC";
+
+    $this->dbManager->prepare($stmt, $sql);
+    $result = $this->dbManager->execute($stmt, array($licenseMatchId));
+
+    $pages = array();
+    while ($row = $this->dbManager->fetchArray($result)) {
+      $pages[] = intval($row['page']);
+    }
+    $this->dbManager->freeResult($result);
+
+    return $pages;
+  }
 }
--- a/src/lib/php/Dao/HighlightDao.php
+++ b/src/lib/php/Dao/HighlightDao.php
@@ -242,4 +242,38 @@ class HighlightDao
     );
     return $row['page'];
   }
+
+  /**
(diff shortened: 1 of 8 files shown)


sakshammishra112:
Hey @AnanyaSH-77,

I tested the changes locally, and I might be misunderstanding the intended behavior, but I don't think the desired outcome is being reflected correctly with the current implementation. Here's what I observed:

1. Your change shows occurrence `#1, #2, ...` based on the individual regex matches nomos records internally, instead of actually showing another occurrence of the same license text. A single license paragraph can match several regex patterns  nomos identifies a license using multiple per-license regex rules (`STRINGS.in`), and each rule that matches (e.g. "anyone to use ... applications", "The origin ... misinterpreted", "Altered source versions .. such.", "This notice ....") is recorded as its own highlight, even though they're all part of the same one occurrence. So the highlight reason here is "which regex rule matched," not "a new occurrence of the license."

2. It is also making monk's output worse. Monk already correctly highlights multiple license text occurrences each real repeat gets its own separate match id, so it already links `#1, #2, ...` correctly today. This change breaks that by additionally treating monk's internal diff-alignment fragments as extra fake occurrences of the same single match.

image


cc: @shaheemazmalmmd

src/lib/php/Dao/HighlightDao.php
```diff
@@ -242,4 +242,35 @@ public function getPageNumberOfHighlightEntry($licenseMatchId)
...
+  /**
+   * Get page numbers for all highlight entries for a given match id.
+   *
+   * The resulting array can contain duplicate page numbers when multiple
+   * highlight fragments exist on the same page.
+   *
+   * @param int $licenseMatchId
+   * @return int[]
+   */
+  public function getPageNumbersOfHighlightEntries($licenseMatchId)
```
sakshammishra112: This function is returning the page number of every highlight row tied to a match id, treating each row as a separate occurrence even though rows can also represent regex sub-matches (nomos)

shaheemazmalmmd:
@AnanyaSH-77 : thank you so much for your efforts, after a brief discussion with the community. we have decided to add the navigators in the legend box instead of adding as highlight id, 

This will ease the workflow preserving the existing behavior. 

image

PR #3738, opened 2026-08-06
feat(api): add OpenAPI v2 endpoint to rename an existing group

### LINKED ISSUE #422: Add OpenAPI v2 endpoint to rename an existing group
#### Problem

The legacy PHP UI provides an Edit Group page (`AdminGroupEdit`) that allows administrators to rename an existing group.

However, OpenAPI v2 currently does not expose any endpoint to update a group's name.

The available Group APIs are:

Image

There is currently no endpoint for renaming a group.

#### Impact

The React/Next.js UI migration cannot fully implement the Admin → Groups → Edit Group page because the required backend functionality is unavailable.

#### Feature Request

Add a REST endpoint to rename an existing group.

harshitg927:
## Description

FOSSologyUI's Admin → Groups → Edit Group page can't be implemented because OpenAPI v2 has no endpoint for renaming a group. The legacy PHP UI has had this (`AdminGroupEdit`) for a long time, so this just exposes the same capability over REST.

Adds `PUT /groups/{name}` (v2) / `PUT /groups/{id}` (v1), following the existing convention where v2 addresses groups by name and v1 by id.

Closes fossology/FOSSologyUI#422

### Changes

- Added `GroupController::updateGroup()` and registered the `PUT` route in `index.php`. Request body is `{"name": "<new name>"}`.
- Name validation reuses `AdminGroupEdit::validateGroupName()` via the plugin instead of duplicating the regex, so the API and the old UI stay in sync.
- Permission check uses `getDeletableAdminGroupMap()`, which is the same set the legacy Edit Group page uses. Personal groups (group name == user name) are excluded, as before.
- `UserDao::editGroup()` had no validation at all, so renaming a group onto an existing name silently created a duplicate. It now does the same case-insensitive uniqueness check `addGroup()` does and throws. The legacy page already catches and displays that exception, so it benefits too.
- Documented the endpoint and an `UpdateGroup` schema in `openapiv2.yaml`.
- Added 7 unit tests to `GroupControllerTest` (v1/v2 success, empty name, invalid name, not found, not group admin, name conflict).

Status codes: 200 on success, 400 for a missing/invalid name, 403 if the caller isn't an admin of the group, 404 if the group doesn't exist, 409 if another group already has that name.

## How to test

Get a token:

```bash
curl -s -X POST http://localhost/repo/api/v2/tokens \
  -H "Content-Type: application/json" \
  -d '{"username":"fossy","password":"fossy","tokenName":"test","tokenScope":"write","tokenExpire":"2026-09-01"}'
```

Create a group and rename it:

```bash
curl -s -X POST "http://localhost/repo/api/v2/groups?name=alpha" -H "Authorization: Bearer <token>"

curl -s -X PUT http://localhost/repo/api/v2/groups/alpha \
  -H "Authorization: Bearer <token>" -H "Content-Type: application/json" \
  -d '{"name":"beta"}'
```

Expect `200 {"code":200,"message":"Group beta updated.","type":"INFO"}`, and `GET /groups` should show the new name.

Error cases worth checking:

- `{"name":"  "}` or `{}` → 400
- `{"name":"12345"}` → 400 (numeric-only)
- `{"name":"bad$name!"}` → 400 (invalid characters)
- renaming a second group onto `beta`, or onto `BETA` → 409
- renaming a group to its own name → 200, should not be treated as a conflict
- `PUT /groups/doesNotExist` → 404
- `PUT /groups/fossy` (personal group) → 403

Unit tests:

```bash
phpunit -c src/phpunit.xml --filter GroupControllerTest
```

### DIFF
--- a/src/lib/php/Dao/UserDao.php
+++ b/src/lib/php/Dao/UserDao.php
@@ -471,9 +471,18 @@ class UserDao
   /**
    * @param int $groupId
    * @param string $newGroupName
+   * @throws \Exception if the name is already taken by another group
    */
   function editGroup($groupId, $newGroupName)
   {
+    $groupAlreadyExists = $this->dbManager->getSingleRow(
+            "SELECT group_pk FROM groups WHERE LOWER(group_name)=LOWER($1) AND group_pk!=$2",
+            array($newGroupName, $groupId),
+            __METHOD__.'.gExists');
+    if ($groupAlreadyExists) {
+      throw new \Exception(_("Group exists. Try different Name, Group-Name checking is case-insensitive and Duplicate not allowed"));
+    }
+
     $this->dbManager->getSingleRow('UPDATE groups SET group_name=$2 WHERE group_pk=$1;',
             array($groupId, $newGroupName),__METHOD__.'.UpdateEditGroup');
   }
--- a/src/www/ui/api/Controllers/GroupController.php
+++ b/src/www/ui/api/Controllers/GroupController.php
@@ -16,6 +16,7 @@ namespace Fossology\UI\Api\Controllers;
 use Fossology\Lib\Auth\Auth;
 use Fossology\Lib\Dao\UserDao;
 use Fossology\UI\Api\Exceptions\HttpBadRequestException;
+use Fossology\UI\Api\Exceptions\HttpConflictException;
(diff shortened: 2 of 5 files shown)


shaheemazmalmmd (approved):
Code looks good.

PR #3740, opened 2026-08-07
Feat/424 license marydone export api v2

GeorgeAlex2004:
## Description

Adds the missing marydone-specific license export endpoints to the v2 REST API and documents the existing JSON import/export endpoints, which were already implemented but missing from the OpenAPI spec. Also fixes a bug where the general license export endpoints only ever returned marydone candidate licenses instead of all licenses.

### Changes

- Add `GET /license/marydone/export-csv` and `GET /license/marydone/export-json` endpoints, which were referenced by the React License Administration UI but never implemented on the backend.
- Document the existing `POST /license/import-json` and `GET /license/export-json` endpoints in the v2 OpenAPI spec.
- Fix `GET /license/export-csv` and `GET /license/export-json` to pass `$allCandidates=true` to `LicenseCsvExport::createCsv()`, so "export all" now actually exports all licenses (main + candidates) as documented, instead of silently only exporting marydone-flagged candidates.
- Move marydone-only export semantics onto the new dedicated `/license/marydone/*` routes.
- Extract the shared CSV/JSON export logic into a private `exportLicenseFile()` helper on `LicenseController` to avoid duplicating the four near-identical endpoint implementations.
- Add unit tests covering all four export endpoints in `LicenseControllerTest.php`.

## How to test

1. Run the unit tests:
vendor/bin/phpunit -c src/www/ui_tests/api/tests.xml --filter LicenseControllerTest

2. Against a running instance, as an admin user:
- `GET /api/v2/license/export-csv` and `/export-json` → response should include both main and candidate licenses.
- `GET /api/v2/license/marydone/export-csv` and `/marydone/export-json` → response should include only marydone-flagged candidate licenses.
- `POST /api/v2/license/import-json` with a previously exported JSON file → licenses should import successfully.
3. Check that the v2 OpenAPI/Swagger docs render correctly for the new and updated `/license/*` paths.

Refs fossology/FOSSologyUI#424

### DIFF
--- a/src/www/ui/api/Controllers/LicenseController.php
+++ b/src/www/ui/api/Controllers/LicenseController.php
@@ -848,7 +848,7 @@ class LicenseController extends RestController
   }
 
   /**
-   * Export licenses to CSV file
+   * Export all licenses (main + candidates) to CSV file
    *
    * @param Request $request
    * @param ResponseHelper $response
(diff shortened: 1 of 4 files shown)


Kaushl2208:
@GeorgeAlex2004, Same issues like https://github.com/fossology/fossology/pull/3751
Two separate endpoints not needed

PR #3741, opened 2026-08-07
feat(api): add OpenAPI v2 endpoints for license compatibility rules

### LINKED ISSUE #423: Add OpenAPI v2 endpoints for License Compatibility Rules
#### Problem

The legacy PHP implementation (AdminLicenseCompatibilityRules) provides functionality to manage License Compatibility Rules, including creating, viewing, updating, and deleting individual rules, as well as importing and exporting rules using YAML files.

These operations are currently handled through the legacy PHP UI and do not have equivalent OpenAPI v2 endpoints.

The missing API operations include:
- Retrieve License Compatibility Rules
- Create a new License Compatibility Rule
- Update an existing License Compatibility Rule
- Delete a License Compatibility Rule
- Import License Compatibility Rules from a YAML file
- Export License Compatibility Rules as a YAML file

#### Impact

The React/Next.js migration cannot fully implement the Admin → License Compatibility Rules page because the required backend APIs for managing and transferring compatibility rules are not available.

The following UI is affected: 
- Admin → License Compatibility Rules → Select Compatibility Rules
- Admin → License Compatibility Rules → Operations   
   - Rules Import
   - Rules Export

#### Feature Request

Expose REST endpoints for:
- List compatibility rules (for example: `GET /license-compatibility-rules` 
     - Retrieve existing compatibility rules, with support for pagination and search/filtering where applicable.
- Create a compatibility rule (for example: `POST /license-compatibility-rules`
     - Add a new compatibility rule with the selected licenses, license types, compatibility result, and rule description.
- Update a compatibility rule `PUT /license-compatibility-rules/{id}`
     - Modify an existing compatibility rule.
- Delete a compatibility rule `DELETE /license-compatibility-rules/{id}`
     - Permanently remove an existing compatibility rule.
- Importing License Compatibility Rules (for example: `POST /license-compatibility-rules/import` )
- Exporting License Compatibility Rules (for example: `GET /license-compatibility-rules/export` )

harshitg927:
## Description

The only way to manage license compatibility rules today is the legacy `AdminLicenseCompatibilityRules` page, which is why the React/Next.js Admin migration is stuck on this screen. This PR adds the missing OpenAPI v2 endpoints for the full CRUD flow plus the YAML import/export.

While hooking the import endpoint up to the existing YAML importer I ran into a few bugs in the code it depends on, so I've fixed those here as well - otherwise the new endpoints would have inherited them. They're listed separately below so they're easy to review on their own.

Closes fossology/FOSSologyUI#423

### Changes

New endpoints, all admin-only and documented in `openapiv2.yaml`:

| Method | Route |
| --- | --- |
| GET | `/license-compatibility-rules` (supports `page`, `limit`, `search`, returns `X-Total-Pages`) |
| POST | `/license-compatibility-rules` |
| PUT | `/license-compatibility-rules/{id}` |
| DELETE | `/license-compatibility-rules/{id}` |
| POST | `/license-compatibility-rules/import` (YAML, `fileInput`) |
| GET | `/license-compatibility-rules/export` (YAML, optional `?id=`) |

New files:

- `src/www/ui/api/Controllers/LicenseCompatibilityRuleController.php`
- `src/www/ui/api/Models/LicenseCompatibilityRule.php` - exposes both the license ids and their short names, so the UI doesn't have to resolve names itself
- tests for the controller, the model, the DAO and the YAML importer

Notes on behaviour:

- `search` filters on the rule description, same as the legacy page.
- A `null` license or license type means "matches any", same as `0` / `---` in the legacy dropdowns.
- License ids are validated against `license_ref` and license types against the `LicenseTypes` sysconfig setting.
- `POST` rejects a rule with no license *and* no type - that row is what `CompatibilityDao` reads as the server's default compatibility, so creating one would shadow it.
- `PUT` only touches the fields present in the body.

Fixes in existing code that the endpoints depend on:

- `CompatibilityDao::getAllRules()` / `getTotalRulesCount()` interpolated the search term straight into an `ILIKE`. It's a bound parameter now. The legacy DataTables search box reaches the same code, and a payload like `x' OR 1=1 OR comment ILIKE 'y` returned every rule, while a bare `'` produced a 500.
- `CompatibilityDao::updateRuleFromArray()` didn't cast PHP booleans. `insertInto()` runs params through `cleanupParamsArray()` but `getSingleRow()` doesn't, so a bound `false` reached Postgres as `""` and the `bool` column rejected it. `"t"`/`"f"` from the legacy form still work.
- `LicenseCompatibilityRulesYamlImport::updateLicense()` wrote to `$rule["compatibility"]`, but `updateRuleFromArray()` only reads `"result"`, and it guarded the update with `count($rule) > 1`. Between the two, a YAML import could never change the compatibility of an existing rule, and a comment-only change was a silent no-op that still logged "updated comment".
- Same class dereferenced `getLicenseByShortName()` without a null check, so a YAML naming an unknown license raised a `TypeError` (not caught by the `catch (\Exception)` in `handleFile()`). It now reports the license name.
- `AdminLicenseFromYAML::handleFileUpload()` is public with the `$fromRest` `[ok, message, code]` return, matching `AdminObligationFromCSV`. The legacy page's output is unchanged.

Known limitation, pre-existing and not touched here: re-importing an export only updates rules that have exactly two of `first_rf_fk`/`second_rf_fk`/`first_type`/`second_type` set. Rules with one, three or four get inserted again, because `handleYamlLicense()` guards its lookup with `count($param) == 2` (dates back to 2021). Two-field rules are what the UI normally produces, so the common round trip is clean. Happy to fix it here if reviewers prefer, but it changes import matching for the legacy page too, so I left it out.

## How to test

Get a token:

```bash
curl -X POST http://localhost/repo/api/v2/tokens \
  -H 'Content-Type: application/json' \
  -d '{"username":"fossy","password":"fossy","tokenName":"rules","tokenScope":"write","tokenExpire":"2026-12-31"}'
export T='Bearer <token>'
export A=http://localhost/repo/api/v2/license-compatibility-rules
```

Create, list, update, delete:

```bash
# needs a real license id, e.g. from GET /repo/api/v2/license/MIT
curl -X POST $A -H "Authorization: $T" -H 'Content-Type: application/json' \
  -d '{"firstLicenseId":125,"secondType":"Strong Copyleft","comment":"A permissive license against a copyleft one","compatibility":true}'

curl -H "Authorization: $T" "$A?limit=10&page=1&search=copyleft" -D -

# partial update, only the fields you send change
curl -X PUT $A/<id> -H "Authorization: $T" -H 'Content-Type: application/json' \
  -d '{"compatibility":false}'

curl -X DELETE $A/<id> -H "Authorization: $T"
```

Export, then import it back:

```bash
curl -H "Authorization: $T" "$A/export" -o rules.yaml
curl -X POST $A/import -H "Authorization: $T" -F "fileInput=@rules.yaml"
```

Worth checking on the error paths too - a bad license id, a license type that isn't in the `LicenseTypes` setting, an empty description, a rule with no license and no type, and a non-`.yaml`/`.yml` upload should all come back as 400s.

Unit tests:

```bash
cd src/www/ui_tests/api && phpunit -c tests.xml
phpunit --bootstrap src/phpunit-bootstrap.php src/lib/php/Application/test
phpunit src/lib/php/Dao/test/CompatibilityDaoTest.php   # needs postgres
```

Since the DAO and the YAML plugin are shared with the old UI, please also sanity-check Admin → License Admin → Compatibility Rules (list, search, add, edit, delete) and the Rules Import / Rules Export pages.

### DIFF
--- a/src/lib/php/Application/LicenseCompatibilityRulesYamlImport.php
+++ b/src/lib/php/Application/LicenseCompatibilityRulesYamlImport.php
@@ -114,14 +114,20 @@ class LicenseCompatibilityRulesYamlImport
     if ($normalizedRow["firstname"] != null) {
       $firstLicense = $this->licenseDao->getLicenseByShortName(
           $normalizedRow["firstname"], null);
-      $firstLicenseId = $firstLicense->getId();
-      $normalizedRow["firstname"] = $firstLicenseId;
+      if ($firstLicense === null) {
+        throw new \UnexpectedValueException("Unable to find license " .
+          $normalizedRow["firstname"] . " from YAML");
+      }
+      $normalizedRow["firstname"] = $firstLicense->getId();
     }
     if ($normalizedRow["secondname"] != null) {
       $secondLicense = $this->licenseDao->getLicenseByShortName(
           $normalizedRow["secondname"], null);
-      $secondLicenseId = $secondLicense->getId();
-      $normalizedRow["secondname"] = $secondLicenseId;
+      if ($secondLicense === null) {
+        throw new \UnexpectedValueException("Unable to find license " .
+          $normalizedRow["secondname"] . " from YAML");
+      }
+      $normalizedRow["secondname"] = $secondLicense->getId();
     }
     return $this->handleYamlLicense($normalizedRow);
   }
@@ -144,14 +150,14 @@ class LicenseCompatibilityRulesYamlImport
 
     $log = [];
     if ($old_comp != $new_comp) {
-      $rule["compatibility"] = $this->dbManager->booleanToDb($new_comp);
+      $rule["result"] = $new_comp;
       $log[] = "updated compatibility";
     }
     if (!empty($row['comment']) && $row['comment'] != $oldRule['comment']) {
       $rule["comment"] = $row["comment"];
       $log[] = "updated comment";
     }
-    if (count($rule) > 1) {
+    if (!empty($rule)) {
       try {
         $this->compatibilityDao->updateRuleFromArray([$lrPk => $rule]);
       } catch (\UnexpectedValueException $e) {
--- a/src/lib/php/Application/test/LicenseCompatibilityRulesYamlImportTest.php
+++ b/src/lib/php/Application/test/LicenseCompatibilityRulesYamlImportTest.php
@@ -1,0 +1,245 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © 2026 Harshit Gandhi <gandhiharshit716@gmail.com>
+
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+
+namespace Fossology\Lib\Application;
+
+use Fossology\Lib\Dao\CompatibilityDao;
+use Fossology\Lib\Dao\LicenseDao;
+use Fossology\Lib\Dao\UserDao;
+use Fossology\Lib\Data\License;
+use Fossology\Lib\Db\DbManager;
+use Fossology\Lib\Test\Reflectory;
+use Mockery as M;
+
+/**
+ * @class LicenseCompatibilityRulesYamlImportTest
+ * @brief Test for LicenseCompatibilityRulesYamlImport
+ */
+class LicenseCompatibilityRulesYamlImportTest extends \PHPUnit\Framework\TestCase
+{
+  /** @var DbManager $dbManager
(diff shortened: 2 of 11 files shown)


shaheemazmalmmd (approved):
Code looks good.

PR #3749, opened 2026-08-10
feat(api): monk agent renewal api v2

### LINKED ISSUE #425: Add OpenAPI v2 endpoint to trigger Monk agent renewal
#### Problem

The legacy PHP administration page renews the Monk agent using:
`$agentDao->renewCurrentAgent('monk');`

There is currently no equivalent OpenAPI v2 endpoint for triggering this administrative operation.

#### Impact

The React/Next.js migration cannot implement the Admin → Monk page because the required backend functionality is unavailable.

#### Feature Request

Expose a REST endpoint that triggers renewal of the current Monk agent, providing functionality equivalent to the existing PHP implementation.

GeorgeAlex2004:
## Description

Adds an OpenAPI v2 endpoint to trigger Monk agent renewal. The legacy PHP admin page renews Monk via `AgentDao::renewCurrentAgent('monk')`, but no REST equivalent existed, which was blocking the React/Next.js Admin → Monk page from being implemented.

### Changes

- Add `POST /agents/monk/renew`, an admin-only REST endpoint that calls `AgentDao::renewCurrentAgent('monk')`, equivalent to the legacy "Manage Monk Revision" admin page (`admin_monk_revision`).
- Return `404` if the Monk agent has never run, instead of calling `renewCurrentAgent()` on a nonexistent agent.
- Document the new endpoint (and a new `Agent` tag) in `openapiv2.yaml`.
- Fix `AgentDao::createAgentRef()` to guard against `getSingleRow()`/`fetchArray()` returning `false` (not an array) when no row matches — this previously emitted spurious PHP warnings on the new endpoint's 404 path (and any other caller hitting a not-yet-run agent), the same bug class already fixed for `maintagent`. The fix is behavior-preserving: still returns `AgentRef(0, null, null)`.
- Add `AgentControllerTest` covering a successful renewal, the 404-when-never-run case, and the 403-for-non-admins case.

## How to test

1. As an admin, `POST /repo/api/v2/agents/monk/renew` (no request body needed) — expect `200` with an `Info` success message, and a new `monk` row in the `agent` table with a bumped `agent_rev`.
2. As a non-admin (or unauthenticated) user, repeat the request — expect `403`.
3. On an install where Monk has never run, repeat the request — expect `404`, with no PHP warnings in the logs.
4. Run the unit tests: `phpunit -c src/phpunit.xml --testsuite "Fossology PhpUnit Test Suite" --filter AgentControllerTest`.

Closes fossology/FOSSologyUI#425

### DIFF
--- a/src/lib/php/Dao/AgentDao.php
+++ b/src/lib/php/Dao/AgentDao.php
@@ -236,11 +236,15 @@ class AgentDao
   }
 
   /**
-   * @param string[] $row
+   * @param string[]|false $row getSingleRow()/fetchArray() return false, not
+   *        an array, when no row matches.
    * @return AgentRef
    */
   private function createAgentRef($row)
   {
+    if (!is_array($row)) {
+      return new AgentRef(0, null, null);
+    }
     return new AgentRef(intval($row['agent_pk']), $row['agent_name'], $row['agent_rev']);
   }
 
--- a/src/www/ui/api/Controllers/AgentController.php
+++ b/src/www/ui/api/Controllers/AgentController.php
@@ -1,0 +1,53 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © 2026 Fossology contributors
+
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+/**
+ * @file
+ * @brief Controller for agent related REST endpoints
+ */
+
+namespace Fossology\UI\Api\Controllers;
+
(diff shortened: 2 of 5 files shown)


Kaushl2208:
Hey @GeorgeAlex2004 , This is not mandatory according to the UI requirements. I am closing this now. Thanks for the efforts.

PR #3754, opened 2026-08-10
Add OpenAPI v2 endpoints for tag create and enable/disable

### LINKED ISSUE #437: Add OpenAPI v2 endpoints for Tag functionalities
#### Problem

These operations of create, enable and disable tag under Admin → Tag currently have no equivalent OpenAPI v2 endpoints.

Missing operations include:
- Create Tag
- Enable Tag
- Disable Tag

#### Impact

The React/Next.js migration cannot implement the following API functionality:
- Admin → Tag → Create Tag
   - Create Tag
- Admin → Tag → Enable/Disable Tag
   - Enable Tag
   - Disable Tag

#### Feature Request

Expose REST endpoints for:
- Create Tag
- Enable Tag
- Disable Tag

GeorgeAlex2004:
## Description

Adds the missing OpenAPI v2 REST endpoints for tag administration: creating a
tag, and enabling/disabling tag display on an upload. These were the last
pieces blocking the FOSSologyUI Admin > Tag pages from being wired up to the
API.

### Changes

- Adds `TagController::createTag` (`POST /tags`) and
  `TagController::setTagDisplayStatus` (`PUT /uploads/{id}/tags/display`),
  backed by the `tag` and `tag_manage` tables, mirroring the behavior of the
  legacy `admin_tag` / `admin_tag_manage` plugins.
- Adds a `Tag` model (`src/www/ui/api/Models/Tag.php`) that validates and
  parses the create-tag request body (name required, ≤32 characters,
  restricted to the same character set as the legacy plugin; non-array/
  non-string input is rejected instead of causing an internal error).
- Registers both routes in `index.php` and documents them in
  `openapiv2.yaml` (new `Tags` tag group, `Tag` schema).
- Fixes `AjaxFolderContents` (`/folders/{id}/contents`) to also return the
  upload's real `upload_pk` as a new `uploadId` field. Previously the `id`
  field on each entry was `foldercontents_pk`, which gave REST clients no
  way to know which upload a listed folder entry actually refers to -
  something the enable/disable tag display page needs in order to target
  the correct upload. This is additive; the existing `id`/`content`/
  `removable` fields are unchanged.
- Adds `TagControllerTest.php` covering both endpoints: success paths,
  admin-only enforcement, validation errors (missing/invalid/oversized
  name, non-string input, malformed body), duplicate-name conflict, and
  upload not-found/not-accessible cases.

## How to test

1. Build/run the stack (`podman-compose up -d` from the repo root) and log
   in as an admin user.
2. Create a tag:
  curl -X POST http://localhost/repo/api/v2/tags 
  
  -H "Authorization: Bearer <token>" 
  
  -H "Content-Type: application/json" 
  
  -d '{"name": "MyTag", "description": "A test tag"}'
  
  
  Expect `201` with the new tag's id. Repeating the same request should
  return `409` (duplicate name).
3. Enable/disable tag display for an upload:
  curl -X PUT http://localhost/repo/api/v2/uploads/<uploadId>/tags/display 
  
  -H "Authorization: Bearer <token>" 
  
  -H "Content-Type: application/json" 
  
  -d '{"enabled": false}'
  
  
  Expect `200`. Verify with `SELECT * FROM tag_manage WHERE upload_fk =
  <uploadId>;` that a disabled row was inserted, and that re-sending with
  `{"enabled": true}` removes it.
4. Run the new unit tests:
  vendor/bin/phpunit -c phpunit.xml --no-coverage 
  
  src/www/ui_tests/api/Controllers/TagControllerTest.php

Fixes fossology/FOSSologyUI#437.

### DIFF
--- a/src/www/ui/admin-tag.php
+++ b/src/www/ui/admin-tag.php
@@ -25,16 +25,66 @@ class admin_tag extends FO_Plugin
   }
 
   /**
-   * \brief Create Tag without tagging anything
+   * \brief Check whether a tag with the given name already exists
    *
-   * \return null for success or error text
+   * \param string $tag_name
+   * \return boolean true if a tag with this name exists
    */
-  function CreateTag()
+  function TagExists($tag_name)
   {
     global $PG_CONN;
 
-    $tag_name = GetParm('tag_name', PARM_TEXT);
-    $tag_desc = GetParm('tag_desc', PARM_TEXT);
+    $sql = "SELECT tag_pk FROM tag WHERE tag = '" . pg_escape_string($tag_name) . "'";
+    $result = pg_query($PG_CONN, $sql);
+    DBCheckResult($result, $sql, __FILE__, __LINE__);
+    $exists = pg_num_rows($result) > 0;
+    pg_free_result($result);
+    return $exists;
+  }
+
+  /**
+   * \brief Get the tag_pk of a tag by name
+   *
+   * \param string $tag_name
+   * \return integer|null pk of the tag, or null if no such tag exists
+   */
+  function GetTagId($tag_name)
+  {
+    global $PG_CONN;
+
+    $sql = "SELECT tag_pk FROM tag WHERE tag = '" . pg_escape_string($tag_name) . "' LIMIT 1;";
+    $result = pg_query($PG_CONN, $sql);
+    DBCheckResult($result, $sql, __FILE__, __LINE__);
+    $tagPk = null;
+    if (pg_num_rows($result) > 0) {
+      $row = pg_fetch_assoc($result);
+      $tagPk = intval($row['tag_pk']);
+    }
+    pg_free_result($result);
+    return $tagPk;
+  }
+
+  /**
+   * \brief Create Tag without tagging anything
+   *
+   * \param array $tag_array Optional array with tag_name/tag_desc, used
+   *        when calling this method outside of the plugin's own POST
+   *        handling (e.g. from the REST API). Falls back to request
+   *        parameters when not given.
+   *
+   * \return null for success or error text
+   */
+  function CreateTag($tag_array = null)
+  {
+    global $PG_CONN;
+
+    if (isset($tag_array)) {
+      $tag_name = $tag_array['tag_name'];
+      $tag_desc = $tag_array['tag_desc'];
+    } else {
+      $tag_name = GetParm('tag_name', PARM_TEXT);
+      $tag_desc = GetParm('tag_desc', PARM_TEXT);
+    }
     if (empty($tag_name)) {
       $text = _("TagName must be specified. Tag Not created.");
       return ($text);
--- a/src/www/ui/api/documentation/openapiv2.yaml
+++ b/src/www/ui/api/documentation/openapiv2.yaml
@@ -57,6 +57,8 @@ tags:
     description: Maintenance operations
   - name: Overview
     description: Overview of FOSSology operations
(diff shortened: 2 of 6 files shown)


src/www/ui/api/Models/Tag.php
```diff
@@ -0,0 +1,154 @@
...
+  {
+    if (! is_array($inputTag)) {
+      return -2;
+    }
+    $inputKeys = array_keys($inputTag);
+    $intersectKeys = array_intersect($inputKeys, self::ALLOWED_KEYS);
+    if (count($inputKeys) > 0 && count($intersectKeys) != count($inputKeys)) {
+      return -1;
+    }
+    if (! array_key_exists('name', $inputTag) || ! is_string($inputTag['name']) ||
```
Kaushl2208: This is pretty much all useless, There is already `createTag()` functionality, Follow the code, Look for the `admin_tag` file, Check how the plugins are loaded.

Follow the pattern. There are a lot of examples.

Kaushl2208 (changes requested):
@GeorgeAlex2004 , PRs are getting real sloppy right about now. 

Spend time with the generated code(REVIEW REVIEW REVIEW), See if that follows the pattern. Dont start creating logics randomly.

src/www/ui/api/index.php
```diff
@@ -311,6 +313,13 @@ function (\Slim\Routing\RouteCollectorProxy $app) {
     $app->any('/{params:.*}', BadRequestController::class);
   });
 
+////////////////////////////TAGS/////////////////////
+$app->group('/tags',
+  function (\Slim\Routing\RouteCollectorProxy $app) {
+    $app->post('', TagController::class . ':createTag');
```
Kaushl2208, line 319: Creating Tag I can see. What about fetching tags? UI also needs to show Tags, Update Tags, Delete Tags as well?

src/www/ui/async/AjaxFolderContents.php
```diff
@@ -51,15 +52,17 @@ public function handle(Request $request)
       $uploadDate = explode(".", $upload['upload_ts'])[0];
```
Kaushl2208, line 52: Is this in the scope of this PR?

src/www/ui/admin-tag.php
Kaushl2208, line 104: If you have already modularised this code, Why not replace it with the function?

src/www/ui/api/documentation/openapiv2.yaml
```diff
@@ -1088,6 +1133,64 @@ paths:
...
+      requestBody:
+        description: Whether tag display should be enabled for the upload
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                enabled:
+                  type: boolean
```
Kaushl2208, line 1162: ENUM would make more sense?

src/www/ui/api/documentation/openapiv2.yaml
```diff
@@ -5897,6 +6000,12 @@ components:
         removable:
           type: boolean
           description: Is the folder removable.
+        uploadId:
+          type: integer
+          description: >
+            Id of the upload (upload_pk), if this content is an upload.
+            Null if the content is a subfolder.
+          nullable: true
```
Kaushl2208, line 6008: Not in the scope.

Kaushl2208 (changes requested):

shaheemazmalmmd:
Closing as there is no response.

PR #3757, opened 2026-08-11
feat(api): add endpoint to revoke REST API tokens

### LINKED ISSUE #447: Add OpenAPI v2 endpoint to revoke REST API tokens
#### Problem

The OpenAPI v2 specification currently exposes the following REST API token endpoints:

- `POST /users/tokens` – Create a new REST API token
- `GET /users/tokens/{type}` – Retrieve active or expired REST API tokens

However, there is no endpoint to revoke an existing REST API token.

The legacy PHP UI provides a Revoke action for tokens, but the functionality is not available through the OpenAPI v2 API.

#### Impact

The React/Next.js migration cannot fully replicate the REST API Tokens page because the Revoke action cannot be implemented without a corresponding backend endpoint.

As a result:

- Users can create and list REST API tokens through the new UI.
- Existing tokens cannot be revoked.
- The migrated page does not provide feature parity with the legacy PHP implementation.

#### Feature Request

Expose a REST endpoint to revoke an existing REST API token.

A possible design would be:

- DELETE `/users/tokens/{tokenId}`

or

- POST `/users/tokens/{tokenId}/revoke`

where `tokenId` uniquely identifies the token to be revoked.

Example successful response:

{
  "message": "Token revoked successfully"
}

Example error response:

{
  "code": 404,
  "message": "Token not found"
}

This endpoint would allow the React/Next.js UI to implement the Revoke action and achieve feature parity with the legacy PHP interface.

GeorgeAlex2004:
## Description

The React/Next.js UI migration cannot replicate the "Revoke" action on the REST API Tokens page because no backend endpoint exists for it - tokens could only be created and listed via the API, not invalidated.

### Changes

- Add `DELETE /users/tokens/{tokenId}`, restricted to the token's own owner.
- Add `DbHelper::invalidateTokenForUser()`: ownership check and update happen in a single DB round trip via a non-modifying CTE that resolves existence/ownership, LEFT-JOINed against a modifying CTE whose `UPDATE` only touches the row when the ownership predicate matches.
- `UserController::revokeRestApiToken()` returns `200` on success, `400` for a malformed token id, `404` if the token doesn't exist, `403` if it belongs to another user.
- Document the new endpoint in both `openapi.yaml` and `openapiv2.yaml`.
- Add unit tests covering all four outcomes (success, invalid id, not found, forbidden).

## How to test

- Run `phpunit --configuration src/phpunit.xml src/www/ui_tests/api/Controllers/UserControllerTest.php` - all pass.
- Run `phpcs --standard=src/fossy-ruleset.xml` on the changed files - clean.
- Manually: create a token via `POST /users/tokens`, list it via `GET /users/tokens/active` to get its `id`, then call `DELETE /users/tokens/{id}` as that same user - the token disappears from the active list. Retrying the delete, or calling it as a different user, returns `404`/`403` respectively.

Fixes fossology/FOSSologyUI#447

### DIFF
--- a/src/www/ui/api/Controllers/UserController.php
+++ b/src/www/ui/api/Controllers/UserController.php
@@ -17,6 +17,7 @@ use Fossology\Lib\Exceptions\DuplicateTokenNameException;
 use Fossology\UI\Api\Exceptions\HttpBadRequestException;
 use Fossology\UI\Api\Exceptions\HttpConflictException;
 use Fossology\UI\Api\Exceptions\HttpErrorException;
+use Fossology\UI\Api\Exceptions\HttpForbiddenException;
 use Fossology\UI\Api\Exceptions\HttpInternalServerErrorException;
 use Fossology\UI\Api\Exceptions\HttpNotFoundException;
 use Fossology\UI\Api\Exceptions\HttpTooManyRequestException;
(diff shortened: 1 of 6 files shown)


Kaushl2208:
Closing this for now. As discussed for #3754

PR #3761, opened 2026-08-12
fix(reuser): initialize select2 on folder in schedule agents page

shaheemazmalmmd:
## Description

initialize select2 on folder in schedule agents page.

## How to test

* In a running instance with multiple folders search and select different folders in schedule agents page and upload from page.


its-sushant (approved):
LGTM!

PR #3762, opened 2026-08-13
fix(nomos): detect name-only licence references and fix over-grants

shaheemazmalmmd:
## Description

Detect name-only license references and fix over-grants, resolve license titles and missed license names

### Changes

Strings.in: Fix nomos footprints for over grants.
Parse.c: use correct names.

## How to test

* General Testing is needed to check license findings of nomos in HEAD and also this commit.
* Compare the results.


Kaushl2208 (approved):
Looks good.

Issue #3763, opened 2026-08-13
Redundant per-row query in CopyrightDao::updateTable() causes N+1 pattern on bulk clearing operations


Aditya30ag:
### Description
`CopyrightDao::updateTable()` fetches a set of rows with a query that already LEFT JOINs the "event" table, then for every single row issues a second, completely redundant query to re-derive information that was already returned by the first query.

### Location
`src/lib/php/Dao/CopyrightDao.php`, function `updateTable()`

**Initial query (line ~487)** already LEFT JOINs `$cpTableEvent` and
selects its primary key:

```php
$sql = "SELECT DISTINCT ON ($cpTablePk, ut.uploadtree_pk) $cpTablePk, ut.uploadtree_pk, ut.upload_fk, ce." . $cpTableEvent . "_pk
FROM $cpTable as cp
INNER JOIN $itemTable AS ut ON cp.pfile_fk = ut.pfile_fk
LEFT JOIN $cpTableEvent AS ce ON ce.$cpTableEventFk = cp.$cpTablePk
  AND ce.upload_fk = ut.upload_fk AND ce.uploadtree_fk = ut.uploadtree_pk
WHERE $withHash ( ut.lft BETWEEN $1 AND $2 ) $agentFilter AND ut.upload_fk = $3";

$rows = $this->dbManager->getRows($sql, $params, $stmt);
```

**Inside the loop over those rows (lines ~496–503)**, an extra query
re-derives the exact same fact using the exact same join key:

```php
foreach ($rows as $row) {
    ...
    $sqlExists = "SELECT exists(SELECT 1 FROM $cpTableEvent WHERE $cpTableEventFk = $1 AND upload_fk = $2 AND uploadtree_fk = $3)::int";
    $rowExists = $this->dbManager->getSingleRow($sqlExists, array($row[$cpTablePk], $row['upload_fk'], $row['uploadtree_pk']), $stmt.'Exists');
    $eventExists = $rowExists['exists'];
```

Since `ce` is joined with a `LEFT JOIN` on precisely `($cpTableEventFk, upload_fk, uploadtree_fk)`, the column `ce.{$cpTableEvent}_pk` already fetched in `$row` is `NULL` **iff** no matching event row exists which is exactly what the second
query recomputes from scratch.

### Why it matters
[...]

PR #3765, opened 2026-08-13
perf(copyright): remove redundant per-row exists query in CopyrightDa…

Aditya30ag:
## Description
Removes a redundant per-row DB query in `CopyrightDao::updateTable()` that caused an N+1 query pattern on bulk clearing operations.

### Changes
- Deleted the extra `SELECT exists(...)` query run for every row in the loop.
- `$eventExists` is now derived from `ce.{$cpTableEvent}_pk`, a column already returned by the initial LEFT JOIN query no new query needed.

## How to test
- Run existing tests: `testUpdateTable`, `testDeleteCopyright` in `CopyrightDaoTest.php` should pass unchanged.
- Manually: run a bulk "mark folder Irrelevant" action on a large upload and compare Postgres query counts/timing before and after.

#3763


shaheemazmalmmd (approved):

PR #3767, opened 2026-08-17
fix(api): handle missing V2 query parameters in FolderController

### LINKED ISSUE #3736: fix(api): Undefined array key warnings in FolderController for missing V2 query parameters
## Description

`FolderController` directly accesses values from `$request->getQueryParams()` without checking whether the corresponding array keys exist.

Under PHP 8.x, omitting certain query parameters causes `Undefined array key` warnings before the controller can apply its normal validation or default handling.

This affects the following V2 REST API methods:

* `createFolder()`
* `editFolder()`
* `copyFolder()`

For example, `createFolder()` reads:

```php
$parentFolder = $queryParams['parentFolder'];
$folderName = trim($queryParams['folderName']);
$folderDescription = trim($queryParams['folderDescription']);
```

If `folderDescription` is omitted, PHP emits an `Undefined array key` warning instead of allowing the request to proceed with the controller's intended behavior.

Similarly, `editFolder()` and `copyFolder()` directly access query parameters that may be absent, resulting in the same runtime warning.

---

### Case 1 – Missing optional parameter

Send a V2 request without `folderDescription`:

```http
POST /api/v2/folders?parentFolder=1&folderName=TestFolder
```

### Case 2 – Missing required parameter

Send a V2 request without `action`:

```http
PUT /api/v2/folders/5?parent=2
```

Run the application with PHP 8.x and observe the server logs.

---

## Expected Behavior

* Missing optional query parameters should be handled without emitting PHP warnings.
* Missing required query parameters should reach the existing validation logic and return the appropriate HTTP 400 response instead of triggering a PHP runtime warning.

---

## Actual Behavior

PHP 8 emits runtime warnings similar to:

```text
PHP Warning: Undefined array key "folderDescription" in FolderController.php
PHP Warning: Undefined array key "action" in FolderController.php
```

The warning occurs before the controller can handle the request according to the API contract.

---

## Environment

* **FOSSology version:** `master`
* **Operating System:** Ubuntu 22.04 (WSL2)
* **PHP Version:** PHP 8.x

---

## Additional Information

Other REST API controllers already perform safer query parameter handling before accessing array elements. Aligning `FolderController` with the same approach would eliminate these PHP 8 runtime warnings while preserving the existing validation behavior for required parameters.

tanmaysachann:
## Summary

`FolderController::createFolder()`, `editFolder()`, and `copyFolder()` read values directly off `getQueryParams()` without checking whether the key exists. On PHP 8.x, omitting an optional or required query parameter on a V2 request triggers an `Undefined array key` warning before the controller's own validation logic runs.

This defaults each of those reads to `''` (matching the pattern already used elsewhere, e.g. `GroupController::createGroup()`), so:
- Missing **optional** parameters (e.g. `folderDescription`) no longer emit a warning.
- Missing **required** parameters (e.g. `action` on copy/move) fall through cleanly to the existing validation and return the intended `400` error instead of a PHP warning.

Fixes #3736

## Changes

- `src/www/ui/api/Controllers/FolderController.php`: `$queryParams['key']` → `$queryParams['key'] ?? ''` in all three methods (7 lines across 3 functions).

## Test plan

Ran against a local Docker Compose instance (PHP 8.2), using the V2 API with a bearer token:

- [x] `POST /api/v2/folders` with `folderDescription` omitted → `201`, no warning (previously: `Undefined array key "folderDescription"` in the logs)
- [x] `PUT /api/v2/folders/{id}` with `action` omitted entirely → `400 "Action can be one of [copy,move]!"`, no warning (previously: `Undefined array key "action"`)
- [x] `PUT /api/v2/folders/{id}` with `name`/`description` omitted → falls through to existing validation, no warning
- [x] Confirmed via `docker logs` that no `Undefined array key` warnings appear across these requests, where they did before the fix
- [x] Confirmed existing behavior for fully-populated requests is unchanged

### DIFF
--- a/src/www/ui/api/Controllers/FolderController.php
+++ b/src/www/ui/api/Controllers/FolderController.php
@@ -92,9 +92,9 @@ class FolderController extends RestController
   {
     if (ApiVersion::getVersion($request) == ApiVersion::V2) {
       $queryParams = $request->getQueryParams();
-      $parentFolder = $queryParams['parentFolder'];
-      $folderName = trim($queryParams['folderName']);
-      $folderDescription = trim($queryParams['folderDescription']);
+      $parentFolder = $queryParams['parentFolder'] ?? '';
+      $folderName = trim($queryParams['folderName'] ?? '');
+      $folderDescription = trim($queryParams['folderDescription'] ?? '');
     } else {
       $parentFolder = $request->getHeaderLine('parentFolder');
       $folderName = trim($request->getHeaderLine('folderName'));
@@ -180,8 +180,8 @@ class FolderController extends RestController
     $folderId = $args['id'];
     if (ApiVersion::getVersion($request) == ApiVersion::V2) {
       $queryParams = $request->getQueryParams();
-      $newName = $queryParams['name'];
-      $newDesc = $queryParams['description'];
+      $newName = $queryParams['name'] ?? '';
+      $newDesc = $queryParams['description'] ?? '';
     } else {
       $newName = $request->getHeaderLine('name');
       $newDesc = $request->getHeaderLine('description');
@@ -216,8 +216,8 @@ class FolderController extends RestController
     $folderId = $args['id'];
     if (ApiVersion::getVersion($request) == ApiVersion::V2) {
       $queryParams = $request->getQueryParams();
-      $newParent = $queryParams['parent'];
-      $action = strtolower($queryParams['action']);
+      $newParent = $queryParams['parent'] ?? '';
+      $action = strtolower($queryParams['action'] ?? '');
     } else {
       $newParent = $request->getHeaderLine('parent');
       $action = strtolower($request->getHeaderLine('action'));


tanmaysachann:
The failing `docker-test` check is unrelated to this change. Root cause from the logs:

```
+ docker compose exec -T scheduler /usr/local/share/fossology/scheduler/agent/fo_cli -S
ERROR: /fossology/src/scheduler/agent/fo_cli.c.85: unable to connect to localhost port: 24693
```

This run started from a fresh Postgres volume (`Volume "fossology_database" Creating` in the build log), so `db` had to run a full `initdb` before `scheduler` could finish its DB handshake and start listening on its control port. `test-cluster.sh` only waits a flat `sleep 15` for all three containers regardless of how long that takes, so on a slower run the `fo_cli -S` check hits the scheduler before it's ready.

This is a pre-existing timing race in `src/testing/docker/test-cluster.sh` (fixed sleep instead of a readiness check), not something introduced by this PR. It has nothing to do with `FolderController.php`: the `web` container that actually ships this change passed both of its checks (homepage curl, `copyright -h`) earlier in the same run, before the unrelated `scheduler` container's check failed. `docker-test` also passed cleanly on the last several unrelated PRs, so this looks like a one-off flake tied to the cold DB volume on this run rather than a systemic break.

Happy to open a separate issue for the `sleep 15` race in `test-cluster.sh` if that's useful, but didn't want to bundle an unrelated CI fix into this PR.

shaheemazmalmmd (approved):
Code looks good.

PR #3769, opened 2026-08-18
chore(docker): validate compose config and wait for services

helmut-leher-magna:
## Description

Avoid redundant image builds and add small improvement in Docker test startup validation.

### Changes

 - Replace _docker compose build_ in the workflow with docker _compose config -q_.
   - Image building is done in test-cluster.sh (full/all) and test-standalone.sh (web)
   - Although the caches are used and the build is fast, the extra build before the sh scripts is not needed, no?
- Replace the hardcoded sleep 15 in the cluster testfile with _docker compose up -d --wait --wait-timeout 60_
   - Uses so the configured health checks in docker-compose.yml.

## How to test

 - Run docker compose config -q.
 - Run test-cluster.sh.
 - Verify that the Docker test workflow completes successfully.

### DIFF
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v5
+      uses: actions/checkout@v7
 
     - name: Set up QEMU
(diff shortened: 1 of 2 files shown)


Kaushl2208 (approved):
The flow looks much cleaner. Thankyou for your contribution.

PR #3770, opened 2026-08-18
fix(ui): accomodate access token data model change in UI

Kaushl2208:
## Description

Due to Recent changes in the Data Model for `Personal Access Token` table, The UI also started showing `created_on` & `expiry` values in raw time stamp. To get back to older view. 

### Changes

Add, `date()` filter on both, created and expiry values.

After change from #3712 :
image

Earlier view/Now:

image


shaheemazmalmmd (approved):
Code looks good

PR #3775, opened 2026-08-20
fix(api): Allow spaces in group name path parameters

tiyasakundu:
## Description

Update the API route parameter patterns to support spaces in group names without changing the behavior of other path parameters.

### Changes

- Added a dedicated $groupPattern for group name route parameters.
- Allowed spaces in group names using the new pattern.
- Kept the existing $pattern unchanged for users and other path parameters.
- Updated group-related routes to use $groupPattern for group names and $pattern for user names.

### Testing

- Verified that group names containing spaces can be used in group-related API routes.
- Verified that user-related path parameters continue using the original pattern.
- Verified that existing routes using $pattern remain unchanged.

### Screenshots

Before:

image

After:

image

### DIFF
--- a/src/www/ui/api/index.php
+++ b/src/www/ui/api/index.php
@@ -168,6 +168,9 @@ if ($dbConnected) {
 // Regex for matching a valid path parameter
 $pattern = "[\\w\\d\\-\\.@_]+";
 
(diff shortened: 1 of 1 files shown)


shaheemazmalmmd (approved):
Code looks good.

PR #3779, opened 2026-08-23
fix(scheduler): add GMutex around database_exec to fix PQexec race

mandar1045:
# fix(scheduler): add GMutex around database_exec to fix PQexec race

Found this while stress testing the scheduler with a bunch of concurrent jobs running. Kept seeing weird connection failures and reconnect attempts that didn't make sense given the DB was clearly up.

Turned out to be a threading issue. The scheduler has one shared PGconn (scheduler->db_conn), but it's getting hit from multiple threads at once:

- main scheduler loop (database_update_event)
- interface.c worker threads handling socket commands
- agent.c threads managing the scan agents

libpq docs are pretty clear that a single PGconn isn't safe to use across threads. So when two threads fire PQexec at roughly the same time, the buffer gets corrupted, the scheduler thinks the connection died, tries to reconnect, and in the process drops job updates. Annoying to debug because it's not consistent - only shows up under load.

Added a GMutex around database_exec and around the connect/teardown paths. Nothing fancy, just serializes access so only one thread touches the connection at a time.

## Changes

- database.c - added static db_mutex + init flag, init it in database_init(), lock/unlock around the PQexec calls in database_exec(), added database_destroy() to clean the mutex up on shutdown
- database.h - prototype for database_destroy()
- scheduler.c - calling database_destroy() from scheduler_destroy() and scheduler_clear_config()

## Testing

Rebuilt and brought the containers up (db, scheduler, web).

Threw 50 concurrent fo_cli -S calls at it while an upload was processing across all 7 agents (ununpack, adj2nest, mimetype, scancode, nomos, monk, ojo, reuser). Before the fix this would occasionally throw a connection error mid-run. After the fix - ran it probably 10+ times, no disconnects, no dropped updates.

Also just checked fo_cli -S normally to make sure the daemon still reports PID/status/revision fine.

## Screenshots & Verification

fix1
fix2
fix3
fix4
fix5

## Checklist
- [x] follows project style
- [x] commit signed off
- [x] tested locally with docker
- [x] no breaking changes

### DIFF
--- a/src/scheduler/agent/database.c
+++ b/src/scheduler/agent/database.c
@@ -763,6 +763,8 @@ static PGconn* database_connect(gchar* configdir)
   return ret;
 }
 
+static GMutex db_mutex;
+
 /**
  * Initializes any one-time attributes relating to the database. Currently this
  * includes creating the db connection and checking the URL of the FOSSology
@@ -785,6 +787,14 @@ void database_init(scheduler_t* scheduler)
   check_tables(scheduler);
 }
 
+/**
+ * Cleans up database static resources such as the mutex lock.
+ */
+void database_destroy(void)
+{
+  g_mutex_clear(&db_mutex);
+}
+
 /* ************************************************************************** */
 /* **** event and functions ************************************************* */
 /* ************************************************************************** */
@@ -797,6 +807,10 @@ void database_init(scheduler_t* scheduler)
  * ever lost we automatically try to reconnect and if we are unable to, the
  * scheduler will die.
  *
+ * Thread safety: Access to database_exec is synchronized with db_mutex so
+ * multiple worker or agent threads cannot execute PQexec on scheduler->db_conn
+ * concurrently.
+ *
  * @param scheduler  the scheduler_t* that holds the connection
  * @param sql        the sql that will be performed
  * @return           the PGresult struct that is returned by PQexec
@@ -807,6 +821,8 @@ PGresult* database_exec(scheduler_t* scheduler, const char* sql)
 
   V_SPECIAL("DATABASE: exec \"%s\"\n", sql);
 
+  g_mutex_lock(&db_mutex);
+
   ret = PQexec(scheduler->db_conn, sql);
   if(ret == NULL || PQstatus(scheduler->db_conn) != CONNECTION_OK)
   {
@@ -815,6 +831,8 @@ PGresult* database_exec(scheduler_t* scheduler, const char* sql)
 
     ret = PQexec(scheduler->db_conn, sql);
   }
+
+  g_mutex_unlock(&db_mutex);
 
   return ret;
 }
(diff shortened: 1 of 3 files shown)


mandar1045:
Quick note for reviewers: The fix uses standard GLib `GMutex` primitives aligned with the scheduler's GLib thread architecture. The lock is scoped strictly to `PQexec` query execution inside `database_exec()`, ensuring zero overhead on background agent IPC. Tested locally on Docker with 50+ concurrent status requests during active multi-agent scans.

src/scheduler/agent/scheduler.c
```diff
@@ -855,6 +856,7 @@ void scheduler_clear_config(scheduler_t* scheduler)
   g_free(scheduler->email_subject);
   g_free(scheduler->email_command);
   PQfinish(scheduler->db_conn);
+  database_destroy();
```
shaheemazmalmmd: I have observed that there are sudden drop/kill of scheduler while jobs are running on a heavy load. perhaps database_exec() is destroyed on reload ??

src/scheduler/agent/database.c
```diff
@@ -772,6 +775,12 @@ void database_init(scheduler_t* scheduler)
 {
   PGresult* db_result;
 
+  if(!db_mutex_initialized)
+  {
+    g_mutex_init(&db_mutex);
+    db_mutex_initialized = TRUE;
```
shaheemazmalmmd: db_mutex_initialized is called concurrently Can this be a new Race ? because one thread will get true and another will be false ?

shaheemazmalmmd (changes requested):

src/scheduler/agent/scheduler.c
mandar1045: You're right. Checking db_mutex_initialized inside database_init() at runtime could introduce a race condition if initialization happens while other threads are already running.

I'll switch to a static GMutex (static GMutex db_mutex;) instead. GLib will zero-initialize it automatically, so we can remove the db_mutex_initialized guard entirely.

src/scheduler/agent/database.c
mandar1045: that explains the drop during reloads under load. scheduler_clear_config() gets triggered by a SIGHUP / reload while worker threads and scan jobs are still running. Destroying the mutex at that point meant database_exec() could end up using it mid-reload and crash.

I'll remove database_destroy() from scheduler_clear_config() so the mutex stays alive across config reloads, and only call it from scheduler_destroy() when the process is actually shutting down.

mandar1045:
Pushed commit 9740f22 with the requested updates:

1. **Static GMutex Initialization**: Removed the `db_mutex_initialized` runtime check in `database.c` in favor of GLib's static `GMutex` allocation, avoiding any potential runtime initialization races.
2. **Retain Mutex Across Config Reloads**: Removed `database_destroy()` from `scheduler_clear_config()` in `scheduler.c`. The mutex now remains active during SIGHUP / reload events while worker threads execute `database_exec()`, and is only destroyed on process exit in `scheduler_destroy()`.

Thanks @shaheemazmalmmd for the great review!

PR #3780, opened 2026-08-25
fix(m2m): configure a new oidc app for m2m

deo002:
## Description

Configure a new app for m2m communication because we cannot reuse the one already being used for login. The one being used for login usually has more restricted permissions into who can get a token.

### Changes

Added configuration for adding a new app and used it for m2m with licensedb. 

## How to test

Setup the entire flow and check whether licenses and obligations are updated or not.


shaheemazmalmmd (approved):
Code looks good.

PR #3784, opened 2026-08-26
fix(api): enforce folder access check in unlinkFolder endpoint

### LINKED ISSUE #3783: REST API: unlinkFolder endpoint lacks authorization check, allowing any user to unlink folder content across groups
### Description

While going through `FolderController.php` I noticed that `unlinkFolder()` is missing the access check that every other write method in that file has. In practice this means any logged in API user, no matter which group they belong to, can unlink folder content that isn't theirs, just by knowing the contentId.

### Location

`src/www/ui/api/Controllers/FolderController.php`
Method: `unlinkFolder()`
Line: 306

### Vulnerable Code

```php
// src/www/ui/api/Controllers/FolderController.php:306-319
public function unlinkFolder($request, $response, $args)
{
  $folderContentId = intval($args['id']);
  if (! $this->folderDao->doesIdExist($folderContentId, "foldercontents", "foldercontents_pk")) {
    ...
  }
  $this->folderDao->removeContent($folderContentId);
  ...
}
```

Compare this to the other methods in the same controller, `editFolder` (193 to 194), `copyFolder` (236 to 242) and `getUnlinkableFolderContents` (283 to 284), each of which calls `isFolderAccessible($folderId, $userId)` before doing anything. `unlinkFolder` skips that call entirely.

I traced it further into `FolderDao::removeContent()` and `isRemovableContent()` in `src/lib/php/Dao/FolderDao.php` (lines 460 to 476), and neither of those does any permission checking either, they only check whether the content link is a duplicate. The auth middleware, `RestAuthMiddleware.php`, only validates the JWT and its read or write scope, not per resource role, so nothing downstream catches this either.

Interestingly, the legacy (non API) UI already handles this correctly, `AdminContentDelete.php`, the plugin behind the same unlink action in the web UI, requires `Auth::PERM_ADMIN` before running. So the REST API is actually less strict than the older UI it's supposed to be replacing.

### Steps to Reproduce

As UserA in one group, create a folder, upload a file, then copy it into a second folder so the content link becomes removable (a foldercontents row with more than one parent).

As UserB, a completely different non admin user with no access to UserA's folder, call `PUT /repo/api/v2/folders/contents/{contentId}/unlink` using UserB's own valid token.

The request succeeds and the link gets removed, even though UserB has no relationship to UserA's folder or group at all.

### Root Cause

`unlinkFolder()` was added in commit bfff708c (2023-07-26) without the accessibility check that every sibling method in the controller enforces, and the DAO method underneath it never enforced its own check either, so there's nothing catching this at any layer.

### Impact

Severity: Medium to High
Any authenticated, non admin user can remove folder content links belonging to other groups just by iterating the contentId, which is a small sequential integer. This is a broken access control issue that lets users tamper with data organization outside their own group.

### Suggested Fix

Resolve the parent folder for the given contentId and call `isFolderAccessible($parentFolderId, $userId)` before calling `removeContent()`, the same pattern already used in `editFolder` and `copyFolder`. If it fails, throw a 403 instead of proceeding.

bhuvan-somisetty:
## Description

`unlinkFolder()` in `FolderController` was missing the folder accessibility check that every other write method in the same controller performs, letting any authenticated API user unlink folder content belonging to folders or groups they have no access to.

### Changes

* `FolderController::unlinkFolder()` now resolves the parent folder of the given `contentId` via `FolderDao::getContent()` and calls `isFolderAccessible()` before removing the link, throwing `HttpForbiddenException` (403) when the user can't access it. This is the same pattern already used by `editFolder`, `copyFolder` and `getUnlinkableFolderContents`.
* Switched from `$this->container->get('dao.folder')` to `$this->restHelper->getFolderDao()`, which is what every other method in this controller already uses to get the same `FolderDao` service, so the fix stays consistent with the rest of the file.
* Added three regression tests to `FolderControllerTest`: unlinking accessible content still succeeds, unlinking content whose parent folder is inaccessible now throws 403 (and `removeContent()` is never called), and the existing "not removable" 400 path still works with the new check in place.

## How to test

1. As UserA, create a folder, upload a file, and copy it into a second folder so the content link becomes removable (a `foldercontents` row with more than one parent).
2. As UserB, a different user with no access to UserA's folder, call `PUT /repo/api/v2/folders/contents/{contentId}/unlink` using UserB's own token.
3. Before this fix: the link is removed regardless of UserB's access. After this fix: the request returns 403 and the link is left untouched.

Ran the full `FolderControllerTest` suite (PHPUnit 10, PHP 8.1): 46/46 passing. Ran `phpcs --standard=fossy-ruleset.xml` on both changed files: no violations introduced by this change (pre-existing issues elsewhere in the file are untouched by this diff).

Fixes #3783

### DIFF
--- a/src/www/ui/api/Controllers/FolderController.php
+++ b/src/www/ui/api/Controllers/FolderController.php
@@ -310,7 +310,11 @@ class FolderController extends RestController
       throw new HttpNotFoundException("Folder content id not found!");
     }
     /** @var FolderDao $folderDao */
-    $folderDao = $this->container->get('dao.folder');
+    $folderDao = $this->restHelper->getFolderDao();
+    $content = $folderDao->getContent($folderContentId);
+    if (! $folderDao->isFolderAccessible($content['parent_fk'], $this->restHelper->getUserId())) {
+      throw new HttpForbiddenException("Folder is not accessible!");
+    }
     if (!$folderDao->removeContent($folderContentId)) {
       throw new HttpBadRequestException("Content cannot be unlinked.");
     }
--- a/src/www/ui_tests/api/Controllers/FolderControllerTest.php
+++ b/src/www/ui_tests/api/Controllers/FolderControllerTest.php
@@ -1168,6 +1168,100 @@ namespace Fossology\UI\Api\Test\Controllers
 
     /**
      * @test
+     * -# Test for FolderController::unlinkFolder() on accessible content
+     * -# Check if the statusCode is 200 and the content gets unlinked
+     */
+    public function testUnlinkFolder()
+    {
+      $contentId = 4;
+      $parentFolderId = 2;
+      $this->dbHelper->shouldReceive("doesIdExist")
+        ->withArgs(array("foldercontents", "foldercontents_pk", $contentId))
+        ->andReturn(true);
+      $this->folderDao->shouldReceive('getContent')
+        ->withArgs(array($contentId))
+        ->andReturn(["foldercontents_pk" => $contentId, "parent_fk" => $parentFolderId]);
+      $this->folderDao->shouldReceive('isFolderAccessible')
+        ->withArgs(array($parentFolderId, $this->userId))->andReturn(true);
+      $this->folderDao->shouldReceive('removeContent')
+        ->withArgs(array($contentId))->andReturn(true);
+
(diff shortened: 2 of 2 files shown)


shaheemazmalmmd (approved):
Code looks good.

PR #3785, opened 2026-08-27
fix(ununpack): bound pdftotext with timeout and memory limit

its-sushant:
## Description

The ununpack agent doesn't parse PDFs itself, it shells out to `pdftotext` purely to extract text for later license scanning. Certain malformed/adversarial PDFs (xref self-loops, exponential XObject loops, bad stream lengths, etc.) make pdftotext either:
- hang indefinitely - with no timeout around the system() call, the process spins forever, pinning a CPU core, and
- balloon in memory - leading to host OOM.

### Changes

- Timeout - wrap the call with timeout -k 200s, overridable via `FOSSOLOGY_CMD_TIMEOUT`
- Memory ceiling - prefix with ulimit -v, computed as a percentage of total RAM (default 50%, via sysinfo()), overridable via `FOSSOLOGY_CMD_MEM_LIMIT_PERCENT` or an absolute `FOSSOLOGY_CMD_MEM_LIMIT_KB`
- Fail soft - on kill/timeout, log and skip the file instead of aborting.
- No partial artifacts - on kill/timeout, unlink() the truncated `pdftotext`

## How to test

- Build this branch
- And try to upload https://github.com/itext/itext-dotnet source. It contains lot of malformed pdfs.
- The `ununpack` agent should not hang the system. It should skip those malformed pdfs and continue extracting the other files.

### DIFF
--- a/src/adj2nest/agent/adj2nest.c
+++ b/src/adj2nest/agent/adj2nest.c
@@ -115,34 +115,38 @@ void	WalkTree	(long Index, long Depth)
 {
   long LeftSet;
   PGresult* pgResult;
+  int i;
 
-  if (agent_verbose)
+  /* Traverse sibling chain iteratively to avoid stack overflow on directories
+     with a large number of files
+  */
+  for (;;)
     {
-    int i;
-    for(i=0; i<Depth; i++) printf(" ");
-    LOG_VERBOSE("%ld\n",Tree[Index].UploadtreePk);
-    }
+    if (agent_verbose)
+      {
+      for(i=0; i<Depth; i++) printf(" ");
+      LOG_VERBOSE("%ld\n",Tree[Index].UploadtreePk);
+      }
 
-  LeftSet = SetNum;
-  SetNum++;
-
-  if (Tree[Index].Child > -1)
-    {
-    WalkTree(Tree[Index].Child,Depth+1);
+    LeftSet = SetNum;
     SetNum++;
-    }
 
-  snprintf(SQL,sizeof(SQL),"UPDATE %s SET lft='%ld', rgt='%ld' WHERE uploadtree_pk='%ld'",
+    if (Tree[Index].Child > -1)
+      {
+      WalkTree(Tree[Index].Child,Depth+1);
+      SetNum++;
+      }
+
+    snprintf(SQL,sizeof(SQL),"UPDATE %s SET lft='%ld', rgt='%ld' WHERE uploadtree_pk='%ld'",
 	uploadtree_tablename,LeftSet,SetNum,Tree[Index].UploadtreePk);
-  pgResult = PQexec(pgConn, SQL);
-  fo_checkPQcommand(pgConn, pgResult, SQL, __FILE__, __LINE__);
-  PQclear(pgResult);
-  fo_scheduler_heart(1);
+    pgResult = PQexec(pgConn, SQL);
+    fo_checkPQcommand(pgConn, pgResult, SQL, __FILE__, __LINE__);
+    PQclear(pgResult);
+    fo_scheduler_heart(1);
 
-  if (Tree[Index].Sibling > -1)
-    {
+    if (Tree[Index].Sibling < 0) break;
     SetNum++;
-    WalkTree(Tree[Index].Sibling,Depth+1);
+    Index = Tree[Index].Sibling;
     }
 
 } /* WalkTree() */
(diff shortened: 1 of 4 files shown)


src/ununpack/agent/ununpack.h
```diff
@@ -45,6 +46,8 @@
 #endif
 
 #define Last(x)	(x)[strlen(x)-1]
+#define DEFAULT_PDF_CMD_TIMEOUT 200
+#define DEFAULT_CMD_MEM_LIMIT_PERCENT 50
```
Kaushl2208: I dont think percentage would be a good thing to go about this, In a 8GB system you allocate 4 and in 32GB you allocate 16.
Thats quite substantial, Instead of percentage can we bound it to some conservative arbitary limit?

Kaushl2208 (changes requested):
The change is needed. Its good to have a deep defensive check. 

Can we also think about a PDF validator?? Which can basically structurally validate the PDF? Like if they are encrypted, OR some other issues that you mentioned in the Description. Before we even send it to `pdftotext` we can have a filter right?

I was not able to find a pdf with issues to `pdftotext` I tried with a few in my CLI. 
Can you also refer a few such PDFs where ununpack needed help?

src/ununpack/agent/ununpack.h
its-sushant: Fair point. I have put the limit to `2GB` now. But if you have any other number in mind I can make the change. Also if user wants to increase or decrease the cap, it can be overwritten by an env variable `FOSSOLOGY_CMD_MEM_LIMIT_KB`

its-sushant:
> The change is needed. Its good to have a deep defensive check.
> 
> Can we also think about a PDF validator?? Which can basically structurally validate the PDF? Like if they are encrypted, OR some other issues that you mentioned in the Description. Before we even send it to `pdftotext` we can have a filter right?
> 
> I was not able to find a pdf with issues to `pdftotext` I tried with a few in my CLI. Can you also refer a few such PDFs where ununpack needed help?

I have added a check using `pdfinfo` and it's quite userful to filter out `corrupted/encrypted/password-protected` pdfs beforehand so that we avoid calling `pdftotext` on those pdf files. 
image

And the corrupted pdfs are present inside the test folder for the repo: https://github.com/itext/itext-dotnet. But for the testing purpose I have collected few pdfs which can be useful.
[corrupted_pdf.zip](https://github.com/user-attachments/files/31738794/corrupted_pdf.zip)

And this is the output of `pdfinfo` command on all the pdfs in `corrupted_pdf.zip`
[pdfinfo_results.txt](https://github.com/user-attachments/files/31738994/pdfinfo_results.txt)

src/adj2nest/agent/adj2nest.c
```diff
@@ -115,34 +115,38 @@ void	WalkTree	(long Index, long Depth)
...
+    snprintf(SQL,sizeof(SQL),"UPDATE %s SET lft='%ld', rgt='%ld' WHERE uploadtree_pk='%ld'",
 	uploadtree_tablename,LeftSet,SetNum,Tree[Index].UploadtreePk);
-  pgResult = PQexec(pgConn, SQL);
-  fo_checkPQcommand(pgConn, pgResult, SQL, __FILE__, __LINE__);
-  PQclear(pgResult);
-  fo_scheduler_heart(1);
+    pgResult = PQexec(pgConn, SQL);
+    fo_checkPQcommand(pgConn, pgResult, SQL, __FILE__, __LINE__);
+    PQclear(pgResult);
+    fo_scheduler_heart(1);
```
Kaushl2208, line 145: have we tried this on larger components? 
This path can run long on big uploads, and heartbeat regression can cause scheduler timeout. Just to make double sure, the hearbeat works.

src/adj2nest/agent/adj2nest.c
```diff
@@ -115,34 +115,38 @@ void	WalkTree	(long Index, long Depth)
...
-    }
 
-  snprintf(SQL,sizeof(SQL),"UPDATE %s SET lft='%ld', rgt='%ld' WHERE uploadtree_pk='%ld'",
+    if (Tree[Index].Child > -1)
+      {
+      WalkTree(Tree[Index].Child,Depth+1);
+      SetNum++;
+      }
+
+    snprintf(SQL,sizeof(SQL),"UPDATE %s SET lft='%ld', rgt='%ld' WHERE uploadtree_pk='%ld'",
```
Kaushl2208, line 140: Due to change in `lft, rgt` by shifting from recursive to iterative. Can we write a regression test or confim as a validation on an upload (old vs new) values of: `(uploadtree_pk, lft, rgt)` to make sure things are correct.

PR #3787, opened 2026-08-27
fix(search): escape filename parameter to prevent SQL injection

### LINKED ISSUE #3786: SQL Injection in Search via unescaped `filename` parameter
## Summary

When you search for a file in FOSSology, the `filename` value you type gets pasted **directly** into a SQL query with no escaping:

```php
// src/lib/php/Dao/SearchHelperDao.php  line 158
$SQLWhere .= " ufile_name ilike '" . $Filename . "'";
```

`$Filename` comes straight from `$_GET['filename']` via `PARM_RAW`, which explicitly returns the raw, untouched value.

## Proof what SQL gets built

Input: `foo' OR '1'='1`

```sql
SELECT * FROM uploadtree WHERE ufile_name ilike 'foo' OR '1'='1'
-- returns every file in the database regardless of access filters
```

UNION payload to read the users table:

```sql
SELECT ... FROM uploadtree WHERE ufile_name ilike 'x'
UNION SELECT usrname,user_pass,NULL,... FROM users--
```

## Fix

```php
// Before
$SQLWhere .= " ufile_name ilike '" . $Filename . "'";

// After
$SQLWhere .= " ufile_name ilike '" . pg_escape_string($Filename) . "'";
```

One-line change, consistent with how the same file handles the other parameters.

Aditya30ag:
## Description

Fix SQL injection in file search by escaping the `filename` parameter before interpolating it into the SQL query.

### Changes

* Escape `$Filename` using `pg_escape_string()` in `SearchHelperDao::GetResults()`.
* Keep the existing search behavior unchanged for normal filename patterns.
* Make filename handling consistent with the existing escaping of `$License` and `$Copyright`.
* Prevent SQL injection through the web UI and REST API search endpoints.

### How to test

1. Apply the change in `src/lib/php/Dao/SearchHelperDao.php`.
2. Search for a normal filename such as `%.php` and verify that the expected results are returned.
3. Search using a filename containing a single quote, such as `foo'bar`, and verify that the search completes without a SQL error.
4. Test the previously vulnerable input `foo' OR '1'='1`.
5. Verify that the input is treated as a literal filename search value rather than modifying the SQL query.
6. Run the relevant FOSSology search/API tests and confirm they pass.

Fixes #3786

### DIFF
--- a/src/lib/php/Dao/SearchHelperDao.php
+++ b/src/lib/php/Dao/SearchHelperDao.php
@@ -155,7 +155,7 @@ class SearchHelperDao
       if ($NeedAnd) {
         $SQLWhere .= " AND";
       }
-      $SQLWhere .= " ufile_name ilike '" . $Filename . "'";
+      $SQLWhere .= " ufile_name ilike '" . pg_escape_string($Filename) . "'";
       $NeedAnd = 1;
     }
 


its-sushant (approved):
Good Catch. LGTM!

PR #3791, opened 2026-08-29
fix(api): check the target group in getGroupMembers, not just any group

### LINKED ISSUE #3790: REST API: getGroupMembers checks the wrong group, letting any user read any group's member list
### Description

I was comparing the group management endpoints in GroupController against each other and noticed that getGroupMembers checks whether the caller administers some group somewhere, but never checks whether they administer the specific group whose members are being requested. Every other write endpoint in the same file gets this right, so it stood out as the odd one out.

### Location

`src/www/ui/api/Controllers/GroupController.php`
Function: `getGroupMembers()`, lines 278 to 312

### Vulnerable Code

```php
// src/www/ui/api/Controllers/GroupController.php:278-291
public function getGroupMembers($request, $response, $args)
{
  $apiVersion = ApiVersion::getVersion($request);
  $userId = $this->restHelper->getUserId();
  $userDao = $this->restHelper->getUserDao();
  $groupMap = $userDao->getAdminGroupMap($userId, $_SESSION[Auth::USER_LEVEL]);

  if (empty($groupMap)) {
    throw new HttpForbiddenException("You have no permission to manage any group.");
  }

  // Get the group name/id form the params and then the group Id
  $groupId = $apiVersion == ApiVersion::V2 ? intval($this->restHelper->getUserDao()->getGroupIdByName($args['pathParam'])) : intval($args['pathParam']);
  ...
```

`$groupMap` is the set of groups the caller actually administers, `getAdminGroupMap()` in UserDao.php confirms it only returns groups where the caller has group_perm ADMIN, unless the caller is a global superadmin. The check right after it only asks whether that set is non-empty, it never asks whether `$groupId`, which comes straight from the URL path and can be any group in the system, is one of the groups in that set. The function then goes ahead and queries `group_user_member` for that arbitrary group and returns the full list of usernames and permission levels.

Compare this with the three sibling endpoints in the exact same file. `deleteGroupMember` (line 217), `changeUserPermission` (line 413) and `addMember` (line 346) all call `isAdvisorOrAdmin($userId, $group_pk)` against the specific target group_pk taken from the path, not just a general "do you admin something" check. getGroupMembers is the only one of the four that skips this.

### Root Cause

getGroupMembers authorizes the caller against the wrong resource. It confirms the caller is an admin of at least one group, then uses that as a stand in for "the caller is allowed to see this particular group's membership," when the two are unrelated. The group id being queried is never checked against the set of groups the caller was actually shown to administer.

### Impact

Severity: Medium

Every account qualifies for the first check by default, account creation always makes a new user the admin of their own personal group (group_perm ADMIN in group_user_member), so there's no special setup needed. Any authenticated user can call this endpoint with any other group's name and get back that group's full member list together with everyone's permission level inside it, including private or unrelated groups the caller has no membership in and no other visibility into. It doesn't let them change anything, but it is a real information disclosure and a horizontal privilege escalation on the read side, since the access decision is based on the wrong identifier.

### Steps to Reproduce

1. As UserA, use any ordinary account, every account is admin of its own personal group by default, so no special setup is needed.
2. As UserB, create a second group that UserA has no membership or admin rights in.
3. As UserA, call `GET /repo/api/v2/groups/{UserB's group name}/members` using UserA's own token.
4. The request succeeds and returns the full list of UserB's group members with their permission levels, even though UserA has no relationship to that group at all.

### Suggested Fix

After resolving `$groupId`, check that it is actually a key in `$groupMap` (or call `isAdvisorOrAdmin($userId, $groupId)` the same way the other three endpoints in this file do) and throw HttpForbiddenException if not, before running the membership query. Keeping it consistent with the pattern already used in deleteGroupMember, changeUserPermission and addMember should make this a small, easy to review change.

bhuvan-somisetty:
## Description

`GroupController::getGroupMembers()` was authorizing against the wrong resource. It checked whether the caller administers some group in the system, but never checked whether the caller administers the specific group whose members were being requested through the URL path.

### Changes

* `getGroupMembers()` now resolves the target group first, confirms it exists (`HttpNotFoundException` if not, matching the other endpoints in this file), and then requires the caller to be a global admin or `isAdvisorOrAdmin()` of that specific group before returning its member list, throwing `HttpForbiddenException` otherwise. This is the exact same pattern already used by `deleteGroupMember`, `changeUserPermission` and `addMember` in the same controller.
* Removed the old `getAdminGroupMap()` based check, since it was only checking "does the caller administer anything" and every account administers its own personal group by default (`group_perm=ADMIN` is set at account creation), so it never actually blocked anyone.
* Updated the existing `testGetGroupMembers` test to mock the new checks instead of the removed one, and added three regression tests: a caller who administers a different group gets a 403 for the target group, a nonexistent group returns 404, and a global admin without a specific role on the target group can still see its members.

## How to test

1. As UserA, use any ordinary account, every account administers its own personal group by default.
2. As UserB, create a second group UserA has no membership or admin rights in.
3. As UserA, call `GET /repo/api/v2/groups/{UserB's group name}/members`.
4. Before this fix: the request succeeds and returns UserB's group membership and permission levels. After this fix: the request returns 403.

Ran the `GroupControllerTest` suite locally (PHPUnit 10, PHP 8.1): 34/34 passing. Ran `phpcs --standard=fossy-ruleset.xml` on both changed files: no violations introduced by this change.

Fixes #3790

### DIFF
--- a/src/www/ui/api/Controllers/GroupController.php
+++ b/src/www/ui/api/Controllers/GroupController.php
@@ -280,14 +280,20 @@ class GroupController extends RestController
     $apiVersion = ApiVersion::getVersion($request);
     $userId = $this->restHelper->getUserId();
     $userDao = $this->restHelper->getUserDao();
-    $groupMap = $userDao->getAdminGroupMap($userId, $_SESSION[Auth::USER_LEVEL]);
-
-    if (empty($groupMap)) {
-      throw new HttpForbiddenException("You have no permission to manage any group.");
-    }
 
     // Get the group name/id form the params and then the group Id
-    $groupId = $apiVersion == ApiVersion::V2 ? intval($this->restHelper->getUserDao()->getGroupIdByName($args['pathParam'])) : intval($args['pathParam']);
+    $groupId = $apiVersion == ApiVersion::V2 ? intval($userDao->getGroupIdByName($args['pathParam'])) : intval($args['pathParam']);
+
+    $userIsAdmin = Auth::isAdmin();
+    $userHasGroupAccess = $userDao->isAdvisorOrAdmin($userId, $groupId);
+
+    if (!$this->dbHelper->doesIdExist("groups", "group_pk", $groupId)) {
+      throw new HttpNotFoundException("Group id not found!");
+    }
+    if (! $userIsAdmin && ! $userHasGroupAccess) {
+      throw new HttpForbiddenException("Not advisor or admin of the group. " .
+        "Can not process request.");
+    }
 
     // The query to get the list of users with corresponding roles from the group.
     $dbManager = $this->dbHelper->getDbManager();
--- a/src/www/ui_tests/api/Controllers/GroupControllerTest.php
+++ b/src/www/ui_tests/api/Controllers/GroupControllerTest.php
@@ -553,7 +553,6 @@ class GroupControllerTest extends \PHPUnit\Framework\TestCase
   private function testGetGroupMembers($version = ApiVersion::V2)
   {
     $userIds = [2];
-    $groupName = 'fossy';
     $groupId = 1;
     $newuser = 3;
     $userPk = 2;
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
@shaheemazmalmmd Sir PTAL Whenever u get chance

shaheemazmalmmd (approved):
Code looks good

PR #3793, opened 2026-08-30
fix(uploadPermission): check the target upload id, not just list emptiness

### LINKED ISSUE #3792: Broken membership check in upload permissions lets any user grant admin access to arbitrary uploads
### Description

I was reading through how the REST API lets a user change who can access an upload, and the membership check inside insertPermission in UploadPermissionPage.php isn't actually checking membership. It loops over the list of uploads the caller is allowed to write to and asks whether each entry has a truthy upload_pk, which is true for basically every row, instead of comparing that entry against the upload actually being modified. The practical effect is that a user only needs write access to any upload anywhere in some folder they control to grant themselves admin permission on a completely unrelated upload they were only ever allowed to read.

### Location

`src/www/ui/page/UploadPermissionPage.php`
Function: `insertPermission()`, lines 175 to 199, the broken check is on line 179

`src/www/ui/api/Controllers/UploadController.php`
Function: `setUploadPermissions()`, lines 834 to 872

### Vulnerable Code

```php
// src/www/ui/page/UploadPermissionPage.php:175-185
function insertPermission($groupId,$uploadId,$permission,$uploadList)
{
  $fileName = false;
  foreach ($uploadList as $uploadEntry) {
    if ($uploadEntry['upload_pk']) {
      $fileName = $uploadEntry['name'];
    }
  }
  if (empty($fileName)) {
    throw new \Exception('This upload is missing or inaccessible');
  }
  ...
  $this->uploadPermDao->insertPermission($uploadId, $groupId, $permission);
}
```

The intent of this loop is obviously to confirm that `$uploadId`, the upload whose permission is about to change, is one of the uploads the caller is actually allowed to touch. What it does instead is walk the list and keep overwriting `$fileName` with whatever the last entry's name happens to be, as long as that entry has any upload_pk at all. It never compares `$uploadEntry['upload_pk']` to `$uploadId`. So as long as `$uploadList` isn't empty, the function proceeds and calls `uploadPermDao->insertPermission($uploadId, ...)` using the real target id from the caller, regardless of whether that id ever appeared in the list.

`$uploadList` itself comes from `FolderListUploads_perm($folder_pk, Auth::PERM_WRITE)` in common-folders.php, which filters to uploads inside `$folder_pk` that the caller's group can already edit (`src/lib/php/common-folders.php:382-384`, calling `UploadDao::isEditable()`). `$folder_pk` is a value the REST caller supplies directly in the request body as `folderId`, with no relationship enforced to the upload id in the URL.

On the REST side, `setUploadPermissions()` only gates entry with `uploadAccessible($upload_pk)` (`src/www/ui/api/Controllers/UploadController.php:853`, defined in `RestController.php:99-108`), which calls `UploadDao::isAccessible()` (`src/lib/php/Dao/UploadDao.php:655-658`), which just delegates to `UploadPermissionDao::isAccessible()`. That check passes for read access or even a merely public upload, not write or admin access to that specific upload:

```php
// src/lib/php/Dao/UploadPermissionDao.php:27-42
public function isAccessible($uploadId, $groupId)
{
  $perm = $this->dbManager->getSingleRow('SELECT perm FROM perm_upload WHERE upload_fk=$1 AND group_fk=$2',
      array($uploadId, $groupId), __METHOD__.'.group_perm');
  if ($perm && $perm['perm'] > Auth::PERM_NONE) {
    return true;
  }
  ...
  $uploadPub = $this->dbManager->getSingleRow('SELECT public_perm FROM upload WHERE upload_pk=$1 AND public_perm>$2',
          array($uploadId,Auth::PERM_NONE),__METHOD__.'.public_perm');
  return !empty($uploadPub);
}
```

### Root Cause

`insertPermission()` never checks that the target upload id is actually a member of the accessible upload list it was handed, it only checks that the list is non-empty. Combined with the fact that the list is built from a caller-supplied `folder_pk` that has no required relationship to the caller-supplied `upload_pk`, and that the REST endpoint's own upfront check only requires read level access to the target upload, there's no point in the whole chain where the caller is actually shown to have write or admin rights over the upload whose permission they're changing.

### Impact

Severity: High

Any user who can read an upload, including one that's merely marked public, and who has write access to at least one upload anywhere in any folder of their own, which includes their own personal uploads by default, can call the permissions endpoint for the unrelated upload and grant their own group admin permission on it. From there they can read, edit, or redistribute clearing data for an upload that was never meant to be theirs to control, and the same broken check also lets them set that upload's public permission. This is a privilege escalation from read only or public access up to admin control over arbitrary uploads.

### Steps to Reproduce

1. As UserA, upload any file into a folder UserA controls, so UserA's group has write access to at least one upload somewhere.
2. As UserB, upload a separate file and either leave the default group permission such that UserA's group only has read access to it, or mark it public.
3. As UserA, call `PUT /repo/api/v2/uploads/{UserB's upload id}/permissions` with a body setting `folderId` to UserA's own folder (the one from step 1), `groupId` to UserA's group id, and `newPermission` to `"admin"`.
4. The call succeeds. UserA's group now has admin permission on UserB's upload, even though UserA never had write or admin access to it.

### Suggested Fix

Make the loop in `insertPermission()` actually check membership, for example set `$fileName` only when `$uploadEntry['upload_pk'] == $uploadId`, and keep the existing "missing or inaccessible" exception as the fallback when no match is found. That alone closes the gap regardless of what folder_pk the caller supplies. It would also be worth having `UploadController::setUploadPermissions()` require edit or admin level access to the specific upload being modified rather than the plain `uploadAccessible()` read level check, so the REST layer doesn't rely entirely on the plugin's internal list filtering to enforce this.

bhuvan-somisetty:
## Description

`UploadPermissionPage::insertPermission()` was checking the wrong thing. It's supposed to confirm the upload id being granted a new permission is one of the uploads the caller has write access to, but the loop only checked whether each entry in that accessible list had a truthy upload_pk, never whether it matched the actual target upload id.

### Changes

* Fixed the loop in `insertPermission()` to compare `$uploadEntry['upload_pk']` against `$uploadId` instead of just checking truthiness, and added a `break` once the match is found. Now the exception is raised whenever the target upload id genuinely isn't present in the caller's accessible list, which is what the surrounding code was already trying to do.
* Added `UploadPermissionPageTest`, a new isolated unit test (following the same `newInstanceWithoutConstructor()` + `Reflectory` pattern already used by `AgentAdderTest` for this kind of legacy plugin class) covering three cases: a target id that is not in the accessible list is rejected and never reaches the DAO, a target id that is genuinely in the list is granted correctly, and the pre-existing empty-list guard still works. Registered the new file in `phpunit.xml` the same way `AgentAdderTest.php` is registered, since `www/ui_tests/Unit` isn't picked up as a directory glob.

## How to test

1. As UserA, upload a file into any folder UserA controls, so UserA's group has write access to at least one upload.
2. As UserB, upload a separate file and leave the default group permission at read-only for UserA's group (or mark it public).
3. As UserA, call `PUT /repo/api/v2/uploads/{UserB's upload id}/permissions` with `folderId` set to UserA's own folder from step 1, `groupId` set to UserA's group, and `newPermission` set to `"admin"`.
4. Before this fix: the call succeeds and UserA's group gets admin permission on UserB's upload. After this fix: it throws "This upload is missing or inaccessible" and no permission is granted.

I verified the new test actually catches the regression by temporarily reverting the fix and confirming `testRejectsUploadNotInAccessibleList` fails against the vulnerable code, then re-ran it green after restoring the fix.

Ran the full `www/ui_tests/api` controller suite plus the `Unit` tests (706 tests, PHPUnit 10, PHP 8.1) locally: all passing except two pre-existing, unrelated environment gaps (a missing `yaml` extension in `LicenseCompatibilityRuleControllerTest`, and a `ReportControllerTest` MIME-guessing test that only needed `fileinfo` enabled, which I did and confirmed green). Ran `phpcs --standard=fossy-ruleset.xml` on the changed and new files: no violations introduced.

Fixes #3792

### DIFF
--- a/src/phpunit.xml
+++ b/src/phpunit.xml
@@ -28,6 +28,7 @@
       <directory suffix="Test.php">lib/php</directory>
       <file>www/ui_tests/startUpTest.php</file>
       <file>www/ui_tests/Unit/AgentAdderTest.php</file>
+      <file>www/ui_tests/Unit/UploadPermissionPageTest.php</file>
       <directory suffix="Test.php">www/ui_tests/api</directory>
       <file>cli/tests/test_fo_copyright_list.php</file>
     </testsuite>
--- a/src/www/ui/page/UploadPermissionPage.php
+++ b/src/www/ui/page/UploadPermissionPage.php
@@ -176,8 +176,9 @@ class UploadPermissionPage extends DefaultPlugin
   {
     $fileName = false;
     foreach ($uploadList as $uploadEntry) {
-      if ($uploadEntry['upload_pk']) {
+      if ($uploadEntry['upload_pk'] == $uploadId) {
         $fileName = $uploadEntry['name'];
+        break;
       }
     }
     if (empty($fileName)) {
--- a/src/www/ui_tests/Unit/UploadPermissionPageTest.php
+++ b/src/www/ui_tests/Unit/UploadPermissionPageTest.php
@@ -1,0 +1,138 @@
+<?php
+/*
+ SPDX-FileCopyrightText: © 2026 Siemens AG
+
+ SPDX-License-Identifier: GPL-2.0-only
+*/
+
+use Fossology\Lib\Auth\Auth;
+use Fossology\Lib\Dao\UploadPermissionDao;
+use Fossology\Lib\Test\Reflectory;
+use Mockery as M;
+
+require_once(dirname(dirname(dirname(__DIR__))) . '/lib/php/Test/Reflectory.php');
+require_once(dirname(dirname(dirname(__DIR__))) . '/lib/php/Plugin/FO_Plugin.php');
+require_once(dirname(dirname(dirname(__DIR__))) . '/lib/php/common-plugin.php');
+require_once(dirname(dirname(dirname(__DIR__))) . '/lib/php/common-menu.php');
+
+/**
+ * @class UploadPermissionPageTest
+ * @brief Test for UploadPermissionPage::insertPermission(), covering the
+ *        broken upload membership check fix
+ * @runTestsInSeparateProcesses
+ * @preserveGlobalState disabled
+ *
+ * Regression coverage for: insertPermission() must confirm the target
+ * upload id is actually present in the caller's accessible upload list,
+ * not merely that the list is non-empty. Before the fix, the loop only
+ * checked truthiness of each entry's own upload_pk, so any non-empty
+ * but unrelated list let the caller change permissions on an arbitrary
+ * upload id they were never shown to have access to.
+ */
+class UploadPermissionPageTest extends \PHPUnit\Framework\TestCase
+{
+  /** @var int */
+  private $assertCountBefore;
+  /** @var UploadPermissionPage */
+  private $page;
+
+  protected function setUp(): void
+  {
+    if (!class_exists('UploadPermissionPage', false)) {
+      $GLOBALS['SysConf'] = [];
+      $GLOBALS['container'] = new class {
+        public function get($name)
+        {
+          return new \stdClass();
+        }
+      };
+      require_once(__DIR__ . '/../../ui/page/UploadPermissionPage.php');
+    }
+
+    $this->assertCountBefore = \Hamcrest\MatcherAssert::getCount();
+
+    global $MenuList, $Plugins;
+    $MenuList = array();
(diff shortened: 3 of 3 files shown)


shaheemazmalmmd (approved):
Code looks good.

PR #3795, opened 2026-08-31
fix(home): escape error_description param to prevent reflected XSS

### LINKED ISSUE #3794: Reflected XSS on home page via `error_description` GET parameter
### What's happening

`HomePage.php` reads `error_description` from the URL and stores it directly into the template message variable with no HTML escaping:

```php
// src/www/ui/page/HomePage.php  line 76-77
if (! empty(GetParm("error", PARM_TEXT))) {
    $vars['message'] = GetParm("error_description", PARM_TEXT);
}
```

`PARM_TEXT` only calls `stripslashes()` no HTML encoding at all.
Twig's `autoescape` is globally set to `false` in `services.xml.in`, so nothing is escaped at render time either. The script reaches the browser and executes.

### Attack URL
http://localhost:8081/repo/?mod=home&error=1&error_description=<script>alert(document.cookie)</script>

### Proof

Image

Tested on a live local FOSSology instance. The alert fired and exposed:

- `token=Bearer eyJ...` full JWT auth token
- `Login=...` active session ID

The attacker needs no account to craft this URL. The **victim** must be
logged in for the stolen cookies to be useful.

### Root cause

Two things combine:

1. `GetParm("error_description", PARM_TEXT)` returns raw input (`stripslashes()` only)
2. `autoescape = false` globally in `services.xml.in` Twig never encodes output unless explicitly told to

### Fix

One line in `HomePage.php`:

```php
// before
$vars['message'] = GetParm("error_description", PARM_TEXT);

// after
$vars['message'] = htmlspecialchars(GetParm("error_description", PARM_TEXT), ENT_QUOTES, 'UTF-8');
```

### Files affected

- `src/www/ui/page/HomePage.php`

Aditya30ag:
### Summary

`HomePage.php` reflected the raw `error_description` GET parameter into the Twig `message` variable without HTML escaping. Because Twig's `autoescape` is globally disabled in `services.xml.in`, the value reached the browser verbatim and executed as JavaScript.

### Impact

An attacker with no account can craft a URL and send it to any logged-in user. When the victim clicks it, the script runs in their browser and exposes their session cookies including the JWT bearer token and session ID to the attacker.

Tested on a live FOSSology instance: `alert(document.cookie)` successfully exposed a valid `Bearer` JWT token and `Login` session ID.

### Root cause

Two things combined to create the vulnerability:

1. `GetParm("error_description", PARM_TEXT)` returns raw input `PARM_TEXT`  only calls `stripslashes()`, no HTML encoding
2. `autoescape = false` globally in `services.xml.in` Twig never escapes output unless explicitly instructed to

### Change

Wrap with `htmlspecialchars()` at the point the value enters `vars['message']`:

```diff
- $vars['message'] = GetParm("error_description", PARM_TEXT);
+ $vars['message'] = htmlspecialchars(GetParm("error_description", PARM_TEXT), ENT_QUOTES, 'UTF-8');
```

Closes #3794

### DIFF
--- a/src/www/ui/page/HomePage.php
+++ b/src/www/ui/page/HomePage.php
@@ -74,7 +74,7 @@ class HomePage extends DefaultPlugin
       }
     }
     if (! empty(GetParm("error", PARM_TEXT))) {
-      $vars['message'] = GetParm("error_description", PARM_TEXT);
+      $vars['message'] = htmlspecialchars(GetParm("error_description", PARM_TEXT), ENT_QUOTES, 'UTF-8');
     }
 
     if ($email !== null) {


shaheemazmalmmd (approved):
Code looks good.

PR #3797, opened 2026-08-31
fix(api): pass authenticated user id to folder access checks and guard root folder lookup

### LINKED ISSUE #3796: [Bug]: Unauthenticated user context in REST folder checks causes 500 fatal error in FolderDao
### Description

While testing folder permission validation in the REST API, I noticed that `UploadController::postUpload()` and `FolderController::getFolders()` call `FolderDao::isFolderAccessible($folderId)` without passing the authenticated user's ID (`$this->restHelper->getUserId()`).

Because `$userId` is omitted, `FolderDao::isFolderAccessible()` falls back to `Auth::getUserId()`. During token-authenticated REST calls, `Auth::getUserId()` returns `0`. This makes `$this->getRootFolder(0)` return `null`, which immediately throws a fatal error when calling `->getId()`:

```text
PHP Fatal error: Uncaught Error: Call to a member function getId() on null in /usr/local/share/fossology/lib/php/Dao/FolderDao.php:523

### PR DESCRIPTION
**Pull request by harshitnub077**
## Description

Enforce authenticated user context in `UploadController::postUpload()` and `FolderController::getFolders()` folder accessibility checks, and add a null-safety guard in `FolderDao::isFolderAccessible()`.

### Motivation

In the REST API, controller methods identify the caller via `$this->restHelper->getUserId()`. While almost all folder endpoints pass this user ID down to `FolderDao::isFolderAccessible($folderId, $userId)`, two locations omitted the argument:
1. `UploadController::postUpload()` at line 510
2. `FolderController::getFolders()` at line 53

When `$userId` is omitted, `FolderDao::isFolderAccessible()` falls back to `Auth::getUserId()`, which returns `0` for token-authenticated REST calls. This causes `$this->getRootFolder(0)` to return `null`, triggering a fatal `Call to a member function getId() on null` error (HTTP 500) instead of properly evaluating the authenticated user's permissions.

Passing `$this->restHelper->getUserId()` ensures permissions are evaluated strictly against the caller, and adding a null check in `FolderDao::isFolderAccessible()` prevents crashes when a user record has no associated root folder.

### Changes

- **`src/lib/php/Dao/FolderDao.php`**:
  - Added null check on `$this->getRootFolder($userId)` before accessing `->getId()`, returning `false` if `$rootFolder` is null.
- **`src/www/ui/api/Controllers/UploadController.php`**:
  - Passed `$this->restHelper->getUserId()` to `$this->restHelper->getFolderDao()->isFolderAccessible($folderId, $this->restHelper->getUserId())`.
- **`src/www/ui/api/Controllers/FolderController.php`**:
  - Passed `$this->restHelper->getUserId()` to `$folderDao->isFolderAccessible($id, $this->restHelper->getUserId())`.
- **`src/www/ui_tests/api/Controllers/UploadControllerTest.php`** & **`src/www/ui_tests/api/Controllers/FolderControllerTest.php`**:
  - Updated Mockery expectations for `isFolderAccessible` to include the user ID parameter across the affected test cases.
- **`src/lib/php/Dao/test/FolderDaoTest.php`**:
  - Added unit test `testIsFolderAccessible_nullRootFolderReturnsFalse` to verify that `isFolderAccessible()` returns `false` safely when no root folder exists.

## How to test

1. Run the unit test suites:
   ```bash
   phpunit -c src/www/ui_tests/phpunit.xml src/www/ui_tests/api/Controllers/UploadControllerTest.php
   phpunit -c src/www/ui_tests/phpunit.xml src/www/ui_tests/api/Controllers/FolderControllerTest.php
   phpunit -c src/lib/php/Dao/test/phpunit.xml src/lib/php/Dao/test/FolderDaoTest.php
   ```
2. Send a token-authenticated `POST /api/v1/uploads` request as a non-admin user targeting an inaccessible folder:
   ```bash
   curl -k -X POST http://localhost/repo/api/v1/uploads \
     -H "Authorization: Bearer <TOKEN>" \
     -H "folderId: <OTHER_USER_FOLDER_ID>" \
     -H "uploadType: url" \
     -H "Content-Type: application/json" \
     -d '{"location": {"url": "https://example.com/package.tar.gz"}}'
   ```
3. Confirm that the API returns HTTP 403 (`folderId <ID> is not accessible!`) rather than an HTTP 500 internal server error.

Fixes #3796

### DIFF
--- a/src/lib/php/Dao/FolderDao.php
+++ b/src/lib/php/Dao/FolderDao.php
@@ -520,8 +520,11 @@ WHERE fc.parent_fk = $1 AND fc.foldercontents_mode = " . self::MODE_UPLOAD . ";"
     if ($userId == null) {
       $userId = Auth::getUserId();
     }
-    $rootFolder = $this->getRootFolder($userId)->getId();
-    return $this->isInFolderTree($rootFolder, $folderId);
+    $rootFolder = $this->getRootFolder($userId);
+    if (!$rootFolder) {
+      return false;
+    }
+    return $this->isInFolderTree($rootFolder->getId(), $folderId);
   }
 
   /**
--- a/src/lib/php/Dao/test/FolderDaoTest.php
+++ b/src/lib/php/Dao/test/FolderDaoTest.php
@@ -197,4 +197,12 @@ class FolderDaoTest extends \PHPUnit\Framework\TestCase
     $badFolder = $this->folderDao->getFolder(987);
     assertThat($badFolder, is(nullValue()));
   }
+
+  public function testIsFolderAccessible_nullRootFolderReturnsFalse()
+  {
+    $this->testDb->createPlainTables(array('users'));
+    $this->folderDao->ensureTopLevelFolder();
+    $isAccessible = $this->folderDao->isFolderAccessible(FolderDao::TOP_LEVEL, 99999);
+    assertThat($isAccessible, is(false));
+  }
 }
--- a/src/www/ui/api/Controllers/FolderController.php
+++ b/src/www/ui/api/Controllers/FolderController.php
@@ -50,7 +50,7 @@ class FolderController extends RestController
     $folderDao = $this->restHelper->getFolderDao();
     if (isset($args['id'])) {
       $id = intval($args['id']);
-      if (! $folderDao->isFolderAccessible($id)) {
+      if (! $folderDao->isFolderAccessible($id, $this->restHelper->getUserId())) {
         throw new HttpForbiddenException("Folder id $id is not accessible");
       }
       if ($folderDao->getFolder($id) === null) {
--- a/src/www/ui/api/Controllers/UploadController.php
+++ b/src/www/ui/api/Controllers/UploadController.php
@@ -493,7 +493,8 @@ class UploadController extends RestController
     if (!in_array($folderId, $allFolderIds)) {
       throw new HttpNotFoundException("folderId $folderId does not exists!");
     }
-    if (!$this->restHelper->getFolderDao()->isFolderAccessible($folderId)) {
+    if (!$this->restHelper->getFolderDao()->isFolderAccessible($folderId,
+        $this->restHelper->getUserId())) {
       throw new HttpForbiddenException("folderId $folderId is not accessible!");
     }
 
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by harshitnub077** [comment]
Thanks @shaheemazmalmmd for triaging and consolidating to this PR updated the branch against latest master. Ready for review whenever you have a moment

File: src/lib/php/Dao/FolderDao.php [hunk 1]
```diff
@@ -520,8 +520,11 @@ public function isFolderAccessible($folderId, $userId = null)
     if ($userId == null) {
       $userId = Auth::getUserId();
     }
-    $rootFolder = $this->getRootFolder($userId)->getId();
-    return $this->isInFolderTree($rootFolder, $folderId);
+    $rootFolder = $this->getRootFolder($userId);
+    if (!$rootFolder) {
+      return false;
+    }
+    return $this->isInFolderTree($rootFolder->getId(), $folderId);
```
its-sushant, line 527: Why do we need this change? Is this really reproducible?

src/lib/php/Dao/test/FolderDaoTest.php
```diff
@@ -197,4 +197,12 @@ public function testGetFolder()
     $badFolder = $this->folderDao->getFolder(987);
     assertThat($badFolder, is(nullValue()));
   }
+
+  public function testIsFolderAccessible_nullRootFolderReturnsFalse()
```
its-sushant: Kindly follow the naming convention for the function name.

its-sushant (changes requested):
@harshitnub077 thanks for the PR. Please look into the inline comments.

src/lib/php/Dao/FolderDao.php
harshitnub077: it was originally reproduced when the controllers did not pass the user id so it defaulted to Auth getUserId which returns 0 in token authenticated rest calls
getRootFolder 0 returned null which triggered a fatal getId on null error and threw a 500

passing the authenticated user id from the controllers solves that for valid users but keeping this null check makes isFolderAccessible safe against non existent or invalid user ids so it returns false instead of crashing
happy to revert this check if you prefer keeping the dao strictly assuming valid root folders

src/lib/php/Dao/test/FolderDaoTest.php
harshitnub077: updated the test name to testIsFolderAccessibleWhenRootFolderIsNull to match the camelCase convention across the suite and pushed the update

PR #3799, opened 2026-09-01
fix(api): enforce upload accessibility check in JobController::getJobs

### LINKED ISSUE #3798: REST API: getJobs endpoint discloses job and upload metadata across groups, no accessibility check on job id or upload filter
### Description

While comparing `JobController.php` against `UploadController.php`, I noticed `getJobs()` never checks whether the caller can actually access the upload a job belongs to. Every job's `id`, `name`, `uploadId`, owning `userName`/`groupName`, `status`, and per-agent `jobQueue` details (start/end times, items processed, log path, error text) can be read by any authenticated API user, regardless of which group they belong to.

### Location

`src/www/ui/api/Controllers/JobController.php`
Function: `getJobs()`, lines 99-136
Helpers: `getAllResults()` (lines 275-290), `getFilteredResults()` (lines 306-322)

### Vulnerable Code

```php
// src/www/ui/api/Controllers/JobController.php:118-132
$id = isset($args['id']) ? intval($args['id']) : null;
if ($id !== null && !$this->dbHelper->doesIdExist("job", "job_pk", $id)) {
  throw new HttpNotFoundException("Job id " . $id . " doesn't exist");
}

if ($id !== null) {
  /* If the ID is passed, don't check for upload */
  return $this->getAllResults($id, $status, $request, $response, $sort, $limit, $page, $apiVersion);
}

if (array_key_exists(self::UPLOAD_PARAM, $queryParams)) {
  /* If the upload is passed, filter accordingly */
  return $this->getFilteredResults(intval($queryParams[self::UPLOAD_PARAM]),
    $status, $request, $response, $sort, $limit, $page, $apiVersion);
}
```

`getFilteredResults()` (line 306) only calls `doesIdExist("upload", "upload_pk", $uploadId)`, an existence check, not an accessibility check, before returning every job for that upload. The `$id` branch is even more direct, the comment literally says "don't check for upload," and `getAllResults()` never resolves or checks the job's upload at all.

Compare this with `UploadController.php`, where practically every method that takes an upload id, `getUploadSummary`, `deleteUpload`, `getUploadLicenses`, `getUploadCopyrights`, `getClearingProgressInfo`, and a dozen others, calls `$this->uploadAccessible($id)` (defined in `RestController.php:99`, which checks both existence and `UploadDao::isAccessible()`) before doing anything else. `JobController::getJobs()` is the one place that takes a job/upload identifier and skips that call entirely, even though `Job::getArray()` (`src/www/ui/api/Models/Job.php:115-144`) already carries the job's `uploadId`, `userName` and `groupName`.

It's worth noting `downloadReport()`/`checkReport()` in `ReportController.php` (lines 265-275) does get this right for the actual report file, it resolves the job's `job_upload_fk` and calls `isAccessible()` before serving the report. So `getJobs()` is inconsistent with both its sibling controller and the endpoint that serves the report generated by the very jobs it lists.

### Root Cause

`getJobs()` authorizes the caller against the wrong thing (or nothing at all). For the `id` path it only confirms the job row exists; for the `upload` path it only confirms the upload row exists. Neither path confirms the requesting user's group actually has read access to that upload, which is the resource the job data is scoped to everywhere else in the API.

### Impact

Severity: Medium

Any authenticated API user can enumerate job ids (`job_pk` is a small sequential integer) or pass an arbitrary `upload` query parameter and get back job history for uploads and groups they have no relationship to at all: the owning user's name, the owning group's name, the upload id, job type/agent (e.g. "wget", "nomos_agent", "Delete"), status, timestamps, item counts, and any log/error text from the jobqueue rows. This leaks the existence and internal state of other groups' uploads and confirms who is scanning what, a real horizontal information disclosure of data FOSSology's per-group upload permission model is meant to keep private. The actual report file download stays protected by `ReportController::checkReport()`'s separate accessibility check, so this doesn't extend to full report content, it's a metadata leak, not a file leak.

### Steps to Reproduce

1. As UserA, in one group, upload a file and let any agent job run against it (e.g. nomos). Note the resulting job id from the response, or the upload id.
2. As UserB, a different user in a different group with no access to UserA's upload (confirmed by `GET /repo/api/v2/uploads/{uploadId}` returning 403 for UserB), call `GET /repo/api/v2/jobs/{jobId}` using UserB's own valid token.
3. The request returns 200 with the full job object, including UserA's `userName`, `groupName`, `uploadId`, `status` and `jobQueue`, even though UserB has no relationship to that upload or group.
4. Alternatively, `GET /repo/api/v2/jobs?upload={UserA's uploadId}` as UserB returns 200 with the full list of jobs for that upload, for the same reason.

### Suggested Fix

After resolving the job (or before returning it), check the job's `uploadId` against the caller's access with `$this->uploadAccessible($job->getUploadId())` (or the DAO's `isAccessible()` call directly, since the CTE result already carries `job_upload_fk`) and throw `HttpForbiddenException` when it fails, the same pattern already used throughout `UploadController.php` and in `ReportController::checkReport()`. This needs to apply to both the `id` branch in `getAllResults()` and the `upload` branch in `getFilteredResults()`.

bhuvan-somisetty:
### Problem

`JobController::getJobs()` serves both `GET /jobs/{id}` and `GET /jobs?upload={id}`, but only checked that the job/upload row existed, never that the requesting user's group could actually access the upload it belongs to. Every method in `UploadController.php` that takes an upload id calls the shared `uploadAccessible()` helper before doing anything else, `getJobs()` was the one place handling a job/upload identifier that skipped it, so any authenticated API user could enumerate job ids (a small sequential integer) or pass an arbitrary `upload` query parameter and read job metadata for uploads and groups they have no relationship to at all: the owning user's name, the owning group's name, the upload id, job/agent type, status, timestamps, and jobqueue log/error text.

### Root cause

The `id` branch in `getJobs()` (line 118-126) resolved straight into `getAllResults()`, which never checked the returned job's upload against the caller's access. The `upload` branch (`getFilteredResults()`) only called `doesIdExist("upload", "upload_pk", $uploadId)`, an existence check, not `UploadDao::isAccessible()`. Note the report file itself stays protected, `ReportController::checkReport()` already resolves the job's `job_upload_fk` and calls `isAccessible()` before serving the report, so this was a metadata leak rather than a content leak, but a real one, it discloses the existence and internal state of other groups' uploads.

### Fix

Call the existing `RestController::uploadAccessible()` helper (already the standard pattern across `UploadController` and `ReportController::checkReport()`) in both places, `getAllResults()` after fetching the single job, checked against the upload id the job already carries (`$jobs[0]->getUploadId()`), no extra query needed, and `getFilteredResults()` before querying, replacing the existence-only check. There is no behavioural change for callers who already have access, and admin-facing `getAllJobs()` (id = null) is untouched since the new check is scoped to the single-job and upload-filter paths only.

### Testing

`JobControllerTest.php` was extended with three new regression tests covering a single job id whose upload the caller can't access (now 403), an `upload` filter for an upload the caller can't access (now 403, and the job list is never queried), and an `upload` filter for a non-existent upload (still 404). The two existing success-path tests (`testGetJobFromId`, `testGetJobsFromUpload`) were updated to mock the new accessibility check.

I ran `php vendor/bin/phpunit -c phpunit.xml --bootstrap phpunit-bootstrap.php www/ui_tests/api/Controllers/JobControllerTest.php`, 9/9 passing, and the broader `www/ui_tests/api/Controllers` directory (306 tests) to confirm no regressions elsewhere; the one pre-existing failure there (`LicenseCompatibilityRuleControllerTest::testExportRules`, missing `yaml` PHP extension in my test environment) is unrelated to this change and was already failing before it. `phpcs --standard=fossy-ruleset.xml` on both changed files reports no new violations, the two findings it does report (CRLF line endings, a class-brace style note) already exist on the unmodified file.

Fixes #3798

### DIFF
--- a/src/www/ui/api/Controllers/JobController.php
+++ b/src/www/ui/api/Controllers/JobController.php
@@ -121,7 +121,8 @@ class JobController extends RestController
     }
 
     if ($id !== null) {
-      /* If the ID is passed, don't check for upload */
+      /* If the ID is passed, ignore any upload query parameter and let
+       * getAllResults() authorize access to the job's own upload. */
       return $this->getAllResults($id, $status, $request, $response, $sort, $limit, $page, $apiVersion);
     }
 
@@ -271,10 +272,17 @@ class JobController extends RestController
    * @param integer $page    Page number required
    * @param integer $apiVersion API version
    * @return ResponseHelper
+   * @throws HttpErrorException If a specific job id is requested and its
+   *         upload does not exist or is not accessible to the caller
    */
   private function getAllResults($id, $status, $request, $response, $sort, $limit, $page, $apiVersion)
   {
     list($jobs, $count) = $this->dbHelper->getJobs($id, $status, $sort, $limit, $page, null);
+    if ($id !== null && !empty($jobs)) {
+      /* A specific job was requested, make sure the caller can access the
+       * upload it belongs to before returning any of its data. */
+      $this->uploadAccessible($jobs[0]->getUploadId());
+    }
     $finalJobs = [];
     foreach ($jobs as $job) {
       $this->updateEta($job);
@@ -301,14 +309,12 @@ class JobController extends RestController
    * @param integer $page Page number required
    * @param integer $apiVersion API version
    * @return ResponseHelper
-   * @throws HttpNotFoundException
+   * @throws HttpErrorException If the upload does not exist or is not
+   *         accessible to the caller
    */
   private function getFilteredResults($uploadId, $status, $request, $response, $sort, $limit, $page, $apiVersion)
   {
-    if (! $this->dbHelper->doesIdExist("upload", "upload_pk", $uploadId)) {
-      throw new HttpNotFoundException("Upload id " . $uploadId .
-        " doesn't exist");
-    }
+    $this->uploadAccessible($uploadId);
     list($jobs, $count) = $this->dbHelper->getJobs(null, $status, $sort, $limit, $page, $uploadId);
     $finalJobs = [];
     foreach ($jobs as $job) {
--- a/src/www/ui_tests/api/Controllers/JobControllerTest.php
+++ b/src/www/ui_tests/api/Controllers/JobControllerTest.php
@@ -14,7 +14,9 @@ namespace Fossology\UI\Api\Test\Controllers;
 
 use Fossology\Lib\Dao\JobDao;
 use Fossology\Lib\Dao\ShowJobsDao;
+use Fossology\Lib\Dao\UploadDao;
 use Fossology\UI\Api\Controllers\JobController;
+use Fossology\UI\Api\Exceptions\HttpForbiddenException;
 use Fossology\UI\Api\Exceptions\HttpNotFoundException;
 use Fossology\UI\Api\Helper\ResponseHelper;
 use Fossology\UI\Api\Models\ApiVersion;
@@ -60,6 +62,18 @@ class JobControllerTest extends \PHPUnit\Framework\TestCase
   private $showJobsDao;
 
   /**
+   * @var UploadDao $uploadDao
+   * UploadDao mock
+   */
+  private $uploadDao;
+
+  /**
+   * @var integer $groupId
+   * Group id of the current user
+   */
+  private $groupId;
+
+  /**
    * @var JobController $jobController
    * JobController object to test
    */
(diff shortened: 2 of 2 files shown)


shaheemazmalmmd (approved):
Code looks good.

bhuvan-somisetty:
Thanks for reviewing! Let me know if there's anything else needed to get this merged.

PR #3802, opened 2026-09-03
fix(api): require write permission for clearing-decision write endpoints

### LINKED ISSUE #3801: REST API: clearing-decision endpoints skip write-permission check, letting read-only users edit license decisions
### Description

While tracing how the REST API enforces upload permissions (following up on the `getJobs`/`unlinkFolder`/`getGroupMembers` accessibility bugs I filed earlier), I noticed `UploadTreeController.php` has a different flavor of the same class of bug: the endpoints that actually *write* clearing decisions only check that the caller can read the upload, never that they can edit it. A group with read-only permission on an upload can create, edit, and delete license clearing decisions on it through the API, something the regular web UI explicitly blocks for the same user.

### Location

`src/www/ui/api/Controllers/UploadTreeController.php`
- `setClearingDecision()`, lines 130-161 (route: `PUT /uploads/{id}/item/{itemId}/clearing-decision`)
- `handleAddEditAndDeleteLicenseDecision()`, lines 611-720 (route: `PUT /uploads/{id}/item/{itemId}/licenses`)

### Vulnerable Code

```php
// src/www/ui/api/Controllers/UploadTreeController.php:130-134
public function setClearingDecision($request, $response, $args)
{
  $uploadId = intval($args['id']);
  $this->uploadAccessible($uploadId);
  ...
```

```php
// src/www/ui/api/Controllers/UploadTreeController.php:611-621
public function handleAddEditAndDeleteLicenseDecision($request, $response, $args)
{
  $body = $this->getParsedBody($request);
  $uploadTreeId = intval($args['itemId']);
  $uploadId = intval($args['id']);
  $uploadDao = $this->restHelper->getUploadDao();
  $errors = [];
  $success = [];

  $this->uploadAccessible($uploadId);
  $this->isItemExists($uploadId, $uploadTreeId);
  ...
```

Both call `$this->uploadAccessible($uploadId)` (`RestController.php:99`), which only calls `UploadDao::isAccessible()`. That method (`UploadPermissionDao::isAccessible()`, `src/lib/php/Dao/UploadPermissionDao.php:27-42`) returns true as soon as `perm_upload.perm > Auth::PERM_NONE`, i.e. it's satisfied by `Auth::PERM_READ` (=1) alone. Neither method ever calls `UploadDao::isEditable()`, which is the one that actually checks `perm >= Auth::PERM_WRITE` (=3) (`UploadPermissionDao::isEditable()`, same file, lines 44-56).

After that check, `handleAddEditAndDeleteLicenseDecision()` goes straight to `$this->clearingDao->insertClearingEvent(...)` / `updateClearingEvent(...)` (`ClearingDao.php:449`, `509`), which write to `clearing_event` unconditionally, no permission check inside the DAO either. `setClearingDecision()` goes through `ClearingView::updateLastItem()` (`src/www/ui/ui-clearing-view.php:387-408`), which calls `ClearingDecisionProcessor::makeDecisionFromLastEvents()` just as unconditionally.

### Root Cause

`read_only` is a first-class, admin-assignable permission level in FOSSology's own model, `UploadController::setUploadPermissions()` maps it straight to `Auth::PERM_READ` (`UploadController.php:877-878`), and it's meant to let a group view an upload without being able to touch its clearing state. The regular web UI respects this: `ClearingView::Output()` computes `$hasWritePermission = $this->uploadDao->isEditable($uploadId, $groupId)` (`ui-clearing-view.php:279`) and, for a read-only user, sets `auditDenied` instead of rendering the license-editing controls (line 302-304). But `updateLastItem()` itself, the method the REST API calls directly, never re-checks that permission, it just trusts the caller not to invoke it. The REST controller is exactly that untrusted caller, and it skips the check the HTML page uses to decide whether to even show the button.

### Impact

Severity: Medium-High

Any authenticated API user in a group that's been granted read-only access to an upload, a normal, intentional sharing configuration, not a misconfiguration, can call these two endpoints directly and: add or remove concluded licenses on any file in that upload, edit report text/acknowledgement/comment on existing decisions, and flip the upload's global clearing decision type (WIP/final/etc.) via `globalDecision`. This corrupts the legal clearing record FOSSology exists to produce, and it's done by a user the upload owner explicitly restricted to viewing. There's no trace in the UI that this happened outside the normal audit flow since the DB rows look like any other clearing event.

### Steps to Reproduce

1. As UserA (upload owner), upload a file and share it with GroupB as `read_only` via `PUT /repo/api/v2/uploads/{uploadId}/permissions` (`groupId`, `newPermission: "read_only"`).
2. As UserB, a member of GroupB with GroupB as active group, open the upload's license view in the web UI, confirm the license-editing controls are hidden (`auditDenied`).
3. As UserB, call `PUT /repo/api/v2/uploads/{uploadId}/item/{itemId}/licenses` with body `[{"shortName":"MIT","add":true}]` using UserB's own token.
4. The request returns 200 and a new `clearing_event` row is created for that license on that item, even though UserB only has read access to the upload.
5. Same result calling `PUT /repo/api/v2/uploads/{uploadId}/item/{itemId}/clearing-decision` with a `decisionType`/`globalDecision` body.

### Suggested Fix

Add `$this->uploadAccessible($uploadId)` *and* a write check, e.g. `if (!$this->restHelper->getUploadDao()->isEditable($uploadId, $this->restHelper->getGroupId())) { throw new HttpForbiddenException(...); }`, at the top of both `setClearingDecision()` and `handleAddEditAndDeleteLicenseDecision()`, before any DAO write happens. Worth checking `scheduleBulkScan()` in the same file for the same gap while in there, since it's another item-level write reachable with only `uploadAccessible()`.

bhuvan-somisetty:
## Description

UploadTreeController::setClearingDecision(), handleAddEditAndDeleteLicenseDecision() and scheduleBulkScan() only called uploadAccessible(), which is satisfied by Auth::PERM_READ alone (UploadPermissionDao::isAccessible() just checks perm > PERM_NONE). None of the three ever checked UploadDao::isEditable(), which requires Auth::PERM_WRITE.

read_only is a first class, admin assignable permission level, UploadController::setUploadPermissions() maps it straight to Auth::PERM_READ, so this is not a misconfiguration, it is normal upload sharing. A group with read only access could still create, edit and delete license clearing decisions and schedule bulk scans on that upload through the REST API. The legacy web UI gets this right, ClearingView::Output() computes $hasWritePermission = $this->uploadDao->isEditable(...) and hides the license editing controls (auditDenied) for a read only user, but updateLastItem(), the method the REST API calls directly, never re-checks that permission itself, and neither does ClearingDao::insertClearingEvent()/updateClearingEvent() underneath handleAddEditAndDeleteLicenseDecision().

## Root cause

Missing write permission check at the REST controller layer, the one place both the UI page and the DAO actually rely on to gate the write.

## Changes

* Added RestController::uploadEditable($id), which calls the existing uploadAccessible($id) (existence and read check) and then throws HttpForbiddenException unless UploadDao::isEditable($id, $groupId) (Auth::PERM_WRITE) is true.
* Replaced uploadAccessible($uploadId) with uploadEditable($uploadId) in the three write endpoints, setClearingDecision(), handleAddEditAndDeleteLicenseDecision(), scheduleBulkScan(). Every read only endpoint in the file (viewLicenseFile, getNextPreviousItem, getBulkHistory, getClearingHistory, getHighlightEntries, getTreeView, getLicenseDecisions) is untouched and still only requires read access, as it should.
* Updated the existing PHPUnit coverage for the three endpoints to mock isEditable() => true on the happy path, and added one regression test per endpoint asserting HttpForbiddenException (and that the DAO or plugin write call never happens) when isEditable() is false.

## How to test

1. As UserA (upload owner), upload a file and share it with GroupB as read_only via PUT /repo/api/v2/uploads/{uploadId}/permissions.
2. As UserB, a member of GroupB, open the upload's license view in the web UI and confirm the license editing controls are hidden.
3. As UserB, call PUT /repo/api/v2/uploads/{uploadId}/item/{itemId}/licenses with body [{"shortName":"MIT","add":true}].
   Before this fix: 200, a new clearing_event row is created.
   After this fix: 403 Forbidden, nothing is written.
4. Same before and after for PUT /repo/api/v2/uploads/{uploadId}/item/{itemId}/clearing-decision and POST /repo/api/v2/uploads/{uploadId}/item/{itemId}/bulk-scan.

I verified the new tests actually catch the regression by temporarily reverting the uploadEditable calls back to uploadAccessible and confirming all three ...ReturnsForbiddenForReadOnlyUpload tests fail against the vulnerable code, then re-ran green after restoring the fix.

Ran www/ui_tests/api/Controllers/UploadTreeControllerTest.php (PHPUnit 8.5, PHP 8.5), 20/20 passing, including the 3 new tests. Also ran FolderControllerTest, JobControllerTest and GroupControllerTest (other controllers sharing the RestController base) to confirm nothing else regressed, all passing. Ran phpcs with the fossy-ruleset.xml standard on the changed files, no violations introduced. Ran php -l on the changed files, no syntax errors.

Fixes #3801

### DIFF
--- a/src/www/ui/api/Controllers/RestController.php
+++ b/src/www/ui/api/Controllers/RestController.php
@@ -108,6 +108,26 @@ class RestController
   }
 
   /**
+   * Check if upload is accessible and writable by the current group.
+   *
+   * Read access (Auth::PERM_READ) alone is not enough to create or modify
+   * clearing decisions, license findings or scans on an upload; the calling
+   * group needs at least Auth::PERM_WRITE.
+   *
+   * @param integer $id Upload ID
+   * @throws HttpNotFoundException Upload not found
+   * @throws HttpForbiddenException Upload not accessible or read-only for the caller
+   */
+  protected function uploadEditable($id): void
+  {
+    $this->uploadAccessible($id);
+    if (! $this->restHelper->getUploadDao()->isEditable($id,
+        $this->restHelper->getGroupId())) {
+      throw new HttpForbiddenException("Upload is not editable");
+    }
+  }
+
+  /**
    * Check if upload tree is accessible
    *
    * @param int $uploadId
--- a/src/www/ui/api/Controllers/UploadTreeController.php
+++ b/src/www/ui/api/Controllers/UploadTreeController.php
@@ -130,7 +130,7 @@ class UploadTreeController extends RestController
   public function setClearingDecision($request, $response, $args)
   {
     $uploadId = intval($args['id']);
-    $this->uploadAccessible($uploadId);
+    $this->uploadEditable($uploadId);
 
     $body = $this->getParsedBody($request);
     $decisionType = $body['decisionType'];
@@ -617,7 +617,7 @@ class UploadTreeController extends RestController
     $errors = [];
     $success = [];
 
-    $this->uploadAccessible($uploadId);
+    $this->uploadEditable($uploadId);
     $this->isItemExists($uploadId, $uploadTreeId);
 
     if (empty($body)) {
@@ -735,7 +735,7 @@ class UploadTreeController extends RestController
     $uploadId = intval($args['id']);
     $licenseDao = $this->container->get('dao.license');
 
-    $this->uploadAccessible($uploadId);
+    $this->uploadEditable($uploadId);
     $this->isItemExists($uploadId, $uploadTreeId);
 
     $isValid = true;
--- a/src/www/ui_tests/api/Controllers/UploadTreeControllerTest.php
+++ b/src/www/ui_tests/api/Controllers/UploadTreeControllerTest.php
@@ -321,6 +321,8 @@ namespace Fossology\UI\Api\Test\Controllers {
       $this->uploadDao->shouldReceive("getUploadtreeTableName")->withArgs([$upload_pk])->andReturn("uploadtree");
       $this->uploadDao->shouldReceive('isAccessible')
         ->withArgs([$upload_pk, $this->groupId])->andReturn(true);
+      $this->uploadDao->shouldReceive('isEditable')
+        ->withArgs([$upload_pk, $this->groupId])->andReturn(true);
       $this->dbHelper->shouldReceive('doesIdExist')
         ->withArgs(["upload", "upload_pk", $upload_pk])->andReturn(true);
       $this->dbHelper->shouldReceive('doesIdExist')
@@ -369,6 +371,8 @@ namespace Fossology\UI\Api\Test\Controllers {
       $this->uploadDao->shouldReceive("getUploadtreeTableName")->withArgs([$upload_pk])->andReturn("uploadtree");
       $this->uploadDao->shouldReceive('isAccessible')
         ->withArgs([$upload_pk, $this->groupId])->andReturn(true);
+      $this->uploadDao->shouldReceive('isEditable')
+        ->withArgs([$upload_pk, $this->groupId])->andReturn(true);
       $this->dbHelper->shouldReceive('doesIdExist')
         ->withArgs(["upload", "upload_pk", $upload_pk])->andReturn(true);
       $this->dbHelper->shouldReceive('doesIdExist')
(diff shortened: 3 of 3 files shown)


bhuvan-somisetty:
@shaheemazmalmmd PTAL whenever u get chance

PR #3804, opened 2026-09-06
fix(api): require admin permission for deleteObligation endpoint

### LINKED ISSUE #3803: REST API: deleteObligation is missing the admin check every sibling endpoint in the same controller has
### Description

I was going through `ObligationController.php` after fixing a couple of similar missing-authz endpoints elsewhere in the API (#3784, #3790, #3792, #3794), and checked whether the same pattern existed here. Every write/admin action in this controller calls `throwNotAdminException()` at the top, except `deleteObligation`, which doesn't call it at all.

### Location

`src/www/ui/api/Controllers/ObligationController.php`
Function: `deleteObligation()`, lines 131-140

### Vulnerable Code

```php
// src/www/ui/api/Controllers/ObligationController.php:131-140
function deleteObligation($request, $response, $args)
{
  $obligationId = intval($args['id']);
  if (!$this->dbHelper->doesIdExist("obligation_ref", "ob_pk", $obligationId)) {
    throw new HttpNotFoundException("Obligation does not exist");
  }
  $this->obligationMap->deleteObligation($obligationId);
  $returnVal = new Info(200, "Successfully removed Obligation.", InfoType::INFO);
  return $response->withJson($returnVal->getArray(), $returnVal->getCode());
}
```

Compare this with the other four methods in the exact same file, `importObligationsFromCSV`, `exportObligationsToCSV`, `importObligationsFromJSON` and `exportObligationsToJSON`, every one of them starts with `$this->throwNotAdminException();` before doing anything else. `deleteObligation` is the only mutating endpoint in the controller that skips it.

The route is wired up in `src/www/ui/api/index.php`:

```php
$app->delete('/{id:\d+}', ObligationController::class . ':deleteObligation');
```

and `RestAuthMiddleware` only checks that the request carries a valid, "write"-scoped token, it has no concept of admin at all, that check is left entirely to each controller method. So there is nothing upstream of `deleteObligation` that stops a non-admin caller.

For context on what this data actually is: the web UI equivalent, `admin-obligation-file.php`, sets `$this->DBaccess = PLUGIN_DB_ADMIN;`, meaning obligation/risk management is intentionally admin-only in the UI. `obligation_ref` rows aren't per-user or per-group data, they're the shared obligation/risk definitions referenced by license clearing and reporting across the whole instance.

### Root Cause

`deleteObligation` never checks the caller's privilege level before deleting the row. Every sibling method in the same class enforces `Auth::isAdmin()` via `throwNotAdminException()`, this one just forgot to.

### Impact

Severity: Medium

Any authenticated user can generate their own REST API token with write scope for their own account, no special privilege is needed to do that, and then call `DELETE /repo/api/v2/obligations/{id}` for any obligation id and have it deleted, even though the equivalent action is admin-only everywhere else (UI and the rest of this same controller). Since obligation rows are shared, instance-wide reference data used in license clearing and obligation reporting, a non-admin user can silently remove entries other users and admins rely on, which is a broken access control issue with real data-integrity impact, not just an information leak.

### Steps to Reproduce

1. As a non-admin user, log in and generate a personal REST API token with write scope (`POST /users/tokens`, `tokenScope: write`), this is available to every account, no admin action required.
2. Find any existing obligation id, e.g. via `GET /obligations/list`.
3. Call `DELETE /repo/api/v2/obligations/{id}` using the non-admin token.
4. The obligation is deleted and a 200 is returned, even though the same user gets a 403 if they try `POST /obligations/import-csv` or any of the other admin-gated endpoints in this controller.

### Suggested Fix

Add `$this->throwNotAdminException();` as the first line of `deleteObligation`, matching the pattern already used by the other four methods in this file. Should be a small, self-contained fix, and it needs a regression test in `ObligationControllerTest` asserting a non-admin caller gets `HttpForbiddenException` from this endpoint, since none of the current tests appear to cover that case.

bhuvan-somisetty:
## Description

ObligationController::deleteObligation() was missing the admin check that every other mutating endpoint in the same file enforces. importObligationsFromCSV, exportObligationsToCSV, importObligationsFromJSON and exportObligationsToJSON all call $this->throwNotAdminException() as their first line, deleteObligation did not.

RestAuthMiddleware only validates the token's scope (read vs write), it has no concept of admin, so nothing upstream of the controller stopped a non admin caller. Obligation rows are shared, instance wide reference data used in license clearing and reporting, and the web UI equivalent (admin-obligation-file.php) sets DBaccess = PLUGIN_DB_ADMIN, so deleting them is meant to be admin only, same as it already is for every other endpoint in this controller.

## Changes

Added $this->throwNotAdminException(); as the first line of deleteObligation(), matching the pattern already used by the other four methods in the file.

## How to test

Added ObligationControllerTest.php, there was no existing test file for this controller. Covers:

* testDeleteObligationAsAdmin: admin deleting an existing obligation still succeeds (200).
* testDeleteObligationNotFound: admin deleting a non existent id still gets HttpNotFoundException.
* testDeleteObligationNotAdmin: regression test for the bug, a non admin, write scoped caller now gets HttpForbiddenException, and ObligationMap::deleteObligation() is asserted to never be called.
* testDeleteObligationNotAdminNonexistentId: a non admin caller gets HttpForbiddenException even for an id that does not exist, so the endpoint cannot be used to probe which obligation ids are valid before the admin check.

I do not have a working PHP or Docker toolchain in my current environment to run phpunit or phpcs locally, so I was not able to execute this suite myself before opening the PR. I did verify by hand that every mocked method name, signature and constant (DbHelper::doesIdExist, ObligationMap::deleteObligation, Auth::PERM_ADMIN and PERM_WRITE) matches the real source, and the test file follows the same Mockery and PHPUnit structure as the existing, passing LicenseCompatibilityRuleControllerTest.php. Please let CI's phpunit and phpcs run confirm this, and flag anything that does not line up.

## Impact

Before this change, any authenticated user could generate their own write scoped REST API token (no special privilege needed) and call DELETE /repo/api/v2/obligations/{id} to delete shared obligation or risk data, even though the identical action is admin gated in the web UI and in every sibling endpoint of this same controller. This is a broken access control and missing function level authorization issue with a data integrity impact, not just information disclosure.

Fixes #3803

### DIFF
--- a/src/www/ui/api/Controllers/RestController.php
+++ b/src/www/ui/api/Controllers/RestController.php
@@ -108,6 +108,26 @@ class RestController
   }
 
   /**
+   * Check if upload is accessible and writable by the current group.
+   *
+   * Read access (Auth::PERM_READ) alone is not enough to create or modify
+   * clearing decisions, license findings or scans on an upload; the calling
+   * group needs at least Auth::PERM_WRITE.
+   *
+   * @param integer $id Upload ID
+   * @throws HttpNotFoundException Upload not found
+   * @throws HttpForbiddenException Upload not accessible or read-only for the caller
+   */
+  protected function uploadEditable($id): void
+  {
+    $this->uploadAccessible($id);
+    if (! $this->restHelper->getUploadDao()->isEditable($id,
+        $this->restHelper->getGroupId())) {
+      throw new HttpForbiddenException("Upload is not editable");
+    }
+  }
+
+  /**
    * Check if upload tree is accessible
    *
    * @param int $uploadId
--- a/src/www/ui/api/Controllers/UploadTreeController.php
+++ b/src/www/ui/api/Controllers/UploadTreeController.php
@@ -130,7 +130,7 @@ class UploadTreeController extends RestController
   public function setClearingDecision($request, $response, $args)
   {
     $uploadId = intval($args['id']);
-    $this->uploadAccessible($uploadId);
+    $this->uploadEditable($uploadId);
 
     $body = $this->getParsedBody($request);
     $decisionType = $body['decisionType'];
@@ -617,7 +617,7 @@ class UploadTreeController extends RestController
     $errors = [];
     $success = [];
 
-    $this->uploadAccessible($uploadId);
+    $this->uploadEditable($uploadId);
     $this->isItemExists($uploadId, $uploadTreeId);
 
     if (empty($body)) {
@@ -735,7 +735,7 @@ class UploadTreeController extends RestController
     $uploadId = intval($args['id']);
     $licenseDao = $this->container->get('dao.license');
 
-    $this->uploadAccessible($uploadId);
+    $this->uploadEditable($uploadId);
     $this->isItemExists($uploadId, $uploadTreeId);
 
     $isValid = true;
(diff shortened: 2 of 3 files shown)


shaheemazmalmmd:
PR Contains fixes for multiple things that are not needed

PR #3810, opened 2026-09-08
fix(api): add missing groupMap check in deleteGroup

### LINKED ISSUE #3809: deleteGroup returns wrong HTTP status on permission error and ignores built groupMap
## What's happening

`GroupController::deleteGroup()` builds a `$groupMap` of groups the current user is allowed to delete but never checks it before calling `deleteGroup()`:

```php
$groupMap = $userDao->getDeletableAdminGroupMap($userId, $_SESSION[Auth::USER_LEVEL]);

// $groupMap is never checked here
try {
    $userDao->deleteGroup($groupId);  // called without checking groupMap
    unset($groupMap[$groupId]);       // $groupMap never used again after this
} catch (\Exception $e) {
    throw new HttpBadRequestException($e->getMessage(), $e); // wrong type
}
```

## Three problems

**1. Wrong HTTP status on permission error**
Permission errors are caught and re-thrown as `400 Bad Request` instead of `403 Forbidden`. API clients can't distinguish bad input from not allowed.

**2. Inconsistent with `updateGroup()` in the same file**
`updateGroup()` correctly checks the map before acting `deleteGroup()` doesn't.

```php
// updateGroup() correct pattern
if (!array_key_exists($groupId, $groupMap)) {
    throw new HttpForbiddenException("Not admin of the group.");
}
```

Note:  The actual deletion is still blocked by an inner check inside
`UserDao::deleteGroup()` this is not a security bypass, but an API contract bug.

## Fix

```php
if (!array_key_exists($groupId, $groupMap)) {
    throw new HttpForbiddenException("Not admin of the group. Can not process request.");
}
```

## Files affected

- `src/www/ui/api/Controllers/GroupController.php`

Aditya30ag:
## Description

`deleteGroup()` built a `$groupMap` of groups the current user is allowed to delete but never used it to gate the deletion. Permission errors were also returned as `400 Bad Request` instead of `403 Forbidden`, giving API clients incorrect error semantics.

### Changes

- Added missing `array_key_exists($groupId, $groupMap)` check in `deleteGroup()` before calling `userDao->deleteGroup()`, consistent with the existing pattern in `updateGroup()` in the same file
- Permission errors now correctly throw `HttpForbiddenException` (HTTP 403) instead of being swallowed into `HttpBadRequestException` (HTTP 400)

## How to test

1. Create two groups `GroupA` and `GroupB`
2. Log in as a non-admin user who is group admin of `GroupA` only
3. Call `DELETE /api/v1/groups/{GroupB_id}` with that user's token
4. **Before fix:** returns `400 Bad Request` with message "Permission Denied."
5. **After fix:** returns `403 Forbidden` with message "Not admin of the group."
6. Confirm `GroupB` is not deleted in either case
7. Call `DELETE /api/v1/groups/{GroupA_id}` should still return `202` and
   delete the group successfully

Closes #3809


Aditya30ag:
The docker-test failure (fo_cli unable to connect on port 24693) is unrelated to this change. This PR only modifies GroupController.php and GroupControllerTest.php (pure PHP). The scheduler socket failure 
is a CI infrastructure/timing issue all 1377 PHPUnit tests pass. 
Please re-run the docker-test workflow.

PR #3820, opened 2026-09-15
fix(api): add missing admin check in deleteObligation

### LINKED ISSUE #3819: Missing admin permission check in `deleteObligation` endpoint
## What's happening

`DELETE /api/v1/obligations/{id}` permanently deletes an obligation from the database but has no admin permission check:

```php
function deleteObligation($request, $response, $args)
{
  // ← throwNotAdminException() missing here
  $obligationId = intval($args['id']);
  $this->obligationMap->deleteObligation($obligationId);
```

Every other write endpoint in the same controller requires admin:

```php
public function importObligationsFromCSV()  { $this->throwNotAdminException(); }
public function exportObligationsToCSV()    { $this->throwNotAdminException(); }
public function importObligationsFromJSON() { $this->throwNotAdminException(); }
public function exportObligationsToJSON()   { $this->throwNotAdminException(); }
```

## Why admin is required

- The OpenAPI spec tags this endpoint under `Admin`
- The web UI page that does the same deletion sets `$this->DBaccess = PLUGIN_DB_ADMIN`
- All sibling write endpoints in the same controller enforce admin

Aditya30ag:
## Description

`deleteObligation` permanently deletes an obligation from the database but was missing the admin permission check present in every other write endpoint in the same controller.

### Changes

- Added `$this->throwNotAdminException()` at the start of `deleteObligation()` consistent with `importObligationsFromCSV`, `exportObligationsToCSV`, `importObligationsFromJSON` and `exportObligationsToJSON` in the same file

## How to test

1. Generate a write-scope API token for a **non-admin** user
2. Call `DELETE /api/v1/obligations/{id}` with that token
3. **Before fix:** returns `200` obligation deleted
4. **After fix:** returns `403 Forbidden` "Only admin can access this endpoint."
5. Repeat with an **admin** token should still return `200` and delete successfully

Closes #3819


Aditya30ag:
Hello @shaheemazmalmmd,
Could you please review this PR? It fixes a missing admin permission check in the `deleteObligation` endpoint a one-line fix consistent with the existing pattern in the same controller.

shaheemazmalmmd (approved):
Code looks good.

PR #3822, opened 2026-09-15
fix(ui): compute unfiltered recordsTotal for compatibility rules table

### LINKED ISSUE #3821: License Compatibility Rules page reports wrong recordsTotal when a search filter is active
### Description

While looking through the server-side pagination endpoint for the License Compatibility Rules admin page, I noticed `recordsTotal` and `recordsFiltered` are set to the same value, and that value is actually the filtered count, not the true total row count.

### Location

`src/www/ui/page/AdminLicenseCompatibilityRules.php`, method `fetchRules()`

```php
private function fetchRules(Request $request)
{
  $offset = intval($request->query->get('start', 0));
  $limit = intval($request->query->get('length', 10));
  $draw = intval($request->query->get('draw', 1));
  $searchQuery = $_GET['search']['value'] ?? '';

  if (!empty($searchQuery)) {
    $searchQuery = '%' . $searchQuery . '%';
  }

  $totalCount = $this->compatibilityDao->getTotalRulesCount($searchQuery);
  $ruleArray = $this->compatibilityDao->getAllRules($limit, $offset, $searchQuery);

  return new JsonResponse([
    "draw" => $draw,
    "recordsTotal" => $totalCount,
    "recordsFiltered" => $totalCount,
    "data" => $ruleArray,
  ], JsonResponse::HTTP_OK);
}
```

### Root Cause

`getTotalRulesCount()` in `src/lib/php/Dao/CompatibilityDao.php` takes the search term and, when it's non-empty, adds a `WHERE comment ILIKE $1` clause:

```php
public function getTotalRulesCount($searchTerm = '')
{
  $stmt = __METHOD__;
  $params = [];
  $query = "SELECT COUNT(*) as count FROM license_rules";
  if (!empty($searchTerm)) {
    $params[] = $searchTerm;
    $query .= ' WHERE comment ILIKE $' . count($params);
    $stmt .= '.search';
  }
  $count = $this->dbManager->getSingleRow($query, $params, $stmt);
  return $count ? intval($count['count']) : 0;
}
```

`fetchRules()` always passes `$searchQuery` into this call, so whenever a user has typed something in the DataTable search box, `$totalCount` already reflects the filtered result set, not the full table. That same filtered number is then reused for both `recordsTotal` and `recordsFiltered` in the JSON response, which isn't what DataTables' server-side protocol expects: `recordsTotal` should be the unfiltered row count, and `recordsFiltered` the count after the search is applied.

You can see the correct pattern right next door in `src/www/ui/page/AdminCustomTextList.php`, which implements basically the same kind of endpoint:

```php
$totalCount = $this->getTotalPhrasesCount('');
$filteredCount = empty($filteredQuery) ? $totalCount : $this->getTotalPhrasesCount($filteredQuery);
...
"recordsTotal" => $totalCount,
"recordsFiltered" => $filteredCount,
```

`fetchRules()` should be doing the same thing: call `getTotalRulesCount()` with an empty string for `recordsTotal`, and only use the filtered count for `recordsFiltered`.

### Impact

Nothing security-sensitive, just incorrect UI data. With a search term active, the DataTable's "Showing 1 to N of N entries (filtered from X total entries)" footer shows the wrong X, so there's no way to tell how many compatibility rules exist in total while searching. It also feeds into the "add rule" button's pagination-jump logic in `src/www/ui/template/admin_license_compatibility_rules.js.twig`, which reads `pageInfo.recordsTotal` to figure out which page to jump to after a new rule is added, so that can land on the wrong page if a search filter happens to be active at the time.

Severity: Low

### Steps to Reproduce

1. Go to Admin > License Admin > Compatibility Rules, with more than a handful of rules in the table.
2. Type a search term into the DataTable search box that matches only some of the rows.
3. Look at the "Showing ... (filtered from X total entries)" text at the bottom of the table. X is the filtered row count, not the actual total number of rules.

bhuvan-somisetty:
## Description

Fixes the wrong `recordsTotal` value returned by the License Compatibility Rules admin page when a search filter is active.

### Changes

`fetchRules()` in `AdminLicenseCompatibilityRules.php` was passing the search query straight into `getTotalRulesCount()` and then reusing that single result for both `recordsTotal` and `recordsFiltered` in the JSON response sent to DataTables. Since `getTotalRulesCount()` applies the search term as a `WHERE comment ILIKE` filter when one is given, `recordsTotal` ended up holding the filtered count instead of the true total number of rules any time a user typed something into the search box.

This splits the two values: `recordsTotal` is now always fetched with an empty search term, and `recordsFiltered` only re-queries with the search term when one is actually present. This mirrors the same pattern already used in `AdminCustomTextList.php` for its own server-side pagination endpoint.

## How to test

1. Go to Admin > License Admin > Compatibility Rules, with more than a few rules already in the table.
2. Type a search term into the DataTable search box that matches only some of the rows.
3. Check the "Showing 1 to N of N entries (filtered from X total entries)" text at the bottom of the table. Before this change, X equaled the filtered row count. After this change, X should equal the real total number of compatibility rules regardless of the search term.
4. Clear the search box and confirm the table still loads and paginates normally.

Fixes #3821


shaheemazmalmmd (approved):

PR #3824, opened 2026-09-16
fix(scandeps): accept null PyPI project URLs

Shubham-Padkonde:
## Description

Valid PyPI metadata can produce misleading component-parsing errors during dependency scanning. PyPI returns `project_urls: null` when a release has no URLs, but the parser calls `.items()` on it.

PyPI's [JSON API implementation](https://github.com/pypi/warehouse/blob/main/warehouse/legacy/api/json.py) explicitly serializes `release.urls or None`.

### Changes

Treat absent or null project URLs as an empty mapping. Keep source/homepage extraction and distribution download URLs unchanged. Add tests for null metadata across multiple components and for a populated URL mapping.

## How to test

From `utils/automation`, with `requirements.txt` installed:

```sh
python -m unittest discover -s tests -v
```

The null-metadata regression fails before the change because valid responses produce error logs. Both tests pass afterward. HTTP responses are mocked; the parser and SBOM classification use their real implementations.

The full FOSSology server suite was not run. Prepared with Codex assistance.


Shubham-Padkonde:
Corrected the missing copyright header on the new test file and added the missing DCO sign-off in cca8cc2c1. Both parser tests pass. Full REUSE 5.0.2 lint passes with no missing copyright/license information or read errors. The parser logic is unchanged.

shaheemazmalmmd (approved):
Code looks good.

PR #3826, opened 2026-09-16
fix(ui): compute unfiltered iTotalRecords for SH file browser table

### LINKED ISSUE #3825: Software Heritage file browser total record count is wrong when filtering
### Description

Right after tracking down a similar bug in the License Compatibility Rules table (#3821), I went looking for other DataTables endpoints with the same pattern and found one in the Software Heritage file browser. The ajax handler in `src/softwareHeritage/ui/AjaxSHDetailsBrowser.php` returns the same value for both `iTotalRecords` and `iTotalDisplayRecords`, and that value is already the filtered count rather than the true total number of files in the directory.

In `handle()`:

```php
return new JsonResponse(array(
  'sEcho' => intval($request->get('sEcho')),
  'aaData' => $vars['fileData'],
  'iTotalRecords' => $vars['iTotalDisplayRecords'],
  'iTotalDisplayRecords' => $vars['iTotalDisplayRecords']
));
```

`$vars['iTotalDisplayRecords']` comes from `createFileListing()`, where it's computed after the search term from `sSearch` has already been folded into `$options`:

```php
$descendantView = new UploadTreeProxy($uploadId, $options, $itemTreeBounds->getUploadTreeTableName(), 'uberItems');
$vars['iTotalDisplayRecords'] = $descendantView->count();
```

So that count is already scoped to whatever the user typed into the search box, which is correct for `iTotalDisplayRecords`, but `handle()` reuses that same number for `iTotalRecords`, which per the DataTables server-side protocol should be the unfiltered total row count. This endpoint never reports the real total once a filter is applied.

Severity: Low, no security impact, just a wrong number shown in the UI.

#### How to reproduce

1. Upload something with a decent number of files in one directory and let it scan.
2. Open the Software Heritage view for that upload and browse into a directory with several files.
3. Type something into the file browser's search box that only matches some of the files.
4. Look at the "Showing ... (filtered from X total entries)" text at the bottom of the table. X is the filtered count, not the real total number of files in that directory, because `iTotalRecords` always equals `iTotalDisplayRecords`.

#### Screenshots

None, this is a logic bug found by reading the source, not something that shows up visually beyond the wrong number in the table footer text.

### Versions

* Last commit id on master: 6345fd6f74a84d07a1fa427d10fa9dc672a90d74
* Operating System (lsb_release -a): Windows 11

### Logs

Not applicable, this doesn't produce any errors or warnings, the endpoint just returns a JSON response with the wrong count.

#### FOSSology logs

None.

#### Apache logs

None.

#### Job logs

None.

bhuvan-somisetty:
## Description

Fixes the Software Heritage file browser reporting the wrong total record count once a search filter is applied, same bug class as #3821/#3822 in the compatibility rules table.

### Changes

`handle()` in `src/softwareHeritage/ui/AjaxSHDetailsBrowser.php` was reusing `iTotalDisplayRecords` for `iTotalRecords`, so once the file browser search box had a term in it, the DataTables footer never showed the real unfiltered total. Now `createFileListing()` builds an unfiltered `UploadTreeProxy` first to get `iTotalRecords`, and only builds a second, filtered proxy for `iTotalDisplayRecords` when the `ext` or `head` search options are actually present, reusing the unfiltered count otherwise instead of running an identical query twice.

## How to test

1. Upload something with several files in one directory and let it scan.
2. Open the Software Heritage view for that upload and browse into a directory with several files.
3. Type something into the file browser's search box that only matches some of the files.
4. Check the "Showing 1 to N of N entries (filtered from X total entries)" text at the bottom of the table. X should now be the real total number of files in the directory, not the filtered count.

Fixes #3825


shaheemazmalmmd (approved):
Code looks good. Needs test.

shaheemazmalmmd:
@bhuvan-somisetty : the testing if the same doesn't work. Please check the screenshot, thanks for your efforts. 

image

PR #3827, opened 2026-09-16
fix(api): paginate copyright findings by file

### LINKED ISSUE #3823: GET /uploads/{id}/licenses returns duplicate file entries when license=false&copyright=true
### Description

When the `/uploads/{id}/licenses` endpoint is called with `license=false` and `copyright=true`, each file appears in the response once for every copyright statement found in it. Each of those duplicate entries contains all of the file's copyright statements.

A file with N copyright statements produces N identical entries, which means N² copyright strings in total. This also inflates the `X-Total-Pages` header and breaks pagination, because the same file can span pages.

Issue #2272 reported this in 2022. It was closed by PR #2273, which added the separate `/uploads/{id}/copyrights` endpoint, but the copyright-only branch of `/licenses` was never fixed.

**Root cause:** `UIExportList::getCopyrights()` (`src/www/ui/ui-export-list.php`) returns a flat list with one row per copyright statement. In `UploadHelper::getUploadLicenseList()` (`src/www/ui/api/Helper/UploadHelper.php`), the copyright-only branch does this:

```php
} elseif (!$boolLicense && $boolCopyright) {
  foreach ($copyrightList as $copyFilepath) {          // one iteration per statement
    $copyrightContent = array();
    foreach ($copyrightList as $copy) {                // collects ALL statements of that file
      if (($copyFilepath['filePath'] == $copy['filePath']) === true) {
        $copyrightContent[] = $copy['content'];
      }
    }
    ...
    $responseList[] = $responseRow->getArray($apiVersion);   // appended every time, no de-dup
  }
}
```

The outer loop never skips a `filePath` it has already handled, so one entry is added per statement instead of per file. The nested loop also makes this O(n²) for uploads with many copyright findings.

#### How to reproduce

1. Upload a package where at least one file has two or more copyright statements, for example a source file with `Copyright (C) 2020 Foo` and `Copyright (C) 2021 Bar`.
2. Run the `copyright` agent and one license agent, such as `nomos`.
3. Call:
   ```
   GET /repo/api/v2/uploads/<upload_id>/licenses?agent=nomos&license=false&copyright=true
   ```
   For v1, pass `page` and `limit` as headers instead.
4. **Actual:** the same `filePath` appears several times, and each entry has the full `copyright` array.
5. **Expected:** each `filePath` appears once, with its copyright statements listed once, and `X-Total-Pages` is based on the number of unique files.

#### Suggested fix

Group the rows by `filePath` in one pass, then build one `FileLicenses` entry per file:

```php
$grouped = [];
foreach ($copyrightList as $copy) {
  $grouped[$copy['filePath']][] = $copy['content'];
}
foreach ($grouped as $filePath => $contents) {
  $findings = new Findings();
  $findings->setCopyright(array_values(array_unique($contents)));
  $responseList[] = (new FileLicenses($filePath, $findings))->getArray($apiVersion);
}
```

### Versions

* Last commit id on master: `5e6b6d2710fe93c44eb9f420029c93a156c9cefb`

---

If you want, I can also prepare a PR with this fix.

Shubham-Padkonde:
## Description

Fixes #3823. With `license=false&copyright=true`, `/uploads/{id}/licenses` currently emits one entry per copyright finding instead of one per file. A file with two findings appears twice, and the duplicate inflates the total used for pagination.

### Changes

Group copyright rows by file path before constructing responses. Each file appears once with its findings in their original order; pagination and the returned total operate on those unique file entries. This also removes the nested scan of all findings for every row. The license-enabled branch is unchanged.

Add helper regression tests for API v1 and v2, non-adjacent findings for the same file, numeric filenames, all page boundaries, empty results, and exporter warnings.

## How to test

From `src`, after `composer install`:

```
vendor/bin/phpunit --no-configuration --bootstrap phpunit-bootstrap.php www/ui_tests/api/Helper/UploadHelperTest.php
vendor/bin/phpcs --standard=fossy-ruleset.xml www/ui/api/Helper/UploadHelper.php www/ui_tests/api/Helper/UploadHelperTest.php
```

On PHP 8.1: 10 regression failures and 4 passing cases before the fix; 14 tests and 40 assertions pass after. Both changed PHP files pass the project coding standard.

Developed and tested with OpenAI Codex assistance.

Broader validation: `vendor/bin/phpunit www/ui_tests/api` passed all 723 tests (1048 assertions) after installing the YAML extension. The existing PHPUnit configuration emits validation warnings for `cacheDirectory` and `coverage` with the repository-pinned PHPUnit 8.5.53; no configuration changes are included. A full application/database integration run was not performed.


anushkagupta200615-jpg:
i have already mentioned over pr that i will solve the issue @Shubham-Padkonde why u are submitting pr without asking

Shubham-Padkonde:
You are right: your issue said you were preparing the fix, and I should have coordinated before submitting. Sorry for duplicating your work. I am closing this PR in favor of your #3828.

PR #3829, opened 2026-09-17
fix(api): correct api version passing for scheduling scan

Kaushl2208:
## Description

With migration to API V2, We started following the new format for the payload. And tried to keep the v1 and v2 formatting separate, But for `POST /uploads` , SchedulingAnalysis was always considering every payload to be v1 (by design)

### Changes

Passing the correct `API_VERSION` to `handleScheduleAnalysis()` (which by default thinks it v1, if nothing is passed.
Add a Schema Validator, if there is an issue with the JSON Schema of `scan_options` throw exception

## How to test

1. Use `curl` to upload a file and try to run analysis for `nomos` and `copyrightEmailAuthor` with these exact keys. 
2. Analysis should be scheduled and these agents should be triggered: `ununpack, adj2nest, nomos and copyright`

Before this code, if you try the same curl request, it will not fail but wont schedule. Copyright or Nomos.


shaheemazmalmmd (approved):
Code looks good. Needs test

PR #3833, opened 2026-09-17
fix(ui): reject negative iDisplayStart/iDisplayLength in file browsers

### LINKED ISSUE #3832: Negative iDisplayLength breaks file browser ajax queries instead of showing all rows
### Description

A DataTables `iDisplayLength` of `-1` (the standard DataTables convention for "show all entries") breaks the file browser ajax endpoints instead of showing all rows. `GetParm('iDisplayLength', PARM_INTEGER)` just does an `intval()` with no lower bound, so a negative value passes straight through, and the code that appends it to the SQL query only checks truthiness, not that it's positive.

I found this in `createFileListing()`:

```php
$offset = GetParm('iDisplayStart', PARM_INTEGER);
$limit = GetParm('iDisplayLength', PARM_INTEGER);
if ($offset) {
  $orderString .= " OFFSET $offset";
}
if ($limit) {
  $orderString .= " LIMIT $limit";
}
```

In PHP `-1` is truthy, so `if ($limit)` lets it through and `LIMIT -1` gets appended straight into the raw SQL string that's executed a few lines down. Postgres rejects a negative `LIMIT` outright, so the query throws instead of running.

Same code shows up in three places:

* `src/softwareHeritage/ui/AjaxSHDetailsBrowser.php`, around line 172-179
* `src/www/ui/async/AjaxFileBrowser.php`, around line 156-163
* `src/www/ui/async/AjaxExplorer.php`, around line 214-221 (this one uses `intval($request->get(...))` directly instead of `GetParm`, but hits the same truthiness problem)

#### How to reproduce

1. Log in and open any upload's file listing so one of the three ajax endpoints above is being called (mod=ajax_sh_browser, mod=ajax_file_browser, or mod=ajax_explorer depending on where you are).
2. Open browser dev tools, find the ajax request to one of these endpoints, and re-send it with `iDisplayLength=-1` added to the query string (or just build the URL by hand with that param).
3. The request fails instead of returning all rows, because the generated SQL contains `LIMIT -1`.

#### Screenshots

Not applicable, this only shows up in the failed ajax response / server log, there's nothing to screenshot in the UI itself since the table just fails to load for that request.

### Versions

* Last commit id on master: 66fe75f23480660e6405872f11866a122a08ff70
* Operating System (lsb_release -a): Windows 11

### Logs

Any logs (if any) generated in

#### FOSSology logs

Would show a Postgres error along the lines of `ERROR: LIMIT must not be negative` bubbling up from the query built in one of these three files.

#### Apache logs

None.

#### Job logs

Not applicable, this is a UI ajax call, not a job.

bhuvan-somisetty:
## Description

Fixes negative iDisplayLength/iDisplayStart breaking the file browser ajax queries instead of showing all rows.

### Changes

`AjaxSHDetailsBrowser.php`, `AjaxFileBrowser.php` and `AjaxExplorer.php` all appended OFFSET/LIMIT to the SQL query based on a bare truthiness check of the iDisplayStart/iDisplayLength params. In PHP -1 is truthy, and DataTables uses -1 as the iDisplayLength value for its All entries option, so that value went straight into the raw SQL string as LIMIT -1, and Postgres rejects a negative LIMIT with an error instead of running the query. Changed the checks in all three files from `if ($limit)`/`if ($offset)` to `if ($limit > 0)`/`if ($offset > 0)` so only positive values get appended.

## How to test

1. Log in and open any upload's file listing.
2. Find the ajax request going to one of ajax_sh_browser, ajax_file_browser or ajax_explorer and resend it with iDisplayLength=-1 added to the query string.
3. Before this change the request fails with a Postgres error about a negative LIMIT. After this change it succeeds and just returns all rows without a LIMIT clause.

Fixes #3832


shaheemazmalmmd:
Check #3832

PR #3835, opened 2026-09-17
fix(scheduler): prevent buffer overflow in meta_agent_init

### LINKED ISSUE #3834: Bug: Heap buffer overflow in scheduler's meta_agent_init()
In `src/scheduler/agent/agent.c`, the length check currently allows commands up to `MAX_CMD` (1023 bytes), but `strcat(ma->raw_cmd, " --scheduler_start")` then appends another 18 bytes. This can overflow the 1024-byte buffer when the command is between 1006 and 1023 bytes long.

To reproduce, initialize the scheduler with a command that is exactly 1010 characters long. It passes the length check but overflows when `strcat` is called.

Proposed fix: change the validation to `strlen(cmd) > (MAX_CMD - 18)` and use `snprintf` instead of `strcat`.

mandar1045:
### TL;DR

Fixes a heap buffer overflow in `meta_agent_init()` caused by `strcat` appending 18 bytes beyond the `raw_cmd` buffer.

* Adjusts the command length check to account for the appended string.
* Replaces `strcpy`/`strcat` with `snprintf`.
* Adds a unit test for oversized commands.

Fixes #3834


src/scheduler/agent/agent.c
```diff
@@ -904,7 +904,7 @@ meta_agent_t* meta_agent_init(char* name, char* cmd, int max, int spc)
   }
 
   /* confirm valid inputs */
-  if (strlen(name) > MAX_NAME || strlen(cmd) > MAX_CMD)
+  if (strlen(name) > MAX_NAME || strlen(cmd) > (MAX_CMD - 18))
```
shaheemazmalmmd: @mandar1045 : can you tell us the reason behind magic number 18 ?? is it depends on filezise/Argument size ??

shaheemazmalmmd (changes requested):

src/scheduler/agent/agent.c
mandar1045: Hi @shaheemazmalmmd, the `18` is just `strlen(" --scheduler_start")`.

Since `" --scheduler_start"` gets appended to `cmd` a few lines later, we need to reserve those 18 bytes to avoid a buffer overflow. I can replace the number with `strlen(" --scheduler_start")` or a constant if you'd prefer.

shaheemazmalmmd: Sure replace with strlen is better

mandar1045: okay

PR #3838, opened 2026-09-17
fix(ui): reject negative iDisplayStart/iDisplayLength in folder browse listing

### LINKED ISSUE #3837: Browse page: negative iDisplayLength breaks the folder upload listing instead of being ignored
### Description

The main Browse page (the one that lists uploads inside a folder) sends its pagination params straight into the SQL query with no lower bound check, so a negative `iDisplayLength` breaks the page instead of just being ignored.

This is in `getListOfUploadsOfFolder()` in `src/www/ui/async/AjaxBrowse.php`, around lines 288-292:

```php
$rangedFilterParams = $this->filterParams;
$rangedFilterParams[] = intval($request->get('iDisplayStart'));
$statementString .= ' OFFSET $' . count($rangedFilterParams);
$rangedFilterParams[] = intval($request->get('iDisplayLength'));
$statementString .= ' LIMIT $' . count($rangedFilterParams);
```

`iDisplayStart`/`iDisplayLength` come straight from the DataTables ajax request via `intval($request->get(...))`, with nothing checking they're actually positive, and they get appended unconditionally, there isn't even a truthy check like the OFFSET/LIMIT code in some of the other browsers in this codebase has. If `iDisplayLength` is negative, it gets bound straight into `LIMIT $n` and Postgres rejects a negative LIMIT outright, so the whole folder listing fails to load instead of just falling back to some default page size.

#### How to reproduce

1. Log in and open the main Browse page (`mod=browse`) so it loads a folder's upload list.
2. Open browser dev tools, find the ajax call to `?mod=browse-processPost` that the table makes, and resend it with `iDisplayLength=-1` in the request (everything else the same, e.g. `folder`, `iDisplayStart`, `sEcho`, etc).
3. Instead of returning rows, the response comes back empty/erroring because the generated SQL has `LIMIT -1` bound in, which Postgres rejects.

#### Screenshots

Not applicable, this only shows up as a failed ajax response / a Postgres error in the log, there's nothing to screenshot in the UI besides the table just not loading for that request.

### Versions

* Last commit id on master: 6345fd6f74a84d07a1fa427d10fa9dc672a90d74
* Operating System: Windows 11

#### FOSSology logs

Would show a Postgres error like `ERROR: LIMIT must not be negative` coming from the query built in `getListOfUploadsOfFolder()`.

#### Apache logs

None.

#### Job logs

Not applicable, this is a UI ajax call, not a job.

bhuvan-somisetty:
## Description

The main Browse page's folder listing (`getListOfUploadsOfFolder()` in `AjaxBrowse.php`) bound `iDisplayStart`/`iDisplayLength` straight into the query's OFFSET/LIMIT parameters with no lower bound check. DataTables uses `-1` for `iDisplayLength` as its "All entries" option, and that went straight into `LIMIT $n`, which Postgres rejects outright, so the whole listing failed to load instead of returning all rows.

### Changes

Clamped `iDisplayStart` to `0` when it's not positive, and bind `iDisplayLength` as `NULL` when it's not positive. Postgres treats `LIMIT NULL` the same as no LIMIT clause at all, so a negative value (like DataTables' `-1`) now actually returns all rows instead of erroring out. Same fix pattern already used in #3833 for the other file browsers.

## How to test

1. Log in and open the main Browse page (`mod=browse`) so it loads a folder's upload list.
2. Open browser dev tools, find the ajax call to `?mod=browse-processPost`, and resend it with `iDisplayLength=-1` added (keep `folder`, `iDisplayStart`, `sEcho`, etc. the same).
3. Before this change, the request fails because the generated SQL has `LIMIT -1` bound in, which Postgres rejects.
4. After this change, the request succeeds and returns all rows for the folder without a LIMIT.

Fixes #3837


shaheemazmalmmd:
SEE: #3837

PR #3843, opened 2026-09-19
fix(api): pass authenticated user id to folder access checks and guard root folder lookup

### LINKED ISSUE #3842: Bug: Fatal Error in FolderDao::isFolderAccessible() when called without session user in REST API
### Description

In the REST API, calling `GET /folders/{id}` or `POST /uploads` can result in an unhandled 500 error (`TypeError: Call to a member function getId() on null`) instead of performing authorization checks against the authenticated user.

This occurs because both `FolderController::getFolders()` and `UploadController::postUpload()` invoke `FolderDao::isFolderAccessible($folderId)` without providing the caller's authenticated user ID (`$this->restHelper->getUserId()`).

#### Technical Details

In `src/lib/php/Dao/FolderDao.php` (lines 518–525):
```php
  public function isFolderAccessible($folderId, $userId = null)
  {
    if ($userId == null) {
      $userId = Auth::getUserId();
    }
    $rootFolder = $this->getRootFolder($userId)->getId();
    return $this->isInFolderTree($rootFolder, $folderId);
  }

### PR DESCRIPTION
**Pull request by harshitttt077**
## Description

In the REST API, calling `GET /folders/{id}` or `POST /uploads` without an active UI session caused `FolderDao::isFolderAccessible()` to throw an unhandled 500 fatal error (`TypeError: Call to a member function getId() on null`).

This happened because `FolderController::getFolders()` and `UploadController::postUpload()` called `FolderDao::isFolderAccessible($folderId)` without passing the caller's authenticated user ID (`$this->restHelper->getUserId()`). In token-authenticated requests, `Auth::getUserId()` returns `0`, which caused `getRootFolder(0)` to return `null`.

### Changes

- **`src/www/ui/api/Controllers/FolderController.php`**: Passed `$this->restHelper->getUserId()` to `$folderDao->isFolderAccessible()`.
- **`src/www/ui/api/Controllers/UploadController.php`**: Passed `$this->restHelper->getUserId()` to `$this->restHelper->getFolderDao()->isFolderAccessible()`.
- **`src/lib/php/Dao/FolderDao.php`**: Added a null check on `$this->getRootFolder($userId)` so requests for non-existent users safely return `false` instead of calling `getId()` on `null`.
- **`src/lib/php/Dao/test/FolderDaoTest.php`**: Added `testIsFolderAccessible_nullRootFolderReturnsFalse` to verify that calls with non-existent user IDs safely return `false`.
- **`src/www/ui_tests/api/Controllers/FolderControllerTest.php` & `UploadControllerTest.php`**: Updated Mockery expectations to verify that the user ID parameter is forwarded to `isFolderAccessible()`.

## How to test

1. Send an authenticated API request with a token:
   ```bash
   curl -k -s -X GET "https://<fossology-host>/repo/api/v1/folders/2" \
     -H "Authorization: Bearer <token>"
   ```
2. Before this fix: the server threw an unhandled 500 `TypeError` in `FolderDao::isFolderAccessible()` when session state was not set.
3. After this fix: access is checked against the authenticated user token and returns 200 (if accessible) or 403 (if not accessible).
4. Run unit tests:
   - `FolderControllerTest`
   - `UploadControllerTest`
   - `FolderDaoTest`

Fixes #3842

### CODE REVIEW

File: src/lib/php/Dao/FolderDao.php [hunk 1]
```diff
@@ -520,8 +520,11 @@ public function isFolderAccessible($folderId, $userId = null)
     if ($userId == null) {
       $userId = Auth::getUserId();
     }
-    $rootFolder = $this->getRootFolder($userId)->getId();
-    return $this->isInFolderTree($rootFolder, $folderId);
+    $rootFolder = $this->getRootFolder($userId);
```
shaheemazmalmmd, line 523: Can you please explain why ->getId(); is removed ??

shaheemazmalmmd (changes requested):

src/lib/php/Dao/FolderDao.php
harshitttt077: Previously, chaining $this->getRootFolder($userId)->getId() directly caused an unhandled fatal error (TypeError: Call to a member function getId() on null) whenever $this->getRootFolder($userId) returned null (for instance, in token-authenticated API calls where Auth::getUserId() resolves to 0).

sSplitting the line allows checking if (!$rootFolder) first so it can safely return false before calling ->getId() on line 527

shaheemazmalmmd:
Puplicate of https://github.com/fossology/fossology/pull/3797

Issue #3846, opened 2026-09-19
Fix remaining spelling mistakes in user-facing messages (succesfully, occured, nonexistant)


bhuvan-somisetty:
### Description

While reading through the UI and REST API code I noticed a few user-facing messages that are still misspelled on master. PR #3605 fixed a batch of similar typos but these were missed, and I couldn't find an open issue for them.

The ones I found:

* `src/www/ui/api/Helper/UserHelper.php:63`: `"User updated succesfully!"` is returned by the REST API when a user is edited, so API clients see the typo in the response body.
* `src/www/ui/api/Controllers/ReportController.php:156`: `"Some error occured!"` in the HttpInternalServerErrorException message.
* `src/www/ui/scripts/user-edit.js:42` and `:78`: `"Some error occured"` shown in the user edit page when the ajax call fails.
* `"Invalid URL, nonexistant item"` in `src/www/ui/template/ui-view-info.html.twig:10`, `src/lib/php/Dao/SearchHelperDao.php:54` and `src/buckets/ui/ui-buckets.php:180`. This string is also a msgid in `src/www/locale/de_DE` and `nl_NL` `messages.po`, so those two entries need to be updated together with the source or the translations stop matching.

It should be `successfully`, `occurred` and `nonexistent`.

#### How to reproduce

1. Run `git grep -n "succesfully\|error occured\|nonexistant" -- src` on current master and you will see the lines above.
2. To see the API one live, send a valid `PUT /users/{id}` request as an admin and read the `message` in the 200 response.

### Versions

* Last commit id on master: 6345fd6f7
* Operating System (lsb_release -a): not OS specific, found by reading the source

### Logs

Not applicable, this is text only.

PR #3605, opened 2026-04-20
fix: correct typos in user-facing messages

mandar1045:
## Summary
- correct typo errors in user-facing/runtime messages
- include the common-sysconfig JWKS description typo fix
- keep comment-only and copied/generated text out of the PR

## Verification
- git diff origin/master --check
- checked PR diff for secret-looking OIDC/client/hex values

PR #3847, opened 2026-09-19
fix(ui): correct spelling in user update and error messages

bhuvan-somisetty:
## Description

Fixes a few spelling mistakes in user-facing messages that were left over after #3605: `succesfully` and `occured`.

### Changes

* `UserHelper.php`: `User updated succesfully!` is now `User updated successfully!` (this is the message the REST API returns after a user update).
* `ReportController.php`: `Some error occured!` is now `Some error occurred!`.
* `user-edit.js`: `Some error occured` is now `Some error occurred` in both places it is shown on the user edit page.

I left the `Invalid URL, nonexistant item` string from the issue out of this PR on purpose. It is a msgid in the de_DE and nl_NL `messages.po` files and the compiled `de_DE/messages.mo` is checked in, so changing it needs the translation catalogs regenerated together with the source. I'd rather do that as a separate change than leave the German translation out of sync.

## How to test

1. Run `git grep -n "succesfully\|error occured" -- src/www/ui` and check that nothing is left in these files.
2. Send a valid user update request to the REST API as an admin (`PUT /users/{id}`) and check that the response message reads `User updated successfully!`.

Part of #3846


shaheemazmalmmd:
@bhuvan-somisetty : thank you for all your efforts this PR https://github.com/fossology/fossology/pull/3848/changes have more corrections compared to this.

bhuvan-somisetty:
Thanks for the note! Makes sense, #3848 covers more, so I'll leave it to that one.

PR #3850, opened 2026-09-21
fix: restrict REG_EXCEPTION_COPY camelCase pattern to case-sensitive …

its-sushant:
## Description

`REG_EXCEPTION_COPY` contains a `To[A-Z][a-zA-Z]{4,}` branch to skip camelCase identifiers like `Copyright.ToString()`. Since the regex is compiled with icase, [A-Z] loses its meaning and the branch matches any word starting with to, so `Copyright TOSHIBA/Toshiba CORPORATION, 2022` was flagged as an exception and deactivated.

Fixed by making the prefix case-sensitive: `(?-i:To[A-Z])[a-zA-Z]{4,}`.

Test file: 
[copyright_corpus.txt](https://github.com/user-attachments/files/32457586/copyright_corpus.txt)


shaheemazmalmmd (approved):
Code looks good.

PR #3760, opened 2026-08-11
fix: fail dependent jobs when a prerequisite agent fails + docker permission fix

dineshr93:
Fixes #3759

## Problem

Two related issues made URL/git uploads fail and leave jobs stuck:

1. Missing execute bit on repository directories - libfossrepo creates directories under a restrictive umask that can strip the owner/group execute bits (leaving mode 2660 instead of 2770). The scheduler runs as the fossy user, so without the execute bit it cannot create log files inside such directories, and agents (e.g. wget_agent) cannot traverse/create work directories, causing failures like: Permission denied failed to open [logs/00/05/65/000565] and fatal: Invalid path .../.git: Permission denied.

2. Stuck dependent jobs - when wget_agent fails, its dependent ununpack and adj2nest jobs never become runnable, because the scheduler checkout query requires every dependency to have completed successfully (jq_endtime IS NOT NULL AND jq_end_bits < 2). A failed prerequisite can never satisfy that, so dependents stay queued forever.

## Changes

- docker-entrypoint.sh: enforce drwxrws--- (2770) on every repository directory after fo-postinstall, so the missing execute bit can never make jobs fail or get stuck.
- src/scheduler/agent/sqlstatements.h: add jobsql_fail_dependents, a recursive CTE marking every queue entry transitively depending on a failed jq_pk as failed (depth-guarded, leaves finished entries untouched).
- src/scheduler/agent/database.c: on JB_FAILED transition, also fail dependent queue entries so ununpack/adj2nest fail immediately instead of getting stuck.

## Verification

The recursive SQL runs cleanly against a live FOSSology database; in a rollback test a failed wget_agent correctly marks its dependent ununpack and adj2nest entries failed (jq_end_bits=2, jq_endtext=dependency failed) instead of leaving them stuck. The chmod 2770 find is idempotent and repairs the broken 2660 directory.

Issue #3759, opened 2026-08-11
unpack and adj2nest remains stuck without fail if dependent root job wget fails


dineshr93:
Version: [4.7.1.89], Branch: [master], Commit: [#7b9177] 2026/08/07 08:19 +00:00 built @ 2026/08/10 07:29 +00:00 (docker compose up -d)

**what??**
whenever the wget fails the dependency jobs unpack and adj2nest job gets stuck. Shouldnt it fail as soon as the root job **wget** fails due to various reasons (in my case its permission denied)??

<img width="1645" height="165" alt="Image" src="https://github.com/user-attachments/assets/9487b10d-3d9c-41d0-9f40-7dc687e6ebd3" />

when clicking **579**
wget job
<img width="1648" height="710" alt="Image" src="https://github.com/user-attachments/assets/c1f57bec-7d57-4c4a-a19f-db76df3ba07a" />

with following error
```
2026-08-11 19:41:49 wget_agent [0] :: JOB[579].wget_agent[158245.localhost]: "Cloning into '/srv/fossology/repository/58eae198b897/wget/wget.158245.dir'..."
2026-08-11 19:41:49 wget_agent [0] :: JOB[579].wget_agent[158245.localhost]: "fatal: Invalid path '/srv/fossology/repository/58eae198b897/wget/wget.158245.dir/.git': Permission denied"
2026-08-11 19:41:49 wget_agent [0] :: JOB[579].wget_agent[158245.localhost]: "FATAL /fossology/src/wget_agent/agent/wget_agent.c.632: [/fossology/src/wget_agent/agent/wget_agent.c:632] Error exit: 128, CMD:git clone https://github.com/psf/requests.git   /srv/fossology/repository/58eae198b897/wget/wget.158245.dir && rm -rf /srv/fossology/repository/58eae198b897/wget/wget.158245.dir/.git"
2026-08-11 19:41:49 wget_agent [0] :: JOB[579].wget_agent[158245.localhost]: "FATAL /fossology/src/wget_agent/agent/wget_agent.c.637: please make sure the URL of repo is correct, also add correct proxy for your version control system, command is:git clone https://github.com/psf/requests.git   /srv/fossology/repository/58eae198b897/wget/wget.158245.dir && rm -rf /srv/fossology/repository/58eae198b897/wget/wget.158245.dir/.git, GlobalTempFile is:/srv/fossology/repository/58eae198b897/wget/wget.158245, rc is:32768. "
[...]

PR #3851, opened 2026-09-21
fix(scheduler): sweep for dependent jobs stuck by an insertion race

### LINKED ISSUE #3852: Some dependent jobs still remain stuck (not auto-failed) after a prerequisite fails, despite #3760
## Description

Follow-up to #3759 / #3760.

#3760 fixed the common stuck-dependent-job case by adding a one-shot
cascade (`jobsql_fail_dependents`) that fails a job's dependents at the
exact moment it transitions to `JB_FAILED`. In production, this still
leaves some dependent jobs stuck, requiring manual kill.

## Root cause

`JobQueueAdd()` (`src/lib/php/common-job.php`) inserts one `jobqueue` +
`jobdepends` row per selected agent, each in its own transaction, in a
loop. If the upstream job fails very fast - e.g. `wget_agent` hitting a
corrupt/missing URL - it can finish and run its one-shot cascade before
every sibling agent row (`pkgagent`, `copyright`, `ecc`, `keyword`,
`mimetype`, `monk`, `ojo`, `reuser`, `decider`, ...) has even been
inserted into `jobdepends`.

Those late-inserted rows are invisible to the cascade that already ran,
and are never retried: `basic_checkout` only ever releases a job once its
dependency finished *successfully*, never once it failed, so the row sits
in the queue forever with no agent ever spawned for it.

## Observed behavior

For a single failed `wget_agent` job, some direct dependents of the
(cascade-)failed `adj2nest` job correctly flipped to `dependency failed`
immediately, while structurally-identical siblings (same single
dependency edge) sat queued for 1-3 minutes and had to be killed manually:

Image

```
274121 / 274120  dependency failed   adj2nest
274122 / 274121  dependency failed   nomos        <- caught by the #3760 cascade
274123 / 274121  killed by user      pkgagent     <- inserted after the cascade ran, stuck
274125 / 274121  killed by user      copyright    <- same
274126 / 274121  killed by user      ecc          <- same
...

```

## Fix

Proposed in #3851: add `jobsql_fail_stuck_dependents`, a lightweight,
unparametrized sweep run on every scheduler poll cycle in
`database_update_event()` (alongside `basic_checkout`). It fails any
still-pending `jobqueue` row whose direct dependency is already marked
failed, so late-inserted rows get caught on the very next poll instead of
depending on a single cascade that ran before they existed.

dineshr93:
Fixes #3852. Follow-up to #3759 / #3760.

## Problem

#3760 fixed the common stuck-dependent-job case by adding a one-shot cascade
(`jobsql_fail_dependents`) that fails a job's dependents at the exact moment
it transitions to `JB_FAILED`.

That closes the gap for the case it tested, but there is still a race: the
PHP side (`JobQueueAdd()` in `src/lib/php/common-job.php`) inserts one
`jobqueue` + `jobdepends` row per selected agent, each in its own
transaction, in a loop. If the upstream job fails very fast - e.g.
`wget_agent` hitting a corrupt/missing URL, which can fail in well under a
second - it can finish and run its one-shot cascade before every sibling
agent row (`pkgagent`, `copyright`, `ecc`, `keyword`, `mimetype`, `monk`,
`ojo`, `reuser`, `decider`, ...) has even been inserted into `jobdepends`.

Those late-inserted rows are invisible to the cascade that already ran, and
they are never retried: `basic_checkout` only ever releases a job once its
dependency finished *successfully*, never once it failed, so the row is
stuck in the queue forever with no agent ever spawned for it - until a user
manually kills it.

Observed in production: for a single failed `wget_agent` job, some direct
dependents of the (cascaded-)failed `adj2nest` job correctly flipped to
`dependency failed` immediately, while structurally-identical siblings
(same single dependency edge) sat queued for 1-3 minutes and had to be
killed manually. This is exactly what you'd expect from the insertion race
above, and matches the reporter's observation that "some dependencies gets
cancelled automatically...but some gets stuck."

## Changes

- `src/scheduler/agent/sqlstatements.h`: add `jobsql_fail_stuck_dependents`,
  an unparametrized sweep query. It fails any still-pending (`jq_endtime IS
  NULL`) `jobqueue` row whose direct dependency (`jobdepends`) already has
  the failed bit set.
- `src/scheduler/agent/database.c`: run that sweep on every scheduler poll
  cycle inside `database_update_event()`, right alongside the existing
  `basic_checkout` poll.

This is a light, self-healing safety net rather than a replacement for the
existing one-shot cascade (which still gives instant feedback for the rows
that do exist in time): it only scans currently pending rows and joins
through indexed `jobqueue`/`jobdepends` primary keys, so cost is bounded by
the size of the pending queue, not job history. Multi-level dependency
chains resolve within a few poll cycles since polling runs every few
seconds, closing the race regardless of insertion order or timing.

## Verification

Reviewed against the production log sequence that exposed the bug (single
`wget_agent` failure with ~12 sibling scan-agent dependents, only a subset
auto-failed); the new sweep independently catches any dependent whose
prerequisite already shows failed, regardless of when its `jobqueue`/
`jobdepends` rows were inserted relative to the original failure event.

No C build toolchain (libpq-dev) was available in this environment to run
the scheduler's own test suite; please rebuild `fo_scheduler` and validate
against a fast-failing URL upload with several selected scan agents before
merging.


dineshr93:
Tested manually with uploading invalid file url and it succesfully fails. 
image

